Issue #155, opened 2014-06-12
New rule: enforce header style (often for copyright)


derekcicerone:
Just a simple rule to enforce a certain header style.  Ideally it would support regex so that things like a copyright year can vary.

alexeagle:
@ScottSWu want to do this one?

PR #221, opened 2014-10-07
header rule

fent:
Takes care of #155

Not sure if this needs a walker, is there a better way to access only the first token?

And as of now, only headers in multiline comments `/* */` will be accepted, should it also allow several lines of single line comments?


ashwinr:
closing the PR, we're rewriting all rules for 1.4

PR #245, opened 2014-11-22
Make type requirement for call-signatures more explicit

park9140:
Make type requirement for call-signatures more explicit, also don't require type for constructors

We were getting additional type errors for constructors and the tests didn't catch it, I made the tests also verify no extra errors were found.


park9140:
@ashwinr, any idea when we might get a new release?

ashwinr:
sorry for the delay, you should be able to see the latest release now.

park9140:
@ashwinr, thanks for the relase :).

PR #265, opened 2015-01-22
Migrate variable name rule to 1.4

adidahiya:
- also fix some compilation / linting errors in the source


adidahiya:
for some reason it's not compiling `quoteMarkRule` and its tests...

adidahiya:
rebased & squashed

PR #267, opened 2015-01-22
Migrate use strict rule to 1.4

adidahiya:


ashwinr:
geoff, please review this.

src/rules/useStrictRule.ts
```diff
@@ -34,61 +35,51 @@ class UseStrictWalker extends Lint.ScopeAwareRuleWalker<{}> {
...
         // current depth is 2: global scope and the scope created by this module
-        // but skip declare module statements
-        if (this.getCurrentDepth() === 2 &&
-            !(childCount > 0 && TypeScript.childAt(node.modifiers, 0).kind() === TypeScript.SyntaxKind.DeclareKeyword)) {
-
-            if (this.hasOption(UseStrictWalker.OPTION_CHECK_MODULE)) {
-                this.checkUseStrict(node, node.moduleElements);
+        if (this.getCurrentDepth() === 2 && !hasDeclareModifier) {
+            if (node.body != null && this.hasOption(UseStrictWalker.OPTION_CHECK_MODULE)) {
+                this.handleBlock(node, <ts.Block> node.body);
```
gscshoyru: How do you know this isn't a ModuleDeclaration here? (Is it because of the declaration modifier? If so, that's fine, I'm just confused/curious. Also, minor detail -- this should probably be ModuleBlock and not Block, to match up with how body is typed.

src/rules/useStrictRule.ts
```diff
@@ -34,61 +35,51 @@ class UseStrictWalker extends Lint.ScopeAwareRuleWalker<{}> {
...
+                node.body != null &&
+                this.hasOption(UseStrictWalker.OPTION_CHECK_FUNCTION)) {
+            this.handleBlock(node, node.body);
         }
 
         super.visitFunctionDeclaration(node);
     }
 
-    private checkUseStrict(node: TypeScript.ISyntaxNode, syntaxList: TypeScript.ISyntaxElement[]) {
+    private handleBlock(node: ts.Declaration, block: ts.Block) {
```
gscshoyru: Oh I see that's why you typed it that way in the cast, above. I think that makes sense.

src/rules/useStrictRule.ts
adidahiya: `this.handleBlock` can't take a `ModuleBlock`, it must be cast to `Block`

gscshoyru: Right, I commented on the thing down below. Although now that I think about it, is there a reason why you can't use the union type in the handleBlock now that we have it, since we're using 1.4?

adidahiya: yep, union type works, updating now

adidahiya:
removed union type so that old tslint will work on our src

PR #270, opened 2015-01-27
Migrate one line rule to 1.4

gscshoyru:


src/rules/oneLineRule.ts
```diff
@@ -27,80 +25,212 @@ export class Rule extends Lint.Rules.AbstractRule {
...
-
-                var lastLine = this.getLine(lastState.position);
-                var currentLine = this.getLine(this.getPosition());
-                var position = this.getPosition() + TypeScript.leadingTriviaWidth(token);
-
-                if (this.hasOption(OPTION_BRACE) && currentLine !== lastLine) {
-                    failure = this.createFailure(position, TypeScript.width(token), Rule.BRACE_FAILURE_STRING);
-                } else if (this.hasOption(OPTION_WHITESPACE) && !this.hasTrailingWhiteSpace(lastState.token)) {
-                    failure = this.createFailure(position, TypeScript.width(token), Rule.WHITESPACE_FAILURE_STRING);
+class BraceWalker extends Lint.RuleWalker {
```
ashwinr: while you're at it, rename this to OneLineWalker for consistency

src/rules/oneLineRule.ts
```diff
@@ -27,80 +25,212 @@ export class Rule extends Lint.Rules.AbstractRule {
...
+        var sourceFile = node.getSourceFile();
+        var thenStatement = node.thenStatement;
+        if (thenStatement.kind === ts.SyntaxKind.Block) {
+            var expressionCloseParen = node.getChildAt(3);
+            var thenOpeningBrace = thenStatement.getChildAt(0);
+            this.handleOpeningBrace(expressionCloseParen, thenOpeningBrace);
+        }
+
+        var elseStatement = node.elseStatement;
+        if (node.elseStatement != null) {
```
ashwinr: just elseStatement, no node.

ashwinr: just elseStatement, no node.

ashwinr: also do an early return here if elseStatement is null, it's cleaner to read code below because it won't be nested

gscshoyru: We can't do an early return statement; we need to call super at the end.

src/rules/oneLineRule.ts
```diff
@@ -27,80 +25,212 @@ export class Rule extends Lint.Rules.AbstractRule {
...
+            // find the else keyword
+            var elseKeyword = node.getChildren().filter((child) => child.kind === ts.SyntaxKind.ElseKeyword)[0];
+            if (elseStatement.kind === ts.SyntaxKind.Block) {
+                var elseOpeningBrace = elseStatement.getChildAt(0);
+                this.handleOpeningBrace(elseKeyword, elseOpeningBrace);
+            }
+            if (this.hasOption(OPTION_ELSE)) {
```
ashwinr: why don't you move this out to the outer if check?

ashwinr: oh, i see. nm

src/rules/oneLineRule.ts
```diff
@@ -27,80 +25,212 @@ export class Rule extends Lint.Rules.AbstractRule {
...
+        if (body.kind === ts.SyntaxKind.ModuleBlock) {
+            var openBraceToken = body.getChildAt(0);
+            this.handleOpeningBrace(nameNode, openBraceToken);
+        }
+        super.visitModuleDeclaration(node);
+    }
+
+    public visitEnumDeclaration(node: ts.EnumDeclaration) {
+        var nameNode = node.name;
+        var openBraceToken = node.getChildren().filter((child) => child.kind === ts.SyntaxKind.OpenBraceToken)[0];
```
ashwinr: use a private utility method that filters children based on the kind and gives you back the first child. also, this is slightly buggy because it'll crash TSLint when indexing into [0] if the child is not present. we shouldn't be crashing even in incorrect code.

src/rules/oneLineRule.ts
```diff
@@ -27,80 +25,219 @@ export class Rule extends Lint.Rules.AbstractRule {
...
+        }
+        if (failure) {
+            this.addFailure(failure);
+        }
+    }
 
-        var kind = trivia.syntaxTriviaAt(0).kind();
-        return (kind === TypeScript.SyntaxKind.WhitespaceTrivia);
+    private static getFirstChildOfKind(node: ts.Node, kind: ts.SyntaxKind) {
+        return node.getChildren().filter((child) => child.kind === kind)[0];
```
ashwinr: this is still going to crash TSLint if there's no first child

gscshoyru: No it won't; I put safeguards in handleOpeningBrace in case we get an undefined node back.

PR #271, opened 2015-01-28
Migrate typedef rule to 1.4

adidahiya:


adidahiya:
Looks like `tslint` couldn't handle the `protected` keyword, therefore the travis CI build failed. I've updated the commit.

test/rules/typedefRuleTests.ts
```diff
@@ -34,92 +34,133 @@ describe("<typedef, not enabled>", () => {
...
+            [36, 12],
+            "expected member-variable-declaration: 'Member' to have a typedef");
+
+        assertFailures(expectedFailure);
     });
 
     it("only has the expected failures", function() {
         if (actualFailures.length > 0) {
-            assert(false, "got these extra errors " + JSON.stringify(actualFailures, null, 2));
+            assert(false, "got " + actualFailures.length + " extra errors");
```
adidahiya: `JSON.stringify` no longer works on the failures because there's a circular reference in the `Node` object structure

src/language/walker/syntaxWalker.ts
```diff
@@ -253,6 +265,15 @@ module Lint {
                     this.visitFunctionDeclaration(<ts.FunctionDeclaration> node);
                     break;
 
+                case ts.SyntaxKind.FunctionExpression:
+                    this.visitFunctionExpression(<ts.FunctionExpression> node);
+                    break;
+
+                case ts.SyntaxKind.GetAccessor:
```
gscshoyru: It is plasuible/likely the case that some future rule will want to walk these differently than regular method declarations. These should probably be their own functions -- though you can make a helper that all three call in your rule.

adidahiya: ok, I'll split them out.

it sort of pains me to have to write out the following:
```ts
    public visitFunctionDeclaration(node: ts.FunctionDeclaration) {
        this.handleCallSignature(node);
        super.visitFunctionDeclaration(node);
    }

    public visitGetAccessor(node: ts.AccessorDeclaration) {
        this.handleCallSignature(node);
        super.visitGetAccessor(node);
    }

    public visitMethodDeclaration(node: ts.MethodDeclaration) {
        this.handleCallSignature(node);
        super.visitMethodDeclaration(node);
    }

    // ... 2 more identical functions
```

but I can't think of anything better given the current architecture...

adidahiya:
rebased & squashed

PR #275, opened 2015-01-29
Migrate typedef whitespace rule to 1.4

adidahiya:


src/rules/typedefWhitespaceRule.ts
```diff
@@ -14,129 +14,102 @@
...
+    public visitMethodDeclaration(node: ts.MethodDeclaration) {
+        this.checkSpace("call-signature", node, node.type, node.parameters.end + 1);
+        super.visitMethodDeclaration(node);
     }
 
-    public visitPropertySignature(node: TypeScript.PropertySignatureSyntax): void {
-        this.checkSpace("property-signature", node, node.typeAnnotation);
+    public visitParameterDeclaration(node: ts.ParameterDeclaration) {
+        // a parameter's "type" could be a string literal, for example `fn(option: "someOption", anotherOption: number)`
+        if (node.type != null && node.type.kind !== ts.SyntaxKind.StringLiteral) {
```
gscshoyru: You should probably check for equality rather than checking it isn't the other one here, it's cleaner. Also, your comment above checkSpace below implies it can handle string literals here -- why are you doing this, if that's the case?

src/rules/typedefWhitespaceRule.ts
```diff
@@ -14,129 +14,102 @@
...
+        return options == null || options[option] != null;
     }
 
-    private getOption(option: string): string {
+    private getOption(option: string) {
         var allOptions = this.getOptions();
-        if (!allOptions || allOptions.length === 0) {
-            return undefined;
+        if (allOptions == null || allOptions.length === 0) {
+            return;
```
gscshoyru: Why this change, and the change of the return type? It seems cleaner to have it return null and be typed as string, than to not do this.

adidahiya: I thought we prefer implicit return types? And I could change this to `return null;` instead of `return;` – the main reasoning here is that you never _really_ need to use the `undefined` keyword in TS, and if we can completely remove the use of a particular concept, the language becomes simpler to use.

src/rules/typedefWhitespaceRule.ts
adidahiya: good point; I allowed `StringLiteral` to go through when I did `IndexSignatureDeclaration` and forgot that I filtered it out here

src/rules/typedefWhitespaceRule.ts
gscshoyru: We do prefer those -- but with the return statement as it is, I'm pretty sure the implicit return type isn't string (though correct me if I'm wrong). Return null is better and probably the way it should have been -- I agree with you on the undefined. If the implicit works with the return null, then you don't need to make it explicit.

adidahiya: i'll make it return null. it works with the implicit return type.

adidahiya:
@gscshoyru updated

PR #283, opened 2015-02-02
Catch errors for files that don't compile

gscshoyru:


src/tslint.ts
```diff
@@ -63,24 +63,32 @@ module Lint {
...
+                                failures.push(ruleFailure);
+                            }
+                        });
+                    }
                 }
+            } catch (e) {
+                return {
+                    failureCount: 1,
+                    format: "error",
+                    output: "This file did not successfully compile; please fix compilation errors and re-run tslint"
```
ashwinr: this is incorrect. this should not be a string because this will throw an exception for people externally parsing the output via the JSON formatter or some other custom format. instead i think you should just create a single RuleFailure with an empty rule name and at position 0 with the error message "unable to compile file" and pass that into the formatter.

src/tslint.ts
```diff
@@ -63,24 +63,29 @@ module Lint {
...
+                        ruleFailures.forEach((ruleFailure) => {
+                            if (!this.containsRule(failures, ruleFailure)) {
+                                failures.push(ruleFailure);
+                            }
+                        });
+                    }
                 }
+            } catch (e) {
+                failures = [new RuleFailure(sourceFile, 0, 1,
+                    "This file did not successfully compile; please fix compilation errors and re-run tslint", "compilation")];
```
ashwinr: let's make the error message simpler -- just "unable to compile" should be enough.

PR #285, opened 2015-02-02
Remove check for compilation issues, fix infinite loop bug

gscshoyru:


src/language/utils.ts
```diff
@@ -39,7 +39,9 @@ module Lint {
             getNewLine: () => "\n"
         };
 
-        return ts.createProgram([normalizedName], compilerOptions, compilerHost).getSourceFile(normalizedName);
+        var program = ts.createProgram([normalizedName], compilerOptions, compilerHost);
+        program.getTypeChecker(true);
```
ashwinr: comment please

PR #299, opened 2015-02-07
Comments on the beginning of the file are invalid

mwe:
/**
     * jsdoc at start of file
     */ 

Above command is invalid according jsdocFormatRule.ts. 
Pull request for the change i made to fix it. It is probably a HACK but i am not able to find the cause.

### DIFF
--- a/src/rules/jsdocFormatRule.ts
+++ b/src/rules/jsdocFormatRule.ts
@@ -53,8 +53,8 @@ class JsdocWalker extends Lint.RuleWalker {
                 return;
             }
 
-            // Correct position for comments on the start of the file
-            if (currentPosition == 1) {
+            // Fix position for comments on the start of the file
+            if (currentPosition === 1) {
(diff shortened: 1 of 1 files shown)


gscshoyru:
I'm not able to reproduce this bug here -- putting the comment you made at the top of a file, or even in a file on its own does not seem to cause this error. I'm not sure if I'm missing something here or not -- can you give me the text of an example file that creates this bug, and which error it is you're getting?

ashwinr:
can't reproduce it either. @mwe closing this PR, please reopen with repro steps.

mwe:
After looking further, it seems to be related to UTF8 files. 

The file:

    /**
     * @file filename
     */ 

will be validated as // tslint : (jsdoc-format) client/tslinterror.ts[2, 1]: asterisks in jsdoc must be aligned.

    /**
      * @file filename
      */ 
is validated as correct.

Issue #292, opened 2015-02-05
tslint 2.1.0 has regression when using variables-before-functions member ordering rule


ericvw:
Given the following `tslint.json`:

```json
{
    "rules": {
        "member-ordering": [true,
            "variables-before-functions"
        ]
    }
}
```

and `test.ts`:

```typescript
class Foo {
    private bar(): any {
        var bla: { a: string } = {a: '1'};
    }
}
```

`tslint 1.2.0` reports nothing
`tslint 2.1.0` reports: `test.ts[3, 20]: Declaration of public instance member variable not allowed to appear after declaration of private instance member function`

Using `git-bisect` indicates that b26988695f263cda8f518178ae074a7c4d075d9d introduced the regression, but I haven't fully understood why yet.

ericvw:
The key part here that the internal variable is an object literal where `a` is following `bar` that `tslint` is complaining about.

ashwinr:
@gscshoyru can you look into this? thanks.

ericvw:
Thanks!  When may expect a next release that will include this fix?

ashwinr:
we're still fixing a couple of other reported bugs, so tentatively it'll be sometime early next week

PR #300, opened 2015-02-08
fix member ordering regression w.r.t. interfaces

ghatchue:
This addresses a regression introduced by same commit (b269886) that caused #292. Basically, the member ordering rule needs to be aware of interface definitions. Currently, a property declaration within an interface triggers an error if a previously defined class contains methods.


ghatchue:
Also fixed by #302, which is a slightly better fix. Closing this.

PR #301, opened 2015-02-08
bug fix: don't issue typedef variable-declaration warning with for-in statements

ghatchue:
This fix prevents tslint from issuing the typedef variable-declaration warning when declaring an iteration variable within for-in statements. 
e.g
```typescript
for (var p in [1, 2]) {
// "p" cannot have a type annotation.
}
```
Type annotations cannot be used for variable declarations within a for-in statement.


adidahiya:
lgtm @ashwinr

Issue #293, opened 2015-02-06
Error while checking comments style within a string interpolation


SergioMorchon:
If I write this line:
``` typescript
`${location.protocol}//${location.hostname}`
```
tslint writes: comment must start with a space
But it is a part of the new TypeScript 1.4 feature: the ES6 template strings, and not a `//` comment prefix.

PR #302, opened 2015-02-08
walk interface and reset previous state for member-ordering rules

ericvw:
This change introduces resetting the previous `IModifiers` state when
visiting an interface declaration.  Without doing so, subsequent
interface declarations would be conflated as if they were the same
interface when using the `variables-before-functions` rule option.

Version `1.2.0` did not have this problem and `2.1.0` introduced this
regression.  The `memberordering-method.test.ts` has been updated
appropriately to test for this regression and an additional test has
been provided to ensure that interface declarations should trigger
similarly to class declarations.

### DIFF
--- a/src/rules/memberOrderingRule.ts
+++ b/src/rules/memberOrderingRule.ts
@@ -62,12 +62,13 @@ export class MemberOrderingWalker extends Lint.RuleWalker {
     }
 
     public visitClassDeclaration(node: ts.ClassDeclaration): void {
-        this.previous = {
-            isMethod: false,
-            isPrivate: false,
-            isInstance: false
-        };
+        this.resetPreviousModifiers();
         super.visitClassDeclaration(node);
+    }
+
+    public visitInterfaceDeclaration(node: ts.InterfaceDeclaration): void {
+        this.resetPreviousModifiers();
+        super.visitInterfaceDeclaration(node);
     }
 
     public visitMethodDeclaration(node: ts.MethodDeclaration): void {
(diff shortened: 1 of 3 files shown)


ericvw:
#300 also addresses this similarly as well I just found out.  We must have been working on this in parallel :).  The addition to the test case that is provided in this PR also touches on the interface conflation issue, which is what originally motivated me to investigate this further.

ghatchue:
@ericvw Yeah, seems like we were working on this in parallel. I closed my pull request, your fix is slightly better. Less code duplication when resetting the state.

src/rules/memberOrderingRule.ts
```diff
@@ -57,19 +57,28 @@ function toString(modifiers: IModifiers): string {
 export class MemberOrderingWalker extends Lint.RuleWalker {
     private previous: IModifiers;
 
-    constructor(sourceFile: ts.SourceFile, options: Lint.IOptions) {
-        super(sourceFile, options);
-    }
-
-    public visitClassDeclaration(node: ts.ClassDeclaration): void {
+    private resetPrevious(): void {
```
ashwinr: please move this method towards the end of the class (after all public methods). also i think "resetModifiers" or "resetPreviousModifiers" is a better name.

ashwinr: otherwise, looks good! i can merge after you update the PR

ericvw:
Updated PR based on feedback regarding placement of private method and renamed as suggested.

ashwinr:
thanks!

ericvw:
I know I asked this in #292, but is there any updates on an ETA for the next `npm` release?

ashwinr:
we're looking into a fix for https://github.com/palantir/tslint/issues/293, but if that doesn't seem likely by end-of-day tomorrow, i can make a release anyway.

ericvw:
Awesome - thanks for the update.

ashwinr:
released

ericvw:
Thanks!

PR #303, opened 2015-02-09
Fixes #295

### LINKED ISSUE #295: Complains about duplicate variable even though I use let instead of var
The following code generates a `duplicate variable: something` warning even though the code is perfectly fine.

```javascript
if (condition) {
    let something = 1;
} else {
    let something = 2;
}
```

gscshoyru:
Now differentiates between lets and vars

### DIFF
--- a/src/rules/noDuplicateVariableRule.ts
+++ b/src/rules/noDuplicateVariableRule.ts
@@ -35,16 +35,18 @@ class NoDuplicateVariableWalker extends Lint.ScopeAwareRuleWalker<ScopeInfo> {
         var declarationIsLet = (Math.floor(node.flags / ts.NodeFlags.Let) % 2) === 1;
 
         var failureString = Rule.FAILURE_STRING + variableName + "'";
-        if (currentScope.varVariableNames.indexOf(variableName) >= 0) {
+        if (currentScope.varNames.indexOf(variableName) >= 0) {
+            // if there was a previous var declaration with the same name, this declaration is invalid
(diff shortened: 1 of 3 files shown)


src/rules/noDuplicateVariableRule.ts
```diff
@@ -44,5 +52,6 @@ class NoDuplicateVariableWalker extends Lint.ScopeAwareRuleWalker<ScopeInfo> {
 }
 
 class ScopeInfo {
-    public variableNames: string[] = [];
+    public varVariableNames: string[] = [];
```
ashwinr: let's call them letNames and varNames. variable seems unnecessarily long here.

src/rules/noDuplicateVariableRule.ts
```diff
@@ -31,12 +31,20 @@ class NoDuplicateVariableWalker extends Lint.ScopeAwareRuleWalker<ScopeInfo> {
         var propertyName = node.name;
         var variableName = propertyName.text;
         var currentScope = this.getCurrentScope();
+        // determine if the appropriate bit is set, which indicates this is a "let"
+        var declarationIsLet = (Math.floor(node.flags / ts.NodeFlags.Let) % 2) === 1;
 
-        if (currentScope.variableNames.indexOf(variableName) >= 0) {
-            var failureString = Rule.FAILURE_STRING + variableName + "'";
+        var failureString = Rule.FAILURE_STRING + variableName + "'";
+        if (currentScope.varVariableNames.indexOf(variableName) >= 0) {
```
ashwinr: comment this block

PR #311, opened 2015-02-10
Fixes #293

### LINKED ISSUE #293: Error while checking comments style within a string interpolation

gscshoyru:
Comment format rule should skip templates (and other things). And the
whitespace rule should skip templates too.

### DIFF
--- a/src/rules/commentFormatRule.ts
+++ b/src/rules/commentFormatRule.ts
@@ -44,6 +44,7 @@ class CommentWalker extends Lint.RuleWalker {
                 scanner.setTextPos(this.tokensToSkipStartEndMap[startPos]);
                 return;
             }
+
             if (scanner.getToken() === ts.SyntaxKind.SingleLineCommentTrivia) {
                 var commentText = scanner.getTokenText();
                 var startPosition = scanner.getTokenPos() + 2;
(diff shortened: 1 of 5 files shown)


src/rules/commentFormatRule.ts
```diff
@@ -27,8 +27,23 @@ export class Rule extends Lint.Rules.AbstractRule {
...
     public visitSourceFile(node: ts.SourceFile): void {
+        super.visitSourceFile(node);
         Lint.scanAllTokens(ts.createScanner(ts.ScriptTarget.ES5, false, node.text), (scanner: ts.Scanner) => {
+            var startPos = scanner.getStartPos();
+            if (this.tokensToSkipStartEndMap[startPos] != null) {
+                // tokens to skip are places where the scanner gets confused about what the token is, without the proper context
+                // (specifically, regex, identifiers, and templates). So skip those tokens.
+                scanner.setTextPos(this.tokensToSkipStartEndMap[startPos]);
+                return;
+            }
```
adidahiya: why not `else if`?

gscshoyru: There should probably be a newline between the two if blocks here -- but it shouldn't be an else if, the two things are different concepts (the first block is saying "if this is true, continue", which doesn't make sense to group with the next block.)

adidahiya: ok, fair :+1:

ashwinr: Yes new line please

src/rules/commentFormatRule.ts
```diff
@@ -47,7 +62,30 @@ class CommentWalker extends Lint.RuleWalker {
...
+
+    public visitIdentifier(node: ts.Identifier) {
+        if (node.getStart() < node.getEnd()) {
+            // only add to the map nodes whose end comes after their start, to prevent infinite loops
+            this.tokensToSkipStartEndMap[node.getStart()] = node.getEnd();
+        }
+        super.visitIdentifier(node);
+    }
+
+    public visitTemplateExpression(node: ts.TemplateExpression) {
```
ashwinr: There's a lot of repetition in these 3 methods. Split out this block into a private addNode method and call it from all these methods

PR #323, opened 2015-02-16
Skip checking indentation inside multi-line comments

Nemo157:
Eventually a full implementation of [smart-tabs](http://www.emacswiki.org/emacs/SmartTabs) would be nice, but the only case I've seen failing so far is using an initial space to align the asterisks in a multi-line comment. This allows that use-case (along with other multi-line comment formats) by simply turning off indentation checking inside multi-line comments.

[Review on Reviewable](https://reviewable.io/reviews/palantir/tslint/323)

### DIFF
--- a/src/rules/indentRule.ts
+++ b/src/rules/indentRule.ts
@@ -49,8 +49,14 @@ class IndentWalker extends Lint.RuleWalker {
             return;
         }
 
+        let endOfComment = -1;
         const scanner = ts.createScanner(ts.ScriptTarget.ES5, false, node.text);
         for (let lineStart of node.getLineStarts()) {
+            if (lineStart < endOfComment) {
+                // skip checking lines inside multi-line comments
+                continue;
+            }
+
             scanner.setTextPos(lineStart);
 
             let currentScannedType = scanner.scan();
@@ -66,6 +72,11 @@ class IndentWalker extends Lint.RuleWalker {
 
                 fullLeadingWhitespace += scanner.getTokenText();
                 currentScannedType = scanner.scan();
+            }
+
+            const commentRanges = ts.getTrailingCommentRanges(node.text, lineStart);
+            if (commentRanges) {
+                endOfComment = commentRanges[commentRanges.length - 1].end;
             }
 
             if (currentScannedType === ts.SyntaxKind.SingleLineCommentTrivia
--- a/test/files/rules/indentwith_tabs.test.ts
+++ b/test/files/rules/indentwith_tabs.test.ts
@@ -1,5 +1,8 @@
 // valid code
 module TestModule {
+	/**
+	 * Some documentation
+	 */
 	var func = () => {
 		console.warn("hi");
 	};
--- a/test/rules/indentRuleTests.ts
+++ b/test/rules/indentRuleTests.ts
@@ -37,45 +37,45 @@ describe("<indent>", () => {
         });
 
         it("enforces variable indentation", () => {
-            expectFailure(Lint.Test.createFailure(fileName, [58, 1], [58, 5], failureStringTabs));
-            expectFailure(Lint.Test.createFailure(fileName, [62, 1], [62, 9], failureStringTabs));
-            expectFailure(Lint.Test.createFailure(fileName, [66, 1], [66, 5], failureStringTabs));
+            expectFailure(Lint.Test.createFailure(fileName, [61, 1], [61, 5], failureStringTabs));
+            expectFailure(Lint.Test.createFailure(fileName, [65, 1], [65, 9], failureStringTabs));
+            expectFailure(Lint.Test.createFailure(fileName, [69, 1], [69, 5], failureStringTabs));
         });
 
         it("enforces class method indentation", () => {
-            expectFailure(Lint.Test.createFailure(fileName, [68, 1], [68, 5], failureStringTabs));
-            expectFailure(Lint.Test.createFailure(fileName, [69, 1], [69, 8], failureStringTabs));
-            expectFailure(Lint.Test.createFailure(fileName, [70, 1], [70, 5], failureStringTabs));
+            expectFailure(Lint.Test.createFailure(fileName, [71, 1], [71, 5], failureStringTabs));
+            expectFailure(Lint.Test.createFailure(fileName, [72, 1], [72, 8], failureStringTabs));
+            expectFailure(Lint.Test.createFailure(fileName, [73, 1], [73, 5], failureStringTabs));
         });
 
         it("enforces object literal indentation", () => {
-            expectFailure(Lint.Test.createFailure(fileName, [74, 1], [74, 5], failureStringTabs));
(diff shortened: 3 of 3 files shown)


ashwinr:
@gscshoyru is looking at a better way to handle template strings, we'll look into this one

ashwinr:
@gscshoyru isn't this fixed in master now?

ashwinr:
ping @gscshoyru?

gscshoyru:
Apologies, @Nemo157 , I thought this was covered by a different thing that I was working on, but it seems to be something separate. I'll go comment on it now.

test/helper.ts
```diff
@@ -80,4 +80,21 @@ module Lint.Test {
             assert(false, "expected " + stringifiedNeedle + " within " + stringifiedHaystack);
         }
     }
+
+    // assert that a failure array does not contain any error on a set of lines
+    export function assertNoFailuresWithin(fileName: string, failures: Lint.RuleFailure[], startLine: number, endLine: number) {
```
gscshoyru: The way we've actually been doing this is assert.lengthOf with actual failiures and however many failures there should be -- so that's the test you should add, rather than adding this code here.

test/rules/indentRuleTests.ts
```diff
@@ -72,6 +76,10 @@ describe("<indent>", () => {
             actualFailures = Lint.Test.applyRuleOnFile(fileName, IndentRule, [true, "spaces"]);
         });
 
+        it("doesn't fail good code", () => {
+            Lint.Test.assertNoFailuresWithin(fileName, actualFailures, 1, 55);
```
gscshoyru: Do assert.lengthOf(actualFailures, N), here, please.

src/rules/indentRule.ts
```diff
@@ -50,7 +50,13 @@ class IndentWalker extends Lint.RuleWalker {
         }
         var scanner = ts.createScanner(ts.ScriptTarget.ES5, false, node.text);
         var lineStarts = node.getLineStarts();
+        var endOfComment = -1;
```
adidahiya: This branch is behind `master`, you'll need to rebase to pick up some new linting rules. For example, we've turned on the `no-var-keyword` rule, so this should be a `const.

src/rules/indentRule.ts
```diff
@@ -50,7 +50,13 @@ class IndentWalker extends Lint.RuleWalker {
...
         lineStarts.forEach((lineStart) => {
+            if (lineStart < endOfComment) {
+                // Skip checking lines inside multi-line comments
+                return;
```
adidahiya: this loop has changed to a `for-of` loop in master, so this line will have to change to be `continue;`

src/rules/indentRule.ts
```diff
@@ -50,7 +50,13 @@ class IndentWalker extends Lint.RuleWalker {
         }
         var scanner = ts.createScanner(ts.ScriptTarget.ES5, false, node.text);
         var lineStarts = node.getLineStarts();
+        var endOfComment = -1;
         lineStarts.forEach((lineStart) => {
+            if (lineStart < endOfComment) {
+                // Skip checking lines inside multi-line comments
```
adidahiya: single line comments should start with a lowercase char

ashwinr: shouldn't TSLint have complained here since we turned on the comment-format rule?

adidahiya: the build is old; also I believe Github does pre-merge build verification

Nemo157:
Cool, I'll rebase and fix up the issues sometime this week. Thanks for taking a look.

test/files/rules/indentwith_tabs.test.ts
```diff
@@ -1,5 +1,8 @@
 // valid code
 module TestModule {
+	/**
+	 * Some documentation
+	 */
```
adidahiya: I think we need more tests here. Here's an example where lines 1 and 3 use tabs:

```ts
–⇥/**
   * More docs
–⇥ */
```

This should fail because of spaces used for indentation on line 2, right?

If possible, this PR should address #485 as well.

Nemo157: Not with the current implementation, this just disables all indentation checking between `/*` and `*/`. This will allow #485 I believe, the test case rendered with `·` for spaces and `⇥` for tab is:

```
⇥/**
⇥·*·Some·documentation
⇥·*/
```

I would prefer to allow this via [smart-tabs](http://www.emacswiki.org/emacs/SmartTabs), but that would be a lot more work to implement.

test/rules/indentRuleTests.ts
```diff
@@ -83,6 +87,10 @@ describe("<indent>", () => {
             actualFailures = Lint.Test.applyRuleOnFile(fileName, IndentRule, [true, "spaces"]);
         });
 
+        it("doesn't fail good code", () => {
+            assert.lengthOf(actualFailures, 21);
+        });
```
adidahiya: this test is a duplicate, please remove

test/files/rules/indentwith_tabs.test.ts
adidahiya: @gscshoyru thoughts on my example here? Maybe we should continue checking indentation within multiline comments, and only allow a single extra space before the `*`?

test/rules/indentRuleTests.ts
Nemo157: Where is the duplicate? L35-38 is for a different test file.

adidahiya: right you are. my mistake.

test/files/rules/indentwith_tabs.test.ts
gscshoyru: This seems to overlap somewhat with the jsdoc rule -- I'm not sure that it makes sense for the indent rule to be aware of jsdoc comments vs other kinds -- and jsdocs aren't the only kind of multiline comments. I think it's likely that simply ignoring indentation inside multiline comments makes the most sense, and is most correct.

adidahiya: No, I don't mean for it to be aware of jsdoc comments. I'm suggesting that we continue checking indentation in _all_ multiline comments and only allow a single space before the `*` for alignment purposes if the indent rule is configured to use tabs. This kind of alignment applies to all multiline comments, I think.

gscshoyru: Right -- but I'm not sure it does apply to all multiline comments. Not all multiline comments have asterisks each line. If you don't have asterisks and still want to align with text that you put on the very first line, for instance. I'm not sure that text-inside-multiline-comments should be considered for the same kind of indentation rules.

adidahiya: Ok, that makes sense. I'll go ahead and merge this.

Nemo157:
@adidahiya @gscshoyru Thanks :smile:

PR #334, opened 2015-02-23
Do not check return type for set accessors since they don't have one.

thmiceli:
This change prevents tslint from reporting a defect for the absence of a return type in a set accessor signature.


thmiceli:
By the way the typescript compiler will report an error if a return type is specified for a set accessor.

adidahiya:
good catch. lgtm :+1: @ashwinr

PR #335, opened 2015-02-24
Oneline rule: allow exception for multiline expression.

thmiceli:
This is a change to the one-line rule that fits the coding style standard at my work place.
I don't think it is an uncommon pattern since it is the second company where I have seen this. So, here it is in case you want to integrate it in tslint in some way.

For an 'if', 'while' or 'for' statement when the expression spans over multiple lines it may be difficult to see where the expression ends. For example with an 'if' statement and a 4 space identation the expression and the 'then' statements are aligned. Example:

if (cond1 &&
    cond2 &&
    cond3 &&) {
    blah;
}

In this situation it is desirable to tolerate the placement of the opening brace on a different line aligned with the start of the statement:

if (cond1 &&
    cond2 &&
    cond3 &&) 
{
    foo();
}

[Review on Reviewable](https://reviewable.io/reviews/palantir/tslint/335)

### DIFF
--- a/src/rules/oneLineRule.ts
+++ b/src/rules/oneLineRule.ts
@@ -38,7 +38,11 @@ class OneLineWalker extends Lint.RuleWalker {
         if (thenStatement.kind === ts.SyntaxKind.Block) {
             var expressionCloseParen = node.getChildAt(3);
             var thenOpeningBrace = thenStatement.getChildAt(0);
-            this.handleOpeningBrace(expressionCloseParen, thenOpeningBrace);
+            if (!this.multilineAndAlignedBrace(node,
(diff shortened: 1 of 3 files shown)


thmiceli:
Here is my example again with the intended vertical alignment:

if (cond1 &&
....cond2 &&
....cond3 &&) {
....blah;
}

And the more readable form:

if (cond1 &&
....cond2 &&
....cond3 &&) 
{
....blah;
}

ashwinr:
sorry, i think this is unnecessary for the core. i'm abandoning it -- if you feel strongly about this rule, please file an issue first.

PR #337, opened 2015-02-26
Added rule to enforce vertical alignment

thmiceli:
This is a new rule that checks for vertical alignment for one of the three options: "parameters", "arguments" and "statements".

### DIFF
--- a/README.md
+++ b/README.md
@@ -106,6 +106,10 @@ Supported Rules
 
 A sample configuration file with all options is available [here](https://github.com/palantir/tslint/blob/master/docs/sample.tslint.json).
 
+* `align` enforces vertical alignment. Rule options:
+  * `"parameters"` checks alignment of function parameters.
+  * `"arguments"` checks alignment of function call arguments.
+  * `"statements"` checks alignment of statements.
 * `ban` bans the use of specific functions. Options are ["object", "function"] pairs that ban the use of object.function()
 * `class-name` enforces PascalCased class and interface names.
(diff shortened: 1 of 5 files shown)


src/rules/alignRule.ts
```diff
@@ -0,0 +1,86 @@
...
+        this.checkArguments(node.arguments);
+        super.visitCallExpression(node);
+    }
+
+    public visitBlock(node: ts.Block) {
+        this.checkStatements(node.statements);
+        super.visitBlock(node);
+    }
+
+    private checkParameters(parameters: ts.Node[]) {
```
gscshoyru: I don't think you need these three private methods here -- you can probably just inline "this.checkAlignment("parameters", node.parameters)", etc, in all the above functions.

README.md
```diff
@@ -106,6 +106,10 @@ Supported Rules
 
 A sample configuration file with all options is available [here](https://github.com/palantir/tslint/blob/master/docs/sample.tslint.json).
 
+* `align` enforces vertical alignment. Rule options:
```
gscshoyru: Thanks for updating the README!
Please update the sample.tslint.json (in docs) with this new rule and its options as well.

src/rules/alignRule.ts
thmiceli: Sure, I refactored the code a little and removed the functions.

README.md
thmiceli: Done.

adidahiya:
:+1: lgtm

PR #338, opened 2015-02-26
Added rule for enforcing 'default' case in 'switch' statement.

thmiceli:
This new rule checks that a 'default' case is present in a 'switch'.
I made it a new rule but it could be an option of a "switch" rule with "no-switch-case-fall-through" made also an option of "switch".

### DIFF
--- a/src/rules/switchDefaultRule.ts
+++ b/src/rules/switchDefaultRule.ts
@@ -25,13 +25,8 @@ export class Rule extends Lint.Rules.AbstractRule {
 export class SwitchDefaultWalker extends Lint.RuleWalker {
 
     public visitSwitchStatement(node: ts.SwitchStatement) {
-        var hasDefaultCase = false;
-        var switchClauses = node.clauses;
-        switchClauses.forEach((child, i) => {
-            if (child.kind === ts.SyntaxKind.DefaultClause) {
-                hasDefaultCase = true;
-            }
-        });
+        var hasDefaultCase = node.clauses.some((clause) =>
+                                               clause.kind === ts.SyntaxKind.DefaultClause);
         if (!hasDefaultCase) {
             this.addFailure(this.createFailure(node.getStart(),
                                                node.getWidth(),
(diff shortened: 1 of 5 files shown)


src/rules/switchDefaultRule.ts
```diff
@@ -0,0 +1,42 @@
...
+export class SwitchDefaultWalker extends Lint.RuleWalker {
+
+    public visitSwitchStatement(node: ts.SwitchStatement) {
+        var hasDefaultCase = false;
+        var switchClauses = node.clauses;
+        switchClauses.forEach((child, i) => {
+            if (child.kind === ts.SyntaxKind.DefaultClause) {
+                hasDefaultCase = true;
+            }
+        });
```
adidahiya: This can be simpler:
`var hasDefaultCase = switchClauses.some((clause) => clause.kind === ts.SyntaxKind.DefaultClause);`

thmiceli: Yes that is much nicer. I updated my code.

src/rules/switchDefaultRule.ts
```diff
@@ -0,0 +1,37 @@
+/*
+ * Copyright 2013 Palantir Technologies, Inc.
```
ashwinr: please change the year as well :)

src/rules/switchDefaultRule.ts
```diff
@@ -0,0 +1,37 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+*/
+
+export class Rule extends Lint.Rules.AbstractRule {
+    public static FAILURE_STRING = "Switch statement doesn't include a 'default' case";
```
ashwinr: error messages start with a lower case conventionally

ashwinr:
also modify the README to add this rule please

thmiceli:
I believe I addressed the feedback. I updated the PR.

ashwinr:
cool thanks!

Issue #349, opened 2015-03-25
ES6 template strings have missing whitespace error


stevenfan:
```
 `
<sl-property-group label=${uppercaseName} Axes">
  <div class="repeat"
`;
```

"Missing whitespace" error on the first double quote (")

Issue #332, opened 2015-02-23
Templated strings


SergioMorchon:
Hi.

When linting this function:
``` typescript
function makeHeader(pkg: Package) {
	return [
		`\t/**`,
		`\t * ${pkg.name} v${pkg.version}`,
		`\t */`
	].join("\r\n");
}
```

i got the message ```jsdoc is not formatted correctly on this line``` in 2nd (``` `\t * ${pkg.name} v${pkg.version}`,```) and 3th (``` `\t */` ```) lines that are missinterpreted as comments.

arnaudpourbaix:
It is the same issue I reported yesterday: see https://github.com/palantir/tslint/issues/331
It is very annoying because I like template strings but linter will report many issues and even VS doesn't properly display template strings content (especially when I use " or ' inside).

adidahiya:
@leeavital, can you please add a regression test case for this in the jsdoc rule?

PR #362, opened 2015-04-13
Fix things that walk tokens to properly handle templates, etc.

### LINKED ISSUE #357: Backtick in regexp causes following comment rule to be ignored
If a regular expression literal containing a backtick character preceeds a comment rule then that rule is ignored, for example:

```
var  re;
re = /`/;   // Backtick in regexp causes following 'disable:no-unused-variable' rule to be ignoredj.

/* tslint:disable:no-unused-variable */
var s = 'xxx';
```

Running tslint on the above file generates _unused variable_ error:

```
$ tslint --version
2.1.1

$ tslint -f t.ts
t.ts[5, 5]: unused variable: 's'
```

gscshoyru:
Fixes #357, #349, #332, and possibly a couple others I missed

### DIFF
--- a/lib/tslint.d.ts
+++ b/lib/tslint.d.ts
@@ -164,7 +164,19 @@ declare module Lint {
     function findFormatter(name: string, formattersDirectory?: string): any;
 }
 declare module Lint {
-    class EnableDisableRulesWalker extends Lint.RuleWalker {
+    class SkipableTokenAwareRuleWalker extends RuleWalker {
+        protected tokensToSkipStartEndMap: {
+            [start: number]: number;
+        };
+        constructor(sourceFile: ts.SourceFile, options: Lint.IOptions);
+        protected visitRegularExpressionLiteral(node: ts.Node): void;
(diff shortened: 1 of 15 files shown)


ashwinr:
failing builds... :(

gscshoyru:
Yup, forgot to actually include the new walker. Added it and repushed, the build is running again.

lib/tslint.d.ts
```diff
@@ -164,7 +164,19 @@ declare module Lint {
     function findFormatter(name: string, formattersDirectory?: string): any;
 }
 declare module Lint {
-    class EnableDisableRulesWalker extends Lint.RuleWalker {
+    class SkipableTokenAwareRuleWalker extends RuleWalker {
```
adidahiya: let's spell it `SkippableTokenAwareRuleWalker`

ashwinr: +1

gscshoyru: Ok, renamed and pushed.

ashwinr:
is the build taking much longer than usual?

PR #366, opened 2015-04-20
No duplicate variable rule no longer walks type literals

gscshoyru:


adidahiya:
:+1: lgtm

test/files/rules/duplicate-variable.test.ts
```diff
@@ -103,3 +103,6 @@ function testArguments2(x: number, y: number): void {
     var x = 1;
     let y = 2;
 }
+
+var references: {[vertex: string]: any};
```
ashwinr: also add a test for parameters:

```
function blah(arg1: {[key: string]: any, arg2: {[key:string]: any}) {
}
```

gscshoyru: Done.

ashwinr:
this is not fully fixed. here's an exception that's still causing false positives:

```
    export interface IClipboard {
        copy(key: string, state: any): void;
        paste(key: string): any;
    }
```

duplicate "key"

ashwinr:
another one:

`findMaxOrMin(values: any[], defaultValue: number, operation: (...values: any[]) => number)`

duplicate "values"

ashwinr:
one more:

```
try {
    //
} catch (e) {
    e.blah();
    //
}

try {
    //
} catch (e) {
    e.blah();
    //
}
```

duplicate "e"

adidahiya:
made a new issue for these false positives: #372

Issue #321, opened 2015-02-11
no-duplicate-variable allows duplicate let declarations in the same block scope


wereHamster:
The following is accepted even though it should generate a warning.

```javascript
function test() {
    let x = 1;
    let x = 2;
}
``` 

Related issues: #295, #303.

adidahiya:
Compiling that particular example via `tsc` throws an error. We hope to avoid duplicating the work of the compiler in TSLint, so I don't think we're going to directly target that example.

PR #374, opened 2015-04-21
#321 no-duplicate-let rule

adidahiya:


ashwinr:
hm, the CI looks broken..

adidahiya:
travis is having downtime: http://www.traviscistatus.com/

does the code here look good?

gscshoyru:
I'm not sure we want to split this out into another rule -- can't we make a walker that's aware of both scopes and blocks, and works accordingly? 

Also, this should probably wait until CI is working again. And wait until after this is working on our code with 1.5, I think, before merging this. Though I could be wrong.

Gruntfile.js
```diff
@@ -35,18 +36,18 @@ module.exports = function(grunt) {
...
       options: {
-        configuration: grunt.file.readJSON("tslint.json")
+        configuration: grunt.file.readJSON('tslint.json')
       },
       src: [
-        "src/*.ts",
-        "src/formatters/**/*.ts",
-        "src/language/**/*.ts",
-        "src/rules/**/*.ts"
+        'src/*.ts',
```
gscshoyru: Why did all these change?

src/rules/noDuplicateLetRule.ts
```diff
@@ -0,0 +1,56 @@
...
+        return this.applyWithWalker(new NoDuplicateLetWalker(sourceFile, this.getOptions()));
+    }
+}
+
+class NoDuplicateLetWalker extends Lint.BlockScopeAwareRuleWalker<ScopeInfo> {
+    public createScope(): ScopeInfo {
+        return new ScopeInfo();
+    }
+
+    public visitVariableDeclaration(node: ts.VariableDeclaration): void {
```
gscshoyru: This is going to hit the same issue that the noDuplicateVariableRule hit, which is that the "variable declaration" in catch blocks seems to have its own special rules. You should probably try to skip that catch-block variable declaration.

Gruntfile.js
adidahiya: single quotes

test/files/rules/no-duplicate-variable.test.ts
```diff
@@ -107,7 +107,7 @@ function testArguments2(x: number, y: number): void {
 var references: {[vertex: string]: any};
 var dependents: {[vertex: string]: any};
 
-function blah(arg1: {[key: string]: any, arg2: {[key:string]: any}) {
+function blah(arg1: {[key: string]: any}, arg2: {[key:string]: any}) {
```
gscshoyru: Oops. Thank you.

Gruntfile.js
gscshoyru: Ah, makes sense, +1

adidahiya:
Do you mean make `ScopeAwareRuleWalker` aware of block scopes as well? So it'll have additional methods like `getCurrentBlockScope()`, `createBlockScope()`, etc. that mirror the existing methods for regular scopes? Feels a little messy but I could be convinced to go that way...

re: CI stuf -- sure, we could wait until Travis is working, but we at least know the appveyor build is passing. Also I'm not sure what you mean by that last part about waiting until this is working with our 1.5 code...

gscshoyru:
Sorry if I wasn't more clear -- we're planning to release a build soon that (should) work with existing 1.4 code and 1.5 typescript -- and I feel like we shouldn't be adding new functionality, just fixing broken things, until that release happens. (Which was going to be soon but looks like we need to wait on 1.5.0-beta for reasons). We could possibly merge to a feature branch instead, though. 

As for scope-aware stuff -- I actually meant making a new walking on top of scope-aware that is also aware of block scopes -- but integrating it into the existing walker also makes sense. I agree that I'm not certain if that's the best way, though -- I just think it might be confusing to have a "no duplicate variable" rule that people may expect to work on lets, but won't.

adidahiya:
Ok, I can merge this into a feature branch instead.

I think I see what you mean now -- I should make `NoDuplicateVariableWalker` extend `BlockScopeAwareRuleWalker<T>`, which will have `getCurrentBlockScope()` in addition to `getCurrentScope()`, etc. so that we can interact with regular scopes as well as block scopes when checking the `no-duplicate-variable` rule.

adidahiya:
abandoning in favor of #375 (PR against 1.5 branch instead of master)

PR #375, opened 2015-04-21
Create no-shadowed-variable rule; support let and const syntax

adidahiya:


src/rules/noDuplicateVariableRule.ts
```diff
@@ -60,11 +64,12 @@ class NoDuplicateVariableWalker extends Lint.ScopeAwareRuleWalker<ScopeInfo> {
         var propertyName = <ts.Identifier> node.name;
         var variableName = propertyName.text;
         var currentScope = this.getCurrentScope();
+        var currentBlockScope = this.getCurrentBlockScope();
         // determine if the appropriate bit in the parent (VariableDeclarationList) is set, which indicates this is a "let"
         var declarationIsLet = (Math.floor(node.parent.flags / ts.NodeFlags.Let) % 2) === 1;
 
         if (currentScope.varNames.indexOf(variableName) >= 0) {
-            // if there was a previous var declaration with the same name, this declaration is invalid
+            // if there was a previous var declaration with the same name, this declaration is invalid (regardless of let)
```
gscshoyru: I'm not sure you need this parenthetical, and if you do want it you should probably explain that it'll be picked up later

src/rules/noDuplicateVariableRule.ts
```diff
@@ -75,6 +80,12 @@ class NoDuplicateVariableWalker extends Lint.ScopeAwareRuleWalker<ScopeInfo> {
             }
         } else {
```
adidahiya: now that I think about it, I think these branches should be flipped around. I always prefer to have the positive conditional first, and the negative branch in the `else`

src/rules/noDuplicateVariableRule.ts
adidahiya: hmm... that's not exactly true; I don't think the fact that we deal with `let`s later has anything to do with this line. I'm basically trying to say that any kind of `VariableDeclaration` is invalid if the `currentScope` has a `var` of the same name.

gscshoyru: Oh, I see what you mean. Maybe expand it a little then, "regardless of whether it's a let or not" or something? Though admittedly I was probably just being confused, this is likely fine as it stands.

test/files/rules/no-shadowed-variable.test.js
```diff
@@ -0,0 +1,57 @@
+function letTesting() {
```
adidahiya: oops, this file should be removed

ashwinr:
This is mixing two things in the same PR -- this is a burden on reviewers, each PR should be as simple as possible. Let's split it out into a PR for no-duplicate-variable and another one for just adding the new rule.

adidahiya:
well, this PR also moves certain functionality out of `NoDuplicateVariableWalker` and into `NoShadowedVariableWalker`... so they're not entirely independent.

I could make the addition of `NoShadowedVariableWalker` _first_ (which would be duplicating functionality in `NoDuplicateVariableWalker`) and _then_ make a PR with edits/removals from `NoDuplicateVariableWalker`... would you prefer that?

ashwinr:
sounds perfect, let's do that!

Issue #365, opened 2015-04-15
Semicolon not being enforced on type definition


goldcaddy77:
I've run into an issue that seems similar to #254, but have confirmed with @gscshoyru that it's different:

Should the private resolvedCampaign : Campaign below be throwing a lint error?

```typescript
///<reference path="../../../../typings/tsd.d.ts" />

class MyClass {
  private resolvedCampaign : Campaign

  constructor(resolvedCampaign : Campaign) {
    this.resolvedCampaign = resolvedCampaign;
  }
}
```

Stack overflow question: http://stackoverflow.com/questions/29654799/tslint-semicolon-option-not-yelling-after-type-definition/

gscshoyru:
I think this likely has the same or a similar root cause, but is totally a bug that should be fixed. Thanks for bringing it to our attention, in case the root cause isn't the same, we should track it separately, I think.

Issue #254, opened 2014-12-18
Semicolon rule doesn't check interfaces


gscshoyru:
```typescript
interface ITest {
    foo: string
}
```

should trigger missing semi-colon, but doesn't.

awerlang:
The main reason against ASI is avoiding gotchas with expressions and statements. It is not a concern in class and interfaces declarations.

And according to http://www.typescriptlang.org/Content/TypeScript%20Language%20Specification.pdf, semicolons are separators and not terminators. But I think even tsc doesn't handle that as per spec.

```
TypeMemberList:
    TypeMember
    TypeMemberList ; TypeMember
```

goldcaddy77:
Note: I found a similar (but different) semicolon linting issue #365

PR #377, opened 2015-04-24
add linting for missing semicolons after type definitions in classes and interfaces (#365 and #254)

gdcohan:


test/files/rules/semicolon.test.ts
```diff
@@ -36,3 +36,13 @@ function useStrictMissingSemicolon() {
     "use strict"
     return null;
 }
+
+class MyClass {
+    public name : string
+    private index : number
```
adidahiya: please add a member with _correct_ semicolon placement to `MyClass` and `ITest` so that we can test for false positives.

adidahiya:
:+1: lgtm

ashwinr:
awesome!

PR #380, opened 2015-04-30
Add tsconfig.json project files & remove reference paths

adidahiya:
This allows us to compile TSLint incrementally via editors like Sublime & Atom
without having to run the whole `grunt ts` task.


Gruntfile.js
```diff
@@ -38,26 +38,30 @@ module.exports = function(grunt) {
         configuration: grunt.file.readJSON("tslint.json")
       },
       src: [
-        "src/*.ts",
-        "src/formatters/**/*.ts",
-        "src/language/**/*.ts",
-        "src/rules/**/*.ts"
+        'src/*.ts',
```
ashwinr: why did these change to single quotes?

Gruntfile.js
```diff
@@ -19,7 +19,7 @@ module.exports = function(grunt) {
         dest: 'lib/tslint.js'
       },
       test: {
-        src: ['<%= typescriptBin %>', 'build/tslint-tests.js'],
+        src: ['lib/tslint.js', 'build/tslint-tests.js'],
```
ashwinr: why did this change?

Gruntfile.js
```diff
@@ -67,7 +71,13 @@ module.exports = function(grunt) {
           declaration: true,
           module: 'commonjs'
         },
-        src: ['src/tslint.ts'],
+        src: [
+          'typings/*.d.ts',
+          'src/language/walker/syntaxWalker.ts',
```
ashwinr: why is syntaxWalker special? if this is for ordering, we need a better way than this.

test/files/rules/whitespace.test.js
```diff
@@ -1,2 +1,39 @@
-
-
```
ashwinr: what file is this?

adidahiya: this should not be in the repo. will remove

Gruntfile.js
adidahiya: to make the file consistently use single quotes

Gruntfile.js
adidahiya: Tests were getting compiled with the source included because of a reference to `../src/tslint.ts` in `test/helper.ts`. Since we removed reference paths, that part of the built `build/tslint-tests.js` file got lost. I didn't want to have to re-compile `src` every time `grunt test` is run, so I changed this line. Open to other ideas

ashwinr: cool

Gruntfile.js
ashwinr: let's consistently use double quotes -- that's our standard everywhere else. feel free to file an issue and revert this file to what it was using originally.

test/files/rules/whitespace.test.js
ashwinr: :+1:

Gruntfile.js
adidahiya: yes, it's for ordering. I'll investigate if there's a way to avoid this by using external modules.

adidahiya: haven't found a good alternative to this manual glob ordering that doesn't involve reference paths

ashwinr:
@adidahiya can we merge this into master?

adidahiya:
can we do a little more gitflow here? I'd be comfortable merging this into a `develop` branch and working off that

ashwinr:
Sure if you want to.

Sent from a phone

On May 6, 2015, at 11:33, Adi Dahiya <notifications@github.com<mailto:notifications@github.com>> wrote:


can we do a little more gitflow here? I'd be comfortable merging this into a develop branch and working off that

PR #391, opened 2015-05-09
Add JSCS to build

### LINKED ISSUE #381: Use double quotes in Gruntfile.js
perhaps also apply JSHint as part of the build

adidahiya:
- Fixes #381


ashwinr:
actually, why is that going into develop? Let's do `master`. It's not popular to use gitflow on github (e.g. angular, react, typescript etc.). I'd rather we follow the convention of popular github projects for consistency.

PR #393, opened 2015-05-11
Fix the custom rule example.

panuhorsmalahti:
I noticed the custom rule example didn't work with the latest version of tslint. This is probably due to TypeScript compiler API changes.


gscshoyru:
Whoops. Thanks!

PR #395, opened 2015-05-12
No var keyword rule

### LINKED ISSUE #378: new rule to disallow `var` declarations in favor of `let` and `const`

DickvdBrink:
Added no-var-keyword rule.

I couldn't think of a better name for this rule so if you have a better name I will change it :)
fix #378
Feel free to close this PR if this isn't the way you want it :)

Btw: I'm currently on vacation so feedback might take till around Sunday :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -172,7 +172,7 @@ A sample configuration file with all options is available [here](https://github.
     * `"check-parameters"` disallows unused function and constructor parameters.
         * NOTE: this option is experimental and does not work with classes that use abstract method declarations, among other things. Use at your own risk.
 * `no-use-before-declare` disallows usage of variables before their declaration.
-* `no-var-keyword` disallows ussage of the `var` keyword, use `let` or `const` instead.
+* `no-var-keyword` disallows usage of the `var` keyword, use `let` or `const` instead.
 * `no-var-requires` disallows the use of require statements except in import statements, banning the use of forms such as `var module = require("module")`
 * `one-line` enforces the specified tokens to be on the same line as the expression preceding it. Rule options:
   * `"check-catch"` checks that `catch` is on the same line as the closing brace for `try`
--- a/README.md
+++ b/README.md
@@ -172,6 +172,7 @@ A sample configuration file with all options is available [here](https://github.
     * `"check-parameters"` disallows unused function and constructor parameters.
         * NOTE: this option is experimental and does not work with classes that use abstract method declarations, among other things. Use at your own risk.
 * `no-use-before-declare` disallows usage of variables before their declaration.
+* `no-var-keyword` disallows ussage of the `var` keyword, use `let` or `const` instead.
 * `no-var-requires` disallows the use of require statements except in import statements, banning the use of forms such as `var module = require("module")`
 * `one-line` enforces the specified tokens to be on the same line as the expression preceding it. Rule options:
   * `"check-catch"` checks that `catch` is on the same line as the closing brace for `try`
--- a/src/rules/noVarKeywordRule.ts
+++ b/src/rules/noVarKeywordRule.ts
@@ -1,0 +1,71 @@
+/*
+ * Copyright 2015 Palantir Technologies, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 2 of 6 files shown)


README.md
```diff
@@ -172,6 +172,7 @@ A sample configuration file with all options is available [here](https://github.
     * `"check-parameters"` disallows unused function and constructor parameters.
         * NOTE: this option is experimental and does not work with classes that use abstract method declarations, among other things. Use at your own risk.
 * `no-use-before-declare` disallows usage of variables before their declaration.
+* `no-var-keyword` disallows ussage of the `var` keyword, use `let` or `const` instead.
```
ash211: Typo: ussage -> usage

src/rules/noVarKeywordRule.ts
```diff
@@ -0,0 +1,71 @@
...
+
+        super.visitVariableStatement(node);
+        this.skipVariableDeclaration = false;
+    }
+
+    public visitVariableDeclaration(node: ts.VariableDeclaration) {
+
+        if (!this.skipVariableDeclaration) {
+            var flags = node.parent.flags;
+            var name = (<ts.Identifier> node.name);
```
adidahiya: `node.name` might be a `BindingPattern` (a.k.a. destructuring), so you'll have to handle both cases.

src/rules/noVarKeywordRule.ts
```diff
@@ -0,0 +1,71 @@
...
+            this.hasModifier(node.modifiers, ts.SyntaxKind.DeclareKeyword)) {
+            this.skipVariableDeclaration = true;
+        }
+
+        super.visitVariableStatement(node);
+        this.skipVariableDeclaration = false;
+    }
+
+    public visitVariableDeclaration(node: ts.VariableDeclaration) {
+
```
adidahiya: remove this newline

src/rules/noVarKeywordRule.ts
```diff
@@ -0,0 +1,71 @@
...
+        if (modifiers == null) {
+            return false;
+        }
+        for (var i = 0; i < modifiers.length; i++) {
+            var modifier = modifiers[i];
+            if (modifier.kind === modifierKind) {
+                return true;
+            }
+        }
+        return false;
```
adidahiya: L62–L68 can be similified:

```ts
return modifiers.some((modifier) => modifier.kind === modifierKind);
```

src/rules/noVarKeywordRule.ts
```diff
@@ -0,0 +1,71 @@
...
+        return this.applyWithWalker(noVarKeywordWalker);
+    }
+}
+
+class NoVarKeywordWalker extends Lint.RuleWalker {
+    private skipVariableDeclaration = false;
+
+    public visitVariableStatement(node: ts.VariableStatement): void {
+        if (this.hasModifier(node.modifiers, ts.SyntaxKind.ExportKeyword) ||
+            this.hasModifier(node.modifiers, ts.SyntaxKind.DeclareKeyword)) {
```
adidahiya: please move the `||` operator to the next line

test/files/rules/novarkeyword.test.ts
```diff
@@ -0,0 +1,25 @@
...
+
+let bar;
+const qux;
+
+let k,
+    l;
+
+module quz {
+    export var i;
+}
```
adidahiya: let's add some more tests for 1.5 syntax, like `export default`, `import`, destructuring (`var [a, b] = [1, 2]`)...

src/rules/noVarKeywordRule.ts
adidahiya: After you rebase on the latest `master`, you can use the new util function `Lint.hasModifier` instead.

adidahiya:
Any estimate on when you'll be able to make these fixes? Should I take up this PR?

DickvdBrink:
Oops sorry, will fix some stuff tonight. Actually forgot about this a bit...

I do have a small question though!
The location where this error is thrown might be a bit odd, it is now thrown twice on the code below and on the location of the name but there is only one `var` keyword so that might be a bit weird. Do you agree?

    var i,
        j;

adidahiya:
I agree, it's excessive to report 2 failures in the case of a variable declaration list of length 2 like that. You could try making the walker visit `VariableDeclarationList` nodes to get the desired behavior.

DickvdBrink:
Fixed most of the feedback stuff (hopefully in the right way) and rebased on master.

What I didn't fix (yet) is the `export default` stuff because I don't know how to make it usable in a single TypeScript file..

PR #398, opened 2015-05-13
Don't check `declared` vars in the unused-variable-rule

### LINKED ISSUE #190: no-unused-variable doesn't properly handle "declare" vars
For example, this should not be flagged as unused:
```
declare var assert: chai.Assert;
```

DickvdBrink:
fix #190 
`declare var x: any;`

^^ Above code isn't an error anymore wen x isn't used

### DIFF
--- a/src/rules/noUnusedVariableRule.ts
+++ b/src/rules/noUnusedVariableRule.ts
@@ -93,9 +93,10 @@ class NoUnusedVariablesWalker extends Lint.RuleWalker {
         this.skipParameterDeclaration = false;
     }
 
-    // skip exported variables
+    // skip exported and declared variables
     public visitVariableStatement(node: ts.VariableStatement): void {
-        if (this.hasModifier(node.modifiers, ts.SyntaxKind.ExportKeyword)) {
+        if (this.hasModifier(node.modifiers, ts.SyntaxKind.ExportKeyword)
+            || this.hasModifier(node.modifiers, ts.SyntaxKind.DeclareKeyword)) {
             this.skipVariableDeclaration = true;
         }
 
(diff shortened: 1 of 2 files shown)


src/rules/noUnusedVariableRule.ts
```diff
@@ -95,7 +95,8 @@ class NoUnusedVariablesWalker extends Lint.RuleWalker {
 
     // skip exported variables
     public visitVariableStatement(node: ts.VariableStatement): void {
-        if (this.hasModifier(node.modifiers, ts.SyntaxKind.ExportKeyword)) {
+        if (this.hasModifier(node.modifiers, ts.SyntaxKind.ExportKeyword) ||
+            this.hasModifier(node.modifiers, ts.SyntaxKind.DeclareKeyword)) {
```
adidahiya: please put the boolean operator on the next line for consistency.
yes, we ought to have a tslint rule that checks this :)

src/rules/noUnusedVariableRule.ts
```diff
@@ -95,7 +95,8 @@ class NoUnusedVariablesWalker extends Lint.RuleWalker {
 
     // skip exported variables
```
adidahiya: let's update this comment as well

src/rules/noUnusedVariableRule.ts
DickvdBrink: Will change it!
though it seems it isn't really consistent as for example here it is written the same way I did: https://github.com/palantir/tslint/blob/b08678fe3a2ea28368e6c319336c12a4e1a81064/src/rules/labelPositionRule.ts#L30

PR #402, opened 2015-05-21
Update noEmptyRule for AccessModifier

yyoshiki41:
## Update noEmptyRule.

```typescript
class testClass {
    constructor(protected allowed: any) {
    }
}
```
^^
Access modifier : `protected`
is not allowed to have empty blocks.
So I created function : `isAccessModifier(modifier: string)`. 
As a result, empty blocks will be ignored when using `protected` .

## Versions
- tslint
2.2.0-beta

- typescript
1.5.0

[Review on Reviewable](https://reviewable.io/reviews/palantir/tslint/402)

### DIFF
--- a/src/rules/noEmptyRule.ts
+++ b/src/rules/noEmptyRule.ts
@@ -52,9 +52,8 @@ class BlockWalker extends Lint.RuleWalker {
             for (var j = 0; param.modifiers != null && j < param.modifiers.length; j++) {
                 var modifier = param.modifiers[j].kind;
 
-                if (modifier === ts.SyntaxKind.PublicKeyword || modifier === ts.SyntaxKind.PrivateKeyword) {
+                if (this.isAccessModifier(param.modifiers[j].kind)) {
                     isSkipped = true;
-
                     this.ignoredBlocks.push(node.body);
 
                     break;
@@ -67,5 +66,18 @@ class BlockWalker extends Lint.RuleWalker {
         }
 
         super.visitConstructorDeclaration(node);
+    }
+
+    private isAccessModifier(modifier: string): boolean {
+        if (modifier === ts.SyntaxKind.PrivateKeyword) {
+            return true;
+        }
+        if (modifier === ts.SyntaxKind.ProtectedKeyword) {
+            return true;
+        }
+        if (modifier === ts.SyntaxKind.PublicKeyword) {
+            return true;
+        }
+        return false;
     }
 }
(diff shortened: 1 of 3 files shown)


src/rules/noEmptyRule.ts
```diff
@@ -68,4 +67,17 @@ class BlockWalker extends Lint.RuleWalker {
...
+            return true;
+        }
+        if (modifier === ts.SyntaxKind.ProtectedKeyword) {
+            return true;
+        }
+        if (modifier === ts.SyntaxKind.PublicKeyword) {
+            return true;
+        }
+        return false;
+    }
```
adidahiya: simplify to:

```ts
return modifier === ts.SyntaxKind.PrivateKeyword,
    || modifier === ts.SyntaxKind.ProtectedKeyword,
    || modifier === ts.SyntaxKind.PublicKeyword;
```

src/rules/noEmptyRule.ts
```diff
@@ -68,4 +67,17 @@ class BlockWalker extends Lint.RuleWalker {
 
         super.visitConstructorDeclaration(node);
     }
+
+    private isAccessModifier(modifier: string): boolean {
```
adidahiya: a better name might be `isPropertyAccessModifier`

adidahiya:
thanks for the bugfix! should be good to go after those small style changes

yyoshiki41:
thanks for reviewing my p-r!!
Simplified function:`isPropertyAccessModifier` :)

src/rules/noEmptyRule.ts
```diff
@@ -68,4 +67,10 @@ class BlockWalker extends Lint.RuleWalker {
 
         super.visitConstructorDeclaration(node);
     }
+
+    private isPropertyAccessModifier(modifier: string): boolean {
+        return modifier === ts.SyntaxKind.PrivateKeyword ||
+            modifier === ts.SyntaxKind.ProtectedKeyword ||
+            modifier === ts.SyntaxKind.PublicKeyword;
```
adidahiya: can you please put the `||` operator on the next lines?

(sorry, we should probably have a tslint rule for this)

```ts
return modifier === ts.SyntaxKind.PrivateKeyword
    || modifier === ...
```

yyoshiki41:
> can you please put the || operator on the next lines?
(sorry, we should probably have a tslint rule for this)

fixed!!
followed a tslint rule :)

PR #404, opened 2015-05-22
New rule: classMemberNoDefaultAccess.

meros:
Disallows public access modifiers that are not explicit in the class.

### DIFF
--- a/docs/sample.tslint.json
+++ b/docs/sample.tslint.json
@@ -13,7 +13,7 @@
     "curly": true,
     "eofline": true,
     "forin": true,
-    "indent": [true, 4],
+    "indent": [true, "spaces"],
     "interface-name": true,
     "jsdoc-format": true,
     "label-position": true,


leeavital:
Can you add this rule to the README and `sample.tslint.json`?

meros:
Sure thing.

adidahiya:
Sorry this has taken so long to get reviewed. Can you please rebase these changes on the latest `master` branch? There are a couple changes I'd like you to pick up and incorporate into this code:

1. We now use `tsconfig.json` project files in this project. So you don't have to use reference paths anywhere. You do, however, have to keep the relevant config files up to date when you add new source files to the project. See `src/tsconfig.json`, `src/rules/tsconfig.json`, `test/tsconfig.json`
2. `utils.ts` now provides a function called `Lint.hasModifier()` which should be able to replace much of the code in this new rule walker.

src/rules/classMemberNoDefaultAccessRule.ts
```diff
@@ -0,0 +1,69 @@
...
+    }
+
+    public visitPropertyDeclaration(node: ts.PropertyDeclaration): void {
+        this.checkModifiers(node, getModifiers(node.modifiers));
+        super.visitPropertyDeclaration(node);
+    }
+
+    private checkModifiers(node: ts.Node, current: IModifiers): void {
+        if (!this.followsRules(current)) {
+            var message = "Default access modifier on member/method not allowed";
```
adidahiya: change this to be a `public static FAILURE_STRING` property on the `Rule` class in this file -- similar to how other rule walkers are structured.

test/rules/classMemberNoDefaultAccessTest.ts
```diff
@@ -0,0 +1,32 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/// <reference path='../references.ts' />
+
+describe("<class-member-no-default-access>", () => {
+    it("disallow default access on member", () => {
+        var fileName = "rules/classmembernodefaultaccess.test.ts";
+        var ClassMemberNoDefaultAccess = Lint.Test.getRule("class-member-no-default-access");
```
adidahiya: `ClassMemberNoDefaultAccessRule`

test/rules/classMemberNoDefaultAccessTest.ts
```diff
@@ -0,0 +1,32 @@
...
+        var actualFailures = Lint.Test.applyRuleOnFile(fileName, ClassMemberNoDefaultAccess, []);
+
+        Lint.Test.assertFailuresEqual(actualFailures, [
+            Lint.Test.createFailure(fileName, [8, 5], [8, 15],
+                "Default access modifier on member/method not allowed"),
```
adidahiya: after addressing the above comments, you should be able to reference this via `Rule.FAILURE_STRING`

test/rules/classMemberNoDefaultAccessTest.ts
```diff
@@ -0,0 +1,32 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/// <reference path='../references.ts' />
+
+describe("<class-member-no-default-access>", () => {
+    it("disallow default access on member", () => {
+        var fileName = "rules/classmembernodefaultaccess.test.ts";
```
adidahiya: use `const` instead of `var`

README.md
```diff
@@ -127,6 +127,7 @@ A sample configuration file with all options is available [here](https://github.
   * `"arguments"` checks alignment of function call arguments.
   * `"statements"` checks alignment of statements.
 * `ban` bans the use of specific functions. Options are ["object", "function"] pairs that ban the use of object.function()
+* `class-member-no-default-access` enforces using explicit visibility on class members
```
ashwinr: this is too verbose. let's call this `member-access`. in general, we should be limiting rule name verbosity.

adidahiya:
@meros I like this rule, and I want it in tslint :)

Will you have time to address the PR comments soon? Or can I take the code and make a new PR?

ashwinr:
@adidahiya let's take this up when you get time and merge it in. also let's rename the rule to `member-no-default-access` to be consistent with `member-ordering`.

adidahiya:
I think `member-access` is slightly better

ashwinr:
What would the value be? Just a Boolean?

Sent from a phone

On Jul 22, 2015, at 09:20, Adi Dahiya <notifications@github.com<mailto:notifications@github.com>> wrote:


I think member-access is slightly better

meros:
Hi guys, sorry I have been a little under the radar a while. I have been in vacation mode, and in sweden that is a non-negligible part of the year  :) I will look into this again soon, probably next week. 

I see no problem in the comments so far. About the name, member-access doesn't really give a hint on what setting the flag would give. I like member-no-default-access the best, it's longer but the behaviour is spot on with what the module enforces.

meros:
Please see this pull request instead: https://github.com/palantir/tslint/pull/552

Issue #324, opened 2015-02-16
tslint throws "jsdoc is not formatted correctly" only on Windows for typedoc blank asterisk


pellekrogholt:
## versions in use

- ts 1.4.1
- grunt-ts: 3.0.0
- grunt-tslint: 2.0.0
- tslint 2.1.1
- typedoc 0.1.0

## problem 

i have typedoc here and there - some times using a blank line for readability or because markdown requires it - lets have an example:

    /*
     * Test file
     *
     * Test a file documentation text
     *
     *     Some code that should be markdown code formatted
     *
     * @param test    Some text about test param.
     * @param path    Some text about path param.
     */
    function testFile(test: any, path:any) {
    ...

on OS like ubuntu 14.* or osx mavericks this passes with my tslint task

    grunt tslint:all

but on Windows (Windows 8.0 / Windows Server 2012) the empty `*` lines raises

{code}
jsdoc is not formatted correctly 
{code}

i could change the empty lines with `* ` and the i pass. but i'm not super happy about this because various white space cleaners cleans that additional extra space away.

pardon me if I post the issue in the wrong spot. i'm a bit unsure if this is a tslint or grunt-tslint issue or me that are one a wrong path here. 

thx!

//

pelle

mwe:
I raised probably the same issue, If your problem is related to UTF8 files. ( see #324 )

eonarheim:
I believe the problem is this line of code right here when we split the lines on "\n" windows line endings are "\r\n"

```js
private findFailuresForJsdocComment(commentText: string, startingPosition: number, sourceFile: ts.SourceFile) {
        var currentPosition = startingPosition;
        var lines = commentText.split("\n"); // Windows file line endings are \r\n!
        var jsdocPosition = currentPosition;
        var firstLine = lines[0];
...
```
which then causes this regex to fail, since there is an extra "\r" from the split

```js
// regex is: start of string, followed by any amount of whitespace, followed by *,
// followed by either a space or the end of the string
var asteriskMatch = line.match(/^\s*\*( |$)/);
if (asteriskMatch == null) {
    this.addFailureAt(jsdocPosition, line.length, Rule.FORMAT_FAILURE_STRING);
}
```

I have raised a pull request fixing this with tests :)

adidahiya:
looks like the fix got merged. please comment if this bug still occurs in the latest tslint.

PR #405, opened 2015-05-22
Fix issue #324 windows line endings not respected

eonarheim:
I came across this issue today as I was putting tslint rules into place on our code base. All of our development occurs on windows, and I was surprised to see so many jsdoc formatting errors. I dug in and found that #324 exactly described what was happening.

After digging further I found that the `jsdocFormatRule` did not support windows line endings! How funny for a TypeScript library I thought :)

I've gone ahead and implemented a fix. The reason I'm solving this with a regular expression and not detecting the host OS with `require('os').EOL` or `process.platform`, is because the line ending is dependent on where the file was originally authored, not where tslint is running.

Let me know if you would like me to update any comments, test, code for this PR.

Cheers!


adidahiya:
nice, thanks for the explanation and the bugfix!

Issue #371, opened 2015-04-20
ES6 let and const syntax


adidahiya:
https://github.com/Microsoft/TypeScript/pull/904

adidahiya:
In order to properly address #321, it seems like we need a `BlockScopeAwareRuleWalker<T>` and a new rule to lint duplicate `let` declarations in the same block scope. I don't see a clean way to integrate this into `NoDuplicateVariableWalker` while still supporting valid syntax like:

```ts
let a = 1;
if (true) {
    let a = 2;
}
```

... so I propose a new `NoDuplicateLetWalker`. Thoughts @gscshoyru, @ashwinr?

PR #409, opened 2015-05-25
Add no-shadowed-variable rule

adidahiya:
part of https://github.com/palantir/tslint/issues/371

this PR is a follow-up of the discussion here: https://github.com/palantir/tslint/pull/375#issuecomment-96439510

`no-shadowed-variable` shares a lot of functionality with the `no-duplicate-variable` rule.
"Shadowing" is the more semantically correct idea here, so we're moving these
checks into this new rule. `no-duplicate-variable` will be changed in an upcoming
PR to only check for duplicates within the same block scope.


adidahiya:
this PR is coming next: https://github.com/adidahiya/tslint/compare/feature-no-shadowed-variable...adidahiya:feature-no-duplicate-variable

src/language/utils.ts
```diff
@@ -76,4 +76,11 @@ module Lint {
             callback(scanner);
         }
     }
+
+      export function isBlockScopedVariable(node: ts.VariableDeclaration): boolean {
+          // determine if the appropriate bit in the parent (VariableDeclarationList) is set,
```
gscshoyru: If you're making this a Lint-wide utility function, you might want to check and see if the parent is a CatchClause, since this is a thing that can happen. I am not sure that counts as block-scoped though, depends on how you're using it.

src/language/walker/blockScopeAwareRuleWalker.ts
```diff
@@ -0,0 +1,85 @@
...
+            return super.isScopeBoundary(node)
+                || node.kind === ts.SyntaxKind.DoStatement
+                || node.kind === ts.SyntaxKind.WhileStatement
+                || node.kind === ts.SyntaxKind.ForStatement
+                || node.kind === ts.SyntaxKind.ForInStatement
+                || node.kind === ts.SyntaxKind.ForOfStatement
+                || node.kind === ts.SyntaxKind.WithStatement
+                || node.kind === ts.SyntaxKind.SwitchStatement
+                || (node.parent != null
+                    && (node.parent.kind === ts.SyntaxKind.TryStatement
```
gscshoyru: I assume this is how you're handing if/else blocks and try/catch/finally clauses? This might act a little weird in the cases of the expression in if statements, though. I'm not sure if there's a better way to handle this, though -- so I think this is ok -- just flagging.

src/language/walker/blockScopeAwareRuleWalker.ts
```diff
@@ -0,0 +1,85 @@
...
+
+        constructor(sourceFile: ts.SourceFile, options?: any) {
+            super(sourceFile, options);
+
+            // initialize stack with global scope
+            this.blockScopeStack = [this.createBlockScope()];
+        }
+
+        public createBlockScope(): U {
+            throw Lint.abstract();
```
leeavital: Is there a reason we can't just make BlockScopeAwareRuleWalker abstract?

gscshoyru: Abstract classes aren't a thing in typescript, which is why we're doing this.

adidahiya:
:+1:

test/rules/noShadowedVariableRuleTests.ts
```diff
@@ -0,0 +1,44 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+describe("<no-shadowed-variable>", () => {
+    it("ensures that variable declarations are not shadowed", () => {
+        var fileName = "rules/no-shadowed-variable.test.ts";
+        var NoShadowedVariableRule = Lint.Test.getRule("no-shadowed-variable");
+        var failureString = NoShadowedVariableRule.FAILURE_STRING + "shadowed'";
```
adidahiya: this variable is unused. look like our test files aren't being linted. @leeavital can you please investigate?

test/rules/noShadowedVariableRuleTests.ts
```diff
@@ -0,0 +1,44 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+describe("<no-shadowed-variable>", () => {
+    it("ensures that variable declarations are not shadowed", () => {
+        var fileName = "rules/no-shadowed-variable.test.ts";
```
adidahiya: let's update these all to be `const`

test/rules/noShadowedVariableRuleTests.ts
adidahiya: if you find the problem, please open a new PR to fix it. I'll take care of the small fixes necessary in this PR.

src/rules/noShadowedVariableRule.ts
```diff
@@ -0,0 +1,99 @@
...
+        return new ScopeInfo();
+    }
+
+    public createBlockScope(): ScopeInfo {
+        return new ScopeInfo();
+    }
+
+    public visitParameterDeclaration(node: ts.ParameterDeclaration): void {
+        // Treat parameters as block-scoped variables
+        var propertyName = <ts.Identifier> node.name;
```
adidahiya: let's use `const` in this file as well

src/rules/noShadowedVariableRule.ts
```diff
@@ -0,0 +1,99 @@
...
+class NoShadowedVariableWalker extends Lint.BlockScopeAwareRuleWalker<ScopeInfo, ScopeInfo> {
+    public createScope(): ScopeInfo {
+        return new ScopeInfo();
+    }
+
+    public createBlockScope(): ScopeInfo {
+        return new ScopeInfo();
+    }
+
+    public visitParameterDeclaration(node: ts.ParameterDeclaration): void {
```
adidahiya: we can drop the return type specifiers everywhere

Issue #368, opened 2015-04-20
ES6 destructuring syntax


adidahiya:
https://github.com/Microsoft/TypeScript/issues/240

some test cases, from [this Mozilla blog post](https://hacks.mozilla.org/2015/05/es6-in-depth-destructuring/):
```ts
var [,,third] = ["foo", "bar", "baz"];
var [head, ...tail] = [1, 2, 3, 4];
```

```ts
var robotA = { name: "Bender" };
var robotB = { name: "Flexo" };

// renaming properties
var { name: nameA } = robotA;
var { name: nameB } = robotB;

// default values
var { x = 3 } = {};

// default values in a function param
jQuery.ajax = function (url, {
  async = true,
  beforeSend = noop,
  cache = true,
  complete = noop,
  crossDomain = false,
  global = true,
  // ... more config
}) {
  // ... do stuff
};
```

PR #410, opened 2015-05-26
Add destructuring syntax & support to nounusedvariable-var.test.ts

adidahiya:
#368

### DIFF
--- a/src/rules/noUnusedVariableRule.ts
+++ b/src/rules/noUnusedVariableRule.ts
@@ -29,8 +29,9 @@ export class Rule extends Lint.Rules.AbstractRule {
 }
 
 class NoUnusedVariablesWalker extends Lint.RuleWalker {
-    private skipVariableDeclaration: boolean;
+    private skipBindingElement: boolean;
     private skipParameterDeclaration: boolean;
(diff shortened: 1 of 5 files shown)


src/rules/noUnusedVariableRule.ts
```diff
@@ -40,6 +41,17 @@ class NoUnusedVariablesWalker extends Lint.RuleWalker {
...
+        if (isSingleVariable && !this.skipBindingElement) {
+            var variableIdentifier = <ts.Identifier> node.name;
+            this.validateReferencesForVariable(variableIdentifier.text, variableIdentifier.getStart());
+        }
+
+        super.visitBindingElement(node);
+    }
+
     public visitImportDeclaration(node: ts.ImportDeclaration): void {
         if (!this.hasModifier(node.modifiers, ts.SyntaxKind.ExportKeyword)) {
```
gscshoyru: I don't think you're handling destructuring in here.

adidahiya: yep, that is implemented in a separate PR: https://github.com/palantir/tslint/pull/411

Issue #399, opened 2015-05-18
ES6 destructuring on import throws an error


bryanforbes:
In the project I'm working on, we're using ES6 imports. tslint 2.2.0-beta is throwing an error attempting to parse it. For instance:

```ts
import { foo } from 'bar';
```

Produces the following backtrace:

```
/Users/bryan/Projects/dojo/dojo2/dojo-core/node_modules/tslint/build/rules/noUnusedVariableRule.js:52
                this.validateReferencesForVariable(name.text, name.getStart())
                                                       ^
TypeError: Cannot read property 'text' of undefined
    at NoUnusedVariablesWalker.visitImportDeclaration (/Users/bryan/Projects/dojo/dojo2/dojo-core/node_modules/tslint/build/rules/noUnusedVariableRule.js:52:56)
    at NoUnusedVariablesWalker.SyntaxWalker.visitNode (/Users/bryan/Projects/dojo/dojo2/dojo-core/node_modules/tslint/bin/tslint-cli.js:41611:26)
    at /Users/bryan/Projects/dojo/dojo2/dojo-core/node_modules/tslint/bin/tslint-cli.js:41707:67
    at visitEachNode (/Users/bryan/Projects/dojo/dojo2/dojo-core/node_modules/tslint/bin/tslint-cli.js:5928:30)
    at Object.forEachChild (/Users/bryan/Projects/dojo/dojo2/dojo-core/node_modules/tslint/bin/tslint-cli.js:6070:24)
    at NoUnusedVariablesWalker.SyntaxWalker.walkChildren (/Users/bryan/Projects/dojo/dojo2/dojo-core/node_modules/tslint/bin/tslint-cli.js:41707:16)
    at NoUnusedVariablesWalker.SyntaxWalker.visitSourceFile (/Users/bryan/Projects/dojo/dojo2/dojo-core/node_modules/tslint/bin/tslint-cli.js:41495:18)
    at NoUnusedVariablesWalker.SyntaxWalker.visitNode (/Users/bryan/Projects/dojo/dojo2/dojo-core/node_modules/tslint/bin/tslint-cli.js:41671:26)
    at NoUnusedVariablesWalker.SyntaxWalker.walk (/Users/bryan/Projects/dojo/dojo2/dojo-core/node_modules/tslint/bin/tslint-cli.js:41348:18)
[...]

PR #411, opened 2015-05-26
Add destructuring syntax & support for no-unused-variable imports

adidahiya:
#368, #399


test/files/rules/nounusedvariable-imports.test.ts
```diff
@@ -10,3 +13,5 @@ $(_.xyz());
 module S {
   var template = '';
 }
+
+import * as foo from "lib";
```
gscshoyru: Can you do a little false positive checking here too, please? Make sure this doesn't trigger if it is actually used.

PR #416, opened 2015-05-27
Added rule to sort object literal keys alphabetically

CalvinFernandez:
This rule lints an object literal's keys and validates that the user has declared them in alphabetical order.

### DIFF
--- a/src/rules/sortObjectLiteralKeysRule.ts
+++ b/src/rules/sortObjectLiteralKeysRule.ts
@@ -1,0 +1,55 @@
+/*
+ * Copyright 2013 Palantir Technologies, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+*/
+
+export class Rule extends Lint.Rules.AbstractRule {
+    public static FAILURE_STRING = "unsorted key '";
+
+    public apply(sourceFile: ts.SourceFile): Lint.RuleFailure[] {
+        return this.applyWithWalker(new SortedKeyWalker(sourceFile, this.getOptions()));
+    }
+}
+
+class SortedKeyWalker extends Lint.RuleWalker {
+    private lastSortedKeyStack: string[] = [];
+    private sortedStateStack: boolean[] = [];
+
+    public visitObjectLiteralExpression(node: ts.ObjectLiteralExpression): void {
+        this.lastSortedKeyStack.push(""); // Char code 0; every string should be >= to this
+        this.sortedStateStack.push(true); // Sorted state is always initially true
+        super.visitObjectLiteralExpression(node);
+        this.lastSortedKeyStack.pop();
+        this.sortedStateStack.pop();
+    }
+
+    public visitPropertyAssignment(node: ts.PropertyAssignment): void {
+        var sortedState = this.sortedStateStack[this.sortedStateStack.length - 1];
+        if (sortedState) {
+            var lastSortedKey = this.lastSortedKeyStack[this.lastSortedKeyStack.length - 1];
+            var keyNode = node.name;
+            if (keyNode.kind === ts.SyntaxKind.Identifier) {
+                var key = (<ts.Identifier> keyNode).text;
+                if (key < lastSortedKey) {
+                    var failureString = Rule.FAILURE_STRING + key + "'";
+                    this.addFailure(this.createFailure(keyNode.getStart(), keyNode.getWidth(), failureString));
+                    this.sortedStateStack[this.sortedStateStack.length - 1] = false;
+                } else {
+                    this.lastSortedKeyStack[this.lastSortedKeyStack.length - 1] = key;
+                }
+            }
+        }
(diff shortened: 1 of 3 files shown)


src/rules/sortObjectLiteralKeysRule.ts
```diff
@@ -0,0 +1,55 @@
...
+    public static FAILURE_STRING = "unsorted key '";
+
+    public apply(sourceFile: ts.SourceFile): Lint.RuleFailure[] {
+        return this.applyWithWalker(new SortedKeyWalker(sourceFile, this.getOptions()));
+    }
+}
+
+class SortedKeyWalker extends Lint.RuleWalker {
+    private lastSortedKeyStack: string[] = [];
+    private sortedStateStack: boolean[] = [];
```
adidahiya: this array only ever has one value. how about we change this to a boolean flag `isCurrentObjectLiteralSorted`?

src/rules/sortObjectLiteralKeysRule.ts
```diff
@@ -0,0 +1,55 @@
...
+        this.lastSortedKeyStack.push(""); // Char code 0; every string should be >= to this
+        this.sortedStateStack.push(true); // Sorted state is always initially true
+        super.visitObjectLiteralExpression(node);
+        this.lastSortedKeyStack.pop();
+        this.sortedStateStack.pop();
+    }
+
+    public visitPropertyAssignment(node: ts.PropertyAssignment): void {
+        var sortedState = this.sortedStateStack[this.sortedStateStack.length - 1];
+        if (sortedState) {
```
adidahiya: add a comment explaining that if a sorting failure was already reported for this object literal, we don't need to keep checking

src/rules/sortObjectLiteralKeysRule.ts
```diff
@@ -0,0 +1,55 @@
...
+export class Rule extends Lint.Rules.AbstractRule {
+    public static FAILURE_STRING = "unsorted key '";
+
+    public apply(sourceFile: ts.SourceFile): Lint.RuleFailure[] {
+        return this.applyWithWalker(new SortedKeyWalker(sourceFile, this.getOptions()));
+    }
+}
+
+class SortedKeyWalker extends Lint.RuleWalker {
+    private lastSortedKeyStack: string[] = [];
```
adidahiya: this can just be a `string`, right? `lastSortedKey`?

src/rules/sortObjectLiteralKeysRule.ts
```diff
@@ -0,0 +1,55 @@
...
+        super.visitObjectLiteralExpression(node);
+        this.lastSortedKeyStack.pop();
+        this.sortedStateStack.pop();
+    }
+
+    public visitPropertyAssignment(node: ts.PropertyAssignment): void {
+        var sortedState = this.sortedStateStack[this.sortedStateStack.length - 1];
+        if (sortedState) {
+            var lastSortedKey = this.lastSortedKeyStack[this.lastSortedKeyStack.length - 1];
+            var keyNode = node.name;
```
adidahiya: since we're on TS 1.5 now, we can use `let` and `const`. prefer `const` whenever possible.

test/rules/sortObjectLiteralKeysRuleTests.ts
```diff
@@ -0,0 +1,39 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+describe("<sort-object-literal-keys>", () => {
+    it("forbids unsorted keys in object literals", () => {
+        var fileName = "rules/sortedkey.test.ts";
```
adidahiya: same comment as above; all these `var`s can be `const`s

src/rules/sortObjectLiteralKeysRule.ts
CalvinFernandez: Sure, can do that, this will of cause the entire recursive object literal check to fail so something like 
```
{
  b: "b",
  a: "a",
  c: {
    "b": "b",
    "a": "a"
  }
}
```

won't throw an error on the inner object literal error. Is that okay?

src/rules/sortObjectLiteralKeysRule.ts
CalvinFernandez: Not quite, the recursive case is going to cause a valid object like this to fail, 
```
var valid = {
  a: 'a',
  b: {
    dd: 'dd'
  },
  c: 'c'
}
```
where the nested object literal will step on the lastSortedKey value of the outer object literal and cause the key walker to compare 'c' to 'dd' and throw an unsorted exception.

adidahiya: Right on, I wasn't thinking about nested literals. Both stacks make sense now. Please add a brief comment above where those member vars are declared explaining that the stacks handle nested object literals.

src/rules/sortObjectLiteralKeysRule.ts
adidahiya: let's keep this as a stack; see my other comment.

test/files/rules/sortedkey.test.ts
```diff
@@ -0,0 +1,61 @@
...
+    b: {
+        aa: 1,
+        bb: 2
+    }
+};
+
+var failC = {
+    a: 1,
+    b: {
+        bb: 2,
```
gscshoyru: Can you add a nesting test that's like, say, {a: 1, c: {aa: 1, bb: 2}, b: 3} or somesuch -- that is, a test that checks the thing from before we went into the next layer down?

CalvinFernandez: done, github didn't close this comment

adidahiya:
:+1:

PR #418, opened 2015-06-01
Fix error when no default configuration exists

gtanner:
Will now return undefined for getConfiguration calls.

I was using tslint programatically in http://www.bithound.io/ I have a case where I want to check if the user has a default config in their repo. My servers do not have a default config in the home dir. This fixes that.


adidahiya:
we've gone ahead and implemented a default config if no config file is found (https://github.com/palantir/tslint/pull/437), so I think this is no longer needed. thanks though!

PR #429, opened 2015-06-11
Prepare for version 2.3.0

leeavital:


CHANGELOG.md
```diff
@@ -1,5 +1,15 @@
 Change Log
 ===
+v2.3.0
+---
+* [bugs] #401 #367 #324 #352
+* [new-rule] `no-var-keyword` disallows `var` in favor of `let` and `const`
+* [new-rule] `sort-object-literal-keys` forces object-literal keys to be sorted alphabetically
+* Update the `whitespace` rule to work with ES6 module syntax
+* Update `variable-name`, `no-use-before-declare`, and `no-unused-variable` to support destructuring
```
adidahiya: let's combine the above 2 bullet points:

* Add support for ES6 destructuring and module syntax (affects `whitespace`, `variable-name`, `no-unused-variable`, `no-use-before-declare` rules)

CHANGELOG.md
```diff
@@ -1,5 +1,15 @@
...
+* Add tsconfig.json & JSCS to the build system
```
adidahiya: how about "Use ..." instead of "Add ..."

CHANGELOG.md
leeavital: We should also mention that the destructuring syntax is incomplete (no field renaming as of now), or wait for that to be done before this release.

adidahiya:
`src/tslint.ts` also contains a reference to "2.2.0-beta" which you'll need to update

PR #431, opened 2015-06-11
Add test for field rename destructuring in no-unused-variable

leeavital:


test/files/rules/nounusedvariable-var.test.ts
```diff
@@ -43,3 +43,9 @@ export function testUnusedSpread() {
 for(let e of [1,2,3]) {
 
 }
+
+export function testRenamingDestructure() {
+  var a = 2;
+  let {a: b} = {a: 4}; // 1 error (b is unused)
+  return a;
+}
```
adidahiya: Let's add a false positive check as well, where the renamed var _is_ used

PR #432, opened 2015-06-12
Introduce no-require-imports rule

leeavital:
This disallows the use of require() style imports

````typescript
var lib = require("lib"); // bad

import {lib} from "lib"; // good
````

This works by reporting an error on every declaration where the right-hand-side is an call expression that calls `require`.

This addresses #388

### CODE REVIEW

File: test/rules/noRequireImportsTests.ts [hunk 1]
```diff
@@ -0,0 +1,37 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+describe("<no-require-imports>", () => {
+
```
adidahiya: remove this newline

test/rules/noRequireImportsTests.ts
```diff
@@ -0,0 +1,37 @@
...
+    it("forbids CommmonJS style imports", () => {
+        var fileName = "rules/norequireimports.test.ts";
+        var Rule = Lint.Test.getRule("no-require-imports");
+
```
adidahiya: remove newline

test/rules/noRequireImportsTests.ts
```diff
@@ -0,0 +1,37 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+describe("<no-require-imports>", () => {
+
+    it("forbids CommmonJS style imports", () => {
+        var fileName = "rules/norequireimports.test.ts";
```
adidahiya: use `const`

test/rules/noRequireImportsTests.ts
```diff
@@ -0,0 +1,37 @@
...
+
+        var actualFailures = Lint.Test.applyRuleOnFile(fileName, Rule);
+
+        assert(actualFailures.length === 4, "Expected four failures");
+        var failure1 = Lint.Test.createFailuresOnFile(fileName, Rule.FAILURE_STRING + "'lib'")([1, 5], [1, 25]);
+        var failure2 = Lint.Test.createFailuresOnFile(fileName, Rule.FAILURE_STRING + "'lib2'")([3, 5], [3, 27]);
+        var failure3 = Lint.Test.createFailuresOnFile(fileName, Rule.FAILURE_STRING + "'lib5'")([11, 5], [11, 27]);
+        var failure4 = Lint.Test.createFailuresOnFile(fileName, Rule.FAILURE_STRING + "'lib6'")([12, 5], [12, 27]);
+
+
```
adidahiya: extra newline --- this should be picked up by the linter; we should investigate why that's not the case.

test/files/rules/norequireimports.test.ts
```diff
@@ -0,0 +1,13 @@
+var lib = require('lib'); // error
+
+let lib2 = require('lib2'); // error
+
```
adidahiya: also add a test for syntax like this:

```
import foo = require('foo');
```

test/files/rules/norequireimports.test.ts
```diff
@@ -0,0 +1,13 @@
...
+import {l} from 'lib';
+
+var lib3 = load('not_an_import');
+
+var lib4 = lib2.subImport;
+
+var lib5 = require('lib5'),
+    lib6 = require('lib6'),
```
adidahiya: these are errors too, right? include `// error` comments consistently through the file

src/rules/noRequireImportsRule.ts
```diff
@@ -0,0 +1,48 @@
...
+export class Rule extends Lint.Rules.AbstractRule {
+    public static FAILURE_STRING = "require() style import is forbidden for module: ";
+
+    public apply(sourceFile: ts.SourceFile): Lint.RuleFailure[] {
+        return this.applyWithWalker(new NoRequireImportsWalker(sourceFile, this.getOptions()));
+    }
+}
+
+class NoRequireImportsWalker extends Lint.RuleWalker {
+
```
adidahiya: remove extra newline

src/rules/noRequireImportsRule.ts
```diff
@@ -0,0 +1,48 @@
...
+    public visitVariableStatement(node: ts.VariableStatement) {
+        var declarations : ts.VariableDeclaration[] = node.declarationList.declarations;
+        declarations.forEach(this.processDeclaration.bind(this));
```
adidahiya: we generally don't use `Function.prototype.bind` in TS because you lose out on some type checking.

perhaps you could use the new `for-of` syntax here? if not, then just expand this to:

```
declarations.forEach((declaration) => this.processDeclaration(declaration));
```

src/rules/noRequireImportsRule.ts
```diff
@@ -0,0 +1,48 @@
...
+
+    public apply(sourceFile: ts.SourceFile): Lint.RuleFailure[] {
+        return this.applyWithWalker(new NoRequireImportsWalker(sourceFile, this.getOptions()));
+    }
+}
+
+class NoRequireImportsWalker extends Lint.RuleWalker {
+
+    public visitVariableStatement(node: ts.VariableStatement) {
+        var declarations : ts.VariableDeclaration[] = node.declarationList.declarations;
```
adidahiya: is this type declaration necessary?

if it is, remove the space before the `:` symbol

src/rules/noRequireImportsRule.ts
```diff
@@ -0,0 +1,48 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+export class Rule extends Lint.Rules.AbstractRule {
+    public static FAILURE_STRING = "require() style import is forbidden for module: ";
```
adidahiya: I don't think you need to provide the name of the module they're trying to import in the error message. Imports are almost always grouped at the top of a file and the user will get position information about the failure anyway, so it should be easy to locate. Adding this feature is just more work for us.

src/rules/noRequireImportsRule.ts
```diff
@@ -0,0 +1,48 @@
...
+    public visitVariableStatement(node: ts.VariableStatement) {
+        var declarations : ts.VariableDeclaration[] = node.declarationList.declarations;
+        declarations.forEach(this.processDeclaration.bind(this));
+
+        super.visitVariableStatement(node);
+    }
+
+    private processDeclaration(decl: ts.VariableDeclaration) : void {
+        // make sure the RHS is a call expression.
+        var call = <ts.CallExpression>(decl.initializer);
```
adidahiya: 1. use `const` and `let` throughout file
2. include a space after the typecast `<ts.CallExpression> (decl.initializer);`

src/rules/noRequireImportsRule.ts
```diff
@@ -0,0 +1,48 @@
...
+        if (call && call.arguments && call.expression) {
+            var fnExpression : string = call.expression.getText(this.getSourceFile());
```
adidahiya: you probably don't need this `: string` type declaration

src/rules/noRequireImportsRule.ts
```diff
@@ -0,0 +1,48 @@
...
+            if (fnExpression === "require") {
+                var libName = call.arguments[0].getText(this.getSourceFile());
+                var errorString = Rule.FAILURE_STRING + libName;
+                var failure = this.createFailure(decl.getStart(), decl.getWidth(), errorString);
+                this.addFailure(failure);
```
adidahiya: inline the `failure` variable

src/rules/noRequireImportsRule.ts
```diff
@@ -0,0 +1,48 @@
...
+    }
+
+    private processDeclaration(decl: ts.VariableDeclaration) : void {
+        // make sure the RHS is a call expression.
+        var call = <ts.CallExpression>(decl.initializer);
+        if (call && call.arguments && call.expression) {
+            var fnExpression : string = call.expression.getText(this.getSourceFile());
+            if (fnExpression === "require") {
+                var libName = call.arguments[0].getText(this.getSourceFile());
+                var errorString = Rule.FAILURE_STRING + libName;
```
adidahiya: you won't need this variable once you remove "libName" from the reported error string

README.md
```diff
@@ -162,6 +162,7 @@ A sample configuration file with all options is available [here](https://github.
 * `no-duplicate-variable` disallows duplicate variable declarations.
 * `no-empty` disallows empty blocks.
 * `no-eval` disallows `eval` function invocations.
+* `"no-require-imports"` disallows require() style imports
```
adidahiya: no quotes around the rule identifier

adidahiya:
Adding commits to PRs rather than `--amend`ing seems to work better on Github, so let's try that for PRs moving forward.

One more task: update `tsconfig.json` files as necessary since you're adding new files to the src

leeavital:
I've made the fixes you recommended. it's probably best if I rebase before merging; we can avoid having 9 commits for a fairly small feature.

test/files/rules/norequireimports.test.ts
adidahiya: missed this one?

test/rules/noRequireImportsTests.ts
```diff
@@ -0,0 +1,37 @@
...
+describe("<no-require-imports>", () => {
+    it("forbids CommmonJS style imports", () => {
+        const fileName = "rules/norequireimports.test.ts";
+        const Rule = Lint.Test.getRule("no-require-imports");
+
+        const actualFailures = Lint.Test.applyRuleOnFile(fileName, Rule);
+
+        assert(actualFailures.length === 5, "Expected five failures");
+        let failure1 = Lint.Test.createFailuresOnFile(fileName, Rule.FAILURE_STRING)([1, 5], [1, 25]);
```
adidahiya: these can be `const` as well

test/rules/noRequireImportsTests.ts
```diff
@@ -0,0 +1,37 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+describe("<no-require-imports>", () => {
+    it("forbids CommmonJS style imports", () => {
+        const fileName = "rules/norequireimports.test.ts";
+        const Rule = Lint.Test.getRule("no-require-imports");
+
```
adidahiya: remove extra newline

src/rules/noRequireImportsRule.ts
```diff
@@ -0,0 +1,45 @@
...
+
+    public apply(sourceFile: ts.SourceFile): Lint.RuleFailure[] {
+        return this.applyWithWalker(new NoRequireImportsWalker(sourceFile, this.getOptions()));
+    }
+}
+
+class NoRequireImportsWalker extends Lint.RuleWalker {
+    public visitVariableStatement(node: ts.VariableStatement) {
+        let declarations = node.declarationList.declarations;
+        for (let decl of declarations) {
```
adidahiya: `const` on these 2 lines?

src/rules/noRequireImportsRule.ts
```diff
@@ -0,0 +1,45 @@
...
+            this.processDeclaration(decl);
+        }
+        super.visitVariableStatement(node);
+    }
+
+    private processDeclaration(decl: ts.VariableDeclaration) : void {
```
adidahiya: you can omit the `: void` here

src/rules/noRequireImportsRule.ts
```diff
@@ -0,0 +1,45 @@
...
+        // make sure the RHS is a call expression.
+        let call = <ts.CallExpression> (decl.initializer);
```
adidahiya: always prefer `const` when you can

src/rules/noRequireImportsRule.ts
```diff
@@ -0,0 +1,45 @@
...
+        if (call && call.arguments && call.expression) {
+            let fnExpression = call.expression.getText(this.getSourceFile());
+            if (fnExpression === "require") {
+                let errorString = Rule.FAILURE_STRING;
```
adidahiya: don't need this variable, just inline it

adidahiya:
sure, you can squash these commits after all the feedback is addressed

src/rules/noRequireImportsRule.ts
```diff
@@ -0,0 +1,49 @@
...
+        }
+        super.visitVariableStatement(node);
+    }
+
+    public visitImportEqualsDeclaration(node: ts.ImportEqualsDeclaration) {
+        this.addFailure(this.createFailure(node.getStart(), node.getWidth(), Rule.FAILURE_STRING));
+        super.visitImportEqualsDeclaration(node);
+    }
+
+    private processDeclaration(decl: ts.VariableDeclaration) : void {
```
adidahiya: in a few other places, we use the verb "handle", so let's make this `handleDeclaration`. also omit the return type

src/rules/noRequireImportsRule.ts
```diff
@@ -0,0 +1,49 @@
...
+        // make sure the RHS is a call expression.
+        const call = <ts.CallExpression> (decl.initializer);
+        if (call && call.arguments && call.expression) {
+            const fnExpression = call.expression.getText(this.getSourceFile());
```
adidahiya: let's rename this to `callExpressionText`

src/rules/tsconfig.json
```diff
@@ -43,6 +43,7 @@
         "./noDuplicateVariableRule.ts",
         "./noEmptyRule.ts",
         "./noEvalRule.ts",
+        "./noRequireImportsRule.js",
```
adidahiya: `.ts`, not `.js`

leeavital: Good catch! Thanks!

adidahiya:
:+1: looks good, let's squash and merge!

PR #435, opened 2015-06-16
Skip checking declared functions for no-unused-variable (fixes #434)

### LINKED ISSUE #434: unused variable error when declaring a function
declare function foo(): any;

this is causing a warning: "unused variable: 'foo'"

adidahiya:
- also update tsconfig project files


adidahiya:
@leeavital please review

leeavital:
Codewise this looks good, but I'm not so sure ignoring declared functions is inline with this rule. Ignoring exports makes sense, but isn't declaring a function and then not using it undesirable? 

no-unused-var will catch:
````typescript
function func3() : any {
    //
}
````

it should probably catch
````
declare func3(): any
````

Perhaps if makes sense to introduce a no-unused-declaration rule, but I think no-unused-var should just catch declarations.

adidahiya:
We made a similar change to skip checking declared vars recently (https://github.com/palantir/tslint/issues/190), so our checking of functions should work the same.

I agree that we should consider adding a `no-unused-declaration` rule that only checks declared vars and functions. That checking should be optional, hence the separate rule. Declaring vars/functions in one file and then using them in other files is apparently a TS pattern.

leeavital:
I'm not familiar with that pattern, but it sounds like we shouldn't be catching undeclared functions either. In that case the no-unused-variable rule needs to be modified further to ignore function declarations.

adidahiya:
Function declarations (NOT `declare foo(): ...`, but rather `function foo () { ... implementation ... }`) generally _should_ be used within a file and if they're not, they ought to be exported, no?

The idea of skipping a `declare foo(): ...` is that you want to declare a shared typing in one file which can then be referenced where that function is actually used.

leeavital:
Thanks, that clears things up. This LGTM.

Issue #137, opened 2014-05-08
Is it OK that tslint doesn't work without tslint.json?


thorn0:
To give this tool a try, I've just installed it globally and am trying to check a file with it. Here's what I'm getting. Not the most friendly first impression.

```
d:\dev\tsfoo\tsfoo\src>tslint -f app.ts

fs.js:427
  return binding.open(pathModule._makeLong(path), stringToFlags(flags), mode);
                 ^
Error: ENOENT, no such file or directory 'C:\Users\thorn\tslint.json'
    at Object.fs.openSync (fs.js:427:18)
    at Object.fs.readFileSync (fs.js:284:15)
    at Object.findConfiguration (C:\Users\thorn\AppData\Roaming\npm\node_modules\tslint\bin\tslint-cli.js:2758
5:47)
    at Object.<anonymous> (C:\Users\thorn\AppData\Roaming\npm\node_modules\tslint\bin\tslint-cli.js:68199:40)
    at Module._compile (module.js:456:26)
    at Object.Module._extensions..js (module.js:474:10)
    at Module.load (module.js:356:32)
    at Function.Module._load (module.js:312:12)
    at Module.require (module.js:364:17)
    at require (module.js:380:17)
    at Object.<anonymous> (C:\Users\thorn\AppData\Roaming\npm\node_modules\tslint\bin\tslint:3:1)
    at Module._compile (module.js:456:26)
    at Object.Module._extensions..js (module.js:474:10)
    at Module.load (module.js:356:32)
    at Function.Module._load (module.js:312:12)
    at Function.Module.runMain (module.js:497:10)
    at startup (node.js:119:16)
    at node.js:902:3
```

PR #437, opened 2015-06-19
tslint-cli uses a default configuration if none is found

leeavital:
The default options enabled are:

curly: true,
indent: [true, 4],
no-duplicate-key: true,
no-duplicate-variable: true,
no-empty: true,
no-eval: true,
no-trailing-whitespace: true,
no-unreachable: true,
no-use-before-declare: true,
quotemark: [true, "double"],
semicolon: true

### DIFF
--- a/src/configuration.ts
+++ b/src/configuration.ts
@@ -48,7 +48,11 @@ module Lint.Configuration {
 
         configFile = findup(CONFIG_FILENAME, { cwd: inputFileLocation, nocase: true }) || defaultPath;
 
-        return configFile ? JSON.parse(fs.readFileSync(configFile, "utf8")) : undefined;
+        if (fs.existsSync(configFile)) {
+            return JSON.parse(fs.readFileSync(configFile, "utf8"));
+        } else {
+            return getDefaultConfiguration();
+        }
     }
 
     function getHomeDir() {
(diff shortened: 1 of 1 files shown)


leeavital:
This fixes #137

src/configuration.ts
```diff
@@ -65,4 +69,22 @@ module Lint.Configuration {
             }
         }
     }
+
+    function getDefaultConfiguration(): any {
+        return {
```
adidahiya: let's just make this a `const`, right after `CONFIG_FILENAME` on L22. call it `DEFAULT_CONFIG`

adidahiya:
Should we notify the user somehow about the fact that we're using a default config in this case and they are encourage to supply their own config file?

ashwinr:
Not in stdout -- it'll break command-line piping.

leeavital:
I was thinking about using console.warn() and couldn't decide one way or
another. For what it's worth, eslint has no such warning.

ashwinr:
I’d just ignore it.

From: Lee Avital
Reply-To: palantir/tslint
Date: Friday, June 19, 2015 at 12:54 PM
To: palantir/tslint
Cc: Ashwin Ramaswamy
Subject: Re: [tslint] tslint-cli uses a default configuration if none is found (#437)

I was thinking about using console.warn() and couldn't decide one way or
another. For what it's worth, eslint has no such warning.

PR #440, opened 2015-06-19
Style improvements to no-var-keyword rule

adidahiya:
Lint.isBlockScopedVariable now supports VariableStatements.

### DIFF
--- a/src/language/utils.ts
+++ b/src/language/utils.ts
@@ -21,7 +21,7 @@ module Lint {
         const normalizedName = path.normalize(fileName).replace(/\\/g, "/");
         const compilerOptions = createCompilerOptions();
 
-        const compilerHost = {
+        const compilerHost: ts.CompilerHost = {
             getCanonicalFileName: (filename: string) => filename,
             getCurrentDirectory: () => "",
             getDefaultLibFileName: () => "lib.d.ts",
@@ -95,6 +95,19 @@ module Lint {
 
         return isNodeFlagSet(parentNode, ts.NodeFlags.Let)
             || isNodeFlagSet(parentNode, ts.NodeFlags.Const);
+    }
+
+    export function isBlockScopedBindingElement(node: ts.BindingElement): boolean {
+        let currentParent = node.parent;
+        while (currentParent.kind !== ts.SyntaxKind.VariableDeclaration) {
+            if (currentParent.parent == null) {
+                // if we didn't encounter a VariableDeclaration, this must be a function parameter, which is block scoped
+                return true;
+            } else {
+                currentParent = currentParent.parent;
+            }
+        }
+        return isBlockScopedVariable(<ts.VariableDeclaration> currentParent);
     }
 
     /**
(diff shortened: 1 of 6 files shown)


src/language/utils.ts
```diff
@@ -90,10 +90,16 @@ module Lint {
...
+    /**
+     * Determines if the appropriate bit in the parent (VariableDeclarationList) is set,
+     * which indicates this is a "let" or "const".
+     */
+    export function isBlockScopedVariable(node: ts.VariableDeclaration | ts.VariableStatement): boolean {
+        const flags = (node.kind === ts.SyntaxKind.VariableDeclaration)
+            ? (<ts.VariableDeclaration> node).parent.flags
+            : (<ts.VariableStatement> node).declarationList.flags;
+
+        return (Math.floor(flags / ts.NodeFlags.Let) % 2) === 1
```
leeavital: I've never seen this style of checking flags before. Is there a reason it's preferable to something like:

````
(flags & ts.NodeFlags.Let) || (flags & ts.NodeFlags.Const)
````

**Comment by leeavital** [comment]
LGTM, pending the flags thing.

File: src/language/utils.ts (hunk 1 above)
Comment (DickvdBrink) [inline src/language/utils.ts]: @leeavital, I believe It is because of the `noBitwiseRule`

Comment (adidahiya) [inline src/language/utils.ts]: to get this to typecheck & pass tslint, you have to do something like:

```ts
/* tslint:disable:no-bitwise */
return ((flags & ts.NodeFlags.Let) !== 0) || ((flags & ts.NodeFlags.Const) !== 0);
/* tslint:enable:no-bitwise */
```

is that preferable to the current implementation?

adidahiya:
I tried to improve the readability of `Lint.isBlockScopedVariable` by adding a private helper function.

adidahiya:
failing because of nested binding patterns... darn. will try and fix

adidahiya:
fixed now

test/files/rules/no-duplicate-variable.test.ts
```diff
@@ -151,4 +151,9 @@ function testDestructuring() {
 
     var [a2, [b2, c2]] = [1, [2, 3]];
     var [{a2, d2}] = [{a2: 1, d2: 4}]; // failure
+
+    function myFunc2([a, b]) {
+        var a; // not a failure; caught by no-shadowed-variable
```
leeavital: Shouldn't it be caught by both?

leeavital:
I'm not sure about the test case you added, other than that looks good.

test/files/rules/no-duplicate-variable.test.ts
adidahiya: hmm it looks like no-duplicate-variable has some issues in its current state... let's take that up in a separate PR since this one is already getting out of scope.

for example, line 98 in this file really ought to report a failure (and there's no such test for that syntax for `no-shadowed-variable`).

adidahiya: in addition: we don't want any overlap in our rules, otherwise errors would be reported twice. `no-duplicate-variable` now checks for duplicates in the same block scope only, rather than across scopes.

PR #442, opened 2015-06-21
Don't check shadowed variables in no-duplicate-variable rule

adidahiya:
Now that we have `no-shadowed-variable`, this functionality should be removed from `no-duplicate-variable`.


ashwinr:
:+1:

PR #443, opened 2015-06-21
Set up TSLint on VSCode

ashwinr:


ashwinr:
going to merge this as I have other commits on top. This only touches typings and tsconfig.json, so it's very straightforward.

src/tslint-cli.ts
```diff
@@ -14,6 +14,10 @@
  * limitations under the License.
  */
 
+/// <reference path="../typings/node.d.ts"/>
+/// <reference path="tslint.ts"/>
+/// <reference path="configuration.ts"/>
+
```
adidahiya: why are these reference paths needed?

ashwinr: because this is not hooked up to tsconfig

Issue #425, opened 2015-06-09
no-duplicate-variable does not handle destructuring


leeavital:
```typescript
var a : number = 1;
var [a, b] : number[] = [2,3];
````

Should warn about duplicating `a`.

adidahiya:
I have code for this somewhere in my fork of tslint. I'll investigate its status.

adidahiya:
yep, I've added destructuring support @ this branch: https://github.com/adidahiya/tslint/tree/destructuring-no-duplicate-variable

but it depends on a currently open PR #409

adidahiya:
sorry, will close when #452 merges

PR #444, opened 2015-06-21
Add destructuring support to no-duplicate-variable rule

adidahiya:
looks like we get nested binding patterns for free!


leeavital:
👍

adidahiya:
fixes #425

PR #445, opened 2015-06-21
shorten error string

ashwinr:
- the no-var-keyword error is too verbose, the convention
  elsewhere is "forbidden blah" (no-eval, no-bitwise) which is
  succint enough. detailed rule information is available on the
  readme.


adidahiya:
:+1:

ashwinr:
straightforward enough to merge.

PR #450, opened 2015-06-22
compile all files using the right tsconfig.json

ashwinr:
- tsc-wrapper.js is just a wrapper on top of tsc for VS Code,
  that walks up the dirtree until a tsconfig.json is found and
  compiles at that path. necessary for tasks.json configuration
- test compilation will also work with this, but there's an issue with
  test/tsconfig.json

### DIFF
--- a/.settings/settings.json
+++ b/.settings/settings.json
@@ -13,9 +13,9 @@
 		".git",
 		".tscache",
 		"bower_components",
-        "bin",
+                "bin",
 		"build",
-        "lib",
+                "lib",
 		"node_modules"
 	]
 }
--- a/.settings/settings.json
+++ b/.settings/settings.json
@@ -13,6 +13,7 @@
 		".git",
 		".tscache",
 		"bower_components",
+        "bin",
 		"build",
         "lib",
 		"node_modules"
--- a/scripts/tsc-wrapper.js
+++ b/scripts/tsc-wrapper.js
@@ -1,0 +1,38 @@
+#!/usr/bin/env node
+
+/* global process */
+
+var execSync = require("child_process").execSync;
+var findup = require("findup-sync");
+var path = require("path");
+var spawnSync = require("child_process").spawnSync;
+
+/* print a buffer to a stream */ 
+function print(buffer, stream) {
+    if (buffer != null) {
+        var output = buffer.toString("utf8");
+        stream.write(output);
+    }
+}
+
+var argv = process.argv;
+if (argv.length < 4) {
+    console.error("insufficient arguments to tsc wrapper");
+    process.exit(1);
+}
+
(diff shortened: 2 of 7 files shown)


.settings/settings.json
```diff
@@ -13,6 +13,7 @@
 		".git",
 		".tscache",
 		"bower_components",
+        "bin",
```
leeavital: indentation

scripts/tsc-wrapper.js
```diff
@@ -0,0 +1,38 @@
+#!/usr/bin/env node
```
leeavital: A comment describing the command line arguments would be very helpful here!

scripts/tsc-wrapper.js
```diff
@@ -0,0 +1,38 @@
...
+var filePath = argv[3];
+
+var configPath = path.dirname(findup("tsconfig.json", { cwd: filePath, nocase: true }));
+if (configPath == null) {
+    console.error("tsconfig.json not found");
+    process.exit(2);
+}
+
+var tscPath = path.join(workspacePath, "node_modules", "typescript", "bin", "tsc");
+var tsc = spawnSync(tscPath, ["-p", configPath]);
```
leeavital: Cool!

test/formatters/externalFormatterTest.ts
```diff
@@ -17,13 +17,14 @@
 describe("External Formatter", () => {
     var TEST_FILE = "formatters/externalFormatter.test.ts";
     var TEST_MODULE = "../test/files/formatters/simple";
-    var sourceFile, formatter;
+    var sourceFile: ts.SourceFile, formatter: Lint.IFormatter;
```
adidahiya: declare multiple vars on new lines.
also use `const` when possible

adidahiya:
is this because VSCode only supports one `tsconfig.json` file per project?

what's the issue with `test/tsconfig.json`? is it fixed by this PR?

test/formatters/externalFormatterTest.ts
ashwinr: this was just changing existing code, but sure

ashwinr:
VSCode can support multiple configs, but it can't look up the dir-tree for the closest `tsconfig.json`. `test/tsconfig.json` is broken somehow -- try running `tsc -p test` to see the errors. I haven't had the time to investigate, so would love it if one of you could.

test/formatters/externalFormatterTest.ts
adidahiya: ok, don't worry about it then; i'm doing some cleanup in #454 anyway

adidahiya:
LGTM :+1:

PR #452, opened 2015-06-22
Add destructuring support to no-shadowed-variable rule

adidahiya:


src/rules/noShadowedVariableRule.ts
```diff
@@ -31,15 +31,37 @@ class NoShadowedVariableWalker extends Lint.BlockScopeAwareRuleWalker<ScopeInfo,
         return new ScopeInfo();
     }
 
+    public visitBindingElement(node: ts.BindingElement) {
+        // TODO: handle node.dotDotDotToken?
+        var isSingleVariable = node.name.kind === ts.SyntaxKind.Identifier;
```
leeavital: I think this is worth extracting into a utility method. I've seen it at least four times in your recent PRs.

adidahiya: which part? just this one-line check for syntax kind? I disagree; having too many utilities is bad and it's pretty readable here as-is

adidahiya:
waiting on #440 so I can rebase and get `Lint.isBlockScopedBindingElement`

src/rules/noShadowedVariableRule.ts
```diff
@@ -50,38 +72,35 @@ class NoShadowedVariableWalker extends Lint.BlockScopeAwareRuleWalker<ScopeInfo,
...
-    public visitVariableDeclaration(node: ts.VariableDeclaration) {
-        const propertyName = <ts.Identifier> node.name;
-        const variableName = propertyName.text;
+    private handleSingleVariableIdentifier(variableIdentifier: ts.Identifier, isBlockScoped: boolean) {
+        const variableName = variableIdentifier.text;
         const currentScope = this.getCurrentScope();
         const currentBlockScope = this.getCurrentBlockScope();
 
-        // this var is shadowing if there's already a var of the same name in any available scope AND
+        // this const is shadowing if there's already a const of the same name in any available scope AND
```
adidahiya: ugh these should be const

adidahiya:
ready for review

adidahiya:
@leeavital ping for review

test/files/rules/no-shadowed-variable.test.ts
```diff
@@ -75,3 +75,23 @@ function testDoStatement() {
         let foo = 2; // failure
     } while (true);
 }
+
+function testDestructuring() {
+    var x, y, z;
```
leeavital: Can you add some destructuring bindings to make sure they get added to the scope? e.g.

````typescript
var x;
var {y, z} = {y: 2, z: 3}
````

leeavital:
Other than the test, this looks good.

test/files/rules/no-shadowed-variable.test.ts
adidahiya: good call

PR #449, opened 2015-06-22
small cleanup

ashwinr:
- redirect bin verification output to /dev/null
- remove unnecessary rule disabling comments
- remove strange swp file

PR #462, opened 2015-06-24
Remove -f option from CLI

### LINKED ISSUE #438: Remove "-f" to specify files in CLI
It's unintuitive to have to use the `-f` flag to specify a file to lint, so let's remove that.

leeavital:
Resolves #438 

I also cleaned up and added a relevant test to test/check-bin.sh.

[Review on Reviewable](https://reviewable.io/reviews/palantir/tslint/462)

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,12 +26,11 @@ Please first ensure that the TypeScript source files compile correctly.
 
 ##### CLI
 
-  usage: `tslint`
+  usage: `tslint [options...] [file ...]`
 
   Options:
 
     -c, --config              configuration file
-    -f, --file                file to lint                  [required]
     -o, --out                 output file
     -r, --rules-dir           rules directory
     -s, --formatters-dir      formatters directory
@@ -40,10 +39,6 @@ Please first ensure that the TypeScript source files compile correctly.
 By default, configuration is loaded from `tslint.json`, if it exists in the current path, or the user's home directory, in that order.
 
 tslint accepts the following commandline options:
-
-    -f, --file:
-        The location of the TypeScript file that you wish to lint. This
-        option is required. Multiple files are processed consecutively.
 
     -c, --config:
         The location of the configuration file that tslint will use to
--- a/src/tslint-cli.ts
+++ b/src/tslint-cli.ts
@@ -22,19 +22,19 @@ const fs = require("fs");
 const optimist = require("optimist")
     .usage("usage: $0")
     .check((argv: any) => {
-        // at least one of file or help or version must be present
-        if (!(argv.f || argv.h || argv.v)) {
-            throw "Missing required arguments: f";
+        // at least one of file, help, version or unqualified argument must be present
+        if (!(argv.h || argv.v || argv._.length > 0)) {
+            throw "Missing files";
+        }
+
+        if (argv.f) {
+            throw "-f option is no longer available. Supply files directly to the tslint command instead.";
         }
     })
     .options({
         "c": {
             alias: "config",
             describe: "configuration file"
-        },
-        "f": {
-            alias: "file",
-            describe: "file to lint"
         },
         "h": {
             alias: "help",
@@ -84,10 +84,6 @@ if ("help" in argv) {
     outputStream.write(optimist.help());
     const outputString = `
 tslint accepts the following commandline options:
-
-    -f, --file:
-        The location of the TypeScript file that you wish to lint. This
-        option is required. Muliptle files are processed consecutively.
 
     -c, --config:
         The location of the configuration file that tslint will use to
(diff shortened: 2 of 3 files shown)


src/tslint-cli.ts
```diff
@@ -175,14 +175,14 @@ const processFile = (file: string) => {
...
 
-if (typeof fileOrFiles === "string") {
-    processFile(fileOrFiles);
-} else {
-    for (const ix in fileOrFiles) {
-        if (fileOrFiles.hasOwnProperty(ix)) {
-            processFile(fileOrFiles[ix]);
-        }
-    }
+for (const file of files) {
```
adidahiya: use `let`

This feels like a bug in the TS compiler? `file` shouldn't be allowed to be a `const` since its value changes on every iteration of the loop...

adidahiya:
Sorry, I take that comment back about `const`... I see why it's allowed now. [Babel](https://babeljs.io/) also allows `const` in for-of loop var declarations. At least it prevents you from re-assigning `file` within the loop body.

README.md
```diff
@@ -26,12 +26,12 @@ Please first ensure that the TypeScript source files compile correctly.
 
 ##### CLI
 
-  usage: `tslint`
+  usage: `tslint -cforst [file ...]`
```
adidahiya: let's follow `tsc`'s CLI:

```
usage: `tslint [options] [file ...]`
```

README.md
```diff
@@ -42,8 +42,9 @@ By default, configuration is loaded from `tslint.json`, if it exists in the curr
 tslint accepts the following commandline options:
 
     -f, --file:
-        The location of the TypeScript file that you wish to lint. This
-        option is required. Muliptle files are processed consecutively.
+        The location of the TypeScript file that you wish to lint. Multiple
+        files are processed consecutively. Files can also be tacked on to the
+        end of the command.
```
adidahiya: Let's just remove this option entirely. No point in having multiple ways to specify files.

src/tslint-cli.ts
```diff
@@ -22,8 +22,8 @@ const fs = require("fs");
 const optimist = require("optimist")
     .usage("usage: $0")
     .check((argv: any) => {
-        // at least one of file or help or version must be present
-        if (!(argv.f || argv.h || argv.v)) {
+        // at least one of file, help, version or unqualified argument must be present
+        if (!(argv.f || argv.h || argv.v || argv._.length > 0)) {
```
adidahiya: let's remove `argv.f` here and change the error message to "Missing file(s) to lint"

src/tslint-cli.ts
```diff
@@ -175,14 +175,14 @@ const processFile = (file: string) => {
...
 };
 
-const fileOrFiles = argv.f;
+let files = [];
+if (argv.f instanceof Array) {
+    files = files.concat(argv.f);
+} else if (typeof argv.f === "string") {
+    files.push(argv.f);
+}
+files = files.concat(argv._);
```
adidahiya: this whole section can be collapsed to

```ts
const files = argv._;
```

test/check-bin.sh
```diff
@@ -12,35 +12,42 @@
...
-if [ $out != 0 ]
-then
-  echo "tslint with a good file did not exit with a 0, got $out"
-  exit 1
-fi
-
+expectOut $? 0 "tslint with a good file did not exit correctly"
 
 # make sure calling tslint with a bad file exits correctly
 ./bin/tslint -f test/files/rules/ban.test.ts -c tslint.json
```
adidahiya: remove this test case when you remove `-f`

test/check-bin.sh
```diff
@@ -12,35 +12,42 @@
 # See the License for the specific language governing permissions and
```
adidahiya: heads up, this fill is going to conflict with #449

README.md
leeavital: Would that require us to bump to version 2? It's technically breaking the API.

adidahiya: hmm you're right, that's a fairly breaking change.

@ashwinr @gscshoyru if we do decide to remove the `-f` option, should we bump the major version? Or is a minor version bump sufficient?

Until then, let's add some sort of deprecation notice for this option (I guess most of my other comments don't apply anymore)

ashwinr: a minor version bump is sufficient, the CLI is pretty basic and I anticipate most people using tslint via grunt/gulp wrappers anyway. let's make sure to call it out in the changelog. +1 for deprecation notice.

ashwinr:
:+1: leaving it up to @adidahiya to review and merge

adidahiya:
Looks good, but the commit should be titled "Make -f optional in CLI", not "Remove -f option".

src/tslint-cli.ts
```diff
@@ -22,9 +22,9 @@ const fs = require("fs");
 const optimist = require("optimist")
     .usage("usage: $0")
     .check((argv: any) => {
-        // at least one of file or help or version must be present
-        if (!(argv.f || argv.h || argv.v)) {
-            throw "Missing required arguments: f";
+        // at least one of file, help, version or unqualified argument must be present
```
adidahiya: If we're removing `-f` altogether, let's add a deprecation notice here?
Check for `argv.f` and `throw "-f option is deprecated. Supply files directly to the tslint command instead."`

PR #464, opened 2015-06-24
Declare 'tslint' module

### LINKED ISSUE #460: tslint.d.ts doesn't provide an external module
I'm writing a custom rule in TypeScript, so I need to provide tsc with the typings for superclasses such as `Lint.Rules.AbstractRule`.
The distributed .d.ts file has internal modules, eg.
```ts
declare module Lint.Rules {
}
```
So when I try the obvious thing,
```ts
/// <reference path="../../node_modules/tslint/lib/tslint.d.ts" />
}
```
I get error
`tslint/functionHasReturnTypeRule.ts(7,27): error TS2503: Cannot find namespace 'Lint'.`

IIUC the .d.ts also needs to re-export as an external module, like 
```ts
declare module "tslint" {
  export = Lint;
}
```

adidahiya:
Fixes #460

### DIFF
--- a/lib/tslint.d.ts
+++ b/lib/tslint.d.ts
@@ -258,3 +258,6 @@ declare module Lint {
         private containsRule(rules, rule);
     }
 }
+declare module "tslint" {
+    export = Lint;
+}
--- a/src/tslint.ts
+++ b/src/tslint.ts
@@ -110,3 +110,7 @@ global.ts = ts;
 module.exports = Lint.Linter;
 // also export Lint.Configuration.findConfiguration so implementers can consume
 module.exports.findConfiguration = Lint.Configuration.findConfiguration;
+
+declare module "tslint" {
+    export = Lint;
+}


leeavital:
(y)

adidahiya:
lol I think you want ": +1 :" (without the spaces)

alexeagle:
In my issue, I suggested this as a fix, but had not tested it first.

Do you have any example of using tslint from TypeScript? I see that grunt-tslint and gulp-tslint are both written in JS. I also have an issue that the .d.ts refers to typescript API's with the `ts` prefix, but doesn't declare a dependency on it. I see that `tslint.ts` has
```global.ts = ts;```
which makes me think this can only work by compiling along with the tslint sources, not as a standalone external library.

**Comment by adidahiya** [comment]
You shouldn't have to compile with the TS source, but you need typings for TS language services.

After declaring the module, I tried out code that looked like this and it seemed to compile:

```ts
/// <reference path="lib/tslint.d.ts" />
/// <reference path="typings/typescriptServices.d.ts" />
/// <reference path="typings/typescriptServicesScanner.d.ts" />

class Rule extends Lint.Rules.AbstractRule {
}
```

alexeagle:
Yes, indeed I was referencing `typescript.d.ts` and not `typescriptServices.d.ts`.
So my original issue was a red herring. Sorry for the trouble! Could rollback 7a503a0feea2cd7dd2b0da8a3202c8e92d22d25a if you want.

I very much enjoy your RuleWalker visitor pattern.

adidahiya:
Hmm, your error was that `tsc` couldn't find the namespace `Lint`, so I figured `declare module "tslint"` was the solution... but the code I pasted above seems to work without it. Yes, I think I'll roll back this change since it's not necessary.

PR #468, opened 2015-06-25
Add support for ES6 bare imports

### LINKED ISSUE #451: no-unused-variable rule crashes on ES6 bare imports
Hie everyone,

I'm using tslint with grunt-tslint and i got an issue when i run it, since I moved on ES6 external module. I got this message : 
 
`
Fatal error: Cannot read property 'name' of undefined
`

Can I get some extra output with special parameter ? 
Anyoby have an idea ?

Tslint worked well when i was using internal module with 'module' keyword.

adidahiya:
Fixes #451

This affects the following rules:

- `no-unused-variable`
- `no-use-before-declare`
- `whitespace` (it was already handled here, but I added a regression test)

[Review on Reviewable](https://reviewable.io/reviews/palantir/tslint/468)

### DIFF
--- a/src/rules/noUnusedVariableRule.ts
+++ b/src/rules/noUnusedVariableRule.ts
@@ -57,7 +57,8 @@ class NoUnusedVariablesWalker extends Lint.RuleWalker {
             const importClause = node.importClause;
 
             // named imports & namespace imports handled by other walker methods
-            if (importClause.name != null) {
+            // importClause will be null for bare imports
+            if (importClause != null && importClause.name != null) {
                 const variableIdentifier = importClause.name;
                 this.validateReferencesForVariable(variableIdentifier.text, variableIdentifier.getStart());
             }
--- a/src/rules/noUseBeforeDeclareRule.ts
+++ b/src/rules/noUseBeforeDeclareRule.ts
@@ -42,14 +42,13 @@ class NoUseBeforeDeclareWalker extends Lint.ScopeAwareRuleWalker<VisitedVariable
     }
 
     public visitImportDeclaration(node: ts.ImportDeclaration) {
-        if (node.importClause != null) {
-            const importClause = node.importClause;
+        const importClause = node.importClause;
 
-            // named imports & namespace imports handled by other walker methods
-            if (importClause.name != null) {
-                const variableIdentifier = importClause.name;
-                this.validateUsageForVariable(variableIdentifier.text, variableIdentifier.getStart());
-            }
+        // named imports & namespace imports handled by other walker methods
+        // importClause will be null for bare imports
+        if (importClause != null && importClause.name != null) {
+            const variableIdentifier = importClause.name;
+            this.validateUsageForVariable(variableIdentifier.text, variableIdentifier.getStart());
         }
 
         super.visitImportDeclaration(node);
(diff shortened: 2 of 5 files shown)


src/rules/noUseBeforeDeclareRule.ts
```diff
@@ -42,14 +42,13 @@ class NoUseBeforeDeclareWalker extends Lint.ScopeAwareRuleWalker<VisitedVariable
...
 
-            // named imports & namespace imports handled by other walker methods
-            if (importClause.name != null) {
-                const variableIdentifier = importClause.name;
-                this.validateUsageForVariable(variableIdentifier.text, variableIdentifier.getStart());
-            }
+        // Named imports & namespace imports handled by other walker methods.
+        // importClause will be null for bare imports.
+        if (importClause != null && importClause.name != null) {
+            const variableIdentifier = importClause.name;
```
ashwinr: it's inappropriate to use `const` for these variables. we have no way of enforcing these -- someone can still do `variableIdentifier.setRandomValue(123)` and TypeScript wouldn't know. we should be reserving `const` for literal values only.

adidahiya: Doesn't matter though. I'm strongly in favor of using the strictest specifier whenever possible for variables. It signals to the developer that the variable binding should never change, and devs are expected to know JS/TS semantics enough to know that the values inside that object might change.

adidahiya: Also, that's unrelated to this change. Bring it up in a separate PR.

ashwinr:
Review status: 0 of 5 files reviewed at latest revision, 2 unresolved discussions, all commit checks successful.

---

<sup>**[src/rules/noUnusedVariableRule.ts, line 59 \[r1\]](https://reviewable.io:443/reviews/palantir/tslint/468#-JskNFh__iTWT0CcAorm)** ([raw file](https://github.com/palantir/tslint/blob/21ad3d55cc2d3285550ade55cb8167c6198765f0/src/rules/noUnusedVariableRule.ts#L59)):</sup>
change this back to what it was and same thing down below. single-line comments start with a lower-case as a style convention. i'm going to change all other single-line comments to start lowercase

---

<sup>**[src/rules/noUseBeforeDeclareRule.ts, line 50 \[r1\]](https://reviewable.io:443/reviews/palantir/tslint/468#-JskN1z4AzErre0g51eE-r1-50)** ([raw file](https://github.com/palantir/tslint/blob/21ad3d55cc2d3285550ade55cb8167c6198765f0/src/rules/noUseBeforeDeclareRule.ts#L50)):</sup>
disagree. let's take it up in a separate issue.

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/palantir/tslint/468)

src/rules/noUnusedVariableRule.ts
```diff
@@ -56,8 +56,9 @@ class NoUnusedVariablesWalker extends Lint.RuleWalker {
         if (!Lint.hasModifier(node.modifiers, ts.SyntaxKind.ExportKeyword)) {
             const importClause = node.importClause;
 
-            // named imports & namespace imports handled by other walker methods
-            if (importClause.name != null) {
+            // Named imports & namespace imports handled by other walker methods.
```
ashwinr: fixed all code comments in https://github.com/palantir/tslint/pull/470. revert this comment to its original format. same thing in noUseBeforeDeclareRule.

adidahiya:
addressed feedback and --amended

ashwinr:
:+1: waiting for builds to finish

ashwinr:
why do builds take so long

adidahiya:
TravisCI has been pretty slow lately. :snail:

PR #477, opened 2015-06-30
Add support for a `null` formatter

gtanner:
When using tslint as a library at bitHound, I would like to use the raw object as
the output.

Expose the raw output as well as add a null formatter that just returns
null (since in my case I will not be using the output).

I can live without the nullFormatter, but having access to the raw
output would be awesome for library users like me since I am just going to `json.parse` the output anyways.

[Review on Reviewable](https://reviewable.io/reviews/palantir/tslint/477)


src/tslint.ts
```diff
@@ -22,6 +22,7 @@ module Lint {
         failureCount: number;
         format: string;
         output: string;
+        raw: RuleFailure[];
```
adidahiya: let's call this property `failures`

adidahiya:
hmm I'm not sure we need the null formatter. let's remove it and just expose the raw failures array for your use case.

gtanner:
Thanks @adidahiya, removed the null formatter and renamed the property

src/tslint.ts
```diff
@@ -84,7 +85,8 @@ module Lint {
             return {
                 failureCount: failures.length,
                 format: this.options.formatter,
-                output: output
+                output: output,
+                failures: failures
```
adidahiya: we now alphabetize keys in object literals, and this is enforced by tslint, which is why your build is failing

adidahiya:
thanks @gtanner!

PR #486, opened 2015-07-02
Resolve #471, add tests for spread operator

### LINKED ISSUE #471: Should node.dotDotToken be handled in visitBindingElement?
For rules `noShadowedVariableRule` and `noDuplicateVariableRule`

adidahiya:
Turns out we don't need to look at node.dotDotToken for BindingElements.
It's literally just a text token. We don't need it for any linting rule
right now, but might in the future.

[Review on Reviewable](https://reviewable.io/reviews/palantir/tslint/486)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,4 @@
 language: node_js
 node_js:
     - "0.10"
-    - "0.12.3"
+    - "0.12"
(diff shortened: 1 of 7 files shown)


adidahiya:
ping for review @leeavital

.travis.yml
```diff
@@ -1,4 +1,4 @@
 language: node_js
 node_js:
     - "0.10"
-    - "0.12.3"
+    - "0.12"
```
leeavital: Why are we removing the minor version?

adidahiya: I believe that if the two node versions specified here have different specificity levels, travis can't figure out which version to use. This is what happened in the failed build in the previous commit.

leeavital:
:+1:

Issue #465, opened 2015-06-24
Switch to ES6 modules in tslint source


adidahiya:
Rather than TS internal modules / namespaces, let's try to switch to ES6 modules.

This would also fix #280.

ashwinr:
You can't use ES6 modules, this runs on node.

Sent from a phone

On Jun 24, 2015, at 18:07, Adi Dahiya <notifications@github.com<mailto:notifications@github.com>> wrote:


Rather than TS internal modules / namespaces, let's try to switch to ES6 modules.

This would also fix #280<https://urldefense.proofpoint.com/v2/url?u=https-3A__github.com_palantir_tslint_issues_280&d=BQMCaQ&c=izlc9mHr637UR4lpLEZLFFS3Vn2UXBrZ4tFb6oOnmz8&r=jafYnmwAIy7b95Z05TKJr1hFAxZozsv6_JDDlWyuJk8&m=A1zEfDJcMFAw9wcFb7eDHMq1wWD8s9bOkUB1J3pxUtg&s=8FIxA4rLx2aGdgt4LcvNVMOjcnmtNaa8gClujchFJpk&e=>.

adidahiya:
from https://github.com/Microsoft/TypeScript/issues/2242
> TypeScript supports down-level compilation of external modules using the new ES6 syntax.

ashwinr:
cool!

PR #495, opened 2015-07-14
Fixed small unwanted config

gilamran:
[Review on Reviewable](https://reviewable.io/reviews/palantir/tslint/495)


adidahiya:
Those are all required parts in the config files, which is why your build is failing. We don't have automatic source dependency ordering because we use internal TS modules. This ought to be fixed when we move to es6 module syntax (#465).

PR #508, opened 2015-07-17
Release v2.4.0

adidahiya:
- Update typings
- Fix tsconfig.json build output paths
- Add "concat-and-test" grunt task for running tests after building from an editor like Atom
- Bump to v2.4.0
- Use TS v1.5.3 release
- Update changelog
- Update npm-shrinkwrap.json

[Review on Reviewable](https://reviewable.io/reviews/palantir/tslint/508)

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "tslint",
-  "version": "2.3.1-beta",
+  "version": "2.4.0",
   "description": "a static analysis linter for TypeScript",
   "bin": {
     "tslint": "./bin/tslint"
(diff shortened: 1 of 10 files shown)


adidahiya:
Before we merge this PR, we should wait for TS to release 1.5 so we can point to the npm release in our package.json.

package.json
```diff
@@ -36,7 +36,7 @@
     "grunt-tslint": "^2.3.1-beta",
     "mocha": "^2.2.5",
     "tslint": "^2.3.1-beta",
-    "typescript": "1.5.0-beta"
+    "typescript": "git://github.com/Microsoft/Typescript.git#release-1.5"
```
adidahiya: we also need to bump our own version in this file

PR #510, opened 2015-07-17
Fix false positive on no-switch-case-fall-through

### LINKED ISSUE #509: `no-switch-case-fall-through` fails when `continue`ing outer loop
The following code fails to check, even though it clearly breaks from the switch-case:

```ts
loop:
while (condition) {
  switch (foo) {
  case bar:
    continue loop; // Line with errors
  default:
  }
}
```

The output I got (with every option except for that one above explicitly disabled):

```
tmp.ts[5, 19]: expected a 'break' before 'case'
tmp.ts[5, 19]: expected a 'break' before 'default'
```

It should be able to infer that it's clearly breaking a parent loop. This is with the current npm CLI version.

leeavital:
Fix #509

NoSwitchCaseFallThroughWalker didn't realize continue statements
would break out of case block.

[Review on Reviewable](https://reviewable.io/reviews/palantir/tslint/510)

### DIFF
--- a/test/files/rules/switchdefault.test.ts
+++ b/test/files/rules/switchdefault.test.ts
@@ -47,3 +47,13 @@ case 2:
     break;
 }
 
+// valid
+baz:
+while (true){
+    switch(foo) {
+    case 1:
+        bar();
+        continue baz;
+    default:
+    }
+}
--- a/src/rules/noSwitchCaseFallThroughRule.ts
+++ b/src/rules/noSwitchCaseFallThroughRule.ts
@@ -72,7 +72,8 @@ export class NoSwitchCaseFallThroughWalker extends Lint.RuleWalker {
         return !statements.some((statement) => {
             return statement.kind === ts.SyntaxKind.BreakStatement
                 || statement.kind === ts.SyntaxKind.ThrowStatement
-                || statement.kind === ts.SyntaxKind.ReturnStatement;
+                || statement.kind === ts.SyntaxKind.ReturnStatement
+                || statement.kind === ts.SyntaxKind.ContinueStatement;
         });
     }
 }


test/files/rules/switchdefault.test.ts
```diff
@@ -47,3 +47,13 @@ case 2:
...
 }
 
+// valid
+baz:
+switch(foo) {
+case 1:
+    bar();
+    continue baz;
+default:
+    break;
```
adidahiya: what happens if you remove this line, like in the example in #509 ?

leeavital:
Review status: 0 of 2 files reviewed at latest revision, 1 unresolved discussion, some commit checks pending.

---

<sup>**[test/files/rules/switchdefault.test.ts, line 57 \[r1\]](https://reviewable.io:443/reviews/palantir/tslint/510#-JuRSqTEGxEEfc8gxaPA-r1-57)** ([raw file](https://github.com/palantir/tslint/blob/45e852081cae6e472d7a1e384ebf47778c34ae9e/test/files/rules/switchdefault.test.ts#L57)):</sup>
It still passes, I've update the test case to look more like the example.

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/palantir/tslint/510)

adidahiya:
Reviewed 1 of 2 files at r1, 1 of 1 files at r2.
Review status: :shipit: all files reviewed at latest revision, all discussions resolved, all commit checks successful.

---



Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/palantir/tslint/510)

DickvdBrink:
Meh, sorry I forgot about `continue` when I created this. Thanks for fixing! :) :+1:

Issue #513, opened 2015-07-20
rule suggestion: no-internal-module


vvakame:
https://github.com/Microsoft/TypeScript/issues/2159
TypeScript 1.5.3 released. it has `namespace` keyword.
internal module is confusion name. I think use `namespace` instead of internal module.

if I create the pull request, will you accept it?

ashwinr:
yes please. let's call the rule name `no-module-keyword`.

vvakame:
@ashwinr yeah! I have one question. external module syntax are required in es6 module context.
like this.

```
declare module "foobar" {
	export default function (): string;
}
```

can I use `no-module-keyword` ?

ashwinr:
But that's a module declaration and not a module definition. We should only be forbidding module definitions using the `module` keyword. Maybe the rule name should be `no-module-definition`

adidahiya:
how about `no-internal-module`? I think "module definition" may be confusing / require lookup in our documentation

ashwinr:
sgtm.

PR #517, opened 2015-07-22
Add no-internal-module rule

vvakame:
implement #513

[Review on Reviewable](https://reviewable.io/reviews/palantir/tslint/517)

### DIFF
--- a/src/rules/noInternalModuleRule.ts
+++ b/src/rules/noInternalModuleRule.ts
@@ -24,7 +24,6 @@ export class Rule extends Lint.Rules.AbstractRule {
 
 class NoInternalModuleWalker extends Lint.RuleWalker {
     public visitModuleDeclaration(node: ts.ModuleDeclaration) {
-        let name = node.name.getText();
         if (Lint.isNodeFlagSet(node, ts.NodeFlags.Namespace)) {
             // ok namespace
         } else if (node.name.kind === ts.SyntaxKind.Identifier) {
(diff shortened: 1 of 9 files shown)


src/rules/noInternalModuleRule.ts
```diff
@@ -0,0 +1,37 @@
...
+    public static FAILURE_STRING = "forbidden internal module";
+
+    public apply(sourceFile: ts.SourceFile): Lint.RuleFailure[] {
+        return this.applyWithWalker(new NoInternalModuleWalker(sourceFile, this.getOptions()));
+    }
+}
+
+class NoInternalModuleWalker extends Lint.RuleWalker {
+    public visitModuleDeclaration(node: ts.ModuleDeclaration) {
+        let name = node.name.getText();
```
adidahiya: I don't think you should look at the raw text here. The type of `node.name` is `Identifier | LiteralExpression`. So you should be able to simplify this to:

```ts
        if (Lint.isNodeFlagSet(node, ts.NodeFlags.Namespace)) {
            // ok namespace
        } else if (node.name.syntaxKind === ts.SyntaxKind.Identifier) {
            // for external modules, node.name will be a LiteralExpression instead of Identifier
        }
```

vvakame: cool! I added a commit.

adidahiya: This `name` variable is no longer used. Linting should've picked that up, maybe it's a bug... Regardless, let's remove the ` let name...`

vvakame: oops, sorry!

adidahiya:
sweet, thanks!

Issue #474, opened 2015-06-27
Use typescript nightly package


ashwinr:
This is to keep TSLint up to date with latest TypeScript

adidahiya:
update: the 'publish-nightly' task has merged in the TS repo: https://github.com/Microsoft/TypeScript/pull/3875

adidahiya:
http://blogs.msdn.com/b/typescript/archive/2015/07/27/introducing-typescript-nightlies.aspx

PR #533, opened 2015-07-28
Track latest typescript in master (#474)

gscshoyru:
createScanner's signature changed, updated everything that called it
to reflect that.
Also had to add createScanner to our typings for some reason.
Note that we're still going to want to bump the typescript version manually
All currently existing tests work, but this won't work with new features in
1.6. Yet.

### DIFF
--- a/package.json
+++ b/package.json
@@ -36,7 +36,7 @@
     "grunt-tslint": "^2.3.1-beta",
     "mocha": "^2.2.5",
     "tslint": "^2.3.1-beta",
-    "typescript": "1.5.3"
+    "typescript": "1.6.0-dev.20150728"
   },
   "license": "Apache-2.0"
 }
--- a/package.json
+++ b/package.json
@@ -36,7 +36,7 @@
     "grunt-tslint": "^2.3.1-beta",
     "mocha": "^2.2.5",
     "tslint": "^2.3.1-beta",
-    "typescript": "1.6.0-dev.20150728"
+    "typescript": "next"
   },
   "license": "Apache-2.0"
 }
--- a/src/enableDisableRules.ts
+++ b/src/enableDisableRules.ts
@@ -20,7 +20,8 @@ module Lint {
 
         public visitSourceFile(node: ts.SourceFile) {
             super.visitSourceFile(node);
-            Lint.scanAllTokens(ts.createScanner(ts.ScriptTarget.ES5, false, node.text), (scanner: ts.Scanner) => {
+            const scan = ts.createScanner(ts.ScriptTarget.ES5, false, ts.LanguageVariant.Standard, node.text);
+            Lint.scanAllTokens(scan, (scanner: ts.Scanner) => {
                 const startPos = scanner.getStartPos();
                 if (this.tokensToSkipStartEndMap[startPos] != null) {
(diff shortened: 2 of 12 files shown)


package.json
```diff
@@ -36,7 +36,7 @@
     "grunt-tslint": "^2.3.1-beta",
     "mocha": "^2.2.5",
     "tslint": "^2.3.1-beta",
-    "typescript": "1.5.3"
+    "typescript": "1.6.0-dev.20150728"
```
adidahiya: let's use the `next` tag

adidahiya:
Looks like we have a few options on how to track the latest compiler:

1. _Only_ track the `next` distribution tag (try `npm view typescript` in your console to see what I mean). Any time a dev runs `npm install` or our CI tests run, it picks up the latest compiler and any breaks will show in tests. This will require us to follow the latest compiler more closely. tslint users will always have the latest compiler bundled with the linter (this shouldn't be a problem since the syntax is backwards-compatible and we haven't given users the option to use a different compiler version anyway).
2. Track the `next` distribution tag on a separate branch (like `develop`) and publish two npm distributions, just like TypeScript does. This creates more overhead for us since we have to maintain two branches & distributions.

I'm leaning towards option 1. Thoughts @ashwinr ?

ashwinr:
Does option 1 mean that tslint won't publish new versions to npm? Will we expect users to specify the tslint repo as their url in package.json?

adidahiya:
No, we will still publish new versions, they will just bundle in the latest nightly compiler.

ashwinr:
How will downstream users know that a new version gets them an unstable compiler version?

adidahiya:
Actually, it's not that much overhead to support option 2. We don't need 2 branches.

Our `master` branch will use `typescript@next` and we can tag releases that use this compiler with `-beta` (for example, `v2.5.0-beta.1`, ...). We can attach the same `dist-tag` that TS uses (`next`) to this line of releases. When TS 1.6 is out, we'll remove the `-beta`.

I doubt there will be more releases on the TS 1.5.x line. If there are, we can start a branch from our v2.4.0 release and make any changes / releases needed on that line.

For tslint users this means:

- `tslint@latest` gives them v2.4.0, which uses the latest _stable_ compiler
- `tslint@next` gives them v2.5.0-beta.*, which uses the latest nightly compiler

If that all sounds good, send a +1 and I'll merge this PR.

ashwinr:
Thanks for the clear explanation Adi. I'm in agreement with most things, except that it'll be published on the same npm dist. This means we can't publish 2.6.0 of tslint until typescript releases 1.6. I fear this means we're beholden to typescript's release schedule for tslint releases. Tslint usually has multiple releases for each typescript release.

gscshoyru:
@ashwinr I'm not sure how this prevents us from releasing multiple times for any one typescript release. We should be able to publish to latest multiple times without difficulty -- if there's anything we need to fix in 2.4 we just branch off of our release, fix it, and publish to latest. We don't need to wait on typescript here, unless I'm missing something?

gscshoyru:
Ok -- after discussion, we're going to track latest (the stable releases) on master, since that's where most people will make pull requests to, and track next (future release of typescript) on a branch called "next". Closing this pull request, will recreate it shortly.

PR #540, opened 2015-08-02
Use node 0.12 in AppVeyor

xt0rted:
This will allow appveyor to run the tests on both node 0.10 and 0.12 just like they do on travis


adidahiya:
thanks @xt0rted!

PR #545, opened 2015-08-04
Fix indent rule in sample tslint.json

myitcv:
The indent rule in the sample is incorrect as of `2.4.2` and `master`

Valid options are:

```
"indent": [true, "spaces"]

OR

"indent": [true, "tabs"]
```

I guess in future the third option would be a number? But hopefully this fixes things for now.


adidahiya:
thanks @myitcv!

PR #548, opened 2015-08-05
Add testing for .tsx files

### LINKED ISSUE #535: Add tests for linting .tsx files

adidahiya:
Fixes #535.


gscshoyru:
Looks good to me. We might want to pull out the "runLinterWithConfiguration" bit at some later point to use with other tests? I like it. But this is good for now.

adidahiya:
@gscshoyru yeah that function can be used in `ruleDisableEnableTests.ts` as well.

In general, I think we need more testing utilities for running more than one rule on a file.

PR #560, opened 2015-08-07
Add TypeScript definition

pspeter3:
This will allow `tsd` 0.6 to discover the file


adidahiya:
cool, thanks @pspeter3!

PR #565, opened 2015-08-12
Cleanup language service host

adidahiya:
computeLineStarts is an impl detail we shouldn't be using:
https://github.com/Microsoft/TypeScript/issues/4057#issuecomment-130120572


gscshoyru:
Well this is simpler.

PR #579, opened 2015-08-18
Ensure that extra failues during classNameRuleTests will be detected

JKillian:


test/rules/classNameRuleTests.ts
```diff
@@ -20,11 +20,12 @@ describe("<class-name>", () => {
 
     it("ensures class names are always pascal-cased", () => {
         const createFailure = Lint.Test.createFailuresOnFile(fileName, ClassNameRule.FAILURE_STRING);
-        const expectedFailure1 = createFailure([5, 7], [5, 23]);
-        const expectedFailure2 = createFailure([9, 7], [9, 33]);
+        const expectedFailures = [
+          createFailure([5, 7], [5, 23]),
+          createFailure([9, 7], [9, 33])
```
adidahiya: oops. just realized (post-merge) that this should have 4-space indentation. and we should enhance our `whitespace` rule to enforce that :)

Issue #507, opened 2015-07-16
assignment operation inside an 'if' condition should be marked with a warning


AGiorgetti:
If we want to spot some more bugs at compile time, I believe that something like:

```javascript
if (data.id = id) { ...
```

while being perfectly legal, it can be a coding bug most of the times.
it would be nice if TsLint could spot this type of code and give us a warning instead of using a custom regex that scan all our files.

adidahiya:
sounds good. it could be called... `no-assignment-in-conditional`? cc @ashwinr for thoughts on rule name

ashwinr:
eslint has the same [rule](http://eslint.org/docs/rules/no-cond-assign). even just `no-conditional-assignment` seems good.

JKillian:
ESLint has an option for the rule where they allow assignments that are wrapped in parenthesis. Do we want the same option or should we ust flag all assignments in a conditional no matter what?

adidahiya:
let's just do the simple thing for now and disallow all assignments in a conditional. we can add the other option later if it's requested.

PR #586, opened 2015-08-19
Add no-conditional-assignment rule

JKillian:
Implements feature request #507


src/rules/noConditionalAssignmentRule.ts
```diff
@@ -0,0 +1,78 @@
...
+        super.visitWhileStatement(node);
+    }
+
+    protected visitDoStatement(node: ts.DoStatement) {
+        this.checkExpressionForBinaryExpressions(node.expression);
+        super.visitWhileStatement(node);
+    }
+
+    protected visitForStatement(node: ts.ForStatement) {
+        if (node.condition) {
```
adidahiya: always do explicit null checks, i.e. `node.condition != null`

src/rules/noConditionalAssignmentRule.ts
```diff
@@ -0,0 +1,78 @@
...
+    public static FAILURE_STRING = "assignment in conditional: ";
+
+    public apply(sourceFile: ts.SourceFile): Lint.RuleFailure[] {
+        const walker = new NoConditionalAssignmentWalker(sourceFile, this.getOptions());
+        return this.applyWithWalker(walker);
+    }
+}
+
+class NoConditionalAssignmentWalker extends Lint.RuleWalker {
+    private inConditionalExpression = false;
```
adidahiya: let's abbreviate this... `isInConditional`

src/rules/noConditionalAssignmentRule.ts
```diff
@@ -0,0 +1,78 @@
...
+    private checkExpressionForBinaryExpressions(expression: ts.Expression) {
+        this.inConditionalExpression = true;
+        if (expression.kind === ts.SyntaxKind.BinaryExpression) {
+            this.checkBinaryExpressionForAssignment(<ts.BinaryExpression> expression);
+        }
+        this.walkChildren(expression);
+        this.inConditionalExpression = false;
+    }
+
+    private checkBinaryExpressionForAssignment(expression: ts.BinaryExpression) {
```
adidahiya: abbreviate to `checkForAssignment` (the function type already tells you you're checking binary expressions)

src/rules/noConditionalAssignmentRule.ts
```diff
@@ -0,0 +1,78 @@
...
+        if (this.isAssignmentToken(expression.operatorToken.kind)) {
+            this.addFailure(this.createFailure(expression.getStart(), expression.getWidth(), Rule.FAILURE_STRING));
+        }
+    }
+
+    private isAssignmentToken(token: ts.SyntaxKind): boolean {
```
adidahiya: * rename `token` to `tokenKind`
* prefer type inference when possible; omit `: boolean` here

src/rules/noConditionalAssignmentRule.ts
```diff
@@ -0,0 +1,78 @@
...
+    protected visitBinaryExpression(expression: ts.BinaryExpression) {
+        if (this.inConditionalExpression) {
+            this.checkBinaryExpressionForAssignment(expression);
+        }
+        super.visitBinaryExpression(expression);
+    }
+
+    private checkExpressionForBinaryExpressions(expression: ts.Expression) {
+        this.inConditionalExpression = true;
+        if (expression.kind === ts.SyntaxKind.BinaryExpression) {
```
adidahiya: let's add a comment along the lines of:

```
// check for simple assignment in a conditional, like `if (a = 1) {`
```

and then below, explain why we walk children:

```ts
// walk the children of the conditional expression for nested assignments, i.e. `if ((a = 1) && (b = 1)) {`
this.walkChildren(expression);
```

src/rules/noConditionalAssignmentRule.ts
```diff
@@ -0,0 +1,78 @@
...
+    }
+
+    protected visitBinaryExpression(expression: ts.BinaryExpression) {
+        if (this.inConditionalExpression) {
+            this.checkBinaryExpressionForAssignment(expression);
+        }
+        super.visitBinaryExpression(expression);
+    }
+
+    private checkExpressionForBinaryExpressions(expression: ts.Expression) {
```
adidahiya: let's call this `validateConditionalExpression` and the argument `conditional`

test/files/rules/nocondassign.test.ts
```diff
@@ -0,0 +1,34 @@
...
+while (x !== 2) { }
+
+for (var x = 8; x == 8; ++x) { }
+for (var x = 8; x == 8; x = 12) { }
+for (;;) { }
+
+// invalid cases
+if (x = 5) { }
+if (a && (b = 5)) { }
+else if (x = 2) { }
```
adidahiya: please add another test where there are two assignments in the conditional expression; the linter should report failures on both

src/rules/noConditionalAssignmentRule.ts
```diff
@@ -0,0 +1,78 @@
...
+    private checkBinaryExpressionForAssignment(expression: ts.BinaryExpression) {
+        if (this.isAssignmentToken(expression.operatorToken.kind)) {
+            this.addFailure(this.createFailure(expression.getStart(), expression.getWidth(), Rule.FAILURE_STRING));
+        }
+    }
+
+    private isAssignmentToken(token: ts.SyntaxKind): boolean {
+        return token >= ts.SyntaxKind.FirstAssignment && token <= ts.SyntaxKind.LastAssignment;
+    }
+
```
adidahiya: please remove this newline

src/rules/noConditionalAssignmentRule.ts
adidahiya: alternatively, this could be slightly better:

keep the arg as `token` and actually take in a token node instead of a `SyntaxKind`. then the body of this function would reference `token.kind`

Issue #600, opened 2015-08-23
no-internal-module raises an error for namespaces with nested paths


vmmitev:
Consider the following example:

```typescript
namespace Example.Utils { }
```

with `no-internal-module` set to to `true` an error is thrown starting at the `Utils` fragment. This can be applied to an arbitrary number of nested namespace paths, each of which raises an error of its own. A possible workaround is to rework the code-block as:

```typescript
namespace Example {
    namespace Utils { }
}
```

If this is the intended way to apply the option it seems to add unnecessary verbosity and nesting to code. If it's not, my best bet is that the module/namespace declaration node is split into 2 or more elements which are visited, the first of which passes the namespace check and the rest end up as the error.

adidahiya:
fixed in `v2.5.0-dev.5`

PR #601, opened 2015-08-24
Fix for no-internal-module rule

vmmitev:
Fix proposal for issue #600 

**Personal note:** I think the better approach to fixing this issue is to have the nodes representing the nested sub-paths in the namespace have the correct flag type `ts.NodeFlags.Namespace` instead of having to recurse for their parent which has it. Currently they have `ts.NodeFlags.Export` as their `flags` property's value. However, I wanted to limit the fix to the rule's definition file, and think this is the best solution without having to tamper into the node walker internals.

### DIFF
--- a/src/rules/noInternalModuleRule.ts
+++ b/src/rules/noInternalModuleRule.ts
@@ -24,12 +24,23 @@ export class Rule extends Lint.Rules.AbstractRule {
 
 class NoInternalModuleWalker extends Lint.RuleWalker {
     public visitModuleDeclaration(node: ts.ModuleDeclaration) {
-        if (Lint.isNodeFlagSet(node, ts.NodeFlags.Namespace)) {
-            // ok namespace
-        } else if (node.name.kind === ts.SyntaxKind.Identifier) {
-            // for external modules, node.name will be a LiteralExpression instead of Identifier
+        if (this.isInternalModuleDeclaration(node)) {
             this.addFailure(this.createFailure(node.getStart(), node.getWidth(), Rule.FAILURE_STRING));
         }
         super.visitModuleDeclaration(node);
+    }
+
+    private isInternalModuleDeclaration(node: ts.ModuleDeclaration) {
+        // an internal module declaration is not a namespace or a nested declaration
+        // for external modules, node.name.kind will be a LiteralExpression instead of Identifier
+        return !Lint.isNodeFlagSet(node, ts.NodeFlags.Namespace)
+            && !this.isNestedDeclaration(node)
+            && node.name.kind === ts.SyntaxKind.Identifier;
+    }
+
+    private isNestedDeclaration(node: ts.ModuleDeclaration) {
+        // in a declaration expression like 'module a.b.c' - 'a' is the top level module declaration node and 'b' and 'c' are nested
+        // therefore we can depend that a node's position will only match with its name's position for nested nodes
+        return node.name.pos === node.pos;
     }
 }
--- a/test/files/rules/nointernalmodule.test.ts
+++ b/test/files/rules/nointernalmodule.test.ts
@@ -11,3 +11,42 @@ declare module "hoge" {
 }
 declare module 'fuga' {
 }
+
+namespace foo.bar {
+}
+namespace foo.bar.baz {
+}
+namespace foo {
+    namespace bar.baz {
+    }
+}
+
+namespace foo.bar {
+    module baz {
+        namespace buzz {
+        }
+    }
+}
+
+module foo.bar {
+    namespace baz {
+        module buzz {
+        }
+    }
+}
+
+namespace name.namespace {
+}
+namespace namespace.name {
+}
+
+// intentionally malformed for test cases, do not format
+declare module declare
+.dec{}
+declare  module dec . declare  {
+}
+
+module  mod.module{}
+module module.mod
+{
+}
--- a/test/rules/noInternalModuleTests.ts
+++ b/test/rules/noInternalModuleTests.ts
@@ -23,7 +23,14 @@ describe("<no-internal-module>", () => {
         const actualFailures = Lint.Test.applyRuleOnFile(fileName, NoInternalModule);
         const expectedFailures = [
             Lint.Test.createFailure(fileName, [4, 1], [4, 15], failureString),
-            Lint.Test.createFailure(fileName, [7, 1], [7, 24], failureString)
+            Lint.Test.createFailure(fileName, [7, 1], [7, 24], failureString),
+            Lint.Test.createFailure(fileName, [25, 5], [28, 6], failureString),
+            Lint.Test.createFailure(fileName, [31, 1], [36, 2], failureString),
+            Lint.Test.createFailure(fileName, [33, 9], [33, 32], failureString),
+            Lint.Test.createFailure(fileName, [44, 1], [44, 30], failureString),
+            Lint.Test.createFailure(fileName, [46, 1], [46, 35], failureString),
+            Lint.Test.createFailure(fileName, [49, 1], [49, 21], failureString),
+            Lint.Test.createFailure(fileName, [50, 1], [50, 22], failureString)
         ];
 
         Lint.Test.assertFailuresEqual(actualFailures, expectedFailures);


JKillian:
Thanks for the PR @vmmitev. I agree with you that, ideally, the `ts.NodeFlags.Namespace` flag should be set for the nested namespace, so I mentioned [that issue](https://github.com/Microsoft/TypeScript/issues/4436) to the Typescript team.

Your method is a decent workaround, however it doesn't handle one of the possible cases:
```typescript
namespace A {
    module B {
    }
}
```
In this case, the rule won't catch the error. I think we have two options here:
* Wait for the Typescript team to fix the bug
* In the error case I posted, `ModuleDeclaration` B is a child of a `ModuleBlock`. Whereas when you have `namespace A.B { }`, `ModuleDeclaration B` is a child of a `ModuleDeclaration`. 

I think for the second option, we could change:

 ```typescript
for (; node.parent; node = node.parent)
``` 
to 

```typescript
for (; node.parent != null && node.parent.kind === ts.SyntaxKind.ModuleDeclaration; node = node.parent)
```
And of course, we'd also want a test for this case.

vmmitev:
Thanks for noticing, @JKillian. I did realize shortly after creating the PR that the node information is supplied by the compiler and that, indeed, it has to be fixed on Typescript's side.

I didn't want to give up on the second option just yet. After debugging the newly introduced erronous case it would seem that we can depend that nested namespaces are a node with 'flags' property set to '1' and nested modules have it set to '0'. Therefore, the solution I came up with is to treat nodes with zero values as nested module declarations. That kind of dependency can be problematic in the long run but I've pushed it regardless so that you can review and weigh in your opinion.

Maybe we can improve the check in such a way that when the compiler reports the correct values the rule won't need to be updated again?

JKillian:
I agree that we can probably get this working correctly on the end of tslint for now. I don't think your flags idea will work though, because of what the flag value of 1 means. If you check out the [NodeFlags type](https://github.com/Microsoft/TypeScript/blob/9a6db930b48b2e51216739aa0d9430e13121cc93/src/compiler/types.ts#L362), you can see that they use that to indicate that it is exported.

So when you do `namespace A.B {}`, `B` is automatically exported. What that means is that you'll have the same flag values if you do this: 
```typescript
namspace C {
    export module D {
    }
}
````
and if we're just checking based on the flag value, the rule would say that that is okay. 

I think my previous suggestion (the modification of the for loop) should work though - feel free to give it a try and let me know if you encounter any problems.

**Comment by vmmitev** [comment]
Your approach was the first thing I tried. However, that check wasn't sufficient to differentiate the namespace and module keywords declarations (all nodes return a `kind` property with value `ts.SyntaxKind.ModuleDeclaration = 206`.

Using the `flags === 0` check is horrible but it was the first thing I found that passed the tests as I'm trying to guess my way around the API.

I'm going to continue looking for a definite solution and update the PR once found.

**Comment by vmmitev** [comment]
In the end it seems the most reliable way to determine if a `ModuleDeclaration` node is not nested is to check whether its text starts with `module` or `namespace`. However if we invert this statement then a nested module will have to begin with its `node.name.text` value. This replaces the `node.flags === 0` check and works well:
```typescript
private isNestedDeclaration(node: ts.ModuleDeclaration): boolean {
    // top level declarations begin with "module" or "namespace", nested ones - with their names
    return node.name != null &&
        node.getText().indexOf(node.name.text) === 0;
}
```
Moreover, if we apply this check to the `if` block where we insert the actual failure, then we'll no longer report multiple errors for blocks like:
```typescript
module foo.bar.baz {
}
```
This should help reduce the number of warnings that developers will get when activating this option on their code that depends on internal module declarations to one per module declaration block.

src/rules/noInternalModuleRule.ts
```diff
@@ -24,12 +24,29 @@ export class Rule extends Lint.Rules.AbstractRule {
...
+
+    private isInternalModuleDeclaration(node: ts.ModuleDeclaration): boolean {
+        // for external modules, node.name will be a LiteralExpression instead of Identifier
+        return node.name.kind === ts.SyntaxKind.Identifier &&
+            !(this.isNamespaceDeclaration(node) || this.isNestedDeclaration(node));
+    }
+
+    private isNamespaceDeclaration(node: ts.ModuleDeclaration): boolean {
+        // the declaration may be nested, so recurse until the nesting chain breaks or reaches a namespace node
+        return node != null &&
```
adidahiya: please move the binary operators to the next line:

```ts
return node != null
    && this.isNestedDeclaration(node)
    && ...
```

src/rules/noInternalModuleRule.ts
```diff
@@ -24,12 +24,29 @@ export class Rule extends Lint.Rules.AbstractRule {
...
+            this.isNestedDeclaration(node) &&
+            this.isNamespaceDeclaration(<ts.ModuleDeclaration> node.parent) ||
+            Lint.isNodeFlagSet(node, ts.NodeFlags.Namespace);
+    }
+
+    private isNestedDeclaration(node: ts.ModuleDeclaration): boolean {
+        // top level declarations begin with "module" or "namespace", nested ones - with their names
+        return node.name != null &&
+            node.getText().indexOf(node.name.text) === 0;
```
adidahiya: same comment about `&&`

src/rules/noInternalModuleRule.ts
```diff
@@ -24,12 +24,29 @@ export class Rule extends Lint.Rules.AbstractRule {
...
+
+    private isNamespaceDeclaration(node: ts.ModuleDeclaration): boolean {
+        // the declaration may be nested, so recurse until the nesting chain breaks or reaches a namespace node
+        return node != null &&
+            this.isNestedDeclaration(node) &&
+            this.isNamespaceDeclaration(<ts.ModuleDeclaration> node.parent) ||
+            Lint.isNodeFlagSet(node, ts.NodeFlags.Namespace);
+    }
+
+    private isNestedDeclaration(node: ts.ModuleDeclaration): boolean {
```
adidahiya: please add a comment above this method explaining that "nested" means something of the form "a.b.c"

src/rules/noInternalModuleRule.ts
```diff
@@ -24,12 +24,29 @@ export class Rule extends Lint.Rules.AbstractRule {
...
-            // ok namespace
-        } else if (node.name.kind === ts.SyntaxKind.Identifier) {
-            // for external modules, node.name will be a LiteralExpression instead of Identifier
+        if (this.isInternalModuleDeclaration(node)) {
             this.addFailure(this.createFailure(node.getStart(), node.getWidth(), Rule.FAILURE_STRING));
         }
         super.visitModuleDeclaration(node);
     }
+
+    private isInternalModuleDeclaration(node: ts.ModuleDeclaration): boolean {
```
adidahiya: omit return types. we prefer TS inference whenever possible.

src/rules/noInternalModuleRule.ts
```diff
@@ -24,12 +24,29 @@ export class Rule extends Lint.Rules.AbstractRule {
...
-        if (Lint.isNodeFlagSet(node, ts.NodeFlags.Namespace)) {
-            // ok namespace
-        } else if (node.name.kind === ts.SyntaxKind.Identifier) {
-            // for external modules, node.name will be a LiteralExpression instead of Identifier
+        if (this.isInternalModuleDeclaration(node)) {
             this.addFailure(this.createFailure(node.getStart(), node.getWidth(), Rule.FAILURE_STRING));
         }
         super.visitModuleDeclaration(node);
     }
+
```
adidahiya: please add a comment linking this to the compiler issue:

```
// todo: simplify this implementation when https://github.com/Microsoft/TypeScript/issues/4436 is fixed
```

adidahiya:
thanks @vmmitev! just need some small tweaks to get this merged. looks like your tests are passing; appveyor just had connectivity problems with Github. It should work fine on your next commit

JKillian:
@vmmitev I still think your old method is simpler and will work with a tweak or two: :smiley: 

```typescript
class NoInternalModuleWalker extends Lint.RuleWalker {
    public visitModuleDeclaration(node: ts.ModuleDeclaration) {
        if (this.isNamespaceFlagSetRecurseParents(node)) {
            // ok namespace
        } else if (node.name.kind === ts.SyntaxKind.Identifier) {
            // for external modules, node.name will be a LiteralExpression instead of Identifier
        }
        super.visitModuleDeclaration(node);
    }

    private isNamespaceFlagSetRecurseParents(node: ts.Node): boolean {
        for (; node != null && node.kind === ts.SyntaxKind.ModuleDeclaration; node = node.parent) {
            if (Lint.isNodeFlagSet(node, ts.NodeFlags.Namespace)) {
                return true;
            }
        }
        return false;
    }
}
```
(I had a few unnecessary `.parent`'s last time.) A nested ModuleDeclaration's parent will only be a ModuleDeclaration if it is declared with the `namespace A.B` syntax and not with the `namespace A { namespace B {} }` syntax. So if its parent isn't a ModuleDeclaration, we can just stop traversing the parent chain and return false.

adidahiya:
I'm not a fan of the for-loop syntax to traverse parent nodes. I like the readability of the current PR. If there are simplifications to be made, please suggest them on the current implementation.

JKillian:
@adidahiya Just the syntax of using a for loop, or the whole concept? If just the syntax, a while loop can be used nearly identically.

@vmmitev Check out this test case :wink: 
```typescript
namespace foo {
    module mod {
    }
}
```

vmmitev:
@JKillian that was a low blow. I'll work on ironing this out as well as the review comments.

JKillian:
@vmmitev I know, definitely a cheapshot :grinning:. Thanks for all the effort on this issue!

vmmitev:
I think this is it. Once the compiler issue is closed and TSLint targets a version containing the fix I'll make sure to come back and create a PR that removes the need to recurse parent nodes.

You can tell by some of the new test cases that my battle with the faulty `isNestedDeclaration` logic got pretty messy. It was going fairly well until I realised `getText()` also returns the body of the whole module declaration. That's where I drew the line and started from scratch trying to find **anything** that's not text and can be resolved down to the desired result. Then it hit me! I had written it in the method's comment all along. All I had to do is to switch to checking the `pos` properties of the `node` and its `node.name`. I also figured that there will never be a case where this method is called on nodes that are not a `ts.ModuleDeclaration` (even when recursing and making a type-cast). That's why it no longer checks if the `name` property exists.

The new test cases are a bit redundant for the current state of the code, but since they are valid declarations with interesting properties (duplicating or partial keywords, and malformed whitespace) let's leave them just in case.

src/rules/noInternalModuleRule.ts
```diff
@@ -24,12 +24,32 @@ export class Rule extends Lint.Rules.AbstractRule {
...
+    private isInternalModuleDeclaration(node: ts.ModuleDeclaration) {
+        // an internal module declaration is not a namespace or a nested declaration
+        // for external modules, node.name.kind will be a LiteralExpression instead of Identifier
+        return node.name.kind === ts.SyntaxKind.Identifier
+            && !(this.isNamespaceDeclaration(node) || this.isNestedDeclaration(node));
+    }
+
+    // todo: remove this method and replace it with only the flag check when https://github.com/Microsoft/TypeScript/issues/4436 is fixed
+    private isNamespaceDeclaration(node: ts.ModuleDeclaration): boolean {
```
vmmitev: Sadly, TypeScript cannot infer the return type because of the underlying recursion. Since this method is flagged for removal I hope it's not a problem if it stays like this for now.

vmmitev:
Quick update. I just reviewed the discussion at [TS4436](https://github.com/Microsoft/TypeScript/issues/4436) and it turns out they only add flags to nodes that contain the actual keyword. Correct me if I'm wrong. This can mean 2 things:
1. We can actually depend on cases where `flags === 0` to determine nesting. However, my observations were that nodes nested in a `namespace` have `flags === 1` which contradicts their statement (nodes nested in an internal `module` declaration exhibited the correct? zero value). Maybe we should bring this up for clarification on their side?
2. dasda

adidahiya:
the compiler bug has been fixed upstream and this is no longer an issue in the latest tslint. can you please update this PR to only add tests (revert your changes to `src/` files)? thanks!

vmmitev:
Sure thing. Would you prefer to keep the current history and do a revert commit or shall I just rebase the HEAD?

adidahiya:
thanks for asking. I'd prefer a cleaner history, so please rebase on master and try to make this just 1 commit. you can start a new PR if that's easier for you

vmmitev:
I know you asked for just the tests but as they were, without ignoring nested dot-syntax declarations, 21 errors were produced instead of the expected 9 (only from nodes directly following the `module` keyword). I can rebase again and add assertions for the additional cases if you prefer the more verbose behaviour.

adidahiya:
Sorry, I forgot to mention -- can you make a new PR targeting the `next` branch? That's where we have the latest TS compiler with the fix for the namespace flag. And TS 1.6 is going to release soon so I don't want to maintain this new code (your adjustments to `NoInternalModuleWalker`) on `master` if it'll be fixed automatically with the latest compiler when the 1.6 beta is out.

PR #605, opened 2015-08-25
Implemented optional trailing underscore for variableNameRule.ts

McEoy:
When implementing AngularJS tests the inject methods recommends parameters to follow the convention "\_paramName\_", eg "\_$httpBackend\_".

The respective rule was missing in TSLint, which is now implemented by adding the option "allow-trailing-underscore" to "variable-name".

Cheers

### DIFF
--- a/test/files/rules/varname.test.ts
+++ b/test/files/rules/varname.test.ts
@@ -32,3 +32,4 @@ export function functionWithInvalidSpread(invalid_arg: ...number) { // 1 failure
 }
 
 let optionallyValid_ = "bar";
+let _$httpBackend_ = "leading and trailing";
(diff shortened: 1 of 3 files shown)


test/files/rules/varname.test.ts
```diff
@@ -30,3 +30,5 @@ export function anotherFunctionWithInvalidParamNames ([first_element, SecondElem
 export function functionWithInvalidSpread(invalid_arg: ...number) { // 1 failure
   //
 }
+
+let optionallyValid_ = "bar";
```
adidahiya: can you please add a test that verifies behavior when _both_ options (leading & trailing) are enabled?

adidahiya:
thanks @McEoy, this looks pretty good! just needs some more testing.

McEoy:
@adidahiya thanks for reviewing. New test is done.

Issue #597, opened 2015-08-21
Use correct file extension for .tsx files being processed by LanguageService


JKillian:
In `noUnusedVariableRule.ts` and `noUseBeforeDeclareRule.ts`, a `LanguageServiceHost` object is created and passed a hard-coded file name, `file.ts`. However, if the file we're working with is actually a `.tsx` file, it will be processed incorrectly by the language service. We need to detect what kind of file is being processed and use a correct extension.

Issue #558, opened 2015-08-07
no-unused-variable false positive with references in JSX


adidahiya:
code sample reported by @pepaar:

```ts
/// <reference path="react.d.ts" />
/// <reference path="react-jsx.d.ts" />

import React = require("react");

interface IProps {
    content: string;
}

class MyComponent extends React.Component<IProps, {}> {
    render() {
        this.alertMe();
        return 
            <div>
                {this.props.content}
                <button onClick={() => this.alertMe()}>AlertMe</button>
            </div>;
    }

    private alertMe() {
        let yo = this.props.content + " YO";
        alert(yo);
    }
}

React.render(<MyComponent content="Hello World"/>, document.body);
```

`languageService.getDocumentHighlights` doesn't return the right thing for `alertMe` in this case.

seems to be caused by a compiler bug: https://github.com/Microsoft/TypeScript/issues/4175

...but @pepaar also reported a linter crash, which should be fixed on our end:

```
C:\Users\pepaar\AppData\Roaming\npm\node_modules\tslint\build\rules\noUnusedVari
ableRule.js:161
        if (highlights[0].highlightSpans.length <= 1) {
                      ^
TypeError: Cannot read property '0' of undefined
    at NoUnusedVariablesWalker.validateReferencesForVariable (C:\Users\pepaar\Ap
pData\Roaming\npm\node_modules\tslint\build\rules\noUnusedVariableRule.js:161:23
)
...
```

__UPDATE__: the linter crash has been fixed by #563

PR #610, opened 2015-08-26
Use .tsx extensions when interacting with language services when appropriate

JKillian:
Fixes issue #597

### DIFF
--- a/src/rules/noUnusedVariableRule.ts
+++ b/src/rules/noUnusedVariableRule.ts
@@ -20,25 +20,22 @@ export class Rule extends Lint.Rules.AbstractRule {
     public static FAILURE_STRING = "unused variable: ";
 
     public apply(sourceFile: ts.SourceFile): Lint.RuleFailure[] {
-        const documentRegistry = ts.createDocumentRegistry();
-        const languageServiceHost = Lint.createLanguageServiceHost("file.ts", sourceFile.getFullText());
-        const languageService = ts.createLanguageService(languageServiceHost, documentRegistry);
-
+        const languageService = Lint.createLanguageService(sourceFile.fileName, sourceFile.getFullText());
         return this.applyWithWalker(new NoUnusedVariablesWalker(sourceFile, this.getOptions(), languageService));
     }
 }
 
 class NoUnusedVariablesWalker extends Lint.RuleWalker {
+    private languageService: ts.LanguageService;
     private skipBindingElement: boolean;
     private skipParameterDeclaration: boolean;
     private skipVariableDeclaration: boolean;
-    private languageService: ts.LanguageService;
 
     constructor(sourceFile: ts.SourceFile, options: Lint.IOptions, languageService: ts.LanguageService) {
         super(sourceFile, options);
+        this.languageService = languageService;
         this.skipVariableDeclaration = false;
         this.skipParameterDeclaration = false;
-        this.languageService = languageService;
     }
 
     public visitBindingElement(node: ts.BindingElement) {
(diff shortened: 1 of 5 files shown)


JKillian:
Also seems to fix issue #558

test/tsxTests.ts
```diff
@@ -82,9 +82,7 @@ describe("TSX syntax", () => {
         });
 
         it("with no false positives", () => {
-            // todo (#558): there should only be 3, but there is a false positive no-unused-variable failure due
-            // to a compiler bug
-            assert.lengthOf(actualFailures, 4);
+            assert.lengthOf(actualFailures, 3);
```
adidahiya: nice!

src/rules/noUnusedVariableRule.ts
```diff
@@ -32,13 +28,17 @@ class NoUnusedVariablesWalker extends Lint.RuleWalker {
...
     private languageService: ts.LanguageService;
 
-    constructor(sourceFile: ts.SourceFile, options: Lint.IOptions, languageService: ts.LanguageService) {
+    constructor(sourceFile: ts.SourceFile, options: Lint.IOptions) {
         super(sourceFile, options);
         this.skipVariableDeclaration = false;
         this.skipParameterDeclaration = false;
-        this.languageService = languageService;
+
+        this.fileName = (Lint.isTsxFile(sourceFile.fileName) ? "file.tsx" : "file.ts");
```
adidahiya: let's just use

```ts
`file.${path.extename(sourceFile.fileName)}`
```

and remove `Lint.isTsxFile` and its tests completely

src/rules/noUnusedVariableRule.ts
```diff
@@ -32,13 +28,17 @@ class NoUnusedVariablesWalker extends Lint.RuleWalker {
...
+        this.languageService = Lint.createLanguageService(this.fileName, sourceFile.getFullText());
```
adidahiya: hm, I liked it earlier when we passed a `LanguageService` to the constructor.... that makes the whole thing more testable... but I see why you're doing it this way so that you only have to declare `fileName` once.

why don't we just use `sourceFile.fileName`? why even make a new name at all?

src/rules/noUnusedVariableRule.ts
```diff
@@ -32,12 +29,14 @@ class NoUnusedVariablesWalker extends Lint.RuleWalker {
     private skipBindingElement: boolean;
     private skipParameterDeclaration: boolean;
     private skipVariableDeclaration: boolean;
+
```
adidahiya: please remove this extra newline, and move `private languageService` to the top (alphabetization)

src/rules/noUnusedVariableRule.ts
```diff
@@ -32,12 +29,14 @@ class NoUnusedVariablesWalker extends Lint.RuleWalker {
...
     private languageService: ts.LanguageService;
 
     constructor(sourceFile: ts.SourceFile, options: Lint.IOptions, languageService: ts.LanguageService) {
         super(sourceFile, options);
         this.skipVariableDeclaration = false;
         this.skipParameterDeclaration = false;
+
```
adidahiya: remove extra newline

src/language/utils.ts
```diff
@@ -120,4 +120,15 @@ module Lint {
...
+    /**
+    * Checks if a file ends with ".tsx"
+    */
+    export function isTsxFile(fileName: string): boolean {
+        const extension = ".tsx";
+        const pathLen = fileName.length;
+        const extLen = extension.length;
+        return pathLen > extLen
+            && fileName.substr(pathLen - extLen, extLen) === extension;
+    }
```
adidahiya: please remove this method, its generated typing, and its unit tests.

adidahiya:
great, can you squash this into 1 commit please?

JKillian:
Squashed

Issue #500, opened 2015-07-15
no-shadowed-variable false positive in function typedefs/ functions defs in interfaces


adidahiya:
the following code incorrectly reports a shadowed variable failure for `a` within the typedef of `done`:

```ts
function foo(a: any, done: (a: any) => void) {
    //
}
```

JKillian:
This also happens in interface declarations, such as:

```typescript
interface Surface {
    diffuse: (pos: Vector) => Color;
    specular: (pos: Vector) => Color;
    reflect: (pos: Vector) => number;
    roughness: number;
}
```
I'm going to expand the scope of this issue to include this as well

adidahiya:
:+1:

adidahiya:
another false positive example:

```ts
function foo(bar: (a: any) => void) {
    let a = 1;
}
```

`a` is incorrectly reported as a shadowed variable

PR #616, opened 2015-08-27
Fix no shadowed vars rule

JKillian:
Fixes Issue #500 
Also adds more specific typing for a couple methods in `SyntaxWalker`


src/rules/noShadowedVariableRule.ts
```diff
@@ -48,6 +50,12 @@ class NoShadowedVariableWalker extends Lint.BlockScopeAwareRuleWalker<ScopeInfo,
         this.visitBlock(node.block);
     }
 
+    public visitFunctionType(node: ts.FunctionOrConstructorTypeNode) {
+        this.isInFunctionType = true;
+        super.visitFunctionType(node);
+        this.isInFunctionType = false;
+    }
```
adidahiya: instead of the `isInFunctionType` flag, what if we just don't call `super.visitFunctionType(node)` here? that should prevent the walker from checking nested parameter declarations.

JKillian: Good call, makes things much simpler

src/rules/noShadowedVariableRule.ts
```diff
@@ -23,6 +23,7 @@ export class Rule extends Lint.Rules.AbstractRule {
 }
 
 class NoShadowedVariableWalker extends Lint.BlockScopeAwareRuleWalker<ScopeInfo, ScopeInfo> {
+
```
adidahiya: remove this extra line please

adidahiya:
great, can you squash this into 1 commit please?

JKillian:
Yep, done

PR #627, opened 2015-09-01
Tests and improvement for no-internal-module rule on next branch

vmmitev:
Moved #601 to `next` branch.

### DIFF
--- a/src/rules/noInternalModuleRule.ts
+++ b/src/rules/noInternalModuleRule.ts
@@ -24,12 +24,23 @@ export class Rule extends Lint.Rules.AbstractRule {
 
 class NoInternalModuleWalker extends Lint.RuleWalker {
     public visitModuleDeclaration(node: ts.ModuleDeclaration) {
-        if (Lint.isNodeFlagSet(node, ts.NodeFlags.Namespace)) {
-            // ok namespace
-        } else if (node.name.kind === ts.SyntaxKind.Identifier) {
-            // for external modules, node.name will be a LiteralExpression instead of Identifier
+        if (this.isInternalModuleDeclaration(node)) {
             this.addFailure(this.createFailure(node.getStart(), node.getWidth(), Rule.FAILURE_STRING));
         }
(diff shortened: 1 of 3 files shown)


src/rules/noInternalModuleRule.ts
```diff
@@ -24,12 +24,23 @@ export class Rule extends Lint.Rules.AbstractRule {
...
+        return !Lint.isNodeFlagSet(node, ts.NodeFlags.Namespace)
+            && !this.isNestedDeclaration(node)
+            && node.name.kind === ts.SyntaxKind.Identifier;
+    }
+
+    private isNestedDeclaration(node: ts.ModuleDeclaration) {
+        // in a declaration expression like 'module a.b.c' - 'a' is the top level module declaration node and 'b' and 'c' are nested
+        // therefore we can depend that a node's position will only match with its name's position for nested nodes
+        return node.name.pos === node.pos;
+    }
```
adidahiya: The changes in this file shouldn't be needed now that we're on the 'next' branch

adidahiya: You'll need to run `npm install` to update your dev environment

vmmitev: Let me try to clarify the reasoning behind leaving this change. It's not because I chose to disregard the update to the compiler-related flags. This version no longer contains the "dirty" recursion approach. I ran it with updated npm modules, so the compiler's fix must've been in effect.

However, the `isNestedDeclaration` method was needed not only for recursing nodes' parents but to reduce the repetition of reported failures for declarations like `module foo.bar.baz`. In such cases 3 separate failures were reported for `foo`, `bar` and `baz`. In my opinion it should be only 1 and seemed fitting to push this change along with the test cases.

JKillian: :+1: Agreed that it should only report one failure for `module a.b.c`

adidahiya: my mistake -- I should've read more closely. this looks good.

Issue #379, opened 2015-04-26
multiple files produce invalid json output


saily:
When i want to run `tslint --format=json --file=/a/b/c.ts --file=/b/c/d.ts --out=tslint-result.json` this will produce an invalid json file because it just concats the output from each file. This also happens if you output to stdout.

EarthCitizen:
This is a huge problem for me as well. I need to send the results to Sonar, but the report is invalid.

JKillian:
Agreed that this is quite flawed. @EarthCitizen if you're able to do a little text-processing of the results, if you replace `}{` with `},{` and surround the whole file with `[]`, it should be a valid JSON and usable. Obviously this is not ideal though...

EarthCitizen:
Looking at the problem briefly, it does not *seem* like a difficult fix.

1) `class Linter` constructor needs to take a list of file names instead of just one
2) `class Linter.lint()` needs to iterate over those files and push all the failures to the common stack
3) `tslint-cli.ts` `processFile()` should be changed to `processFiles()` and take a list of file names instead of just one
4) `processFiles()` would be called just once, and do its validations on all the files

DISCLAIMER: of course I do not know the full ramifications of these changes, but it *seems* like it would work

JKillian:
@EarthCitizen Seems like sensible changes.Changing `processFile` to `processFiles` seems fine, it's an internal method so no real harm done. Changing the `Linter` constructor is trickier, as it is a public API and changing it would probably break many tools that use/wrap TSLint. However, I think it might be possible to make it take either a single filename or an array of filenames and handle the two cases appropriately.

I _think_ with these changes the formatter API could stay the same, which is a very good thing.

I'm actually going to close this issue because I see that #656 is a broader description of this same problem. We can continue the discussion there though!

EarthCitizen:
When I originally looked at this and commented in #379, I was not paying attention to the fact that the Linter constructor wants the file name AND the contents of the file. This makes refactoring class Linter a bit difficult. If there needs to be backwards compatibility with a public API, then perhaps mark that class deprecated and start over in a new class.

If that were to happen, the class should get the file contents itself when it is linting and not receive them in the constructor.

PR #641, opened 2015-09-06
JSON format outputs a line delimited JSON.

pocke:
This PR is a solution of #379 .

Currently, if designate json to format and multiple files, output of tslint is not a JSON.

I propose the line delimited JSON.
The line delimited JSON is machine readable than currently.

### DIFF
--- a/README.md
+++ b/README.md
@@ -318,8 +318,8 @@ Just like rules, additional formatters can also be supplied to TSLint via `--for
 
 export class Formatter extends Lint.Formatters.AbstractFormatter {
     public format(failures: Lint.RuleFailure[]): string {
-        var failuresJSON = failures.map((failure: Lint.RuleFailure) => failure.toJson());
-        return JSON.stringify(failuresJSON);
+        const failuresJSON = failures.map((failure) => failure.toJson());
+        return JSON.stringify(failuresJSON) + "\n";
     }
 }
 ```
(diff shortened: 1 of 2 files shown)


adidahiya:
well, it's still not producing valid JSON even after this change, right?

instead of:

```
{ ... }{ ... }
```

we now have

```
{ ... }
{ ... }
```

so what has been solved here? just readability for humans?

pocke:
Yes, still not valid JSON.

I improved readability for humans and machines.

After this change, we can use readline module for parsing the Line delimited JSON.
https://nodejs.org/api/readline.html

for example.

```javascript
// parse.js
var readline = require('readline');

var rl = readline.createInterface({
  input: process.stdin,
  output: process.stdout
});

rl.on('line', function(line) {
  var parsed = JSON.parse(line);
  DoSomething(parsed);
})
```

```sh
$ tslint | node parse.js
```

I think we can handle the output of tslint more easily.

adidahiya:
Is using `tslint` like that with unix pipes a common use case? Wouldn't you be better off using the programmatic API?

lbesson:
Wouldn't it be better to have the formatter handle all the files at once (#680 would enable this?)

adidahiya:
@lbesson yes, it would, which is why i'm opposed to this PR. This change is a band-aid that doesn't address the real problem. Closing for now.

Issue #642, opened 2015-09-07
tslint crashes when an object is created without parens


berickson1:
I've got something similar to the following in one of my ts classes, which crashes tslint

```typescript
class Foo{
    bar(){
        var foobar = new Object;
    }
}
```

The same issues presents itself if i were to do the following:

    var a = new Foo

While this may not be ideal usage of new Object (use 'new Object()' or {} instead) it still shouldn't crash tslint

```typescript
node_modules\tslint\build\rules\alignRule.js:64
        if (nodes.length === 0 || !this.hasOption(kind)) {
                 ^
TypeError: Cannot read property 'length' of undefined
    at AlignWalker.checkAlignment (\node_modules\tslint\build\rules\alignRule.js:64:18)
    at AlignWalker.visitNewExpression (\node_modules\tslint\build\rules\alignRule.js:55:14)
    at AlignWalker.SyntaxWalker.visitNode (\node_modules\tslint\lib\tslint.js:49589:26)
    at \node_modules\tslint\lib\tslint.js:49661:67
    at visitNode (\node_modules\tslint\lib\tslint.js:7184:20)
    at Object.forEachChild (\node_modules\tslint\lib\tslint.js:7238:21)
    at AlignWalker.SyntaxWalker.walkChildren (\node_modules\tslint\lib\tslint.js:49661:16)
    at AlignWalker.SyntaxWalker.visitVariableDeclaration (\node_modules\tslint\lib\tslint.js:49449:18)
    at AlignWalker.SyntaxWalker.visitNode (\node_modules\tslint\lib\tslint.js:49646:26)
    at \node_modules\tslint\lib\tslint.js:49661:67
```

JKillian:
When TSLint detects an error that it really extra egregious, it decides to crash to let the developer know they really shouldn't be doing that...

Just kidding :smile:, agree that this is a problem, I'll look into it.

PR #643, opened 2015-09-08
Fix various bugs with align-rule

JKillian:
* Align rule won't crash on new expression without parens
* Align rule now checks constructor and method declarations
* Align rule test file now uses valid/normal TS syntax


JKillian:
Fixes #642

Issue #570, opened 2015-08-13
no-unused-variable false positive when variable is declared outside of a module


adidahiya:
simple repro code:

```ts
const fs = require("fs");

module Foo {
    const path = require("path");

    console.log(fs);
    console.log(path);
}
```

tslint reports `'fs'` on line 1 is unused.

JKillian:
Possibly related to if there is a surrounding scope or not.  Problem seems to happen somewhere in the `getReferencedSymbolsForNode` function in [services.ts](https://github.com/Microsoft/TypeScript/blob/5bcb3f0be86cf251c37b618040405a654486f961/src/services/services.ts) (see below for excerpt). With a surrounding scope, the code enters the if statement below and the proper number of references are added to `result`. If there's not a surrounding scope, it enters the else branch and only the initial declaration is added to `result`. Unfortunately, unsure why at this point.

```typescript
            // Try to get the smallest valid scope that we can limit our search to;
            // otherwise we'll need to search globally (i.e. include each file).
            let scope = getSymbolScope(symbol);

            // Maps from a symbol ID to the ReferencedSymbol entry in 'result'.
            let symbolToIndex: number[] = [];
            if (scope) {
                result = [];
                getReferencesInNode(scope, symbol, declaredName, node, searchMeaning, findInStrings, findInComments, result, symbolToIndex);
            }
            else {
                let internedName = getInternedName(symbol, node, declarations)
                for (let sourceFile of sourceFiles) {
                    cancellationToken.throwIfCancellationRequested();

                    let nameTable = getNameTable(sourceFile);

                    if (lookUp(nameTable, internedName)) {
                        result = result || [];
                        getReferencesInNode(sourceFile, symbol, declaredName, node, searchMeaning, findInStrings, findInComments, result, symbolToIndex);
                    }
                }
            }
```

JKillian:
Additional failure example:
```typescript
class HelloWorld {
    constructor(public name: string) { }
    sayHello() {
        return `Hello, ${this.name}!`;
    }
};

let hello = new HelloWorld("TSLint");
hello.sayHello();
```

This is a bad bug, going to try to get it fixed ASAP.

myitcv:
@JKillian @adidahiya - can I ask, is this the following an instance of the same issue? Or a problem specific to `.tsx` files?

```typescript
/// <reference path="tsd.d.ts" />

import * as React from "react";
interface IMProps {
	Name: string;
}

[...]

Issue #651, opened 2015-09-10
no-unused-variable with import used in other imports


praveennet:
I'm getting false positive when using this rule 'no-unused-variable' for the below code. Could this be fixed please?

```typescript
import ExternalDefinition = require("ExternalDefinition");
import Def = ExternalDefinition.MyNamespace.MyClass1;

export class MyImplementation
    extends MyBaseClasses.BaseClass1
    implements Def.Contract {

}
```

Error: 
> TsLint: unused variable: 'ExternalDefinition'

JKillian:
Hi @praveennet, thanks for the report. I think your issue is the same as #570. Feel free to comment on that one with your code sample!

praveennet:
But here import is declared in same level and used in next line itself. it looks different from #570 to me. ?

adidahiya:
@praveennet your issue is actually a duplicate of #613. I'm not sure if @JKillian's PR (#655) fixes this issue as well; we'll need to investigate.

JKillian:
Interestingly, I found that #655 fixes both cases of #613 but not this issue. Going to reopen this

JKillian:
Actually, this is a duplicate of #325 :smiley:

PR #655, opened 2015-09-11
Fix bug with false positives in unused variable rule and other problems when using language services

### LINKED ISSUE #613: no-unused-variable false positive when referenced as a type (rather than a value)
code samples below, reported in #191 

from @SenHeng:

```ts
import Foo = this.that.Foo;

module some.module.blah {
 export  class bar {
    private foo: Foo;
  }
}
```

`Foo` is a false positive

---

from @nikklassen:

```ts
import DateTimeOpts = Intl.DateTimeFormatOptions;

interface MyDateTimeOpts extends DateTimeOpts {
    timezoneOffset: number;
}

let opts: MyDateTimeOpts;
console.log(opts.timezoneOffset - 1);
```

`DateTimeOpts` is a false positive

JKillian:
See discussion of https://github.com/Microsoft/TypeScript/issues/4560
Fixes https://github.com/palantir/tslint/issues/570
Fixes #613

### DIFF
--- a/src/language/languageServiceHost.ts
+++ b/src/language/languageServiceHost.ts
@@ -21,7 +21,7 @@ module Lint {
             getCurrentDirectory: () => "",
             getDefaultLibFileName: () => "lib.d.ts",
             getScriptFileNames: () => [fileName],
-            getScriptSnapshot: () => ts.ScriptSnapshot.fromString(source),
+            getScriptSnapshot: (name) => ts.ScriptSnapshot.fromString(name === fileName ? source : ""),
             getScriptVersion: () => "1",
             log: (message) => { /* */ }
         };
@@ -30,8 +30,7 @@ module Lint {
     }
 
     export function createLanguageService(fileName: string, source: string) {
-        const documentRegistry = ts.createDocumentRegistry();
         const languageServiceHost = Lint.createLanguageServiceHost(fileName, source);
-        return ts.createLanguageService(languageServiceHost, documentRegistry);
+        return ts.createLanguageService(languageServiceHost);
     }
 }
--- a/test/files/rules/nounusedvariable-falsepositives.test.ts
+++ b/test/files/rules/nounusedvariable-falsepositives.test.ts
@@ -1,0 +1,42 @@
+// case 1
+const fs = require("fs");
+
+module Foo {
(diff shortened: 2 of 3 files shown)


src/language/languageServiceHost.ts
```diff
@@ -30,8 +30,7 @@ module Lint {
     }
 
     export function createLanguageService(fileName: string, source: string) {
-        const documentRegistry = ts.createDocumentRegistry();
```
JKillian: Doing this ourselves was unnecessary as it is done internally automatically by the language service

src/language/languageServiceHost.ts
```diff
@@ -21,7 +21,7 @@ module Lint {
             getCurrentDirectory: () => "",
             getDefaultLibFileName: () => "lib.d.ts",
             getScriptFileNames: () => [fileName],
-            getScriptSnapshot: () => ts.ScriptSnapshot.fromString(source),
+            getScriptSnapshot: name => ts.ScriptSnapshot.fromString(name === fileName ? source : ""),
```
adidahiya: please wrap the function args list in parens

test/files/rules/nounusedvariable-falsepositives.test.ts
```diff
@@ -0,0 +1,18 @@
+const fs = require("fs");
```
adidahiya: can you investigate whether this change fixes #613 and #651 as well? if so, please add those tests to this file

JKillian:
[Made suggested changes]

test/files/rules/nounusedvariable-falsepositives.test.ts
```diff
@@ -0,0 +1,42 @@
...
+// case 3
+import Bar = whatever.that.Foo;
+
+module some.module.blah {
+	export class bar {
+		private bar: Bar;
+		constructor() {
+			console.log(this.bar);
+		}
+	}
```
adidahiya: don't mix tabs and spaces in this file; just use spaces

JKillian: Fixed

adidahiya:
looks good, just waiting for CI

Issue #604, opened 2015-08-24
Enforce ALL_CAPS variables to be declared as const


aicioara:
ALL_CAPS variables, by convention are constants. It would be nice to have a rule that enforces `const` instead of `let` for those.

This might not be needed if https://github.com/palantir/tslint/issues/491 is active, but I prefer this weaker form.

JoshuaKGoldberg:
I would argue that this complements `prefer-const` quite nicely. Together, they enforce that any variable in UPPER_CASE must be `const` and therefore never modified.

reduckted:
Check out [tslint-consistent-codestyle](https://github.com/ajafff/tslint-consistent-codestyle) and the `naming-convention` rule. It's very flexible and you should be able to get the outcome that you're after. 

The only thing is that it sort of works in the opposite direction to what you wanted. Instead of enforcing `const` for uppercase variable names, it would enforce uppercase variable names for `const` variables.

adidahiya:
fixed by #2936

PR #660, opened 2015-09-16
Add all-caps rule

prendradjaja:
Addresses #604.

Enforces the use of `const` for all-caps variable names. For example:

    let PI = 3.1415926535; // bad
    const SECONDS_IN_A_DAY = 86400; // good

### DIFF
--- a/README.md
+++ b/README.md
@@ -129,6 +129,7 @@ A sample configuration file with all options is available [here](https://github.
   * `"parameters"` checks alignment of function parameters.
   * `"arguments"` checks alignment of function call arguments.
   * `"statements"` checks alignment of statements.
+* `all-caps` ensures that all-caps variable names are const.
 * `ban` bans the use of specific functions. Options are ["object", "function"] pairs that ban the use of object.function()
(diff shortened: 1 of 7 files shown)


JKillian:
Thanks for the PR! I think this should be an option on the `variable-name` rule instead of a whole new rule though.

prendradjaja:
Hmm, are you sure? The two rules seem pretty orthogonal to me.

I can imagine a user wanting to enforce this rule but wanting to use a different capitalization convention than the one `variable-name` enforces (camelCase and UPPER_CASE), such as  lower_case and UPPER_CASE.

JKillian:
@adidahiya What do you think?

adidahiya:
Sorry for the delay on this one. I think it should be a new option on `variable-name` as well.

We should be able to support options like this, right?

```json
"variable-name": {
  "check-format": false,
  "ban-keywords": false,
  "all-caps": true
}
```

chrismbarr:
what about this instead?
```json
"variable-name": {
  "check-format": false,
  "ban-keywords": false,
  "all-caps-for-const": true
}
```

PR #664, opened 2015-09-16
unused variable rule shouldn't error on abstract methods with parameters

### LINKED ISSUE #663: unused variable rule and abstract methods
Code below errors about a unused variable x; but abstract methods can't have a body so they are always unused

```
abstract class AbstractTest {
    abstract foo(x);
}
```

**I might fix this soon ;)**

DickvdBrink:
fixes #663

### DIFF
--- a/src/rules/noUnusedVariableRule.ts
+++ b/src/rules/noUnusedVariableRule.ts
@@ -193,7 +193,13 @@ class NoUnusedVariablesWalker extends Lint.RuleWalker {
             }
         }
 
+        // skip checking parameter declarations for abstract methods
+        // They can't have a body so parameters are always unused
+        if (Lint.hasModifier(node.modifiers, ts.SyntaxKind.AbstractKeyword)) {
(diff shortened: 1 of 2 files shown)


src/rules/noUnusedVariableRule.ts
```diff
@@ -193,7 +193,13 @@ class NoUnusedVariablesWalker extends Lint.RuleWalker {
             }
         }
 
+        // skip checking parameter declarations for abstract methods
+        // They can't have a body so parameters are always unused
```
adidahiya: let's just simplify to a one-line comment (this will fix the lint failure as well)

```ts
// abstract methods can't have a body so their parameters are always unused
```

DickvdBrink:
Oops, sorry about that... I did ran the test and when I checked my commit I decided to change the comment... without executing the test again

adidahiya:
thanks @DickvdBrink!

Issue #631, opened 2015-09-02
Refactor AbstractRule and AbstractFormatter into real abstract classes


adidahiya:
https://github.com/Microsoft/TypeScript/wiki/What%27s-new-in-TypeScript#abstract-classes-and-methods

need to make sure this isn't a breaking change for custom rules.

PR #665, opened 2015-09-16
AbstractFormatter and AbstractRule are now abstract classes

DickvdBrink:
refs #631

Don't think this breaks custom rules. Feel free to close if this isn't something you want:)

**edit: I did keep the `Lint.abstract();` in `utils.ts` because removing that would change the api**; it isn't used though

### DIFF
--- a/lib/tslint.d.ts
+++ b/lib/tslint.d.ts
@@ -92,10 +92,10 @@ declare module Lint {
     }
 }
 declare module Lint {
-    class ScopeAwareRuleWalker<T> extends RuleWalker {
+    abstract class ScopeAwareRuleWalker<T> extends RuleWalker {
         private scopeStack;
         constructor(sourceFile: ts.SourceFile, options?: any);
-        createScope(): T;
+        abstract createScope(): T;
         getCurrentScope(): T;
         getAllScopes(): T[];
         getCurrentDepth(): number;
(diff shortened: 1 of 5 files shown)


adidahiya:
@DickvdBrink thanks, this is definitely a change we want, hence the issue I filed :)

@JKillian I did some basic testing, but can you please verify that this change doesn't break custom rules?

JKillian:
Looks good to me :+1:

adidahiya:
Ok, great.

@DickvdBrink can you add a `console.warn` in `Lint.abstract()` noting that it's now deprecated? we'll remove it in the next major release.

DickvdBrink:
Sorry For taking so long, I've been really busy.. Will add the `console.warn` hopefully tomorrow... do you want me to squash it into one will I'm at it? Branch is out of date so I need to merge or rebase anyway :)

adidahiya:
yep, a rebase + squash would be nice once you make that change

adidahiya:
we'll go ahead and add the deprecation notice to `Lint.abstract()` as a follow-up PR.

DickvdBrink:
Thanks for merging this and sorry for not adding the warning ;)

JKillian:
@DickvdBrink  No worries, we appreciate the contributions!

Issue #581, opened 2015-08-19
indent rule enhancements


adidahiya:
Similar to [ESLint](http://eslint.org/docs/rules/indent), we should be able to enforce 2-space or 4-space indentation via options like so:
- [ ] `[true, "spaces", 2]`
- [ ] `[true, "spaces", 4]`

In addition, it'd be nice to have a `"detect"` option (as suggested in #122) which tries to guess the indentation style in a file and enforce consistency:
- [ ] `[true, "detect"]`

IlanFrumer:
+1

nikklassen:
+1 for enforcing number of spaces.

allenhwkim:
+1

glen-84:
Would this only work if the indentation is divisible by the number of spaces configured?

```js
{
    theObject.something()
             .more(); // 13 spaces.
}
```

The above (align the periods) method wouldn't be supported?

JKillian:
@glen-84 I see two possibilities: 

* A line by line check of indentation (where your example wouldn't be allowed
* A statement by statement check of indentation. Since in your example, `theObject.something().more()` is all one statement, it would only check the indentation level at the beginning of it.

I think they both have pros and cons, but the second one is probably preferred here.

nwhite89:
:+1:

jppellerin:
+1

adidahiya:
+1 option 2, checking statement-by-statement

glen-84:
@JKillian Thanks. The second option looks good. =)

tengri:
+1 option 1

leocaseiro:
A good way is the ESLINT one http://eslint.org/docs/rules/indent

radarsu:
+9000 SSJ3
Please, guys, implement 4-spaces indent option ASAP. I started programming in typescript a lot and 2-spaces default indent feels so bad for me. For some people it's really hard to see indent level of some code with only 2 spaces...

leocaseiro:
@radarsu you might want to try use [.editorconfig](http://editorconfig.org/) for the time being.

JKillian:
@radarsu Right now the indentation rule just figures out if you have an indentation error based upon how your code aligns with the previous statement. You're able to use four space indentation and make the rule happy (in fact, that's what the TSLint source code does)

patsissons:
[...]

PR #687, opened 2015-09-26
Implement trailing-comma rule

### LINKED ISSUE #557: add support for always-trailing-comma
Trailing commas are a problem in IE8 and below, which doesn't seem relevant today.

Pajn:
fixes #557


Pajn:
The common reason for this is diffs and blame in SCM tools and that is only valid for multiline which is why I created the rule to only care about that. Shout if you prefer an option.

Used enforce instead of always because that is what used in the readme to describe other rules. Can change that too of course.

README.md
```diff
@@ -137,6 +137,7 @@ A sample configuration file with all options is available [here](https://github.
     * `"check-lowercase"` enforces the rule that the first non-whitespace character of a comment must be lowercase, if applicable
     * `"check-uppercase"` enforces the rule that the first non-whitespace character of a comment must be uppercase, if applicable
 * `curly` enforces braces for `if`/`for`/`do`/`while` statements.
+* `enforce-trailing-comma` enforces trailing comma within multiline object literals.
```
JKillian: Rule can just be named "trailing-comma" (this will obviously change things in a few other places as well)

src/rules/enforceTrailingCommaRule.ts
```diff
@@ -0,0 +1,43 @@
...
+
+export class Rule extends Lint.Rules.AbstractRule {
+    public static FAILURE_STRING = "missing trailing comma";
+
+    public apply(sourceFile: ts.SourceFile): Lint.RuleFailure[] {
+        return this.applyWithWalker(new EnforceTrailingCommaWalker(sourceFile, this.getOptions()));
+    }
+}
+
+class EnforceTrailingCommaWalker extends Lint.RuleWalker {
```
JKillian: This should apply to object literals and array literals

src/rules/enforceTrailingCommaRule.ts
```diff
@@ -0,0 +1,43 @@
...
+    public visitObjectLiteralExpression(node: ts.ObjectLiteralExpression) {
+        const child = node.getChildAt(1);
+        if (child != null && child.kind === ts.SyntaxKind.SyntaxList) {
+
+            if (child.getText().match(/\n|\r/)) {
```
JKillian: I like the options ESLint uses: http://eslint.org/docs/rules/comma-dangle
"always", "always-newline", "never"

Maybe we could add options similar to those? Enforcing trailing commas for multiline literals but not single line literals might not be what everyone wants

JKillian:
Thanks for the PR @Pajn! I left a few high-level comments/change suggestions for you

adidahiya:
This new rule breaks an invariant in `tslint.json` where no config can be contradictory. With this addition, a user can enable `no-trailing-comma` and `trailing-comma` at the same time, which isn't great. Before this PR is accepted we'll need to resolve the contradiction there -- we should probably consolidate into a single `trailing-comma` rule? Thoughts @gscshoyru / @ashwinr?

Pajn:
I could merge the rules but it would be a breaking change to remove the `no-trailing-comma` rule.
I'm thinking about how to handle the options if I did.
Personally I would like to specify "always for newline, never for oneline". Would this set of options be okay "always", "never", "always-newline", "always-newline-never-oneline" or do we need even more options? Could split into "always-newline", "always-oneline", "never-newline" and "never-oneline" where the used would set one newline option and one oneline option. Could also extend with just "always" and "never" for people that want a consequent behaviour.

JKillian:
I think the rules should be merged into a single `trailing-comma` rule, but as @Pajn mentioned, that's a breaking change so we'd have to delay it until the next major release. 

As far as options, I like actually like the four-option proposal, "always-newline", "always-oneline", "never-newline" and "never-oneline".

src/rules/enforceTrailingCommaRule.ts
JKillian: Also [destructuring assignments](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment) maybe? Things like `let [a, b, c] = [1, 2, 3]`

Pajn:
Will take a look at array literals, they doesn't seem to be supported by the walker so I fixed the other comments first.

I used an object for options, just as typedefWhitespace as it felt cleaner and allows any combination. Is that ok?

Pajn:
Added support for array literals and destructing. Also took the freedom to add named imports as those are pretty much the same as object destructing.

Tell me if I should squash the commits

Pajn:
In PR #726 you mention bumping to 3.0 this week. If possible it would be nice to get this into that as well as this is breaking too. So if you have time to review the updated version I would be happy :)

JKillian:
@Pajn We'll be having breaking changes that relate to TSLint 3.0 for a little while, so this will definitely make it into 3.0

Pajn:
Nice :) Will try to be more speedy if there are other things to fix as well.

README.md
```diff
@@ -198,6 +197,10 @@ A sample configuration file with all options is available [here](https://github.
 * `semicolon` enforces semicolons at the end of every statement.
 * `sort-object-literal-keys` checks that keys in object literals are declared in alphabetical order
 * `switch-default` enforces a `default` case in `switch` statements.
+* `trailing-comma` enforces or disallows trailing comma within array and object literals, destructing assignment and named imports.
```
JKillian: spelling for 'destructuring'

JKillian:
Also, can you add an example usage to `docs/sample.tslint.json`

test/rules/trailingCommaRuleTests.ts
```diff
@@ -0,0 +1,116 @@
...
+            const actualFailures = Lint.Test.applyRuleOnFile(fileName, TrailingCommaRule, options);
+
+            Lint.Test.assertFailuresEqual(actualFailures, [
+              expectedFailure1,
+              expectedFailure2,
+              expectedFailure3,
+              expectedFailure4,
+              expectedFailure5,
+              expectedFailure6,
+              expectedFailure7
```
adidahiya: 4-space indentation please (sorry this isn't enforced; see https://github.com/palantir/tslint/issues/581)

test/rules/trailingCommaRuleTests.ts
```diff
@@ -0,0 +1,116 @@
...
```
adidahiya: 4-space indentation

test/tsconfig.json
```diff
@@ -40,6 +40,7 @@
         "./rules/classNameRuleTests.ts",
         "./rules/commentFormatRuleTests.ts",
         "./rules/curlyRuleTests.ts",
+        "./rules/enforceTrailingCommaRuleTests.ts",
```
adidahiya: this should be `trailingCommaRuleTests.ts`

src/rules/tsconfig.json
```diff
@@ -21,6 +21,7 @@
         "./classNameRule.ts",
         "./commentFormatRule.ts",
         "./curlyRule.ts",
+        "./enforceTrailingCommaRule.ts",
```
adidahiya: this should be `trailingCommaRule.ts`

src/rules/trailingCommaRule.ts
```diff
@@ -0,0 +1,78 @@
...
+        }
+    }
+
+    private getOption(option: string) {
+        const allOptions = this.getOptions();
+        if (allOptions == null || allOptions.length === 0) {
+            return null;
+        }
+
+        const options = allOptions[0];
```
adidahiya: just inline the `allOptions[0]` value on the next line

adidahiya:
looks pretty good overall; I like the rule configuration. just need some small fixes and a rebase on the latest master to get this in.

cauboy:
You can find the corresponding docs here: https://palantir.github.io/tslint/rules/trailing-comma/

Issue #676, opened 2015-09-21
New Rule: no-inferrable-types


adidahiya:
As implemented in the TS compiler's custom lint rules: https://github.com/Microsoft/TypeScript/pull/4458

myitcv:
:+1:

PR #690, opened 2015-09-28
New Rule: no-inferrable-types

JKillian:
Resolves https://github.com/palantir/tslint/issues/676

### DIFF
--- a/README.md
+++ b/README.md
@@ -170,6 +170,7 @@ A sample configuration file with all options is available [here](https://github.
 * `no-shadowed-variable` disallows shadowed variable declarations.
 * `no-empty` disallows empty blocks.
 * `no-eval` disallows `eval` function invocations.
+* `no-inferrable-types` disallows explicit type declarations for variables or parameters initialized to a number, string, or boolean
 * `no-internal-module` disallows internal `module`, use `namespace` instead.
 * `no-require-imports` disallows require() style imports
 * `no-string-literal` disallows object access via string literals.
--- a/test/files/rules/noinferrabletypes.test.ts
+++ b/test/files/rules/noinferrabletypes.test.ts
@@ -1,0 +1,21 @@
+// errors, inferrable type is declared
+let x: number = 7;
+let y: boolean = false;
+let z: string = "foo";
+
+// errors, types are inferrable
+function foo (a: number = 5, b: boolean = true, c: string = "bah") { }
+
+// not errors, inferrable type is not declared
+let _x = 7;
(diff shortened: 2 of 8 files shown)


test/files/rules/noinferrabletypes.test.ts
```diff
@@ -0,0 +1,21 @@
...
+let _z = "foo";
+
+// not error, type is not inferrable
+let weird: any = 123;
+
+// not errors, inferrable type is not declared
+function bar(a = 5, b = true, c = "bah") { }
+
+// not errors, types are not inferrable
+function baz(a: any = 5, b: any = true, c: any = "bah") { }
```
adidahiya: this might be out of scope for this PR, but I wonder if we can lint for inferrable return types from functions / class methods?

for example, if a function doesn't `return`, we infer its return type as `void`. we use this convention in our rule walkers (at one point I did a pass and removed a lot of `: void` annotations).

README.md
```diff
@@ -146,7 +146,7 @@ A sample configuration file with all options is available [here](https://github.
 * `jsdoc-format` enforces basic format rules for jsdoc comments -- comments starting with `/**`
     * each line contains an asterisk and asterisks must be aligned
     * each asterisk must be followed by either a space or a newline (except for the first and the last)
-    * the only characters before the asterisk on each line must be whitepace characters
+    * the only characters before the asterisk on each line must be whitespace characters
```
adidahiya: separate PR please :)

adidahiya:
looks pretty good overall; needs merge conflict resolution

test/files/rules/noinferrabletypes.test.ts
JKillian: I thought about this as well. We could probably add a feature for functions w/ one return statement where the declared return type is the same as the type of the return statement. Or reduce the scope and only do it for void, string, number, and boolean.

README.md
JKillian: :smirk:

adidahiya: no, really though

adidahiya:
still waiting on removal of the typo fix in README

JKillian:
[changes made]

test/files/rules/noinferrabletypes.test.ts
adidahiya: cool, can you file a follow-up issue to track that enhancement?

JKillian: See #695

PR #699, opened 2015-09-30
Minor README fixes

adidahiya:
- use more representative reference paths in sample TS code
- fix code block syntax highlighting


README.md
```diff
@@ -249,8 +249,8 @@ Rule names are always camel-cased and *must* contain the suffix `Rule`. Let us t
 Now, let us first write the rule in TypeScript. At the top, we reference TSLint's [definition file](https://github.com/palantir/tslint/blob/master/lib/tslint.d.ts) and the [definition file](https://github.com/palantir/tslint/blob/master/typings/typescriptServices.d.ts) for TypeScript's language services. The exported class must always be named `Rule` and extend from `Lint.Rules.AbstractRule`.
 
 ```typescript
-/// <reference path='typings/typescriptServices.d.ts' />
-/// <reference path='lib/tslint.d.ts' />
+/// <reference path='../node_modules/typescript/lib/typescriptServices.d.ts' />
```
JKillian: Doesn't it make more sense to use the typings bundled w/ tslint? Since the custom rules will be run by tslint using the version of TypeScript services which is bundled w/ tslint, to stay in sync I think it makes sense to use `../node_modules/tslint/typings/typescriptServices.d.ts`

adidahiya: good catch!

adidahiya:
[changes made]

Issue #90, opened 2014-02-21
Normalize encoding to UTF-8 before parsing tslint.json


thethomaseffect:
Right now tslint falls over when fed a config file that's not UTF-8 (UTF-16 BE and UTF-16 LE at the least. Since mixed line endings also aren't handled this would be a good opportunity to fix that too. I propose a simple fix: before parsing file contents convert it to UTF8 using https://www.npmjs.org/package/utf8 then run a simple string replace to normalize the line endings to unix-style. It's really easy for these errors to creep into code so it's good to fix this now. tslint.json is small enough that this fix should be enough for now.

If this sounds good I'll go ahead and get working on a pull request with these fixes.

coderangerdan:
It actually needs to be UTF8 without the BOM ... its the BOM which causes the problem. As it loads in non-UTF8 fine, but doesnt load UTF8 with BOM

sindresorhus:
You could use use [`load-json-file`](https://github.com/sindresorhus/load-json-file) which strips UTF-8 BOM and throws user-friendly error messages on invalid JSON.

PR #703, opened 2015-09-30
remove BOM from utf8 files if present

JKillian:
(This includes slightly refactoring the `findConfiguration` method so that the file is only read in one place instead of two.)


adidahiya:
fixes #90

adidahiya:
lgtm but has merge conflicts

JKillian:
[got rid of merge conflicts]

Issue #714, opened 2015-10-04
member-ordering rule issue with "public-before-private" set


zdychacek:
Hi,

I was wrong with understanding the issue #712, so I am closing it.

There are actual two issues with `member-ordering` rule.

First one, when I set the rule with `public-before-private` option then following code throws: 

```
class Foo {
  private method() {
    const bar = {
      get() {}
    };
  }
}
```

An error:

```
myTest.ts[4, 7]: Declaration of public instance member function not allowed to appear after declaration of private instance member function
```

I think, the solution of this issue could be in walking into object literal and to not take into account remembered previous modifiers when object's method is encountered.

Second issue is IMHO caused by the fact `MemberOrderingWalker` can't handle nested structures, e.g. nested classes.

Following example demonstrates this behaviour:

```
class Foo {
  private fooo() {
    class Baz {
      public bazz () {}
    }
  }

  public pub() {}
}
```

There are no errors reported by TSLint, although there is one - I am able to define public method `pub` after private method `fooo` is defined. That's caused by `Baz` class defining public method `bazz` which resets previous modifiers and therefore I am able to define any public method again.

Everything works as expected when inner class `Baz` is removed - an error is reported by TSLint.

I think, there should be some stack structure remembering previous modifiers for each level of nesting, e.g. inner classes.

Any thoughts on this?

Thanks,

O.

unional:
I encountered the same problem as the second example as @zdychacek describe, but slightly different:
![image](https://cloud.githubusercontent.com/assets/3254987/13717149/74b337e2-e794-11e5-9056-d7cc2558e7a2.png)

The code is:
```ts
export function DispatcherMixin(baseClass: { new (...args: Array<any>): Object }) {
    return class extends baseClass {
        eventLookup: any = {};
        messageTypes = [];
        ...
```

andy-hanson:
I think this is fixed by #1957.

PR #715, opened 2015-10-04
member-ordering-rule ignores object literal shorthand methods

zdychacek:
This MR should fix first issue described in #714.


adidahiya:
subsumed by #801

PR #718, opened 2015-10-05
Switch off conflicting rules in sample.json

JKillian:
This doesn't address the issue of the presence of conflicting rules, but anyone using the sample config be default will have a workable config.


adidahiya:
:+1: merge when CI passes

Issue #677, opened 2015-09-21
New Rule: no-null-keyword


adidahiya:
Creating this issue to see if there's any interest in a `no-null-keyword` rule like the one used in the TS compiler: https://github.com/Microsoft/TypeScript/pull/4458

myitcv:
:+1:

Issue #625, opened 2015-08-31
New Rule: no-implicit-type-coercion


adidahiya:
I'd like to forbid null checks which use implicit type coercion:

``` ts
if (nonBooleanValue) {
    ...
}
```

and instead enforce checks like this:

``` ts
if (nonBooleanValue != null) {
    ...
}
```

adidahiya:
blocked on complete type information, which will be enabled by #680

adidahiya:
related: #253

andy-hanson:
`strict-boolean-expressions` should do this.

OliverJAsh:
`strict-boolean-expressions` does slightly more than just disallowing boolean coercion. I filed a suggestion for a new rule which only disallows boolean coercion: https://github.com/palantir/tslint/issues/4495.

PR #722, opened 2015-10-09
Define no-null-keyword rule

myitcv:
Addresses #677

Comments/thoughts welcomed. I've duly referenced the [TypeScript repo version](https://github.com/Microsoft/TypeScript/blob/7813121c4d77e50aad0eed3152ef1f1156c7b574/scripts/tslint/noNullRule.ts) on which this is shamelessly based.


test/tsconfig.json
```diff
@@ -20,21 +20,18 @@
         "./rules/*.ts"
     ],
     "files": [
+        "../lib/tslint.d.ts",
         "../typings/node.d.ts",
         "../typings/typescriptServices.d.ts",
-        "../lib/tslint.d.ts",
```
myitcv: This is not a negative diff; I've simply sorted the contents.

test/tsconfig.json
```diff
@@ -65,6 +62,7 @@
         "./rules/noEvalRuleTests.ts",
         "./rules/noInferrableTypesRuleTests.ts",
         "./rules/noInternalModuleRuleTests.ts",
+        "./rules/noNullKeywordRuleTests.ts",
```
myitcv: This is the on addition.

myitcv: *one

JKillian:
Can you add this to `sample.tslint.json` and to the README?

JKillian:
Actually, before you do that, one area of discussion. We're not sure how to handle rules that conflict with each other (we accidentally ran into this problem about a week ago). 

This rule would potentially conflict with: https://github.com/palantir/tslint/issues/625

myitcv:
@JKillian added to `sample.tslint.json` and `README` updated (rebased at the same time)

Yes, this might potentially conflict with #625 (and indeed others) but I think it's obvious why it would conflict. The problem arises where rules conflict and its not immediately obvious why.

JKillian:
Given more thought, I'm happy to merge this. I think the solution is to just check for conflicting rules as mentioned in #711. @adidahiya, good with you?

myitcv:
@JKillian just rebased this (saw your comment in #747)

myitcv:
@adidahiya @JKillian - just rebased this

adidahiya:
:+1:

Issue #698, opened 2015-09-30
[rule option] add "react" option to no-unused-variable


adidahiya:
if enabled, the walker would treat JSX nodes as references to `React.createElement`

this fixes the false positive reported in #689

PR #725, opened 2015-10-10
Implement a "react" option for the no-unused-variable rule

myitcv:
Addresses #698 

```
"no-unused-variable", [true, "react"],
```

With such a valid option, the following is __not__ an error:

```typescript
import * as React from "react";

let x = <span></span>;
console.log(x);
```

Thoughts/comments?

### DIFF
--- a/src/rules/noUnusedVariableRule.ts
+++ b/src/rules/noUnusedVariableRule.ts
@@ -16,7 +16,13 @@
 import * as Lint from "../lint";
 import * as ts from "typescript";
 
+const OPTION_REACT = "react";
 const OPTION_CHECK_PARAMETERS = "check-parameters";
+
+const REACT_MODULES = ["react", "react/addons"];
+const REACT_NAMESPACE_IMPORT_NAME = "React";
+
+const MODULE_SPECIFIER_MATCH = /^["'](.+)['"]$/;
 
 export class Rule extends Lint.Rules.AbstractRule {
     public static FAILURE_STRING = "unused variable: ";
@@ -33,11 +39,50 @@ class NoUnusedVariablesWalker extends Lint.RuleWalker {
     private skipParameterDeclaration: boolean;
     private skipVariableDeclaration: boolean;
 
+    private reactImport: ts.NamespaceImport;
+    private reactUsed: boolean;
+    private hasSeenJsxElement: boolean;
+
     constructor(sourceFile: ts.SourceFile, options: Lint.IOptions, languageService: ts.LanguageService) {
         super(sourceFile, options);
         this.languageService = languageService;
         this.skipVariableDeclaration = false;
         this.skipParameterDeclaration = false;
+        this.reactUsed = false;
+        this.hasSeenJsxElement = false;
+    }
+
+    public visitSourceFile(node: ts.SourceFile) {
+        super.visitSourceFile(node);
+
+        // after a this.visitSourceFile is completed, this.reactImport will be set to a NamespaceImport
+        // iff:
+        //
+        // + a react option has been provided to the rule and
+        // + an import of a module that matches one of OPTION_REACT_MODULES is found, to a
+        //   NamespaceImport named OPTION_REACT_NAMESPACE_IMPORT_NAME
+        //
+        // e.g.
+        //
+        // import * as React from "react/addons";
+        //
+        // If reactImport is defined when a walk is completed, we need to have:
+        //
+        // a) seen another usage of React and/or
+        // b) seen a JSX identifier
+        //
+        // otherwise a a variable usage failure will will be reported
+        if (this.hasOption(OPTION_REACT) && this.reactImport !== undefined) {
+            if (!this.reactUsed && !this.hasSeenJsxElement) {
+                let name = this.reactImport.name.getText();
+                this.addFailure(this.createFailure(this.reactImport.name.getStart(), name.length, `${Rule.FAILURE_STRING}'${name}'`));
+            }
+        }
+    }
+
+    public visitJsxElement(node: ts.JsxElement) {
+        this.hasSeenJsxElement = true;
+        super.visitJsxElement(node);
     }
 
     public visitBindingElement(node: ts.BindingElement) {
@@ -88,7 +133,26 @@ class NoUnusedVariablesWalker extends Lint.RuleWalker {
     }
 
     public visitNamespaceImport(node: ts.NamespaceImport) {
-        this.validateReferencesForVariable(node.name.text, node.name.getStart());
+        let importDeclaration = <ts.ImportDeclaration>node.parent.parent;
+        let moduleSpecifier = importDeclaration.moduleSpecifier.getText();
+
+        // extract the unquoted module being imported
+        let match = moduleSpecifier.match(MODULE_SPECIFIER_MATCH);
+
+        let isReactImport = match && REACT_MODULES.indexOf(match[1]) !== -1;
+        if (this.hasOption(OPTION_REACT)
+                && isReactImport
+                && node.name.text === REACT_NAMESPACE_IMPORT_NAME) {
+            this.reactImport = node;
+            let fileName = this.getSourceFile().fileName;
+            let position = node.name.getStart();
+            let highlights = this.languageService.getDocumentHighlights(fileName, position, [fileName]);
+            if (highlights != null && highlights[0].highlightSpans.length > 1) {
+                this.reactUsed = true;
+            }
+        } else {
+            this.validateReferencesForVariable(node.name.text, node.name.getStart());
+        }
         super.visitNamespaceImport(node);
     }
 
--- a/test/rules/noUnusedVariableRuleTests.ts
+++ b/test/rules/noUnusedVariableRuleTests.ts
@@ -113,4 +113,66 @@ describe("<no-unused-variable>", () => {
 
         assert.lengthOf(actualFailures, 0);
     });
+
+    describe("with react option", () => {
+        describe("set", () => {
+            function shared(importModuleName: string) {
+                const suffix = moduleNameToFilenameSuffix(importModuleName);
+                describe("importing \"${importModuleName}\"", () => {
+                    it("should not fail if only JSX present", () => {
+                        pass(`rules/nounusedvariable-react1-${suffix}.test.tsx`, true);
+                    });
+
+                    it("should not fail if React namespace is used", () => {
+                        pass(`rules/nounusedvariable-react2-${suffix}.test.tsx`, true);
+                    });
+
+                    it("should fail if neither JSX nor React namespace is seen", () => {
+                        fail(`rules/nounusedvariable-react3-${suffix}.test.tsx`, true);
+                    });
+                });
+            }
+
+            shared("react");
+            shared("react/addons");
+        });
+
+        describe("not set", () => {
+            function shared(importModuleName: string) {
+                const suffix = moduleNameToFilenameSuffix(importModuleName);
+                describe("importing \"${importModuleName}\"", () => {
+                    it("should fail if only JSX present", () => {
+                        fail(`rules/nounusedvariable-react1-${suffix}.test.tsx`, false);
+                    });
+
+                    it("should not fail if React namespace is used", () => {
+                        pass(`rules/nounusedvariable-react2-${suffix}.test.tsx`, false);
+                    });
+
+                    it("should fail if neither JSX nor React namespace is seen", () => {
+                        fail(`rules/nounusedvariable-react3-${suffix}.test.tsx`, false);
+                    });
+                });
+            }
+
+            shared("react");
+            shared("react/addons");
+        });
+
+        function pass(fileName: string, isReactOptionSet: boolean) {
+            const actualFailures = Lint.Test.applyRuleOnFile(fileName, Rule, isReactOptionSet ? [true, "react"] : [true]);
+            assert.lengthOf(actualFailures, 0);
+        }
+
+        function fail(fileName: string, isReactOptionSet: boolean) {
+            const failure1 = Lint.Test.createFailuresOnFile(fileName, Rule.FAILURE_STRING + "'React'")([1, 13], [1, 18]);
+            const actualFailures = Lint.Test.applyRuleOnFile(fileName, Rule, isReactOptionSet ? [true, "react"] : [true]);
+            assert.lengthOf(actualFailures, 1);
+            Lint.Test.assertContainsFailure(actualFailures, failure1);
+        }
+
+        function moduleNameToFilenameSuffix(moduleName: string): string {
+            return moduleName.replace(/\//g, "_");
+        }
+    });
 });
--- a/README.md
+++ b/README.md
@@ -181,6 +181,8 @@ A sample configuration file with all options is available [here](https://github.
 * `no-unused-variable` disallows unused imports, variables, functions and private class members.
     * `"check-parameters"` disallows unused function and constructor parameters.
         * NOTE: this option is experimental and does not work with classes that use abstract method declarations, among other things. Use at your own risk.
+    * `"react"` relaxes the rule for a namespace import named `React` (from either the module `"react"` or `"react/addons"`) to also consider JSX expressions
+      uses of the import
 * `no-use-before-declare` disallows usage of variables before their declaration.
 * `no-var-keyword` disallows usage of the `var` keyword, use `let` or `const` instead.
 * `no-var-requires` disallows the use of require statements except in import statements, banning the use of forms such as `var module = require("module")`
--- a/test/files/rules/nounusedvariable-react1-react.test.tsx
+++ b/test/files/rules/nounusedvariable-react1-react.test.tsx
@@ -1,0 +1,3 @@
+import * as React from "react";
+
+console.log(<div></div>);
--- a/test/files/rules/nounusedvariable-react1-react_addons.test.tsx
+++ b/test/files/rules/nounusedvariable-react1-react_addons.test.tsx
@@ -1,0 +1,3 @@
+import * as React from "react/addons";
+
+console.log(<div></div>);
--- a/test/files/rules/nounusedvariable-react2-react.test.tsx
+++ b/test/files/rules/nounusedvariable-react2-react.test.tsx
@@ -1,0 +1,3 @@
+import * as React from "react";
+
+export class MyComponent extends React.Component<{}, {}> {}
--- a/test/files/rules/nounusedvariable-react2-react_addons.test.tsx
+++ b/test/files/rules/nounusedvariable-react2-react_addons.test.tsx
@@ -1,0 +1,3 @@
+import * as React from "react/addons";
+
+export class MyComponent extends React.Component<{}, {}> {}
--- a/test/files/rules/nounusedvariable-react3-react.test.tsx
+++ b/test/files/rules/nounusedvariable-react3-react.test.tsx
@@ -1,0 +1,1 @@
+import * as React from "react";
--- a/test/files/rules/nounusedvariable-react3-react_addons.test.tsx
+++ b/test/files/rules/nounusedvariable-react3-react_addons.test.tsx
@@ -1,0 +1,1 @@
+import * as React from "react/addons";


myitcv:
Rebased against `master` following merge of #726

myitcv:
@JKillian @adidahiya just rebased this

JKillian:
Ideally, it would be nicer to just accept one option named 'react' and not worry about prefixes. Especially since React 0.14 moved addons to their own npm packages (and hopefully bower packages?). Are there any other legit use cases other than 'react' and 'react/addons'?

myitcv:
> Are there any other legit use cases other than 'react' and 'react/addons'?

Not sure, hence why I coded a bit more defensively here... but on reflection it's much simpler to have the modules `"react"` and `"react/addons"` as constants which makes the option just `"react"`.

I'll push up another commit that makes that change... then we can choose between the two (and then squash commits)

adidahiya:
I'd prefer only supporting `"react"`.

JKillian:
Why not support both? I assume using `"react/addons"` is pretty common

adidahiya:
Sorry, I should've been more clear --

I would like the _rule option_ to be named `"react"`. When looking for imports, we should look for `... from "react"` and `... from "react/*"`.

JKillian:
Oh gotchya. I was thinking have the rule named `"react"` only and when looking for imports only look for `"react"` and `"react/addons"`.

jazzik:
hi folks. I just found this discussion as I saw the same thing in my project. I also have `"unused variable: 'defaultProps'"`, do you think it would make sense to 'whitelist' that as well when `"react"` option is set?

JKillian:
@jazzik Can you post an example of your code? I believe as long as it's not private, you shouldn't get that error. Something like this shouldn't give a lint error:

```
    interface Props { foo: number; }
    export class ReactComponent extends React.Component<Props, {}> {
        public static defaultProps = {
            foo: 12
        };
        ...
    }
```

jazzik:
@JKillian oh right, I had it as private :blush:

myitcv:
@JKillian @adidahiya pushed up a second commit with a single `"react"` option. The list of valid React modules is now maintained as a `const` in the rule.

myitcv:
@adidahiya @JKillian - any thoughts on this? Only reason for me pushing is that we've got a whole load of `/* tslint:disable:no-unused-variable */` lines kicking around our code which I'm keen to remove :smile:

src/rules/noUnusedVariableRule.ts
```diff
@@ -16,8 +16,13 @@
 import * as Lint from "../lint";
 import * as ts from "typescript";
 
+const OPTION_REACT = "react";
+const OPTION_REACT_MODULES = ["react", "react/addons"];
+const OPTION_REACT_NAMESPACE_IMPORT_NAME = "React";
```
JKillian: Can these two consts not start with `OPTION_`? I think that should be reserved only for actual options which are provided to the rule

src/rules/noUnusedVariableRule.ts
```diff
@@ -33,18 +38,57 @@ class NoUnusedVariablesWalker extends Lint.RuleWalker {
...
         super(sourceFile, options);
         this.languageService = languageService;
         this.skipVariableDeclaration = false;
         this.skipParameterDeclaration = false;
+        this.reactImport = undefined;
+        this.reactUsed = false;
+        this.seenJsx = false;
+    }
+
+    public walk(node: ts.Node): void {
```
JKillian: Override `visitSourceFile` instead of `walk` for actions that should happen at the end of traversal. The `walk` method is technically more generic any could be called on any node (I guess). At the least, using `visitSourceFile` is more consistent with the other rules

src/rules/noUnusedVariableRule.ts
```diff
@@ -33,18 +38,57 @@ class NoUnusedVariablesWalker extends Lint.RuleWalker {
...
+        if (this.hasOption(OPTION_REACT) && this.reactImport) {
+            if (!this.reactUsed && !this.seenJsx) {
+                let name = this.reactImport.name.getText();
+                this.addFailure(this.createFailure(this.reactImport.pos, name.length, `${Rule.FAILURE_STRING}'${name}'`));
+            }
+        }
+    }
+
+    public visitJsxElement(node: ts.JsxElement): void {
+        this.seenJsx = true;
```
JKillian: call super after this

src/rules/noUnusedVariableRule.ts
```diff
@@ -88,7 +132,22 @@ class NoUnusedVariablesWalker extends Lint.RuleWalker {
     }
 
     public visitNamespaceImport(node: ts.NamespaceImport) {
-        this.validateReferencesForVariable(node.name.text, node.name.getStart());
+        let importDecl = <ts.ImportDeclaration>node.parent.parent;
+        let modSpec = importDecl.moduleSpecifier.getText();
+        let match = modSpec.match(MODULE_SPECIFIER_MATCH);
```
JKillian: maybe add a comment explaining something like "match quoted import names and retrieve name without quotes"? Feel free to reword that, just that the intent of regex matches can be a little hazy w/o further explanation

src/rules/noUnusedVariableRule.ts
```diff
@@ -88,7 +132,22 @@ class NoUnusedVariablesWalker extends Lint.RuleWalker {
...
+        let isReactImport = match && OPTION_REACT_MODULES.indexOf(match[1]) !== -1;
+        if (this.hasOption(OPTION_REACT) &&
+            isReactImport && node.name.text === OPTION_REACT_NAMESPACE_IMPORT_NAME) {
```
JKillian: let's put each condition on its own line, like line 219 below

test/rules/noUnusedVariableRuleTests.ts
```diff
@@ -113,4 +113,70 @@ describe("<no-unused-variable>", () => {
...
+
+            it("should detect usage of React.*", () => {
+                const fileName = "rules/nounusedvariable-react4.test.tsx";
+                const actualFailures = Lint.Test.applyRuleOnFile(fileName, Rule, [true, "react"]);
+
+                assert.lengthOf(actualFailures, 0);
+            });
+        });
+
+        describe("sometingelse", () => {
```
JKillian: spelling

test/rules/noUnusedVariableRuleTests.ts
```diff
@@ -113,4 +113,70 @@ describe("<no-unused-variable>", () => {
 
         assert.lengthOf(actualFailures, 0);
     });
+
+    describe("react option set to", () => {
```
JKillian: these describe/it blocks could use a little refactoring now that there is only one possible react option. I think the cases should be something like:
React Option Set:
  * No error if JSX present
  * No error if React. usage present
  * error if neither is present
React Option Not Set:
  * Failure if no usage of React
  * Failure if only JSX
  * No failure if React. is present

(and then also a couple similar tests for react/addons)

JKillian:
@myitcv Sorry for the delays, things (as always)  have been busy. Feedback left for you!

test/rules/noUnusedVariableRuleTests.ts
myitcv: Good points; actually forgot to refactor the test cases given the change in option. Will do

src/rules/noUnusedVariableRule.ts
myitcv: Thanks for the pointer

src/rules/noUnusedVariableRule.ts
myitcv: Interestingly is there a good way to enforce that `super` is called in certain situations? It not I smell another rule....

myitcv:
@JKillian comments hopefully addressed. Apologies, I missed reworking the specs in my last commit. The shared specs cover the matrix of cases (and can be extended if required for additional modules)

src/rules/noUnusedVariableRule.ts
JKillian: Not that I know of! It's a good rule idea, although there are definitely cases when you want t override a method and _not_ call super as well

myitcv: > although there are definitely cases when you want t override a method and not call super as well

Absolutely; not trying to cover those cases, only trying to cover those situations where you _know_ you must.

src/rules/noUnusedVariableRule.ts
```diff
@@ -16,8 +16,13 @@
 import * as Lint from "../lint";
 import * as ts from "typescript";
 
+const OPTION_REACT = "react";
+const REACT_MODULES = ["react", "react/addons"];
+const REACT_NAMESPACE_IMPORT_NAME = "React";
```
JKillian: I'd also move these two constants below the two OPTION constants

src/rules/noUnusedVariableRule.ts
```diff
@@ -88,7 +133,26 @@ class NoUnusedVariablesWalker extends Lint.RuleWalker {
...
-        this.validateReferencesForVariable(node.name.text, node.name.getStart());
+        let importDecl = <ts.ImportDeclaration>node.parent.parent;
+        let modSpec = importDecl.moduleSpecifier.getText();
+
+        // extract the unquoted module being imported
+        let match = modSpec.match(MODULE_SPECIFIER_MATCH);
+
+        let isReactImport = match && REACT_MODULES.indexOf(match[1]) !== -1;
+        if (this.hasOption(OPTION_REACT) &&
+            isReactImport &&
```
JKillian: we normally put the && operators on the next line instead of trailing the previous line

JKillian:
Also, README needs to be updated with the `react` option. These are just style things though, we can always do them in a follow-up commit. I think @adidahiya has a few comments first though as well. Hopefully we can get this merged in sometime tomorrow for tomorrow's release

myitcv:
@JKillian re-pushed with suggested changes, added to `README` and squashed commits.

src/rules/noUnusedVariableRule.ts
```diff
@@ -33,18 +39,58 @@ class NoUnusedVariablesWalker extends Lint.RuleWalker {
...
+    // e.g.
+    //
+    // import * as React from "react/addons";
+    //
+    // If reactImport is defined when a walk is completed, we need to have:
+    //
+    // a) seen another usage of React and/or
+    // b) seen a JSX identifier
+    //
+    // otherwise a usage error will be reported
```
adidahiya: change this to "variable usage failure"

"error" is ambiguous in JS, so we should stick to the "failure" terminology.

src/rules/noUnusedVariableRule.ts
```diff
@@ -33,18 +39,58 @@ class NoUnusedVariablesWalker extends Lint.RuleWalker {
...
+    private reactImport: ts.NamespaceImport;
+    private reactUsed: boolean;
+    private seenJsx: boolean;
```
adidahiya: `hasSeenJsxElement`

src/rules/noUnusedVariableRule.ts
```diff
@@ -33,18 +39,58 @@ class NoUnusedVariablesWalker extends Lint.RuleWalker {
...
+
     constructor(sourceFile: ts.SourceFile, options: Lint.IOptions, languageService: ts.LanguageService) {
         super(sourceFile, options);
         this.languageService = languageService;
         this.skipVariableDeclaration = false;
         this.skipParameterDeclaration = false;
+        this.reactImport = undefined;
```
adidahiya: no need to set this to undefined; it starts off that way

src/rules/noUnusedVariableRule.ts
```diff
@@ -33,18 +39,58 @@ class NoUnusedVariablesWalker extends Lint.RuleWalker {
...
+        this.reactUsed = false;
+        this.seenJsx = false;
+    }
+
+    public visitSourceFile(node: ts.SourceFile): void {
+        super.visitSourceFile(node);
+
+        if (this.hasOption(OPTION_REACT) && this.reactImport) {
```
adidahiya: don't implicitly typecast; use `this.reactImport != null`

src/rules/noUnusedVariableRule.ts
```diff
@@ -33,18 +39,58 @@ class NoUnusedVariablesWalker extends Lint.RuleWalker {
...
+            if (!this.reactUsed && !this.seenJsx) {
+                let name = this.reactImport.name.getText();
+                this.addFailure(this.createFailure(this.reactImport.name.getStart(), name.length, `${Rule.FAILURE_STRING}'${name}'`));
+            }
+        }
+    }
+
+    public visitJsxElement(node: ts.JsxElement): void {
```
adidahiya: omit the `: void` type annotation, it's unnecessary

src/rules/noUnusedVariableRule.ts
```diff
@@ -33,18 +39,58 @@ class NoUnusedVariablesWalker extends Lint.RuleWalker {
     private skipParameterDeclaration: boolean;
     private skipVariableDeclaration: boolean;
 
+    // after a this.walk is completed, this.reactImport will be set to a NamespaceImport
```
adidahiya: please move this comment block into `visitSourceFile`, where it is more relevant.

src/rules/noUnusedVariableRule.ts
```diff
@@ -33,18 +39,58 @@ class NoUnusedVariablesWalker extends Lint.RuleWalker {
...
         super(sourceFile, options);
         this.languageService = languageService;
         this.skipVariableDeclaration = false;
         this.skipParameterDeclaration = false;
+        this.reactImport = undefined;
+        this.reactUsed = false;
+        this.seenJsx = false;
+    }
+
+    public visitSourceFile(node: ts.SourceFile): void {
```
adidahiya: omit the `: void` type annotation, it's unnecessary

src/rules/noUnusedVariableRule.ts
```diff
@@ -33,18 +39,58 @@ class NoUnusedVariablesWalker extends Lint.RuleWalker {
...
+        this.seenJsx = true;
+        super.visitJsxElement(node);
     }
 
     public visitBindingElement(node: ts.BindingElement) {
         const isSingleVariable = node.name.kind === ts.SyntaxKind.Identifier;
 
         if (isSingleVariable && !this.skipBindingElement) {
-            const variableIdentifier = <ts.Identifier> node.name;
+            const variableIdentifier = <ts.Identifier>node.name;
```
adidahiya: please add a space after the typecast

src/rules/noUnusedVariableRule.ts
```diff
@@ -88,7 +134,26 @@ class NoUnusedVariablesWalker extends Lint.RuleWalker {
...
+        let importDecl = <ts.ImportDeclaration>node.parent.parent;
+        let modSpec = importDecl.moduleSpecifier.getText();
+
+        // extract the unquoted module being imported
+        let match = modSpec.match(MODULE_SPECIFIER_MATCH);
+
+        let isReactImport = match && REACT_MODULES.indexOf(match[1]) !== -1;
+        if (this.hasOption(OPTION_REACT)
+            && isReactImport
+            && node.name.text === REACT_NAMESPACE_IMPORT_NAME) {
```
adidahiya: indent both `&&` one more level

src/rules/noUnusedVariableRule.ts
```diff
@@ -88,7 +134,26 @@ class NoUnusedVariablesWalker extends Lint.RuleWalker {
     }
 
     public visitNamespaceImport(node: ts.NamespaceImport) {
-        this.validateReferencesForVariable(node.name.text, node.name.getStart());
+        let importDecl = <ts.ImportDeclaration>node.parent.parent;
```
adidahiya: avoid abbreviation, call this `importDeclaration`

src/rules/noUnusedVariableRule.ts
```diff
@@ -88,7 +134,26 @@ class NoUnusedVariablesWalker extends Lint.RuleWalker {
...
+        let modSpec = importDecl.moduleSpecifier.getText();
```
adidahiya: avoid abbreviation, call this `moduleSpecifier`

src/rules/noUnusedVariableRule.ts
```diff
@@ -96,7 +161,7 @@ class NoUnusedVariablesWalker extends Lint.RuleWalker {
         const isSingleVariable = node.name.kind === ts.SyntaxKind.Identifier;
 
         if (isSingleVariable && !this.skipVariableDeclaration) {
-            const variableIdentifier = <ts.Identifier> node.name;
+            const variableIdentifier = <ts.Identifier>node.name;
```
adidahiya: space after typecast

src/rules/noUnusedVariableRule.ts
```diff
@@ -161,7 +226,7 @@ class NoUnusedVariablesWalker extends Lint.RuleWalker {
             && isSingleVariable
             && !this.skipParameterDeclaration
             && !Lint.hasModifier(node.modifiers, ts.SyntaxKind.PublicKeyword)) {
-            const nameNode = <ts.Identifier> node.name;
+            const nameNode = <ts.Identifier>node.name;
```
adidahiya: space after typecast pls

src/rules/noUnusedVariableRule.ts
```diff
@@ -173,7 +238,7 @@ class NoUnusedVariablesWalker extends Lint.RuleWalker {
     public visitPropertyDeclaration(node: ts.PropertyDeclaration) {
         if (node.name != null && node.name.kind === ts.SyntaxKind.Identifier) {
             const modifiers = node.modifiers;
-            const variableName = (<ts.Identifier> node.name).text;
+            const variableName = (<ts.Identifier>node.name).text;
```
adidahiya: revert this diff pls

src/rules/noUnusedVariableRule.ts
```diff
@@ -188,7 +253,7 @@ class NoUnusedVariablesWalker extends Lint.RuleWalker {
     public visitMethodDeclaration(node: ts.MethodDeclaration) {
         if (node.name != null && node.name.kind === ts.SyntaxKind.Identifier) {
             const modifiers = node.modifiers;
-            const variableName = (<ts.Identifier> node.name).text;
+            const variableName = (<ts.Identifier>node.name).text;
```
adidahiya: revert this diff pls

src/rules/noUnusedVariableRule.ts
```diff
@@ -88,7 +134,26 @@ class NoUnusedVariablesWalker extends Lint.RuleWalker {
...
+
+        let isReactImport = match && REACT_MODULES.indexOf(match[1]) !== -1;
+        if (this.hasOption(OPTION_REACT)
+            && isReactImport
+            && node.name.text === REACT_NAMESPACE_IMPORT_NAME) {
+            this.reactImport = node;
+            let fileName = this.getSourceFile().fileName;
+            let position = node.name.getStart();
+            let highlights = this.languageService.getDocumentHighlights(fileName, position, [fileName]);
+            if (highlights && highlights[0].highlightSpans.length > 1) {
```
adidahiya: `highlights != null`

test/rules/noUnusedVariableRuleTests.ts
```diff
@@ -113,4 +113,66 @@ describe("<no-unused-variable>", () => {
 
         assert.lengthOf(actualFailures, 0);
     });
+
+    describe("with react option", () => {
+        function pass(fileName: string, reactOptSet: boolean) {
```
adidahiya: I'd prefer these function & arg names:
`function assertNoFailures(fileName: string, isReactOptionSet: boolean) {`

test/rules/noUnusedVariableRuleTests.ts
```diff
@@ -113,4 +113,66 @@ describe("<no-unused-variable>", () => {
...
+            const actualFailures = Lint.Test.applyRuleOnFile(fileName, Rule, reactOptSet ? [true, "react"] : [true]);
+            assert.lengthOf(actualFailures, 0);
+        }
+
+        function fail(fileName: string, reactOptSet: boolean) {
```
adidahiya: I'd prefer these function & arg names:
`function assertSingleFailure(fileName: string, isReactOptionSet: boolean) {`

test/rules/noUnusedVariableRuleTests.ts
```diff
@@ -113,4 +113,66 @@ describe("<no-unused-variable>", () => {
...
+            assert.lengthOf(actualFailures, 1);
+            Lint.Test.assertContainsFailure(actualFailures, failure1);
+        }
+
+        function modToSuffix(mod: string): string {
+            return mod.replace(/\//g, "_");
+        }
+
+        describe("set", () => {
+            function shared(mod: string) {
```
adidahiya: I'd prefer these function & arg names:
`function describeSuite(importModuleName: srting) {`

test/rules/noUnusedVariableRuleTests.ts
```diff
@@ -113,4 +113,66 @@ describe("<no-unused-variable>", () => {
...
+        }
+
+        function fail(fileName: string, reactOptSet: boolean) {
+            const failure1 = Lint.Test.createFailuresOnFile(fileName, Rule.FAILURE_STRING + "'React'")([1, 13], [1, 18]);
+            const actualFailures = Lint.Test.applyRuleOnFile(fileName, Rule, reactOptSet ? [true, "react"] : [true]);
+            assert.lengthOf(actualFailures, 1);
+            Lint.Test.assertContainsFailure(actualFailures, failure1);
+        }
+
+        function modToSuffix(mod: string): string {
```
adidahiya: let's make the name more descriptive: `moduleNameToFilenameSuffix`

test/rules/noUnusedVariableRuleTests.ts
adidahiya: also, please move all helper methods to the bottom of this describe block. Test suites should be read first, then helper methods later.

src/rules/noUnusedVariableRule.ts
myitcv: Happy to go with whatever you prefer here. 

However, in our project we enforce the `"call-signature"` option of the `typedef` rule. Reason being, it's clearer if function signatures are explicit:

```typescript
function A() {
    return 5;
}
```

src/rules/noUnusedVariableRule.ts
myitcv: I look forward to #732 landing and it being enforced as part of the build (which will mean each file can be automatically formatted)!

src/rules/noUnusedVariableRule.ts
myitcv: Apologies, these have crept in because I automatically formatted the file using `tsfmt`

myitcv:
@adidahiya - all points addressed in most recent push. Apologies, a number of the whitespace changes crept in because my editor automatically runs `tsfmt` on save. Look forward to #732 landing and being enforced in the build :smile:

Issue #532, opened 2015-07-27
Expose testing utilities (helper.ts) in a d.ts file and in npm package


HamletDRC:
I have written several custom tslint rules. 
I have some ad-hoc testing utilities written to test them, but I would rather just reuse the framework for testing rules that is defined in the tslint source code. It would make it easier for me to test my own rules and also easier to port back to the main source code and submit as pull requests. 

What I /think/ I need is: 
- the code from helper.ts exposed in a tslint-test.d.ts
- the code from helper.ts (and dependencies) compiled into tslint-test.js

Please note: I may not know what I want technically. From a feature standpoint... I just want my tests to look exactly like your tests!

adidahiya:
This is definitely a useful feature and we do want to expose testing utilities for custom rules.

I think the easiest way to do this would be to refactor our testing utilities into a `TestLib` object which we then export from our current npm module + include its typings in `tslint.d.ts`. The goal would be to allow users to write `import { TestLib } from "tslint";`

Thoughts @gscshoyru, @ashwinr?

gscshoyru:
I feel a little strange exposing test code in the main repository -- but eslint does it, and the fact that we have pluggable rules is meant to be a feature. So I think that your idea makes sense. 

Except: I think we might want to split the testlib typings out into another d.ts file, since those types/functionality should only be available in tests.

adidahiya:
+1 for exposing the typings separately. although that makes our build a little more complicated...

adidahiya:
This is being implemented in #900. With that PR we're just going to do the straightforward thing and leave the typings where they are, accessible to users like this:

```ts
// names of these exports might change
import {runTest, consoleTestResultHandler} from "tslint/lib/test";

const didAllTestsPass = consoleTestResultHandler(runTest("path/to/my/testing/dir/"));
if (!didAllTestsPass) {
    process.exit(1);
}
```

JKillian:
Or, the `--test` command will provide a way to test rules via the CLI

andy-hanson:
Can this be closed then?

Issue #280, opened 2015-01-31
import typescript


ashwinr:
as a node dependency using `require` within tslint

PR #726, opened 2015-10-12
Modularize

### LINKED ISSUE #549: Consider referencing typescriptServices.d.ts in node_modules directly
Instead of having a local copy in our repo in the `typings/` folder, we might want to reference the one included in `node_modules/typescript/lib/` directly.

weswigham:
Implements #465, #280.

This PR rearchitects tslint to be a series of modules rather than a bunch of merged namespaces meant to be concatenated together. What this has enabled is a handful of cool TS 1.6 things - for one, tslint now has typescript 1.6 compatible `typings` built alongside itself - meaning `import * as Linter from "tslint";` (and `import * as Lint from "tslint/lib/lint";`) can now pull associated typings from the node package.

It does, however, _retain_ the namespace structure tslint used before - it just creates it via modules now, rather than TS's namespace keyword. This means that, given the right entrypoint (namely `tslint/lib/lint`), the API surface should look identical to before this change.

This _does_ contain a fairly large breaking change, however, but in doing so it fixes some odd things with the tslint build and makes it a bit less of a chore to integrate into your builds. For one, it no longer concatenates all of typescript into its built js, and for two, it no longer creates a global `ts` or `Lint` object (thereby mucking with people's build environments like [so](https://github.com/Microsoft/TypeScript/issues/4929)). Combined with the typings field support, this means external rules and formatters must replace their lines like this:
```ts
/// <reference path="../../node_modules/tslint/typings/typescriptServices.d.ts" />
/// <reference path="../../node_modules/tslint/lib/tslint.d.ts" />
```
with this:
```ts
import * as Lint from "tslint/lib/lint";
```

Additionally, this fixes #549, allows #532 to be done more easily.


weswigham:
And after looking at the repo convention beforehand, I now realize I should remove the built output from the PR.

weswigham:
I'm seeing some odd behavior on the cli relating to default arguments for the formatter not getting used... I'm not wholly sure what's up with that.

adidahiya:
awesome! we'll try to merge this this week and bump to 3.0 for the breaking change.

Gruntfile.js
```diff
@@ -80,81 +63,37 @@ module.exports = function (grunt) {
...
                 ],
-                out: "build/tslint-tests.js"
+                outDir: "build/"
             }
         }
     });
 
     // load NPM tasks
     grunt.loadNpmTasks("grunt-contrib-clean");
-    grunt.loadNpmTasks("grunt-contrib-concat");
```
adidahiya: let's remove grunt-contrib-concat from package.json as well

src/language/utils.ts
```diff
@@ -14,111 +14,111 @@
...
-        if (modifiers == null || modifierKinds == null) {
-            return false;
-        }
-
-        return modifiers.some((m) => {
-            return modifierKinds.some((k) => m.kind === k);
-        });
+/**
+    * @returns true if any modifier kinds passed along exist in the given modifiers array
+    */
```
adidahiya: fix indentation please

src/language/utils.ts
```diff
@@ -14,111 +14,111 @@
...
+    });
+}
 
-        return isNodeFlagSet(parentNode, ts.NodeFlags.Let)
-            || isNodeFlagSet(parentNode, ts.NodeFlags.Const);
-    }
+/**
+    * Determines if the appropriate bit in the parent (VariableDeclarationList) is set,
+    * which indicates this is a "let" or "const".
+    */
```
adidahiya: indentation

src/language/utils.ts
```diff
@@ -14,111 +14,111 @@
...
-     * Bitwise check for node flags.
-     */
-    export function isNodeFlagSet(node: ts.Node, flagToCheck: ts.NodeFlags): boolean {
-        /* tslint:disable:no-bitwise */
-        return (node.flags & flagToCheck) !== 0;
-        /* tslint:enable:no-bitwise */
-    }
+/**
+    * Bitwise check for node flags.
+    */
```
adidahiya: indentation

src/language/walker/blockScopeAwareRuleWalker.ts
```diff
@@ -13,71 +13,71 @@
...
-    /**
-     * An AST walker that is aware of block scopes in addition to regular scopes. Block scopes
-     * are a superset of regular scopes (new block scopes are created more frequently in a program).
-     */
-    export abstract class BlockScopeAwareRuleWalker<T, U> extends ScopeAwareRuleWalker<T> {
-        private blockScopeStack: U[];
+/**
+    * An AST walker that is aware of block scopes in addition to regular scopes. Block scopes
+    * are a superset of regular scopes (new block scopes are created more frequently in a program).
+    */
```
adidahiya: indentation

src/ruleLoader.ts
```diff
@@ -1,159 +1,158 @@
...
-                if (Rule !== undefined) {
-                    const all = "all"; // make the linter happy until we can turn it on and off
-                    const allList = (all in enableDisableRuleMap ? enableDisableRuleMap[all] : []);
-                    const ruleSpecificList = (ruleName in enableDisableRuleMap ? enableDisableRuleMap[ruleName] : []);
-                    const disabledIntervals = buildDisabledIntervalsFromSwitches(ruleSpecificList, allList);
-                    rules.push(new Rule(ruleName, ruleValue, disabledIntervals));
-                }
+export function loadRules(ruleConfiguration: {[name: string]: any},
+                            enableDisableRuleMap: {[rulename: string]: Lint.IEnableDisablePosition[]},
+                            rulesDirectory?: string): Lint.IRule[] {
```
adidahiya: fix alignment of arguments

src/language/utils.ts
myitcv: Could you add [`typescript-formatter`](https://github.com/vvakame/typescript-formatter) to the build [like this](https://github.com/myitcv/typescript_test/blob/master/scripts/build.sh#L7) and have it fail when `.{ts,tsx}` files are not formatted correctly? Saves manual checking...

src/tsconfig.json
```diff
@@ -6,35 +6,64 @@
...
-        "./language/rule/rule.ts",
-        "./language/utils.ts",
-        "./language/walker/blockScopeAwareRuleWalker.ts",
-        "./language/walker/skippableTokenAwareRuleWalker.ts",
-        "./configuration.ts",
-        "./enableDisableRules.ts",
-        "./formatterLoader.ts",
-        "./ruleLoader.ts",
-        "./tslint.ts"
+        "./typings/tsd.d.ts",
```
adidahiya: this should be `../typings/tsd.d.ts`

test/helper.ts
```diff
@@ -14,70 +14,70 @@
...
-        });
-    }
+// assert on array equality for failures
+export function assertFailuresEqual(actualFailures: Lint.RuleFailure[], expectedFailures: Lint.RuleFailure[]) {
+    assert.equal(actualFailures.length, expectedFailures.length);
+    actualFailures.forEach((actualFailure, i) => {
+        const startPosition = JSON.stringify(actualFailure.getStartPosition().toJson());
+        const endPosition = JSON.stringify(actualFailure.getEndPosition().toJson());
+        assert.isTrue(actualFailure.equals(expectedFailures[i]),
+                        `actual failure at ${startPosition}, ${endPosition} did not match expected failure`);
```
adidahiya: args alignment

test/lint.ts
```diff
@@ -0,0 +1,6 @@
+export * from "../src/lint";
+import * as Test from "./helper";
+if (false) {
+    console.log(Test);
+}
```
adidahiya: this looks like test code that can be removed

src/language/utils.ts
adidahiya: +1 that sounds great

src/tslint.ts
```diff
@@ -14,106 +14,100 @@
...
-// add the Lint and TypeScript modules to global for pluggable formatters/rules
-global.Lint = Lint;
-global.ts = ts;
+import {findConfiguration as config} from "./configuration";
 
-// export Lint.Linter as the API interface for this module
-module.exports = Lint.Linter;
-// also export Lint.Configuration.findConfiguration so implementers can consume
-module.exports.findConfiguration = Lint.Configuration.findConfiguration;
+namespace Linter {
```
adidahiya: it feels weird to have a class and namespace with the same name... while we're refactoring we might as well fix this ambiguity?

how about we make `Linter` the default export from this module and export the `LintResult` and `ILinterOptions` interfaces as normal?

src/tslint.ts
```diff
@@ -14,106 +14,100 @@
...
+    export var findConfiguration = config;
```
adidahiya: I think `findConfiguration` would be better off as a public static member of the `Linter` class

test/lint.ts
weswigham: No.... this is to satisfy tslint itself. `Test` is flagged as an unused variable (since it's only used in the `export {Test}` below, yes, there's already an open issue on this), by having this bit of unreachable code, the erroneous error is squelched. Though I could probably have squelched it with a comment instead.

adidahiya: ok, please add a comment explaining that this

src/tslint.ts
weswigham: It's not n ambiguity - they're merged. This syntax is the only way to make an interface a static member of a class. If we want `import Linter = require('tslint');` to continue working as it does today, this merging is the best (only?) way to do it. `export default` isn't liable to work well in node with TS, since we fixup our exports with `__esModule`, but don't check for it on imports. `LintResult` and `ILinterOptions` are still reexported into the `Lint` namespace elsewhere.

src/tslint.ts
weswigham: Could be - the namespace is just a static extension of the class. (Except class static members can't be types while namespace members can.)

src/tslint.ts
adidahiya: Right, ok, that makes sense. I propose that we only export `LintResult` and `ILinterOptions` once, into the `Lint` namespace. This would be another breaking change (`Linter.LintResult` => `Lint.LintResult`). We'd also have to move these interfaces to `lint.ts` then, since the `export = Linter;` line would have to stay in this file.

src/language/utils.ts
weswigham: I looked into it - its output leaves a lot to be desired:
```cmd
 D:\rep\tslint ⮀ ⭠ modularize ⮀ grunt tsfmt
Running "tsfmt:src" (tsfmt) task
>> src/configuration.ts is not formatted
>> src/enableDisableRules.ts is not formatted
>> src/formatterLoader.ts is not formatted
>> src/formatters.ts is not formatted
>> src/formatters/index.ts is not formatted
>> src/formatters/jsonFormatter.ts is not formatted
>> src/formatters/pmdFormatter.ts is not formatted
>> src/formatters/proseFormatter.ts is not formatted
>> src/formatters/verboseFormatter.ts is not formatted
>> src/language/formatter/abstractFormatter.ts is not formatted
>> src/language/formatter/formatter.ts is not formatted
>> src/language/languageServiceHost.ts is not formatted
>> src/language/rule/abstractRule.ts is not formatted
>> src/language/rule/rule.ts is not formatted
>> src/language/utils.ts is not formatted
>> src/language/walker/blockScopeAwareRuleWalker.ts is not formatted
>> src/language/walker/index.ts is not formatted
>> src/language/walker/ruleWalker.ts is not formatted
>> src/language/walker/scopeAwareRuleWalker.ts is not formatted
>> src/language/walker/skippableTokenAwareRuleWalker.ts is not formatted
>> src/language/walker/syntaxWalker.ts is not formatted
>> src/lint.ts is not formatted
>> src/ruleLoader.ts is not formatted
>> src/rules.ts is not formatted
>> src/rules/alignRule.ts is not formatted
>> src/rules/banRule.ts is not formatted
>> src/rules/classNameRule.ts is not formatted
>> src/rules/commentFormatRule.ts is not formatted
>> src/rules/curlyRule.ts is not formatted
>> src/rules/eoflineRule.ts is not formatted
>> src/rules/forinRule.ts is not formatted
>> src/rules/indentRule.ts is not formatted
>> src/rules/interfaceNameRule.ts is not formatted
>> src/rules/jsdocFormatRule.ts is not formatted
>> src/rules/labelPositionRule.ts is not formatted
>> src/rules/labelUndefinedRule.ts is not formatted
>> src/rules/maxLineLengthRule.ts is not formatted
>> src/rules/memberAccessRule.ts is not formatted
>> src/rules/memberOrderingRule.ts is not formatted
>> src/rules/noAnyRule.ts is not formatted
>> src/rules/noArgRule.ts is not formatted
>> src/rules/noBitwiseRule.ts is not formatted
>> src/rules/noConditionalAssignmentRule.ts is not formatted
>> src/rules/noConsecutiveBlankLinesRule.ts is not formatted
>> src/rules/noConsoleRule.ts is not formatted
>> src/rules/noConstructRule.ts is not formatted
>> src/rules/noConstructorVarsRule.ts is not formatted
>> src/rules/noDebuggerRule.ts is not formatted
>> src/rules/noDuplicateKeyRule.ts is not formatted
>> src/rules/noDuplicateVariableRule.ts is not formatted
>> src/rules/noEmptyRule.ts is not formatted
>> src/rules/noEvalRule.ts is not formatted
>> src/rules/noInferrableTypesRule.ts is not formatted
>> src/rules/noInternalModuleRule.ts is not formatted
>> src/rules/noRequireImportsRule.ts is not formatted
>> src/rules/noShadowedVariableRule.ts is not formatted
>> src/rules/noStringLiteralRule.ts is not formatted
>> src/rules/noSwitchCaseFallThroughRule.ts is not formatted
>> src/rules/noTrailingCommaRule.ts is not formatted
>> src/rules/noTrailingWhitespaceRule.ts is not formatted
>> src/rules/noUnreachableRule.ts is not formatted
>> src/rules/noUnusedExpressionRule.ts is not formatted
>> src/rules/noUnusedVariableRule.ts is not formatted
>> src/rules/noUseBeforeDeclareRule.ts is not formatted
>> src/rules/noVarKeywordRule.ts is not formatted
>> src/rules/noVarRequiresRule.ts is not formatted
>> src/rules/oneLineRule.ts is not formatted
>> src/rules/quotemarkRule.ts is not formatted
>> src/rules/radixRule.ts is not formatted
>> src/rules/semicolonRule.ts is not formatted
>> src/rules/sortObjectLiteralKeysRule.ts is not formatted
>> src/rules/switchDefaultRule.ts is not formatted
>> src/rules/tripleEqualsRule.ts is not formatted
>> src/rules/typedefRule.ts is not formatted
>> src/rules/typedefWhitespaceRule.ts is not formatted
>> src/rules/useStrictRule.ts is not formatted
>> src/rules/variableNameRule.ts is not formatted
>> src/rules/whitespaceRule.ts is not formatted
>> src/tslint-cli.ts is not formatted
>> src/tslint.ts is not formatted
Warning: Task "tsfmt:src" failed. Use --force to continue.

Aborted due to warnings.
```

At least until it can suggest how to _fix_ those formatting errors (beyond allowing it to autoformat the code) I'm not sure it's as fit for purpose as just tslint itself is.

Given that it's just an additional proofing step, I don't think it needs to be in this PR, regardless.

weswigham:
@adidahiya Updated the PR.

src/language/utils.ts
myitcv: @weswigham - agree, `tsfmt` doesn't need to be in this PR. I'll pick up on comments in the PR you created for it.

adidahiya:
lgtm :+1: 

Now there's a `.d.ts` file produced for every `.js` file in lib/. I should include these in the published NPM package, right?

weswigham:
Correct. You shouldn't need to move them from the output location.

JKillian:
> For one, it no longer concatenates all of typescript into its built js
:+1:  :+1:  :+1:  

Going to try out/review this PR in total now.

JKillian:
We'll want to update the readme with appropriate changes in this PR. Specifically the custom rules/formatters section and potentially other sections.

I noticed that for custom rules

```typescript
import * as Lint from "tslint/lib/lint";
```
will fail unless the built `.js` file of custom rule is located in a directory or sub-directory with `node_modules` containting `tslint` and `typescript`. So for example

```
tslint -c someConfig.json -r ../typescript/built/local/tslint some/file.ts
```

will fail because the typescript project doesn't have typescript in `node_modules`, so `import * as ts from "typescript"` (which is `require('typescript')` in the .js file) fails. Is there a way we can do things so that custom rules being run only depend on things included with tslint and don't have to install these packages?

weswigham:
@JKillian In the case of typescript, I'm pretty sure we'll just replace `import * as ts from "typescript";` with `import * as ts from "../../../lib/typescript";` or something along those lines. Additionally, if we use `import {...} from "wherever/typescript"`, the import will be elided at runtime provided we're only using it for types.

(We have an open issue about not eliding `import * as blah from "blah";` yet.)

But, yeah, by starting to actually use proper external modules, we make the fact that rules depend on the tslint lib at runtime explicit.

adidahiya:
@JKillian I think we can do the README changes in a follow-up PR

JKillian:
@weswigham Makes sense, and likely that any other project using tslint also uses the typescript node module. However, there are still some oddities, right? Custom rules would be run using whatever version of typescript the project their located is using while built-in rules would be run using the version of typescript required by tslint. Do you think this could be potentially problem-causing, or not a big deal?

weswigham:
@JKillian This is the same kind of issue people have had in the `npm` community forever. Potential mismatches can be avoided by using the same version of typescript at runtime with both by flattening your dependency tree with npm 3 or `npm dedupe`. What's more pressing is [this issue](https://github.com/Microsoft/TypeScript/issues/5219). Right now, I'll need to _compile_ my rules against the same version of TS as tslint (and this is before this modularization, too), otherwise I can get odd behavior due to mismatching inlined const enum values. Const enums are fine to use internally, but having them in your API I've discovered to be a no-no, as it makes any change which is not to the end the list of them a breaking change.

myitcv:
I hadn't been following this conversation until now when I tried to implement a new rule in the `tslint` repo, branched off `master` (which has this PR merged in). So this is working on a new rule within a clone of the `tslint` repo.

The problem I am seeing is within my editor, specifically using the excellent [tsuquyomi](https://github.com/Quramy/tsuquyomi) plugin. tsuquyomi 'compiles' on save and shows you the build errors in a window below the file. Is understands and uses `tsconfig.json`.

However, the problem is equivalent to trying to compile by hand:

```
$ pwd
/path/to/tslint/src
$ tsc
enableDisableRules.ts(18,21): error TS2307: Cannot find module 'typescript'.
language/languageServiceHost.ts(18,21): error TS2307: Cannot find module 'typescript'.
language/languageServiceHost.ts(26,29): error TS7006: Parameter 'name' implicitly has an 'any' type.
language/languageServiceHost.ts(28,15): error TS7006: Parameter 'message' implicitly has an 'any' type.
language/rule/abstractRule.ts(18,21): error TS2307: Cannot find module 'typescript'.
....
```

This makes sense, because the module `"typescript"` is not defined anywhere transitively referenced by the contents of `tsconfig.json`. i.e. `tsconfig.json` no longer [contains everything needed to compile the rules](https://github.com/Microsoft/TypeScript/wiki/tsconfig.json). 

However the confusing thing is that `grunt` works. 

It seems that the `grunt` build works because it ignores `tsconfig.json` and [builds up its own command line options](https://github.com/TypeStrong/grunt-ts/blob/8ff80ca78ce1173d0d191f3ffb151a8339a8b197/tasks-internal/modules/compile.js#L216-L222) instead, ultimately calling `tsc @tempfilecontainingoptions`. [Here](https://gist.github.com/myitcv/67c32c9b53830cc0a92b) is a link to one such tempfile; I've sanitised paths and included line breaks for readability (the entire file should be on one line)

When invoking `tsc` in that way, it somehow deals with the fact that the `"typescript"` module is not defined. When invoked bare (i.e. falling back to `tsconfig.json`) it does not. This explains why tsuquyomi and plain `tsc` are failing.

@weswigham - feels like there is a bug here somewhere in core for there to be this discrepancy. Or am I missing something else?

JKillian:
@adidahiya Shouldn't this have been merged into `next`?

JKillian:
@weswigham Great point with regards to const enums. There are a few things that TSLint could do: for example, instead of taking a link to a rules directory containing compiled `.js` files, we could accept uncompiled `.ts` files and compile them on the fly when TSLint is executing. Then they would be compiled and executed using the same version of TypeScript as TSLint is using. I don't necessarily think this is what we *should* do, but I think this whole discussion has been helpful (at the least, to me) in identifying some of the issues with custom rules

JKillian:
@myitcv Good point. I think the solution is to just include the typescript services typings and link to either in `tsd.d.ts` or just link to them directly in `tsconfig.json`

myitcv:
@weswigham dug a bit deeper. 

It appears this is related to:

* the `CWD` of the `tsc` command
* the `module` option passed to `tsc`

Here is a minimal repro of the problem. Assume the following setup:

```
.
├── src
│   └── main.ts
└── tsconfig.json
```

```typescript
// src/main.ts

```

```json
/* tsconfig.json */
{
    "version": "1.6.2",
    "compilerOptions": {
        "module": "commonjs",
        "noImplicitAny": true,
        "target": "es5",
        "outDir": "dist"
    },
    "files": [
        "src/main.ts"
    ]
}
```

Running `tsc` from the root, this compile fine. 

If however we switch the `module` option to be `system`, then we get a compile error:

```
src/main.ts(1,21): error TS2307: Cannot find module 'typescript'.
```

If we move the `tsconfig.json` to the src directory, revert `module` back to `commonjs`, adjust `outDir` and `files` according, and run `tsc` from within `src` we see:

```
main.ts(1,21): error TS2307: Cannot find module 'typescript'.
```

And to complete the test if we now switch the `module` option to be `system`, again running `tsc` from within `src`, we get:

```
main.ts(1,21): error TS2307: Cannot find module 'typescript'.
```

To summarise:

* Case 1: `CWD` is root of project, `"module": "commonjs"` - compiles OK
* Case 2: `CWD` is root of project, `"module": "system"` - fails to compile
* Case 3: `CWD` is `src`, `"module": "commonjs"` - fails to compile
* Case 4: `CWD` is `src`, `"module": "system"` - fails to compile

I've [pushed this repro to Github](https://github.com/myitcv/tsc_module_repro). Each of the cases above is a branch. Simply `git checkout case_n` and then run `./build.sh` to repro.

It seems there is some special treatment of the `import * as ts from "typescript"` statement going on when `module` is set to `commonjs` **and** `tsc` is run with a `CWD` of the root directory (I suspect it it walking into `node_modules` in some special way)

This doesn't feel right to me. `tsconfig.json` should be all the compiler needs to build something, and the `CWD` shouldn't matter.

myitcv:
@JKillian - I'm not sure it is. Because none of those typings files defines the module `"typescript"`.

weswigham:
I think its because we only walk up to the CWD/project root when searching
for node_modules folders? Moving the tsconfig into the root of the repo
(which is where the gruntfile and node_modules folders are) could fix it.

myitcv:
@weswigham - but that doesn't deal with the fact the behaviour of the compiler is different when the output module format is different (case 1 and 2 above)

I should add that in my tests the location of the `tsconfig.json` file matches the `CWD` of `tsc`

weswigham:
The compiler behavior is different between module kinds because commonjs
uses node module resolution by default, whereas anything else uses
"classic". The module resolution algorithm can be set with the "
moduleResolution" flag to either "node" or "classic" to override that
default.

myitcv:
[#5102](https://github.com/Microsoft/TypeScript/issues/5102) looks related. And could explain why case 3 above doesn't work. 

@weswigham - thanks. Seems [#5039](https://github.com/Microsoft/TypeScript/issues/5039) is related, and explains why there is the difference between cases 1 and 2; agree?

weswigham:
@myitcv Yes, the module resolution strategy changes between module commonjs and system, which is why commonjs is a success (it looks in node_modues) where system fails (it does not). It was [#4352](https://github.com/Microsoft/TypeScript/pull/4352) which added this - [#5039](https://github.com/Microsoft/TypeScript/issues/5039) builds on that work to that to support requirejs-like shims.

#5102 itself may be a different issue, but @alexeagle's comment on it is certainly the (exact) same situation. As a stopgap until we fix this (hoping we do), at least for now, you can just move the tsconfig in src into the root of the repo and correct the paths to the files inside it.

weswigham:
@JKillian ideally, you shouldn't need to include any dts for the typescript compiler anymore since we can pull them from the node module.

myitcv:
@weswigham - great, thanks for the explanation. 

Not sure the option of moving `tsconfig.json` to the root is feasible; because there exists `src/tsconfig.json` and `test/tsconfig.json`

Thoughts?

alexeagle:
You can leave tsconfig.json where it is, just make sure the current working
directory when running tsc has a node_modules in it, and use the -p
argument to tsc so it knows where the project root is.

myitcv:
@alexeagle thanks, this solves the problem of running `tsc` manually, but does not fix the issue with tsuquyomi.

myitcv:
@weswigham referring to your comment about enums earlier. I've just been bitten by this. 

Nightly build `typescript@1.7.0-dev.20151006` defines `NullKeyword = 91`
Nightly build `typescript@1.7.0-dev.20151014` defines `NullKeyword = 93` 
Nightly build `typescript@1.7.0-dev.20151015` defines `NullKeyword = 93` 

However [`master`](https://github.com/Microsoft/TypeScript/blob/4fedd77c3310eeebb6bc189f99e0be2b2e916766/lib/typescriptServices.d.ts#L124) (at the time of writing) defines `NullKeyword = 91`.

It also shows that [`typescriptServices.d.ts`](https://github.com/Microsoft/TypeScript/commits/master/lib/typescriptServices.d.ts) was last updated on Sep 15. 

So how does it come about that the builds from 14th and 15th define `NullKeyword = 93`?

JKillian:
Fun project I just learned about the other day, [Naughty TypeScript](https://github.com/TypeStrong/ntypescript) which converts const enums to enums and has a few other useful features. Obviously not going to be usable in all contexts, but good to know about!

myitcv:
@adidahiya @JKillian - any chance we can get this merged into `next` and a new `next` version cut?

Thanks

JKillian:
@myitcv Yep, will do today!

JKillian:
There's an issue with const enums and our next branch related to this as well. Currently, the next branch has a devDependency on `"typescript": "next", and with this modularization update, just a regular old dependency. But we have to be careful, if TSLint gets compiled with a TS nightly from Oct.15th and it gets installed on some later date, TSLint would then be using a newer version of TS to analyze files then it was compiled with, which could lead to enum value conflicts.

weswigham:
[Working on it.](https://github.com/Microsoft/TypeScript/pull/5297)

JKillian:
@weswigham Sweet!
@myitcv Release might be delayed a little, Github seems to be having issues for us today

adidahiya:
released [v3.0.0-dev.1](https://github.com/palantir/tslint/releases/tag/3.0.0-dev.1)

alexeagle:
Trying this now, but something has changed with the way the top-level export works:
```
alexeagle-macbookpro2:use-tslint alexeagle$ node
> require('tslint');
{ default:
   { [Function: Linter]
     VERSION: '3.0.0-dev.1',
     findConfiguration: [Function: findConfiguration] } }
> alexeagle-macbookpro2:use-tslint alexeagle$ rm -rf node_modules/
alexeagle-macbookpro2:use-tslint alexeagle$ npm install tslint
tslint@2.5.1 node_modules/tslint
├── underscore.string@3.1.1
├── optimist@0.6.1 (wordwrap@0.0.3, minimist@0.0.10)
└── findup-sync@0.2.1 (glob@4.3.5)
alexeagle-macbookpro2:use-tslint alexeagle$ node
> require('tslint')
{ [Function: Linter]
  VERSION: '2.5.1',
  findConfiguration: [Function: findConfiguration] }
```
this causes gulp-tslint to fail:
```
[17:17:30] Starting 'lint'...

/Users/alexeagle/Projects/angular/node_modules/gulp-tslint/node_modules/rcloader/index.js:68
      cb(void 0, configFile);
      ^
TypeError: linter is not a function
    at /Users/alexeagle/Projects/angular/node_modules/gulp-tslint/index.js:94:22
    at respond (/Users/alexeagle/Projects/angular/node_modules/gulp-tslint/node_modules/rcloader/index.js:68:7)
    at respond (/Users/alexeagle/Projects/angular/node_modules/gulp-tslint/node_modules/rcloader/node_modules/rcfinder/index.js:140:7)
    at next (/Users/alexeagle/Projects/angular/node_modules/gulp-tslint/node_modules/rcloader/node_modules/rcfinder/index.js:164:16)
    at doNTCallback0 (node.js:407:9)
    at process._tickCallback (node.js:336:13)
```

JKillian:
Thanks for the heads-up @alexeagle. The problem is that when you use the `export default` syntax in TS ([as we do][0]) it creates that `default` field (see [the Default Export and Default Import section of this comment][1]). This is all fine and dandy when using TS ES6 imports, but [it's a little messy when using CommonJS][2]. Luckily [there are ways to deal with this][3]. I assume we'll do something like suggested in that blog post so CommonJS users don't have to change their code. What do you think @adidahiya?

[0]: https://github.com/palantir/tslint/blob/master/src/tslint.ts#L22 
[1]: https://github.com/Microsoft/TypeScript/issues/2242#issue-60179656
[2]: https://github.com/Microsoft/TypeScript/issues/2719
[3]: http://www.jbrantly.com/es6-modules-with-typescript-and-webpack/#packagingasalibrary

alexeagle:
Otherwise, this looks good to me. :+1: 
I have a PR out to add this to angular (didn't find any bugs by turning it on, sadly)
https://github.com/angular/angular/pull/4970

Note you may want to document for the next release that custom rules need to use the same typescript version as tslint internally:
https://github.com/angular/angular/pull/4970/files#diff-dbb7291ecb65301ed3220de1c23eb9faR4
otherwise you get errors like
```
equireParameterTypeRule.ts(6,14): error TS2415: Class 'Rule' incorrectly extends base class 'AbstractRule'.
  Types of property 'apply' are incompatible.
    Type '(sourceFile: SourceFile) => RuleFailure[]' is not assignable to type '(sourceFile: SourceFile) => RuleFailure[]'.
      Types of parameters 'sourceFile' and 'sourceFile' are incompatible.
        Type 'ts.SourceFile' is not assignable to type 'ts.SourceFile'.
          Types of property 'statements' are incompatible.
            Type 'NodeArray<Statement>' is not assignable to type 'NodeArray<Statement>'.
              Types of property 'push' are incompatible.
                Type '(...items: Statement[]) => number' is not assignable to type '(...items: Statement[]) => number'.
                  Types of parameters 'items' and 'items' are incompatible.
                    Type 'ts.Statement' is not assignable to type 'ts.Statement'.
                      Types of property 'kind' are incompatible.
                        Type 'ts.SyntaxKind' is not assignable to type 'ts.SyntaxKind'.
```

JKillian:
Yikes, that's a cryptic error. It's a little bit tricky now: when using the `next` branch of tslint, the version of TS used to compile tslint, provide the language services when tslint is running, and compile custom rules can all be different. 

I think we ought to fix TS to a specific version for the `next` branch. So instead of the dependency being `typescript: next`, we ought to fix it to something like `typescript: 1.8.0-dev.20151027` and then update that each week. This would make ensure that (TS version to compile tslint === TS version to run tslint).

That still leaves the issue of the version of TS used to compile custom rules. I think recommending that custom rules be compile with the same version of TS as tslint is using in the docs is a good starting point. If compiling the custom rules is integrated into your build, this is fine. It still makes it hard to use prebuilt-custom rules though. 

We also have the [issue of const enums](https://github.com/Microsoft/TypeScript/pull/5297) that was mentioned earlier. @weswigham, is it correct that the TS `.d.ts` files will start having regular `enum`s instead of `const enum`s soon?

weswigham:
Yes. We have an open PR with the change that'll likely be merged in soon.

JKillian:
(See https://github.com/Microsoft/TypeScript/pull/5422)

alexeagle:
There is probably an issue where this discussion belongs.

Indeed this is tricky - we have similar problems with ts2dart but have the advantage that we control versions at tool-compile-time, user-compile-time, and runtime. Here's my somewhat-informed input:

If the const enum problem is resolved, then I won't have to care what TS you complied tslint with, I only need runtime compatibility with the JS files you distribute.
For type-checking in my compile step, ideally I think I should bring my own version of typescript, and it should be used for compiling my custom rules, along with other things I compile in my build toolchain. I just need to bring a version modern enough to understand the .d.ts files you distribute.
When tslint executes, I would also like it to use the TS language services at the version I bring. I need to use a modern enough version that the language services provide the APIs tslint calls.

myitcv:
@alexeagle - thanks for adding comment about that version mismatch problem... just ran into exactly the same issue.

Just so that I'm clear, for people writing custom rules (incidentally we need a [`README`](https://github.com/palantir/tslint/blob/3df3e6355efae73f1e1d60cdbae028371dd8745d/README.md#custom-rules) update to reflect the merged #726), the advice is currently as follows:

```
import * as Lint from "tslint/lib/lint";
import * as ts from "tslint/node_modules/typescript";

  // ...
```

Is that correct?

weswigham:
> `import * as ts from "tslint/node_modules/typescript";`

I'd avoid this. With `npm3`, there's a good chance that `tslint`'s internal dependency gets flattened out to the top level.
I'd just use `import * as ts from "typescript";` and specify a `typescript` version in your own package.json. `tslint` could even support more explicitly by requiring `typescript` as a peer dependency when used as a library. (Though, this would mean that `tslint-cli` and `tslint` would have different dependencies and would need to be split apart, as `tslint-cli` would need to require its own typescript compiler, and as of npm3, peer dependencies are not installed automatically.)

alexeagle:
Too late, it's in angular master already :)!
https://github.com/angular/angular/commit/c90e1920f506ef832a29712ea1b8ec42d5fed65b

The problem is that 3.0.0-dev.1 is cut from the tslint@next branch so it
includes the TypeScript@next version. I need to depend on tslint@master so
the TS version will match the one in my node_modules.

JKillian:
I need to think about the best solution a little bit more - this discussion has been great so far, but I don't see a clear cut road forwards yet. 

@weswigham Any recommended documentation on previous/upcoming npm changes? I'm not too familiar with npm 3 vs. npm 2 and definitely not familiar with what the changes down the road will look like.

weswigham:
@JKillian [The npm roadmap](https://github.com/npm/npm/wiki/Roadmap), and the [npm 3 changelog](https://github.com/npm/npm/blob/master/CHANGELOG.md) are great references.

For the record: npm 3+ is the latest stable version of npm - anyone doing an `npm install npm` gets it now. (Though node 4 shipped with npm 2.14.)

JKillian:
> With npm3, there's a good chance that tslint's internal dependency gets flattened out to the top level.

Thanks @weswigham, reading over the changelog and some related issues since npm v3.0.0 was helpful. I had initially thought you meant "with possible future changes to npm3" the dependency might get flattened, but in reality, you meant that npm3 can "choose" to install typescript as a sibling of tslint or it might install it as a child of tslint.

Having a non-deterministic location for typescript seems tricky, I like the idea of making it a peer dependency and thus always having it installed as a sibling. If we went down this road, I wonder what versioning requirement would be appropriate. Maybe `>=` to a specific version of TS?

> Though, this would mean that tslint-cli and tslint would have different dependencies and would need to be split apart, as tslint-cli would need to require its own typescript compiler, and as of npm3, peer dependencies are not installed automatically.

Why would this be though? It seems to me they could stay as one package with a peer dependency on typescript. Yes, it would force people to install typescript besides tslint just in order to use the CLI, but this probably happens in almost all cases anyways.

JKillian:
@alexeagle I'm going to change the behavior of the `tslint` library shortly, so you don't have to do the weird `.default` hack. See https://github.com/palantir/tslint/issues/760 This will make it into `3.0.0-dev.2` which will hopefully be released tomorrow

alexeagle:
Any idea when a 3.0.0 will be cut from master (so it depends on TS 1.6.2?)

myitcv:
> I think we ought to fix TS to a specific version for the `next` branch. So instead of the dependency being typescript: `next`, we ought to fix it to something like typescript: `1.8.0-dev.20151027` and then update that each week.

:+1: 

Just been bitten by this... right now I've had to switch our dependency to be on `next`... which is not great because it's a moving target - with multiple developers and a CI server, that's less than ideal.

myitcv:
@JKillian looking forward to `3.0.0-dev.2` because as of TypeScript `1.8.0-dev.20151103` (i.e. today), `tslint 3.0.0-dev.1` breaks. That's because of [this change](https://github.com/Microsoft/TypeScript/pull/5502/) where the `SyntaxKind` `enum` is no longer defined as `const`.

The good news is that as of TypeScript `1.8.0-dev.20151029` (following the merge of https://github.com/Microsoft/TypeScript/pull/5422) rebuilding `tslint 3.0.0-dev.1` just works, and the `SyntaxKind` etc `enum` values are not erased to integers.

JKillian:
@myitcv Should be pretty soon: https://github.com/palantir/tslint/pull/768  :)
@alexeagle We want to make sure the recent changes are stable before 3.0.0 gets released. Release `3.0.0-dev.2` will fix the issue where you had to do `require('tslint').default`. If everything seems good with this dev release we'll release regular `3.0.0` next.

adidahiya:
published 3.0.0-dev.2

myitcv:
@adidahiya @JKillian - thanks very much, works like a charm.

adidahiya:
FYI, the const enum problem is worse than I actually thought (but maybe y'all were aware). See #832.

PR #729, opened 2015-10-12
Improve use-strict rule to work for module declarations with more than one identifier

### LINKED ISSUE #544: use-strict not working on module
I'm trying to use the use-strict rule for modules and I can't make it to work properly. 

I've have a stripped down version of a module with an interface:

```ts
module app.myModule {
	export interface IListener {
		listeners: Array<any>;
	}
}
```

Running grunt tslint doesn't returns the use-strict error.

I've debugged the rule and it looks like when the depth is 2 for that file the node.body.kind is ModuleDeclaration and not ModuleBlock.

Am I missing something?

ArturDorochowicz:
This will make it work for code like this:

```
module A.B.C {
  // use strict required here
}
```

Fixes #544

### DIFF
--- a/src/rules/useStrictRule.ts
+++ b/src/rules/useStrictRule.ts
@@ -34,13 +34,16 @@ class UseStrictWalker extends Lint.ScopeAwareRuleWalker<{}> {
     }
 
     public visitModuleDeclaration(node: ts.ModuleDeclaration) {
-        // current depth is 2: global scope and the scope created by this module
-        if (this.getCurrentDepth() === 2
-                && !Lint.hasModifier(node.modifiers, ts.SyntaxKind.DeclareKeyword)
+        if (!Lint.hasModifier(node.modifiers, ts.SyntaxKind.DeclareKeyword)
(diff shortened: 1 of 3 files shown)


ArturDorochowicz:
FYI, I kept the previous behavior that the failure is basically on the `module` keyword. Not sure if this is appropriate.

src/rules/useStrictRule.ts
```diff
@@ -77,4 +80,14 @@ class UseStrictWalker extends Lint.ScopeAwareRuleWalker<{}> {
             this.addFailure(this.createFailure(node.getStart(), node.getFirstToken().getWidth(), Rule.FAILURE_STRING));
         }
     }
+
+    private getFirstInModuleDeclarationsChain(node: ts.ModuleDeclaration): ts.ModuleDeclaration {
```
adidahiya: since this is static, can you make this a `function` outside of this class, at the end of this file?

adidahiya:
lgtm besides the helper function comment ^

ArturDorochowicz:
Updated the PR.

adidahiya:
great, thanks!

PR #732, opened 2015-10-13
tsfmt task

weswigham:
For your consideration. I find it's output somewhat lacking, though.

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -9,6 +9,35 @@ if (process.platform  === "win32") {
 
 module.exports = function (grunt) {
     // jscs:disable requireCamelCaseOrUpperCaseIdentifiers
+    grunt.registerMultiTask("tsfmt", "Check typescript format", function() {
+        var done = this.async();
+        var fmt = require("typescript-formatter");
+        var options = { verify: true };
+        var completionPromises = [];
+        this.files.forEach(function(file) {
+            var srcs = file.src;
+            srcs.forEach(function(src) {
+                if (!grunt.file.exists(src)) {
+                    grunt.log.warn('Source file "' + src + '" not found.');
+                    return false;
+                }
+
+                var content = grunt.file.read(src);
+                completionPromises.push(fmt.processString(src, content, options));
+            });
+        });
+        Promise.all(completionPromises).then(function(results) {
+            var success = true;
+            results.forEach(function(result) {
+                if (result.error && result.message) {
+                    grunt.log.warn(result.message);
+                    success = false;
+                }
+            });
+            done(success);
+        });
+    });
+
     grunt.initConfig({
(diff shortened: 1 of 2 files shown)


myitcv:
Picking up on [this thread](https://github.com/palantir/tslint/pull/726#discussion_r41796997)

> its output leaves a lot to be desired:
>
> ```
>  D:\rep\tslint ⮀ ⭠ modularize ⮀ grunt tsfmt
Running "tsfmt:src" (tsfmt) task
>> src/configuration.ts is not formatted
>> src/enableDisableRules.ts is not formatted
```

The output is just fine, it doesn't need to be any more verbose. `tsfmt` shouldn't tell you how to fix the problem; you should simply run `tsfmt` on the file itself and it will fix it for you. Why would you want to fix it by hand?

To fix the entire source tree (one off):

```
tsfmt -r *(app|src)/**/*.{ts,tsx}
```

For what it's worth, I have `tsfmt` running whenever I save a file in my editor. That way things never get out of step.

**Comment by JKillian** [comment]
@weswigham Can you rebase this?

I haven't tried tsfmt or the typescript compiler formatting tools which it's based on. From the description though, the formatting they do seems pretty unobtrusive. I wouldn't want to merge in a tool that does overly zealous reformatting (for example, if I have a multiline conditional statement, I wouldn't want a tool that combines it all into one line). 

I'm assuming based on both of your comments that tsfmt does an appropriate level of adjustments?

**Comment by myitcv** [comment]
@JKillian - because there is no standard format for TypeScript code (which I consider to be a weakness) you may find you want to set [`tsfmt` options](https://github.com/myitcv/typescript_test/blob/master/app/tsfmt.json). 

Does this PR actually enforce, as part of the build, that all code is formatted?

Because the build is green, yet when checked out locally I see:

```
tsfmt --verify *(src|test)/**/*.{ts,tsx}
```

shows that 217 files are not formatted. 

Incidentally, formatting those files creates the following diff (ignoring whitespace changes):

https://gist.github.com/myitcv/0f83f6ce02d184f26f69

JKillian:
Interesting catch with running `tsfmt` on the whole project - we would not want it to run on `.test.ts` files. These are intentionally formatted poorly sometimes, and auto-formatting them could mess that up. I like the idea of using it on the source code though, I want to take a look at a diff that includes whitespace changes and only the real source code and see what it looks like.

weswigham:
This PR just contains a tsfmt verification task. It doesn't add it to the
build step.

weswigham:
I'd recommend someone else pick up this patch and update it if you're still interested in the build step (I still think tsfmt's pass/fail output is lacking, though), I've been focusing on other bits lately.

JKillian:
Closing this in favor of #841, thanks for this idea @weswigham!

PR #734, opened 2015-10-13
Fixes #727

### LINKED ISSUE #727: parameter destructuring renaming and shadowed variable
Following code is incorrectly handled by no-shadowed-variable rule
```ts
interface MyInterface  {
  public a: number;
}

function myFunction(
  { a: var1 }: MyInterface,
  { a: var2 }: MyInterface, // shouldn't create failure
  { a: var1 }: MyInterface // should create failure
) {
  var1 = 2;
  var2 = 3;
}
```
In fact, failure seems to be created because when visiting parameter destructuring renaming, variableName is undefined (so we get shadowed variable: 'undefined'

As an aside, above code also raises no unused variable false positives

lbesson:
not really happy with the getBindingElementVariableDeclaration() but...

### DIFF
--- a/src/language/utils.ts
+++ b/src/language/utils.ts
@@ -97,16 +97,21 @@ export function isBlockScopedVariable(node: ts.VariableDeclaration | ts.Variable
 }
 
 export function isBlockScopedBindingElement(node: ts.BindingElement): boolean {
+    const variableDeclaration = getBindingElementVariableDeclaration(node);
+    // if no variable declaration, it must be a function param, which is block scoped
+    return (variableDeclaration == null) || isBlockScopedVariable(variableDeclaration);
+}
+
+export function getBindingElementVariableDeclaration(node: ts.BindingElement): ts.VariableDeclaration {
     let currentParent = node.parent;
     while (currentParent.kind !== ts.SyntaxKind.VariableDeclaration) {
         if (currentParent.parent == null) {
-            // if we didn't encounter a VariableDeclaration, this must be a function parameter, which is block scoped
-            return true;
+            return null; // function parameter, no variable declaration
         } else {
             currentParent = currentParent.parent;
         }
     }
-    return isBlockScopedVariable(<ts.VariableDeclaration> currentParent);
+    return <ts.VariableDeclaration> currentParent;
 }
 
 /**
--- a/src/rules/noShadowedVariableRule.ts
+++ b/src/rules/noShadowedVariableRule.ts
@@ -36,10 +36,15 @@ class NoShadowedVariableWalker extends Lint.BlockScopeAwareRuleWalker<ScopeInfo,
 
     public visitBindingElement(node: ts.BindingElement) {
         const isSingleVariable = node.name.kind === ts.SyntaxKind.Identifier;
-        const isBlockScoped = Lint.isBlockScopedBindingElement(node);
+        const variableDeclaration = Lint.getBindingElementVariableDeclaration(node);
 
         if (isSingleVariable) {
-            this.handleSingleVariableIdentifier(<ts.Identifier> node.name, isBlockScoped);
+            if (variableDeclaration) {
+                const isBlockScopedVariable = Lint.isBlockScopedVariable(variableDeclaration);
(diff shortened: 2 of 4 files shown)


JKillian:
@lbesson Can you rebase this?

lbesson:
Sure.

lbesson:
(rebased)

src/rules/noShadowedVariableRule.ts
```diff
@@ -63,15 +68,12 @@ class NoShadowedVariableWalker extends Lint.BlockScopeAwareRuleWalker<ScopeInfo,
...
-        const variableIdentifier = <ts.Identifier> node.name;
-        const variableName = variableIdentifier.text;
-        const currentScope = this.getCurrentScope();
+        const isSingleParameter = node.name.kind === ts.SyntaxKind.Identifier;
+        const parameterIdentifier = <ts.Identifier> node.name;
 
-        if (this.isVarInAnyScope(variableName)) {
-            this.addFailureOnIdentifier(variableIdentifier);
+        if (isSingleParameter) {
+            this.handleSingleParameterIdentifier(parameterIdentifier);
```
adidahiya: don't need to make another method; please inline its contents here

src/rules/noShadowedVariableRule.ts
```diff
@@ -35,10 +35,15 @@ class NoShadowedVariableWalker extends Lint.BlockScopeAwareRuleWalker<ScopeInfo,
 
     public visitBindingElement(node: ts.BindingElement) {
         const isSingleVariable = node.name.kind === ts.SyntaxKind.Identifier;
-        const isBlockScoped = Lint.isBlockScopedBindingElement(node);
+        const variableDeclaration = Lint.getBindingElementVariableDeclaration(node);
 
         if (isSingleVariable) {
-            this.handleSingleVariableIdentifier(<ts.Identifier> node.name, isBlockScoped);
+            if (variableDeclaration) {
+                const isBlockedScopeVariable = Lint.isBlockScopedVariable(variableDeclaration);
```
adidahiya: `isBlockScopedVariable` (missing "d")

src/rules/noShadowedVariableRule.ts
```diff
@@ -63,15 +68,12 @@ class NoShadowedVariableWalker extends Lint.BlockScopeAwareRuleWalker<ScopeInfo,
     }
 
     public visitParameterDeclaration(node: ts.ParameterDeclaration) {
-        // treat parameters as block-scoped variables
-        const variableIdentifier = <ts.Identifier> node.name;
-        const variableName = variableIdentifier.text;
-        const currentScope = this.getCurrentScope();
+        const isSingleParameter = node.name.kind === ts.SyntaxKind.Identifier;
+        const parameterIdentifier = <ts.Identifier> node.name;
```
adidahiya: inline this value where it's used, no need for the variable

src/language/utils.ts
```diff
@@ -97,16 +97,22 @@ export function isBlockScopedVariable(node: ts.VariableDeclaration | ts.Variable
 }
 
 export function isBlockScopedBindingElement(node: ts.BindingElement): boolean {
+    const variableDeclaration = getBindingElementVariableDeclaration(node);
+    return variableDeclaration ?
+        isBlockScopedVariable(getBindingElementVariableDeclaration(node)) :
+        true; // if we didn't encounter a VariableDeclaration, this must be a function parameter, which is block scoped
+}
```
adidahiya: refactor:

```ts
const variableDeclaration = getBindingElementVariableDeclaration(node);
// if no variable declaration, it must be a function param, which is block scoped
return (variableDeclaration == null) || isBlockScopedVariable(variableDeclaration);
```

adidahiya:
looks pretty good overall. `getBindingElementVariableDeclaration` is not so bad; I can't think of a great alternative. just left some small style comments

lbesson:
Thanks for the review, I updated the code according to your recommendations.

In my emails, I had one comment "don't need to make another method; please inline its contents here", but it doesn't appear on github, I guess it was deleted?

adidahiya:
@lbesson yes, that comment was deleted, sorry for the email :)

adidahiya:
thanks for the PR! :+1:

Issue #620, opened 2015-08-28
Simplify Testing System for Rules


JKillian:
TODO (based off of new testing system explained [here](https://github.com/palantir/tslint/issues/620#issuecomment-144561726)):

- [x] Write code to run new-style tests
- [x] Write tests for code that runs new-style tests
- [x] Convert old-style tests into new-style tests for each rule
- [x] Clean up all of the above
- [ ] Document how to write tests on website
- [x] Make `--test` command so custom rules can be tested the same way?

---
Original Post:
Finding numerical positions of failures is tedious and inflexible (if the test file changes, all the failures move.) Instead, we could use some sort of marking system:

```typescript
/*fail:0*/var/*endfail:0*/ foo;

function tmp(t: any) {
    /*fail:1*/var/*endfail:1*/ x = 3;
}

/*fail:2*/var/*endfail:2*/ i,
    j;

/*fail:3*/var/*endfail:3*/ [a, b] = [1, 2];

declare var tmp2: any;

let bar;
const qux;

let k,
    l;

module quz {
    export var i;
}

let [x, y] = [1, 2];
```

Admittedly this makes the test files slightly harder to read, but overall it seems like a much easier and maintainable system. (This issue is about coming up with a better system in general, it doesn't have to follow the example above)

ashwinr:
I strongly disagree -- this is an unnecessary complication and makes the test files unreadable, sprinkled with magical markers. I would rather allow reading code off a string buffer and not mandate making a separate test file for each test.

JKillian:
Hmm, each test essentially already has a separate source file though, in `test/files/rules`, as well as well as the file that runs the test in `test/rules`.  The files in `test/rules` are essentially unreadable now:

```typescript
describe("<no-trailing-whitespace>", () => {
    const NoTrailingWhitespaceRule = Lint.Test.getRule("no-trailing-whitespace");
    const fileName = "rules/trailing.test.ts";
    const createFailure = Lint.Test.createFailuresOnFile(fileName, NoTrailingWhitespaceRule.FAILURE_STRING);

    it("forbids trailing whitespace", () => {
        const actualFailures = Lint.Test.applyRuleOnFile(fileName, NoTrailingWhitespaceRule);
        // Ahh! Lots of random numbers!!!!
        const expectedFailure1 = createFailure([2, 24], [2, 28]);
        const expectedFailure2 = createFailure([3, 32], [3, 36]);
[...]

PR #735, opened 2015-10-14
Rule to prevent certain keywords being used as variable names

myitcv:
Prevents nasties like:

```typescript
let undefined = 5;
let Number: Number;
function F8(undefined) {}
function G7(boolean: boolean) {}
let [ string ] = [ 5 ];
let { boolean } = { boolean: 1 };
```

Comments/thoughts welcomed.


src/tsconfig.json
```diff
@@ -32,8 +32,8 @@
         "./rules/noConditionalAssignmentRule.ts",
         "./rules/noConsecutiveBlankLinesRule.ts",
         "./rules/noConsoleRule.ts",
-        "./rules/noConstructRule.ts",
```
myitcv: Not a negative diff, simply correcting ordering

test/tsconfig.json
```diff
@@ -55,8 +56,8 @@
         "./rules/noConditionalAssignmentRuleTests.ts",
         "./rules/noConsecutiveBlankLinesRuleTests.ts",
         "./rules/noConsoleRuleTests.ts",
-        "./rules/noConstructRuleTests.ts",
```
myitcv: Not a negative diff, simply correcting ordering

adidahiya:
looks like this would address https://github.com/Microsoft/TypeScript/issues/3081

src/rules/noKeywordNamedVariablesRule.ts
```diff
@@ -0,0 +1,58 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+import * as Lint from "../lint";
+import * as ts from "typescript";
+
+const BAD_NAMES = [ "any", "Number", "number", "String", "string", "Boolean", "boolean", "undefined" ];
```
JKillian: Is this a comprehensive list of keywords/types that tsc will let you redefine? It looks good to me, just want to make sure there isn't anything missing or extraneous

test/tsconfig.json
```diff
@@ -44,6 +44,7 @@
         "./rules/indentRuleTests.ts",
         "./rules/interfaceNameRuleTests.ts",
         "./rules/jsdocFormatRuleTests.ts",
+        "./rules/keywordNamedVariablesRule.ts",
```
JKillian: shouldn't this be `noKeywordNamedVariablesRuleTest.ts`

test/rules/noKeywordNamedVariablesRule.ts
```diff
@@ -0,0 +1,119 @@
+/*
```
JKillian: Can you rename this `noKeywordNamedVariablesRuleTests.ts`

JKillian:
Looks good other than my comments above!

src/rules/noKeywordNamedVariablesRule.ts
myitcv: @JKillian started with this to kick off discussion.

Arguably this list could include any ambient variable from `lib.d.ts`, like [`Number`](https://github.com/Microsoft/TypeScript/blob/55218d64a58a9964229daa0a38bdfb0a6be717e5/lib/lib.d.ts#L519), [`String`](https://github.com/Microsoft/TypeScript/blob/55218d64a58a9964229daa0a38bdfb0a6be717e5/lib/lib.d.ts#L519) and [`Boolean`](https://github.com/Microsoft/TypeScript/blob/55218d64a58a9964229daa0a38bdfb0a6be717e5/lib/lib.d.ts#L457) 

The idea behind this rule is to increase the likelihood of writing code that is unambiguously understood by as many as possible. You could easily chuck `RegExp`, `Array`and other core interfaces that have real class equivalents in there too... 

But then you could probably well argue it goes too far to include [`event`](https://github.com/Microsoft/TypeScript/blob/55218d64a58a9964229daa0a38bdfb0a6be717e5/lib/lib.d.ts#L16738) for example.

JKillian: Gotchya, I'm fine with it like this for this PR. Changing the list is easy and essentially backwards-compatible, so that can be discussed down the road.

JKillian:
Also, perhaps their should be test cases for reassignments? I know `NaN` isn't on this blacklist right now, but if it was, we would want to disallow things like:

```
NaN = 5;
```

myitcv:
@JKillian agree that is totally bizarre but I intentionally didn't include such things in this rule because arguably you could/should make those definitions in `lib.d.ts` `const` (unless I'm missing something?)

test/tsconfig.json
myitcv: Hmmm, I'm surprised nothing threw an error.... this file doesn't even exist.

myitcv: Ah, [this](https://github.com/palantir/tslint/blob/master/test/tsconfig.json#L20) might explain it. Do you still need/want this explicit list of files?

myitcv:
@JKillian fixed the filename issue. Good catch, but per my comment can you tell me why the build didn't fail? Seems like that entry in `tsconfig.json` is redundant.... hence can all of them go?

JKillian:
@myitcv Perhaps the reason to not make them const is to reflect their JS behavior maybe? Since in JS you can theoretically redefine NaN, TS doesn't prevent you from doing that. Maybe there's some TS library out there that depends on redefining NaN haha.

test/tsconfig.json
JKillian: AKAIK, `filesGlob` is a special field used by atom-typescript (and maybe other IDEs) but isn't actually used by TS. The IDEs just auto-populate the files field based off of `filesGlob`.

myitcv:
> Perhaps the reason to not make them const is to reflect their JS behavior maybe? Since in JS you can theoretically redefine NaN

Could well be. Sounds like a candidate for another rule.... agree?

JKillian:
As far as the build not failing, my guess is that your tests just weren't running at all (and also then not failing at all). 

My inclination is that it can be in this same rule - it seems reasonable that a rule that doesn't let you declare variables with the name `String` also doesn't let you reassign `String` (although in practice it will be difficult to reassign `String` because you have to match the typing)

myitcv:
> As far as the build not failing, my guess is that your tests just weren't running at all (and also then not failing at all).

Quite possibly, but the compiler/grunt should probably error if a listed file does not exist, that was more my point. Which then leads you to working out "oh my tests aren't running"...

> My inclination is that it can be in this same rule

Gave this some more thought last night. 

By preventing assignments like `NaN = 5` we are essentially enforcing that `NaN` is `const`. Question is then why you wouldn't do the same for every ambient `var` in `lib.d.ts`?

Consider the following:

```typescript
// marker 1
name = "test";

export function DoIt(): void {
	// marker 2
	let name = 5;
	console.log(name);
}
```

This is valid because of the declaration of `name` in [`lib.d.ts`](https://raw.githubusercontent.com/Microsoft/TypeScript/d8237dc2f39f80d75a882d8813f4d550565f46a5/lib/lib.d.ts):

```typescript
declare var name: string;
```

Couple of questions (at least) come out of this:

1. Should we prevent the assignment at `marker 1`? `name` is an ambient var like `NaN`
2. The declaration of `name` in the function is essentially shadowing the ambient declaration

The `no-shadowed-variable` rule does not pick up point 2 above; not sure why.

In any case, I think preventing the assignment to `NaN` is better achieved by using a [custom `lib.d.ts`](https://github.com/Microsoft/TypeScript/issues/494) (or equivalent) and changing the ambient `var`s to ambient `const`s. Otherwise the rule needs to maintain a list of all the ambient `var`s. 

The intention of this rule should I think be to prevent variables being named with known 'keywords'. The definition of 'keyword' is clearly up for debate with so many types (`Date`, `Array`, etc) you could include in this list. Hence why I've started with only those types which have primitive equivalents.

JKillian:
Sounds good to me, I'm fine with leaving this rule as is then. 

As far as the strangeness with grunt and a lack of errors, I'll try to get that changed when we move to the test system proposed in #620

JKillian:
Can you just do two tweaks before this gets merged in? 
* Add a few false positive lines to the tests just as a sanity check that the rule doesn't start going off on all variable names, prefixed variable names, etc.
* Add the rule to sample.tslint.json and the readme :)

myitcv:
@JKillian - done. The formatting changes in `sample.tslint.json` are just that; correct the ordering and indentation (again would probably be a good to have this automatically formatted)

PR #745, opened 2015-10-19
Fix #743: incorrect typedef behaviour for for-of loops

### LINKED ISSUE #743: 'expected variable-declaration to have a typedef' lint error on for...of loops
When using `"typedef": [true, "variable-declaration"]`,`tslint` complains about variables within for...of loop declarations; however typedefs are not allowed in this context.

The following will result in a `expected variable-declaration: 'thing' to have a typedef ` error from `tslint`:
```Typescript
for (let thing of setOfThings) {
    ...
}
``` 
but specifying a typedef on `thing` results in the following `tsc` compilation error:
`error TS2483: The left-hand side of a 'for...of' statement cannot use a type annotation`

The `typedef` rule should ignore variables declared within for...of loop declarations

JKillian:


adidahiya:
:+1: lgtm

Issue #680, opened 2015-09-22
Lint Programs as a Whole


JKillian:
Right now, TSLint lints files individually. This means that any information in one file (such as type information) is not available while linting other files.

One possible alternative is to, at startup, create a language services instance that has the source for the entire program and pass this to the rules. Full program information could then be used by the rules.

ashwinr:
what additional features does this enable? we're not linting types yet anyway.

JKillian:
Possibly stuff like this: https://github.com/palantir/tslint/issues/577

adidahiya:
basically any rule that needs to inspect types will need all the type information from the compilation path in order to work reliably

dallonf:
It'd at least be nice to pass a glob to `tslint` and lint multiple files.

JKillian:
@dallonf When using it programmatically I assume? (Since obviously a glob will work when using it via the CLI)

dallonf:
@JKillian I didn't know that was supposed to work ;) Created a ticket for my seperate issue: https://github.com/palantir/tslint/issues/688

myitcv:
Picking up  [@aleksey-bykov's comment](https://github.com/Microsoft/TypeScript/issues/7306#issuecomment-192333122), I wonder if we can kick start some sort of effort on this.... We have a long list of rules that we want to write (let's ignore the fact many of these rules should in fact be part of TypeScript core)

@adidahiya @JKillian - are you aware of anyone who has tried to progress this?

As @aleksey-bykov points out, the effort involve in creating a custom linter that does have type information is not hard. The more complicated bit will be working out how to integrate this into `tslint`

JKillian:
Sorry for the delayed response here @myitcv. I'm not aware of any progress on this so far, feel free to go for it if you have the time! I'm happy to discuss how we can integrate this kind of thing into TSLint as it is now

alexeagle:
How will this approach match the compiler options of tsc with those passed to tslint?

For example, we need the same input files, same typings, same moduleResolution flag, same `baseUrl`,`rootDirs` (https://github.com/Microsoft/TypeScript/issues/5039), `--skipDefaultLibCheck` and so on. Otherwise the typecheck will fail when `tslint` runs, and the user has to carefully duplicate these settings.

One option is to read tsconfig.json as an input to `tslint`, but some users may have configuration in another place (we often wrap the compiler API to workaround differences in filesystem layout).

[...]

PR #748, opened 2015-10-20
Integrate no keywords rule into variable-name rule

JKillian:
@myitcv:

@adidahiya and I spoke offline and he suggested merging your rule into the `variable-name` rule.

Basically, we want to have all functionality related to the names of variables in one rule. Two benefits to this: first, it makes it easier to find the rules/options you're looking for when related functionality is grouped together. Second, it is a little bit better performance-wise, as it's one less rule to traverse the AST.

I've made a PR with the modifications, does it look good to you?


test/files/rules/varname-keywords.test.ts
```diff
@@ -0,0 +1,17 @@
...
+
+// good:
+let foo = 2;
+let bar: any;
+function good(baz: any) { }
+let [faz] = [5];
+const {pom} = {pom: 5};
+
+interface Wob {
+	number: string;
```
adidahiya: don't use tabs

src/rules/variableNameRule.ts
```diff
@@ -29,30 +34,48 @@ export class Rule extends Lint.Rules.AbstractRule {
 }
 
 class VariableNameWalker extends Lint.RuleWalker {
+    private checkFormat: boolean;
+    private banKeywords: boolean;
```
adidahiya: alphabetize these members

src/rules/variableNameRule.ts
```diff
@@ -16,11 +16,16 @@
...
+
 const OPTION_LEADING_UNDERSCORE = "allow-leading-underscore";
 const OPTION_TRAILING_UNDERSCORE = "allow-trailing-underscore";
+const OPTION_BAN_KEYWORDS = "ban-keywords";
+const OPTION_CHECK_FORMAT = "check-format";
 
 export class Rule extends Lint.Rules.AbstractRule {
-    public static FAILURE_STRING = "variable name must be in camelcase or uppercase";
+    public static FORMAT_FAILURE = "variable name must be in camelcase or uppercase";
+    public static KEYWORD_FAILURE = "variable nameclashes with keyword/type";
```
adidahiya: space between "name", "clashes"

src/rules/variableNameRule.ts
```diff
@@ -29,30 +34,48 @@ export class Rule extends Lint.Rules.AbstractRule {
...
+
+        this.banKeywords = this.hasOption(OPTION_BAN_KEYWORDS);
+        // check variable name formatting by default if no options are specified
+        this.checkFormat = !this.banKeywords || this.hasOption(OPTION_CHECK_FORMAT);
+    }
+
     public visitBindingElement(node: ts.BindingElement) {
         if (node.name.kind === ts.SyntaxKind.Identifier) {
-            this.handleVariableName(<ts.Identifier> node.name);
+            const identifier = node.name as ts.Identifier;
```
adidahiya: we don't use `as`, please use the old casting syntax

adidahiya: we should look into making some type guards so this isn't so verbose

src/rules/variableNameRule.ts
```diff
@@ -64,11 +87,19 @@ class VariableNameWalker extends Lint.RuleWalker {
         }
     }
 
-    private handleVariableName(name: ts.Identifier) {
+    private handleVariableNameFormat(name: ts.Identifier) {
```
adidahiya: no need to have separate methods, just keep the old method name and have 2 `if` blocks, that'd simplify the code so much

src/rules/variableNameRule.ts
```diff
@@ -16,11 +16,16 @@
 import * as Lint from "../lint";
 import * as ts from "typescript";
 
+const BAD_NAMES = ["any", "Number", "number", "String", "string", "Boolean", "boolean", "undefined"];
```
adidahiya: `BANNED_KEYWORDS`

src/rules/variableNameRule.ts
```diff
@@ -29,30 +34,48 @@ export class Rule extends Lint.Rules.AbstractRule {
 }
 
 class VariableNameWalker extends Lint.RuleWalker {
+    private checkFormat: boolean;
+    private banKeywords: boolean;
+
+    constructor(sourceFile: ts.SourceFile, options: Lint.IOptions) {
+        super(sourceFile, options);
+
+        this.banKeywords = this.hasOption(OPTION_BAN_KEYWORDS);
```
adidahiya: no need to store this as a member; it's only used once, just inline it

src/rules/variableNameRule.ts
adidahiya: alphabetize these members. also add the prefix `should` since they're booleans; it's more readable / straightforward (regardless of the fact that they're already of the `boolean` type).

test/rules/variableNameRuleTests.ts
```diff
@@ -92,4 +92,23 @@ describe("<variable-name>", () => {
             return optionallyValidFailures.some((f) => f.equals(failure));
         }));
     });
+
+    it("ensures keywords are banned optionally", () => {
```
adidahiya: "optionally" should be the first word in this phrase

myitcv:
@JKillian 

> Basically, we want to have all functionality related to the names of variables in one rule

No fundamental objection, and sounds like a reasonable goal. See my comment at the end of this reply.

> Two benefits to this: first, it makes it easier to find the rules/options you're looking for when related functionality is grouped together. 

Agreed. However I will add that it might be worthwhile (in the near future) having a 'registry' of rules, where you can tag each rule with certain keywords (no pun intended), easily navigate to a description of the rule and some examples (examples that are compiled examples so they don't fall out of date) etc. This would obviate the need to group things into the same rule. But I believe your point still stands in this case.

> Second, it is a little bit better performance-wise, as it's one less rule to traverse the AST.

Point taken, although you could say this about every rule.... There are other ways to solve that performance problem.

All that said... The only reason I made this a separate rule in the first place was because what's being enforced here feels like something the compiler should be doing in any case. That's a subjective comment I know, and folks in the TypeScript core team can and do [think that things like this should be linter rules](https://github.com/Microsoft/TypeScript/issues/5187#issuecomment-149785137) ([root issue](https://github.com/Microsoft/TypeScript/issues/3081)). The naming convention for a variable however is more subjective and is definitely not (if you assume the syntax is correct) something the compiler would ever fail on.

Equally, however, it's just as easy to remove an option from the `variable-name` rule so maybe it's six of one, half a dozen the other. 

But hopefully you understand my rationale for making it a separate rule in the first place.

test/rules/variableNameRuleTests.ts
JKillian: "ensure keywords can optionally be banned" is more in parallel with the other tests

src/rules/variableNameRule.ts
JKillian: Not possible w/o adding another parameter to the method. (Both methods aren't always called together)

src/rules/variableNameRule.ts
JKillian: I'm not sure how much type guards could help. They don't work on dotted names so the code probably would be just about as as verbose

JKillian:
Yep, I think either rule or rule option works fine in the end.

Just curious, have you ever run into performance issues in general? If people are, then it's something to think about, perhaps in conjunction with #680

myitcv:
> Just curious, have you ever run into performance issues in general

Not as yet, but our app is not massive. And importantly the per file lint runtime is relatively low, meaning it's quite acceptable to have a separate terminal showing you the lint errors on save (for example)

> If people are, then it's something to think about, perhaps in conjunction with #680

Indeed, and it's at this point we will probably need to consider an architecture whereby the linter can operate in a 'watch' mode, a la `tsc --watch`. Because otherwise per-file lint times start to go through the roof. And if we were to do this we start getting very close to [TypeScript #3003](https://github.com/Microsoft/TypeScript/issues/3003)

adidahiya:
:+1: lgtm

alexeagle:
:+1:  (though technically these are not keywords)
looking forward to the next release

myitcv:
> though technically these are not keywords

Agreed, not all of them are keywords, but I [borrowed from the spec](https://github.com/Microsoft/TypeScript/blob/master/doc/spec.md#321-the-number-type) for the best possible all-encompassing term that captures the spirit of what we're after here. `number` is a keyword according to the language of the spec, with `Number` its primitive type counterpart. Very open to other suggestions.

Incidentally, I think we should add `Undefined` to this list (even though it cannot be referenced as a type)

JKillian:
PR to add `Undefined` for consistency: #749

adidahiya:
@alexeagle @myitcv just published a dev release, 3.0.0-dev.1, with this new rule option

myitcv:
@adidahiya great thanks

PR #749, opened 2015-10-22
Add Undefined to List of Banned Var Names

JKillian:
See #748 

(No tests added because we don't really need to test that every term in the list is banned)


myitcv:
:+1:

adidahiya:
sweet

PR #761, opened 2015-10-29
Rework Build System

### LINKED ISSUE #760: Fix Interop with CommonJS Usage
CommonJS consumers have to use tslint as such right now:

```ts
var tslint = require('tslint').default;
```

This is unexpected behavior and an unnecessary breaking change.

JKillian:
The build flow is now TS [-> via TypeScript ->] ES6 [-> via Babel ->] ES5

There are a few benefits to this commit:
  * It fixes #760. Why? ES6 modules export their default content as a member called `default`. When we switched over in #726 to ES6 modules we introduced this problem for CommonJS consumers. Babel has a special feature that fixes this. So,  [we use TS then Babel](http://www.jbrantly.com/es6-modules-with-typescript-and-webpack/#packagingasalibrary) to get the result we want.

There are a few other minor tweaks, but the points above are the major parts of this PR


src/configuration.ts
```diff
@@ -16,7 +16,10 @@
 
 import * as fs from "fs";
 import * as path from "path";
-import * as findup from "findup-sync";
+
+// hack to correctly import JS module that uses module.exports = function(){...}
+import * as findupSync from "findup-sync";
+const findup = (<any>findupSync).default;
```
JKillian: @weswigham Do you know if there's a better way to do this when targeting ES6 and then using Babel?

JKillian: Nevermind, I think adjusting the typings is the best bet here

JKillian:
Whoops, this should merge into master and not next. Closing this in favor of #762

PR #762, opened 2015-10-30
Rework build system

### LINKED ISSUE #760: Fix Interop with CommonJS Usage

JKillian:
The build flow is now TS [-> via TypeScript ->] ES6 [-> via Babel ->] ES5

There are a few benefits to this commit:
  * It fixes #760. Why? ES6 modules export their default content as a member called `default`. When we switched over in #726 to ES6 modules we introduced this problem for CommonJS consumers. Babel has a special feature that fixes this. So,  [we use TS then Babel](http://www.jbrantly.com/es6-modules-with-typescript-and-webpack/#packagingasalibrary) to get the result we want.

There are a few other minor tweaks, but the points above are the major parts of this PR


JKillian:
Let's use #764 and #765 instead of introducing Babel.

Issue #583, opened 2015-08-19
Exception from tslint with template strings in className attribute in .tsx


benkopolis:
Hey, 
I encountered an intermittent exception from tslint when using template string in className attribute in tsx syntax. After some random changes and some time the exception may be gone, but not always... And I'm not getting it every time I'm using this stuff. 
This is how code looks like when tslint throws an exception exception (it doesn't help to disable / enable tslint here):
```typescript
return <div>
           <img ref="resizableImage" className={`${this.props.wrapperClassName} otherimageclass`} src={this.props.url}/>
           <div className={`resizableThumbnail ${styleName}`}/>
      </div>;
```
Below is a code with workaround, that makes tslint to work (tslint:disable doesn't work!), but I don't like it, it's not as readable and clean as a code above. 
```typescript
let cssClass: string = `${this.props.wrapperClassName} otherimageclass`;
        return <div>
                    <img ref="resizableImage" className={cssClass} src={this.props.url}/>
                    <div className={`resizableThumbnail ${styleName}`}/>
               </div>;
```
The exception I'm getting:
```
ERROR in MyFile.tsx
Module build failed: TypeError: Cannot read property 'length' of undefined
    at NoUseBeforeDeclareWalker.validateUsageForVariable (MySrcDir\node_modules\tslint-loader\node_modules\tslint\build\rules\noUseBeforeDeclareRule.js:91:41)
    at NoUseBeforeDeclareWalker.visitVariableDeclaration (MySrcDir\node_modules\tslint-loader\node_modules\tslint\build\rules\noUseBeforeDeclareRule.js:84:18)
    at NoUseBeforeDeclareWalker.SyntaxWalker.visitNode (MySrcDir\node_modules\tslint-loader\node_modules\tslint\lib\tslint.js:48633:26)
    at NoUseBeforeDeclareWalker.ScopeAwareRuleWalker.visitNode (MySrcDir\node_modules\tslint-loader\node_modules\tslint\lib\tslint.js:48775:40)
    at MySrcDir\node_modules\tslint-loader\node_modules\tslint\lib\tslint.js:48648:67
    at visitEachNode (MySrcDir\node_modules\tslint-loader\node_modules\tslint\lib\tslint.js:7147:30)
    at Object.forEachChild (MySrcDir\node_modules\tslint-loader\node_modules\tslint\lib\tslint.js:7305:24)
[...]

PR #763, opened 2015-11-02
Make sure highlights is not null

JKillian:
See https://github.com/palantir/tslint/issues/583#issuecomment-153128142 and below for relevant discussion


adidahiya:
:+1:

PR #764, opened 2015-11-03
Make src/tslint.ts a commonjs-style module instead of an ES6 module.

JKillian:
Reverts 'tslint' back to being an old-style module so it works better with CommonJS.

### DIFF
--- a/src/lint.ts
+++ b/src/lint.ts
@@ -1,6 +1,8 @@
-import * as formatters from "./formatters";
 import * as configuration from "./configuration";
+import * as formatters from "./formatters";
+import * as linter from "./tslint";
 import * as rules from "./rules";
+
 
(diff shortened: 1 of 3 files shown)


JKillian:
Should fix #760

Downsides: We now have a CommonJS-style module scattered among ES6-style modules

src/lint.ts
```diff
@@ -31,4 +31,4 @@ export interface ILinterOptions {
     rulesDirectory: string;
 }
 
-export {default as Linter} from "./tslint";
+export import Linter = require("./tslint");
```
adidahiya: can we `import * as Linter` in this file as well, instead of `require()`? I'd like to turn on the `no-require-imports` rule...

Issue #617, opened 2015-08-27
Update grunt-ts, use tsconfig.json support


adidahiya:
beta is coming soon: https://github.com/TypeStrong/grunt-ts/issues/202#issuecomment-135495091

PR #765, opened 2015-11-03
Use tsconfig.json to configure build

JKillian:
A subset of the changes listed in #762.

Changes:


adidahiya:
Fixes #617

.gitignore
```diff
@@ -5,3 +5,4 @@ bin/tslint-cli.js
 lib/
 node_modules/
 tscommand*.txt
+typings/.basedir.ts
```
adidahiya: let's add a comment about this

src/tsconfig.json
```diff
@@ -8,63 +8,94 @@
...
+        "rules/trailingCommaRule.ts",
+        "rules/tripleEqualsRule.ts",
+        "rules/typedefRule.ts",
+        "rules/typedefWhitespaceRule.ts",
+        "rules/useStrictRule.ts",
+        "rules/variableNameRule.ts",
+        "rules/whitespaceRule.ts"
     ]
-}
+}
```
adidahiya: put back trailing newline

test/tsconfig.json
```diff
@@ -5,89 +5,165 @@
...
+        "rules/variableNameRuleTests.ts",
+        "rules/whitespaceRuleTests.ts",
+        "formatters/externalFormatterTest.ts",
+        "formatters/jsonFormatterTests.ts",
+        "formatters/pmdFormatterTests.ts",
+        "formatters/proseFormatterTests.ts",
+        "formatters/verboseFormatterTests.ts"
     ]
-}
+}
```
adidahiya: put back trailing newline

PR #775, opened 2015-11-05
Cleanup

adidahiya:
- whitespace fixes
- some import reordering
- `for ... of` loops
- remove implicit type coercion to bool
- convert methods to private helper functions when possible (all static methods)
- rename walker classes to match file names
- some more random issues

### DIFF
--- a/src/language/languageServiceHost.ts
+++ b/src/language/languageServiceHost.ts
@@ -17,18 +17,16 @@
 import * as Lint from "../lint";
 import * as ts from "typescript";
 
-export function createLanguageServiceHost(fileName: string, source: string) {
-    const host: ts.LanguageServiceHost = {
+export function createLanguageServiceHost(fileName: string, source: string): ts.LanguageServiceHost {
+    return {
         getCompilationSettings: () => Lint.createCompilerOptions(),
         getCurrentDirectory: () => "",
         getDefaultLibFileName: () => "lib.d.ts",
         getScriptFileNames: () => [fileName],
-        getScriptSnapshot: (name) => ts.ScriptSnapshot.fromString(name === fileName ? source : ""),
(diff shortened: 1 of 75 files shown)


src/language/languageServiceHost.ts
```diff
@@ -17,18 +17,16 @@
...
+    return {
         getCompilationSettings: () => Lint.createCompilerOptions(),
         getCurrentDirectory: () => "",
         getDefaultLibFileName: () => "lib.d.ts",
         getScriptFileNames: () => [fileName],
-        getScriptSnapshot: (name) => ts.ScriptSnapshot.fromString(name === fileName ? source : ""),
+        getScriptSnapshot: (name: string) => ts.ScriptSnapshot.fromString(name === fileName ? source : ""),
         getScriptVersion: () => "1",
-        log: (message) => { /* */ }
+        log: () => { /* */ }
```
JKillian: We could probably just delete this line all together, it's an optional field. (I don't really know what it does though by default). If we keep it though, I think we should leave the message param in, since the type of `log` is `(s: string) => void`.

src/language/walker/syntaxWalker.ts
```diff
@@ -314,10 +315,6 @@ export class SyntaxWalker {
                 this.visitArrayLiteralExpression(<ts.ArrayLiteralExpression> node);
                 break;
 
-            case ts.SyntaxKind.BindingElement:
```
JKillian: why is this being deleted?

adidahiya: we have a duplicate case on line 326

src/language/languageServiceHost.ts
adidahiya: ok I'll remove this line

JKillian: Feel free to in a follow-up if you want, definitely not a blocking thing for me though. Going to go ahead and merge this

Issue #791, opened 2015-11-12
TypeScript version dependency for tslint@next


myitcv:
Picking up on https://github.com/palantir/tslint/pull/726#issuecomment-151720524

Specifically:

> I think we ought to fix TS to a specific version for the `next` branch. So instead of the dependency being `typescript: next`, we ought to fix it to something like `typescript: 1.8.0-dev.20151027` and then update that each week.

At the moment with `tslint@next` depending on `typescript@next` this forces our project to also depend on `typescript@next`. Otherwise, we see errors like [these](https://github.com/palantir/tslint/pull/726#issuecomment-151715202). 

But being forced to depend on `typescript@next` means our CI build is not stable from one day to the next (because `typescript@next` points to a nightly build)

One option would be to fix the dependency as [suggested by @JKillian](https://github.com/palantir/tslint/pull/726#issuecomment-151720524) and bump on a weekly basis. This has the downside of constraining our project to also depend on that same version and only moving when `tslint@next` updates its dependency.

Would it instead be possible to have `tslint@next` depend on TypeScript as a peer dependency? Would that work?

I'll admit now I don't understand all the factors at play here.. so please dismiss this suggestion if it's a load of rubbish!

cc @weswigham given his previous thoughts/comments on this topic

adidahiya:
Thanks for bringing this up again and the writeup! I'm in favor of making TS a peer dependency. That makes it so TS is installed in a deterministic location for `tslint`, custom rules, and apps to consume.

JKillian:
@myitcv Do you think this should also be done for the master branch?

adidahiya:
I think we should do the same on the master branch, yes. We'll make that change with 3.0, for same reason I mentioned above (deterministic location for tsc).

myitcv:
Sounds good @adidahiya @JKillian

adidahiya:
published `3.0.0-dev.3` with the peer dependency change.

adidahiya:
3.0 release PR: #804

PR #792, opened 2015-11-13
Move to a peer dependency for typescript

JKillian:
Accept all versions of 1.7.0-dev after the jsx spread bugfix, accept all versions of 1.8.0-dev, accept any release versions >= 1.7.0

### DIFF
--- a/package.json
+++ b/package.json
@@ -22,8 +22,7 @@
   "dependencies": {
     "findup-sync": "~0.2.1",
     "optimist": "~0.6.0",
-    "underscore.string": "~3.1.1",
-    "typescript": "next"
+    "underscore.string": "~3.1.1"
   },
   "devDependencies": {
     "chai": "^3.0.0",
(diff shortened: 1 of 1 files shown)


JKillian:
Addresses #791

JKillian:
hmm, this only seems to fail on Node 0.10 with a 'typescript' not found warning. Any ideas @adidahiya?

package.json
```diff
@@ -38,5 +37,8 @@
     "mocha": "^2.2.5",
     "tslint": "latest"
```
adidahiya: I think you might need to keep typescript as a dev dependency

JKillian:
Might be because with Node 0.10 it uses npm 1.x whereas with Node 0.12+ it uses npm 2.x

package.json
```diff
@@ -36,7 +35,11 @@
     "grunt-ts": "^5.1.0",
     "grunt-tslint": "latest",
     "mocha": "^2.2.5",
-    "tslint": "latest"
+    "tslint": "latest",
+    "typescript": ">=1.7.0 || >=1.7.0-dev.20151003 || >=1.8.0-dev"
```
adidahiya: I think the dev dependency should be `next`, like it was before

JKillian: Good call, just pushed an update with that change

PR #793, opened 2015-11-13
Merge `master` into `next`

JKillian:


JKillian:
Let's merge the react rule bugfix before this.

PR #798, opened 2015-11-13
Prepare release v3.0.0-dev.3

JKillian:


adidahiya:
oops, merged too soon. this doesn't actually include the peer dependency change.

PR #817, opened 2015-11-23
Only lint call expressions and external module references in no-require-imports rule

### LINKED ISSUE #816: no-require-imports false positive
```ts
import TestUtils = React.addons.TestUtils;
```

should not report a failure with the `no-require-imports` rule.

adidahiya:
Fixes #816.

### DIFF
--- a/src/rules/noRequireImportsRule.ts
+++ b/src/rules/noRequireImportsRule.ts
@@ -26,27 +26,21 @@ export class Rule extends Lint.Rules.AbstractRule {
 }
 
 class NoRequireImportsWalker extends Lint.RuleWalker {
-    public visitVariableStatement(node: ts.VariableStatement) {
-        const declarations = node.declarationList.declarations;
-        for (let decl of declarations) {
-            this.handleDeclaration(decl);
+    public visitCallExpression(node: ts.CallExpression) {
+        if (node.arguments != null && node.expression != null) {
+            const callExpressionText = node.expression.getText(this.getSourceFile());
+            if (callExpressionText === "require") {
+                this.addFailure(this.createFailure(node.getStart(), node.getWidth(), Rule.FAILURE_STRING));
+            }
         }
-        super.visitVariableStatement(node);
+        super.visitCallExpression(node);
     }
 
     public visitImportEqualsDeclaration(node: ts.ImportEqualsDeclaration) {
-        this.addFailure(this.createFailure(node.getStart(), node.getWidth(), Rule.FAILURE_STRING));
-        super.visitImportEqualsDeclaration(node);
-    }
-
-    private handleDeclaration(decl: ts.VariableDeclaration)  {
-        // make sure the RHS is a call expression.
-        const call = <ts.CallExpression> (decl.initializer);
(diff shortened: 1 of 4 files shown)


src/rules/noRequireImportsRule.ts
```diff
@@ -26,27 +26,20 @@ export class Rule extends Lint.Rules.AbstractRule {
...
-
-    private handleDeclaration(decl: ts.VariableDeclaration)  {
-        // make sure the RHS is a call expression.
-        const call = <ts.CallExpression> (decl.initializer);
-        if (call && call.arguments && call.expression) {
-            const callExpressionText = call.expression.getText(this.getSourceFile());
-            if (callExpressionText === "require") {
-                this.addFailure(this.createFailure(decl.getStart(), decl.getWidth(), Rule.FAILURE_STRING));
-            }
+        const {moduleReference} = node;
```
JKillian: Do we normally use spaces here? `{ moduleReference }`? Not sure

src/rules/noRequireImportsRule.ts
```diff
@@ -26,27 +26,20 @@ export class Rule extends Lint.Rules.AbstractRule {
 }
 
 class NoRequireImportsWalker extends Lint.RuleWalker {
-    public visitVariableStatement(node: ts.VariableStatement) {
-        const declarations = node.declarationList.declarations;
-        for (let decl of declarations) {
-            this.handleDeclaration(decl);
+    public visitCallExpression(node: ts.CallExpression) {
```
JKillian: Hmm, I guess this actually bans the usage of any function named `require` ever, but I'm not sure there's any better way to do it.

src/rules/noRequireImportsRule.ts
```diff
@@ -26,27 +26,20 @@ export class Rule extends Lint.Rules.AbstractRule {
...
+        if (node.arguments != null && node.expression != null) {
+            const callExpressionText = node.expression.getText(this.getSourceFile());
+            if (callExpressionText === "require") {
+                this.addFailure(this.createFailure(node.getStart(), node.getWidth(), Rule.FAILURE_STRING));
+            }
         }
-        super.visitVariableStatement(node);
```
JKillian: We might still want to call super here, imagine something weird like:
```typescript
let foo;
let bar = someMethod(foo = require('foo'));
```

src/rules/noRequireImportsRule.ts
adidahiya: no

src/rules/noRequireImportsRule.ts
adidahiya: yeah, I would consider that bad form if you're not in a node environment. and it's simpler for TSLint not to ask about your project's environment (browser vs. node)

src/rules/noRequireImportsRule.ts
adidahiya: ok good point

adidahiya:
looks like this will not build until #834 goes in.

adidahiya:
rebased & squashed.

PR #819, opened 2015-11-24
Tighten imports across core sources

adidahiya:
This change reduces usage of `import * as Lint` in favor of more granular imports. This reduces cyclical dependencies between our modules.


adidahiya:
related issue which prompted me to do this: https://travis-ci.org/palantir/tslint/jobs/92878481

adidahiya:
rebased

JKillian:
:+1: Diff looks good.

PR #820, opened 2015-11-25
Merge master -> next

adidahiya:


adidahiya:
nevermind, package.json shouldn't be changed; this should be from a new branch

Issue #692, opened 2015-09-28
Better support for linting multiple files


adidahiya:
Related: #680
- [ ] if given a directory, TSLint should recursively search for `.ts{,x}` files in that directory and add them to be linted (similar to ESLint)
  - implemented here: #857 
- if provided the path to a `tsconfig.json` file, TSLint should lint all the files listed in the `"files"` list and files they reference via imports or references
  - Q: should we ignore `.d.ts` files by default? and provide an option to opt-in to linting them?
  - Q: should this be the default if no files are specified? i.e. TSLint searches for a `tsconfig.json` file in the same way it currently searches for a `tslint.json` file
  - **moved to #858**
- [x] support globs in CLI on Windows (#688, #779)
  - implemented here: #827

Pajn:
>Q: should we ignore .d.ts files by default? and provide an option to opt-in to linting them?
Yes please. Getting lint errors because the typings from tsd doesn't follow the same code standard as the project would be pretty messy.
An opt-out would work as well, but I think the most common would be to not wanting to lint `.d.ts` files.

dsherret:
:+1: and I agree with ignoring `.d.ts` files by default.

raveclassic:
> TSLint should lint all the files listed in the "files"
Are you going to support `fileGlobs` from `tsconfig` as well as `files`? 
This would be very helpful, as it is close to impossible to define all the sources in `files` in large projects.

JKillian:
@raveclassic Definitely want to support `filesGlob` and `exclude`. Without those things are significantly more painful

raveclassic:
@JKillian Thanks, waiting for changes. For now will use gulp to gather the files. 
Besides, great work guys!

radek-holy:
I wonder what's the benefit of merging multiple reports into one... I think that three (relatively) small issues have a greater chance to be implemented than one big issue. But sure, it's up to you how you manage your project...

louy:
I have an unfinished solution for task 2 and #680 in [here](https://github.com/louy/tslint/tree/tsconfig-support). I don't have enough time to finish it so if anyone would like to step in feel free to do so.

BTW, after #857 is merged I feel that this one should be closed. Task 2 isn't relevant here.

adidahiya:
moved task 2 (tsconfig.json support) to #858

unional:
> I agree with ignoring .d.ts files by default

Please do not do that, at least not hard coding that. Using tslint to lint `.d.ts` for typings.

[...]

PR #827, opened 2015-11-28
Better support for globs in cli

louy:
Related to #692. Potentially fixed glob issues in windows.

This makes the cli accept a glob as a string and expand it.

It solves a problem with "arguments list is too large" when using regular globs.

### DIFF
--- a/package.json
+++ b/package.json
@@ -21,6 +21,7 @@
   },
   "dependencies": {
     "findup-sync": "~0.2.1",
+    "glob": "^6.0.1",
     "optimist": "~0.6.0",
     "underscore.string": "~3.1.1"
   },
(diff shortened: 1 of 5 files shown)


louy:
@adidahiya I think I'll add support for directories using a glob pattern "[directory]/**/*.ts" for all passed directories. What do you think?

adidahiya:
I don't see a need for automatically expanding directory paths into globs, so what you have here looks fine.

If your shell supports globs, wouldn't it still expand them though? How would you get around the "argument list is too large" issue? Would you have to provide a glob path as a string? `tslint 'src/**/*.ts'`?

louy:
Yeah you'll have to provide the path as a string. However, I'm planing to accept directories and convert them to a glob like "[directory]/**.ts". this will be done in a separate PR, so in the future you'll be able to do "tslint ." or "tslint src/" and it will walk through all .ts files in that directory.

adidahiya:
@louy cool, I like the idea of adding support for directories in a separate PR

PR #841, opened 2015-12-01
[WIP] Enforce tsfmt of tslint code base

myitcv:
[View diff ignoring whitespace changes](https://github.com/palantir/tslint/compare/master...myitcv:tsfmt_code?w=1)

Tasks as part of this PR:

* [ ] Agree on settings for `tsfmt.json`
* [ ] Define scope of which files should be formatted
  * `**/*.@(ts|tsx)`
  * Exclude:
    * `lib`
    * `test/files/**`
* [ ] Format all code (excluding test files) - partially complete subject to previous points
* [ ] Write `grunt` task to enforce formatting of files in scope (fails if `tsfmt --verify` fails) (see #732)
* [ ] Write `grunt` task to format and replace files in scope
* [ ] Add docs for contributors explaining the setup:
  * Reference new `grunt` tasks
  * Suggest they setup `tsfmt` as part of 'save' hook in editor

### DIFF
--- a/src/rules/eoflineRule.ts
+++ b/src/rules/eoflineRule.ts
@@ -32,7 +32,7 @@ export class Rule extends Lint.Rules.AbstractRule {
         if (eofTokenFullText.length === 0 || eofTokenFullText.charAt(eofTokenFullText.length - 1) !== "\n") {
             const start = eofToken.getStart();
             return [
-                new Lint.RuleFailure(sourceFile, start, start, Rule.FAILURE_STRING, this.getOptions().ruleName)
+                new Lint.RuleFailure(sourceFile, start, start, Rule.FAILURE_STRING, this.getOptions().ruleName),
             ];
         }
 
--- a/src/tslint-cli.ts
+++ b/src/tslint-cli.ts
@@ -72,7 +72,7 @@ let processed = optimist
             describe: "output format (prose, json, verbose)",
         },
         "test": {
-            describe: "test that tslint produces the correct output for the specified directory"
+            describe: "test that tslint produces the correct output for the specified directory",
         },
         "v": {
             alias: "version",
--- a/test/rule-tester/testData.ts
+++ b/test/rule-tester/testData.ts
@@ -36,7 +36,7 @@ export const codeStr2 = `
 A file with an error
 `;
 export const resultErrs2: LintError[] = [
-  { startPos: { line: 1, col: 0 }, endPos: { line: 1, col: 5 }, message: "error" }
+  { startPos: { line: 1, col: 0 }, endPos: { line: 1, col: 5 }, message: "error" },
 ];
 
 
@@ -105,7 +105,7 @@ if (code === lint-error-free) {
 
 `;
 export const resultErrs6: LintError[] = [
-    { startPos: { line: 1, col: 13 }, endPos: { line: 1, col: 28 }, message: "A longer error message I didn't want to type every time!" }
+    { startPos: { line: 1, col: 13 }, endPos: { line: 1, col: 28 }, message: "A longer error message I didn't want to type every time!" },
 ];
 
 export const lintStr7 = `


myitcv:
@adidahiya @JKillian kicking this off, your comments welcomed.

JKillian:
Here's my initial thoughts on settings are similar to the ones in the README for typescript-formatter (see below). What settings did you use when you ran it above? 

```
{
  "indentSize": 4,
  "tabSize": 4,
  "newLineCharacter": "\n",
  "convertTabsToSpaces": true,
  "insertSpaceAfterCommaDelimiter": true,
  "insertSpaceAfterSemicolonInForStatements": true,
  "insertSpaceBeforeAndAfterBinaryOperators": true,
  "insertSpaceAfterKeywordsInControlFlowStatements": true,
  "insertSpaceAfterFunctionKeywordForAnonymousFunctions": false,
  "insertSpaceAfterOpeningAndBeforeClosingNonemptyParenthesis": false,
  "insertSpaceAfterOpeningAndBeforeClosingNonemptyBrackets": false,
  "placeOpenBraceOnNewLineForFunctions": false,
  "placeOpenBraceOnNewLineForControlBlocks": false
}
```

src/enableDisableRules.ts
```diff
@@ -19,7 +19,7 @@ import {SkippableTokenAwareRuleWalker} from "./language/walker/skippableTokenAwa
 import * as ts from "typescript";
 
 export class EnableDisableRulesWalker extends SkippableTokenAwareRuleWalker {
-    public enableDisableRuleMap: {[rulename: string]: Lint.IEnableDisablePosition[]} = {};
+    public enableDisableRuleMap: { [rulename: string]: Lint.IEnableDisablePosition[] } = {};
```
JKillian: perhaps disable this tsfmt rule

src/language/rule/rule.ts
```diff
@@ -79,10 +79,10 @@ export class RuleFailure {
     private ruleName: string;
 
     constructor(sourceFile: ts.SourceFile,
-                start: number,
-                end: number,
-                failure: string,
-                ruleName: string) {
+        start: number,
```
JKillian: assuming this is part of the indent rule? I like the previous alignment better, but I also like automation over manual work :)

src/rules/labelPositionRule.ts
```diff
@@ -29,10 +29,10 @@ class LabelPositionWalker extends Lint.RuleWalker {
...
         const statement = node.statement;
         if (statement.kind !== ts.SyntaxKind.DoStatement
-                && statement.kind !== ts.SyntaxKind.ForStatement
-                && statement.kind !== ts.SyntaxKind.ForInStatement
-                && statement.kind !== ts.SyntaxKind.WhileStatement
-                && statement.kind !== ts.SyntaxKind.SwitchStatement) {
+            && statement.kind !== ts.SyntaxKind.ForStatement
+            && statement.kind !== ts.SyntaxKind.ForInStatement
+            && statement.kind !== ts.SyntaxKind.WhileStatement
+            && statement.kind !== ts.SyntaxKind.SwitchStatement) {
```
JKillian: The downside of this alignment is that the conditions kind of blend in with the statements inside the if block. Assuming this is part of the indentation feature as well?

test/formatters/pmdFormatterTests.ts
```diff
@@ -38,22 +38,22 @@ describe("PMD Formatter", () => {
...
-                "</file>" +
-                "<file name=\"formatters/pmdFormatter.test.ts\">" +
-                    "<violation begincolumn=\"3\" beginline=\"6\" priority=\"1\" rule=\"last failure\"> " +
-                    "</violation>" +
-                "</file>" +
-                "<file name=\"formatters/pmdFormatter.test.ts\">" +
-                    "<violation begincolumn=\"1\" beginline=\"1\" priority=\"1\" rule=\"full failure\"> " +
-                    "</violation>" +
-                "</file>" +
+            "<file name=\"formatters/pmdFormatter.test.ts\">" +
```
JKillian: This also isn't great, but this should probably be switched to a template/multiline string anyways

JKillian:
Left a few comments for discussion - they aren't things to fix right away, just points of interest to note as we try to figure out good settings for tsfmt/if it'll do what we want.

src/enableDisableRules.ts
myitcv: There is no option to disable this particular formatting

src/language/rule/rule.ts
myitcv: Indeed, automation is the goal here!

src/rules/labelPositionRule.ts
myitcv: Arguably if readability is an issue here, this massive conditional check could be extracted into a function

test/formatters/pmdFormatterTests.ts
myitcv: I would hate to have to format this by hand, even the first time!

myitcv:
@JKillian - thanks for the comments.

As you alluded to, I think the principal goal here is automation. We shouldn't have to manually format code: it isn't an effective use of anyone's time. 

I've added a commit of the `tsfmt.json` to this PR (apologies I missed that before) which shows the options I used.

In some respects, it's a disappointment that there are any [formatting options](https://github.com/Microsoft/TypeScript/blob/master/lib/typescript.d.ts#L1819) in the first place (you will notice that `typescript-formatter` is simply exposing the core TypeScript formatting options):

```typescript
    interface FormatCodeOptions extends EditorOptions {
        InsertSpaceAfterCommaDelimiter: boolean;
        InsertSpaceAfterSemicolonInForStatements: boolean;
        InsertSpaceBeforeAndAfterBinaryOperators: boolean;
        InsertSpaceAfterKeywordsInControlFlowStatements: boolean;
        InsertSpaceAfterFunctionKeywordForAnonymousFunctions: boolean;
        InsertSpaceAfterOpeningAndBeforeClosingNonemptyParenthesis: boolean;
        InsertSpaceAfterOpeningAndBeforeClosingNonemptyBrackets: boolean;
        PlaceOpenBraceOnNewLineForFunctions: boolean;
        PlaceOpenBraceOnNewLineForControlBlocks: boolean;
        [s: string]: boolean | number | string;
    }
```

Simply because it leads to bike-shedding on which options should be on off, etc. The core TypeScript team should, in my opinion, define how TypeScript code is formatted, then everyone can move on. 

The reality however is that we do have some options (as above) so some scope for bike-shedding :bike:!!

src/rules/labelPositionRule.ts
JKillian: good point, that's always a possibility for any massive conditional, or it could be a variable beforehand as well.

JKillian:
@myitcv After looking at this again, I'm a :+1: for it. @adidahiya what do you think?

src/language/utils.ts
```diff
@@ -89,8 +89,8 @@ export function hasModifier(modifiers: ts.ModifiersArray, ...modifierKinds: ts.S
  */
 export function isBlockScopedVariable(node: ts.VariableDeclaration | ts.VariableStatement): boolean {
     const parentNode = (node.kind === ts.SyntaxKind.VariableDeclaration)
-        ? (<ts.VariableDeclaration> node).parent
-        : (<ts.VariableStatement> node).declarationList;
+        ? (<ts.VariableDeclaration>node).parent
+        : (<ts.VariableStatement>node).declarationList;
```
adidahiya: Is there an option to turn this formatting rule off? I prefer the space here.

src/language/rule/rule.ts
adidahiya: Eh, I don't like this one. It goes against the align rule and I like the readability we had here.

test/ruleDisableEnableTests.ts
```diff
@@ -20,10 +20,12 @@ describe("Enable and Disable Rules", () => {
...
-        const validConfiguration = {rules: {
-            "variable-name": true,
-            "quotemark": [true, "double"]
-        }};
+        const validConfiguration = {
+            rules: {
+                "variable-name": true,
+                "quotemark": [true, "double"]
+            }
+        };
```
adidahiya: Nice, this is a good catch

adidahiya:
I don't agree with a couple of the formatting rules but in general this is a good idea.

src/language/utils.ts
myitcv: No there is no option for this

myitcv:
@JKillian @adidahiya thanks for the comments.

> I don't agree with a couple of the formatting rules...

The options are therefore:

1. ask the core TypeScript team to change the default formatting rules
2. ask the core TypeScript team to add an option

Option 1 is clearly a non-starter given the current behaviour of the formatter; the core TypeScript team would almost certainly agree on that point. 

I very much hope that the core TypeScript team would also dismiss option 2; simply because another formatting option increases by at least a factor of 2 the number of permutations in which TypeScript code can be formatted.

Why?

My mental model here is very much driven by [`gofmt`](https://blog.golang.org/go-fmt-your-code). I'm going to shamelessly steal from the linked page to say that `tsfmt` formatted code is:

* easier to write: never worry about minor formatting concerns while hacking away
* easier to read: when all code looks the same you need not mentally convert others' formatting style into something you can understand
* easier to maintain: mechanical changes to the source don't cause unrelated changes to the file's formatting; diffs show only the real changes
* uncontroversial: never have a debate about spacing or brace position ever again!

Hence why I see the benefit of _not_ having options for the automatic formatting of code. The core TypeScript team should define how code is formatted (just as the Go team did) and everyone be happy with that.

> ... but in general this is a good idea

I think the benefits are clear to everybody. It's just that, understandably, people are coming into this sort of discussion with different mental models/experience in other languages.

I'll rebase and reformat and borrow from @weswigham's PR to get the `grunt` tasks in place too. Then we can take a another look

JKillian:
@myitcv Yes, even if the automatic formatted code is slightly less readable in a couple places, the extra effort to read the code seems to me to be less than the extra effort it takes to ensure all your code is formatted perfectly at every character. 

When I'm writing code, the difference between: `if (foo === bar)`  and `if(foo===bar)`is completely irrelevant because there is no semantic difference. Making me worry about the difference between those two takes away thought and concentration from actual functionality issues and thus make code worse (or more timely/costly to write). 

I think this does point to the advantage of style rules in a linter as well. I don't have to worry about trailing whitespace at the end of lines because the linter will notify me and then I can take care of it in a quick separate pass at the end. (Of course, automatically fixing things is even better.)

In other words, on automatic formatting, I agree! Although I don't really mind there being options in `tsfmt`, it seems to me to be fine either way.

myitcv:
@adidahiya @JKillian - looks like this PR has got itself into a funny state after I force pushed a revised version of these changes against the latest master. I'll open a new PR but reference the comments and discussion in this thread

PR #844, opened 2015-12-01
Don't double-call getRulesDirectories()

SlurpTheo:
Relative path to `--rules-dir` is getting two times the number of `'..\\'` prefixes that it needs.

`getRulesDirectories()` is already called inside `ruleLoader.ts::loadRules()` and each call is currently calling `getRelativePath()`.

### DIFF
--- a/src/tslint.ts
+++ b/src/tslint.ts
@@ -17,7 +17,7 @@
 import {IFormatter} from "./language/formatter/formatter";
 import {RuleFailure} from "./language/rule/rule";
 import {getSourceFile} from "./language/utils";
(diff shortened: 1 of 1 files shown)


adidahiya:
@SlurpTheo nice, thanks for the PR!

adidahiya:
@SlurpTheo I just published v3.2.0-dev.1 to `tslint@next`, can you check it out to verify the fix?

SlurpTheo:
@adidahiya :thumbsup:


I did `npm install tslint@next` & `npm install typescript@1.7` in my test as I had my `package.json` fixed to 3.0/1.6 versions while I waited for this fix to be published.

PR #849, opened 2015-12-03
Standardize license docs as JSDoc.

achew22:
JSDoc has a @license annotation. This change modifies all .ts files and
adds @license to every annotation. It also switches to a real JSDoc
leading comment /** instead of just /*.

This change will make it so that your license is retained when code goes through Google's compiler[0]
[0] https://developers.google.com/closure/compiler/faq?hl=en#license

### DIFF
--- a/src/formatters.ts
+++ b/src/formatters.ts
@@ -1,4 +1,5 @@
 /*
+ * @license
  * Copyright 2013 Palantir Technologies, Inc.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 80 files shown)


src/formatters.ts
```diff
@@ -1,4 +1,5 @@
 /*
```
JKillian: missing a `*` here

JKillian:
Thanks! This seems like a good idea and the PR looks good. Just found one tiny tweak to make if you're able, comment is above.

PR #852, opened 2015-12-05
Cleanup test/

adidahiya:
- Tighten module imports across test sources (no `import *`)
- Allow Atom to rewrite tsconfig.json
- memberAccessTests.ts -> memberAccessRuleTests.ts
- test/helper.ts -> test/utils.ts; import it as TestUtils
- test/references.ts -> test/assert.ts
- Remove test/chaiAssert.d.ts
- Add test/typings/tsd.d.ts
- Remove lib/.gitkeep
- update .gitignore and .npmignore to better match our repo structure now that we don't concatenate to one file during TS compilation


test/formatters/externalFormatterTest.ts
```diff
@@ -1,62 +1,61 @@
...
-    it("handles no failures", () => {
-        const result = formatter.format([]);
-        assert.equal(result, "");
-    });
-
-    function getPositionString(line: number, character: number) {
-        return "[" + line + ", " + character + "]";
-    }
-});
+/*
```
adidahiya: converted line endings to LF in this file.

src/tsconfig.json
```diff
@@ -9,7 +9,7 @@
         "outDir": "../lib"
     },
     "atom": {
-        "rewriteTsconfig": false
+        "rewriteTsconfig": true
```
JKillian: I don't think we should turn this on. It makes it so devs working on tslint have to be using atom to generate the correct tsconfig.json file. Right now it's updated as part of the build - this is better because it works the same for everyone.

.gitignore
```diff
@@ -1,8 +1,7 @@
 .DS_Store
 .tscache/
-build/
-bin/tslint-cli.js
-lib/
+/build/
+/lib/
 node_modules/
```
JKillian: Didn't know exactly how this behaved differently, learned something new! (http://stackoverflow.com/questions/8527597/how-do-i-ignore-files-in-a-directory-in-git) 
Seems like a good change

test/assert.ts
```diff
@@ -14,6 +14,12 @@
  * limitations under the License.
  */
 
+declare namespace NodeJS {
+    interface Global {
+        assert: Chai.Assert;
```
JKillian: Confused why we need this and line 24. (I just don't get what's going on here, can you explain it?)

test/lint.ts
```diff
@@ -1,6 +1,23 @@
+/*
```
JKillian: We probably ought to use the @license annotation and the JSDoc notation here and in other test files as well. Not a blocking thing for this PR, just a thought

test/tsconfig.json
```diff
@@ -8,20 +8,19 @@
...
-        "/*.ts",
+        "./*.ts",
         "./rules/*.ts",
         "./formatters/*.ts"
     ],
     "files": [
         "../typings/tsd.d.ts",
-        "typings/chai/chai.d.ts",
-        "typings/mocha/mocha.d.ts",
+        "./typings/tsd.d.ts",
```
JKillian: I kind of prefer not using tsd.d.ts and instead just having typings referenced here. It feels weird to have one file that contains `///` references while no other files do. But tsd.d.ts is also sort of a standard, so either way is fine

src/tsconfig.json
JKillian: Also, when the grunt build is triggered, it will (I believe) change the files array back to its own style, and we'll have nasty back and forth diffs of tsconfig.json style.

JKillian:
Looks good! Only major things to change is the "Allow Atom to rewrite tsconfig.json" part

test/tsconfig.json
adidahiya: yeah, that makes sense actually -- I was trying to standardize on one style of referencing typings. Seems like the other way is better and both src/ and test/ should remove tsd.d.ts

src/tsconfig.json
adidahiya: hm ok you're right, I'll revert this part. I forgot that our build expands filesGlob. This change would've alleviated a small inconvenience for Atom users (you can switch branches and not have to run `grunt` to keep coding), but it's something I'll have to live with

test/assert.ts
adidahiya: this interface augmentation is necessary to allow assigning `global.assert` on line 25. I'm not exactly sure why we need line 25 though; I can look into it.

test/lint.ts
adidahiya: sure I'll update it for this PR

PR #854, opened 2015-12-06
Prepare release 3.2.0-dev.1

adidahiya:


JKillian:
Looks fine - although feels a little funny to be getting rid of the `3.1.0-dev.1` release notes altogether, since that was an actual release. Seems like they should be kept around in some way at least

PR #855, opened 2015-12-06
Forbid var keyword in for, for-in, and for-of loops

csnover:
no-var-keyword currently only prevents the `var` keyword in a block. This PR ensures var keywords are also correctly forbidden in for loop initializers.


adidahiya:
:+1:

csnover:
Thanks!

PR #857, opened 2015-12-06
Directories support

louy:
Accepting directories as arguments in cli. For example `tslint .`

Related: #692 

Note: `node_modules` and `typings` are now ignore by default, even in subdirectories.

### DIFF
--- a/src/tslint-cli.ts
+++ b/src/tslint-cli.ts
@@ -178,8 +178,20 @@ const processFile = (file: string) => {
     }
 };
 
-const files = argv._;
+const GLOB_IGNORE = 'node_modules';
 
-for (const file of files) {
-    glob.sync(file).forEach(processFile);
-}
+const fileArgs: string[] = argv._;
+const files: string[] = fileArgs.reduce((files: string[], arg: string) => {
+    return files.concat(glob.sync(arg, {ignore: GLOB_IGNORE}));
+}, []);
+
+files.forEach(file => {
+    // Check whether argument is actually a directory
+    const stat = fs.lstatSync(file);
+    if (stat.isDirectory()) {
+        // Just read all ts files in the directory
+        glob.sync(file + "/**/*.ts", {ignore: GLOB_IGNORE}).forEach(processFile);
+    } else {
+        processFile(file);
+    }
+});
--- a/src/tslint-cli.ts
+++ b/src/tslint-cli.ts
@@ -16,6 +16,7 @@
  */
 
 import * as fs from "fs";
+import * as path from "path";
 import * as glob from "glob";
 import * as optimist from "optimist";
 import * as Linter from "./tslint";
@@ -178,20 +179,48 @@ const processFile = (file: string) => {
     }
 };
 
-const GLOB_IGNORE = 'node_modules';
+const GLOB_IGNORE: string[] = ["**/node_modules/**", "**/typings/**"];
 
 const fileArgs: string[] = argv._;
-const files: string[] = fileArgs.reduce((files: string[], arg: string) => {
-    return files.concat(glob.sync(arg, {ignore: GLOB_IGNORE}));
+const files: string[] = fileArgs.reduce((list: string[], arg: string) => {
+    return list.concat(glob.sync(arg, {ignore: GLOB_IGNORE}));
 }, []);
 
-files.forEach(file => {
+files.forEach(arg => {
+    const stat = fs.lstatSync(arg);
+
     // Check whether argument is actually a directory
-    const stat = fs.lstatSync(file);
     if (stat.isDirectory()) {
+        let globIgnore = GLOB_IGNORE;
+
+        // Try finding a tsconfig file
+        const configFile = path.join(arg, "/tsconfig.json");
+        let configStat: fs.Stats;
+        try {
+            configStat = fs.lstatSync(configFile);
+        } catch (e) {
+            // File doesn't exist.
+        }
+
+        if (configStat && configStat.isFile()) {
+            let config: any;
+            try {
+                config = JSON.parse(fs.readFileSync(configFile, "utf8"));
+            } catch (e) {
+                // Invalid file contents?
+                console.warn("tsconfig.json file (" + configFile + ") appears not to be valid");
+                console.error(e);
+            }
+            if (config) {
+                if (config.exclude) {
+                    globIgnore = globIgnore.concat(config.exclude);
+                }
+            }
+        }
+
         // Just read all ts files in the directory
-        glob.sync(file + "/**/*.ts", {ignore: GLOB_IGNORE}).forEach(processFile);
+        glob.sync(path.join(arg, "/**/*.ts"), {ignore: globIgnore}).forEach(processFile);
     } else {
-        processFile(file);
+        processFile(arg);
     }
 });


blakeembrey:
Just some comments since I'm interested in such a feature, is it possible to make `tsconfig.json` a "mode" and have exclude be a CLI flag which overrides the default excludes?

louy:
yeah sure. something like `--project, -p`?

blakeembrey:
That makes sense, so it'll be opt in? I more need something that looks like `tslint "src/**/*.ts" --exclude "__test__/fixtures/**"`. If you don't want this going way out of scope though, I can make another PR. It's just using `tsconfig.json` as the source of truth isn't really correct all the time, plus promoting using exclude might confuse people (the compiler will use files and ignore exclude according to the spec, but there's no such check here).

src/tslint-cli.ts
```diff
@@ -178,8 +179,48 @@ const processFile = (file: string) => {
...
 };
 
-const files = argv._;
+const GLOB_IGNORE: string[] = ["**/node_modules/**", "**/typings/**"];
 
-for (const file of files) {
-    glob.sync(file).forEach(processFile);
-}
+const fileArgs: string[] = argv._;
+const files: string[] = fileArgs.reduce((list: string[], arg: string) => {
```
adidahiya: no need for `: string[]` typedef

src/tslint-cli.ts
```diff
@@ -178,8 +179,48 @@ const processFile = (file: string) => {
...
+}, []);
+
+files.forEach(arg => {
+    const stat = fs.lstatSync(arg);
+
+    // Check whether argument is actually a directory
+    if (stat.isDirectory()) {
+        let globIgnore = GLOB_IGNORE;
+
+        // Try finding a tsconfig file
```
adidahiya: It feels a little strange to add half-baked support for `tsconfig.json` right now -- I'm worried it might be unintuitive. I think I'd rather have that implemented fully with #858 and a `--project` option. And as @blakeembrey mentioned, we should have a separate `--exclude` option for TSLint to override `tsconfig.json`.

I'd be happy to look at another PR that adds an `--exclude` option, but don't think it should be mixed in with this one.

src/tslint-cli.ts
```diff
@@ -178,8 +179,48 @@ const processFile = (file: string) => {
...
 
-for (const file of files) {
-    glob.sync(file).forEach(processFile);
-}
+const fileArgs: string[] = argv._;
+const files: string[] = fileArgs.reduce((list: string[], arg: string) => {
+    return list.concat(glob.sync(arg, {ignore: GLOB_IGNORE}));
+}, []);
+
+files.forEach(arg => {
```
adidahiya: please use the for-of syntax and rename `arg` to `file`, like it was before.

src/tslint-cli.ts
```diff
@@ -178,8 +179,48 @@ const processFile = (file: string) => {
     }
 };
 
-const files = argv._;
+const GLOB_IGNORE: string[] = ["**/node_modules/**", "**/typings/**"];
```
adidahiya: no need for `: string[]` typedef

adidahiya:
Filed #915 to track `--exclude`.

andrenarchy:
Great feature! This brings tslint closer to the behavior of other popular linters, e.g., eslint. Is there any progress on the revision?

adidahiya:
ping @louy -- will you get back to this PR or should someone else take up this feature?

louy:
@adidahiya didn't you say that we should wait for #858 before moving on with this one?

TBH I'm thinking of a different approach now. here's how I suggest the workflow to be:

1. if you pass `tslint` command a file, it lints it.
1. if you pass `tslint` a directory, it looks for a `tsconfig.json` file.
  1. if there's a file, it reads the `files` prop and only lints them
  1. if there's a file and it has an `exclud` option, it lints all files that match `**/*.ts` in that folder except those specified in `exclude`.
  1. if there's no `tsconfig` file, it lints all files that match `**/*.ts`.
1. when searching a directory for files, tslint should look for a `.tslintignore` file. if that file is found, it ignores everything that matches rules inside that file even if they're listed in `tsconfig.json#files`. that file should have the same syntax as a `.gitignore` or `.eslintignore` file.
1. when linting a file, tslint should look for all `.tslintrc` files in the same and parent directories until it reaches a file which has the `"root": true` option in it. it should merge all `.tslintrc` files found with the deepest file having the highest priority.

so in order to exclude `.d.ts` files in a project, all you have to do is create a `.tslintignore` file with `*.d.ts` inside it. you can also add `typings/` to ignore all files in typings folder.

this is how I imagine `tslint` should work ideally, and I think this is how `eslint` and `jshint` work.

blakeembrey:
@louy I don't think `tslint` should use `files` unless it intends to completely replicate `tsc` behaviour and start working in a project context. It's a lot of work and the things I use in `tsconfig.json` are never representative of what I want to lint - so it's implicit behaviour I'll never use and will likely confuse even though I've read this issue. Also, you need to resolve `.tsx`. On top of that, what's the behaviour with `allowJs` enabled?

unional:
> It's a lot of work and the things I use in tsconfig.json are never representative of what I want to lint

I hold a slightly different opinion. For `tsconfig.json` that is consumed by `tsc`, yes I agreed.

But for `tsconfig.json` that is consumed by `ts language service`, it is likely what I want to lint also.

Of course, able to decouple them is a good thing. So how about:
- Read from `tslint.json/fileGlobs` if exists
- If not, then read from `/tsconfig.json` (or add a field `tslint.json/tsconfig` to point to the file)

Just saying on top of my head. :smirk:

blakeembrey:
@unional `filesGlob` is non-standard and the first sentence I wrote is the most important here - if `tslint` doesn't create a project instance and resolve everything as `tsc` does then it's not really useful to work with `files`. The `files` field in `tsconfig.json` is meant as entry points for the compiler to start resolving and using it to mean something else sounds like a frustrating exercise to me.

Isn't TypeScript language services is the same as `tsc` here? I'm not 100% sure about the comment, both `tsc` and the language services resolve the dependency tree which is why it'd be important that `tslint` start working in a "project" basis, not a file basis.

unional:
> if tslint doesn't create a project instance and resolve everything as tsc does then it's not really useful to work with files. The files field in tsconfig.json is meant as entry points for the compiler to start resolving and using it to mean something else sounds like a frustrating exercise to me.

Agree~~  But as mentioned https://github.com/palantir/tslint/pull/857#issuecomment-164667421, I thought we are talking about project here. :smile:




> Isn't TypeScript language services is the same as tsc here?

Code-wise, yet. Application-wise, no.
`tsc` is about compilation. `TypeScript language service` is about type resolution and provide support to IDE.
That's why I have been saying they should consume different `tsconfig.json`.
`tsconfig.json` should use `files` to explicitly point out what to compile. They should be:
- source files
- typings files

`tsconfig.json` should use `exclude` (or `include`) to include ANY files the user may read and edit. These includes:
- source files
- test files
- plumbing files
- typings files
- etc.

adidahiya:
Closing this PR since it's very out of date. Let's continue discussion in #692 if necessary.

PR #868, opened 2015-12-10
Add typescript definition info to packages.json

berickson1:
This allows tools like tsd to auto-pull definition files from the node_modules folder


adidahiya:
thanks @berickson1!

Issue #717, opened 2015-10-05
Add --init command to CLI


adidahiya:
similar to ESLint, this would generate a starter config file for you

myitcv:
On a related not, would it also make sense to generate the documentation (e.g. `README` which lists all the core rules) and the `sample.tslint.json`?

Could have a build step that checks the checked in code has been re-generated (that way you keep the generated output as part of the repo)

Ensures that everything remains in sync.

JKillian:
@myitcv I like that idea, would makes things easier. https://github.com/palantir/tslint/issues/674 and #629 ed to be figured out first here though.

adidahiya:
@myitcv @JKillian that please file that as a separate issue. A very basic `--init` feature is simple to implement right now and not blocked on anything.

JKillian:
:+1:  Moved idea to part of #830

PR #871, opened 2015-12-11
Automatic tslint.json generation with --init command

JKillian:
Implements #717 

A simple implementation of a feature to generate a tslint.json file for users. This PR is very limited in scope - there are lots of ways for this to be improved in the future.

This feature intentionally isn't added to the README yet, that can be done once a better default configuration is settled on.

### DIFF
--- a/src/configuration.ts
+++ b/src/configuration.ts
@@ -19,8 +19,8 @@ import * as fs from "fs";
 import * as path from "path";
 import * as findup from "findup-sync";
 
-const CONFIG_FILENAME = "tslint.json";
-const DEFAULT_CONFIG = {
+export const CONFIG_FILENAME = "tslint.json";
+export const DEFAULT_CONFIG = {
     "rules": {
         "curly": true,
         "indent": [true, 4],
--- a/src/configuration.ts
+++ b/src/configuration.ts
@@ -22,17 +22,33 @@ import * as findup from "findup-sync";
 export const CONFIG_FILENAME = "tslint.json";
 export const DEFAULT_CONFIG = {
     "rules": {
-        "curly": true,
-        "indent": [true, 4],
-        "no-duplicate-key": true,
+        "class-name": true,
+        "comment-format": [true, "check-space"],
+        "indent": [true, "spaces"],
         "no-duplicate-variable": true,
-        "no-empty": true,
         "no-eval": true,
+        "no-internal-module": true,
         "no-trailing-whitespace": true,
-        "no-unreachable": true,
-        "no-use-before-declare": true,
+        "no-var-keyword": true,
+        "one-line": [true, "check-open-brace", "check-whitespace"],
         "quotemark": [true, "double"],
-        "semicolon": true
+        "semicolon": true,
+        "triple-equals": [true, "allow-null-check"],
+        "typedef-whitespace": [true, {
+            "call-signature": "nospace",
+            "index-signature": "nospace",
+            "parameter": "nospace",
+            "property-declaration": "nospace",
+            "variable-declaration": "nospace"
+        }],
+        "variable-name": [true, "ban-keywords"],
+        "whitespace": [true,
+            "check-branch",
+            "check-decl",
+            "check-operator",
+            "check-separator",
+            "check-type"
+        ],
     }
 };
 const moduleDirectory = path.dirname(module.filename);
--- a/src/tslint-cli.ts
+++ b/src/tslint-cli.ts
@@ -44,7 +44,7 @@ let processed = optimist
         },
         "i": {
             alias: "init",
-            describe: "generate a tslint.json config file in the cwd"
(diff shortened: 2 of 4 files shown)


adidahiya:
- [ ] needs test in `check-bin.sh`
- [ ] we should settle on a good default config and add it to the readme before merging this

JKillian:
Added test. As far as a default config, I think it should be pretty conservative in what is enabled

JKillian:
Added a commit with a proposed reasonable default config. I looked at our settings as well as the rules TypeScript used as a baseline, and ruling out rules that are more aggressive (or bug prone, like no-unused-variable).

src/configuration.ts
```diff
@@ -19,20 +19,35 @@ import * as fs from "fs";
...
         "no-eval": true,
+        "no-internal-module": true,
         "no-trailing-whitespace": true,
-        "no-unreachable": true,
-        "no-use-before-declare": true,
-        "quotemark": [true, "double"],
-        "semicolon": true
+        "no-var-keyword": true,
+        "one-line": [true, "check-open-brace", "check-whitespace"],
+        "quotemark": [true, "single"],
```
adidahiya: why single? we use double and so does the TS team. let's change it back.

src/configuration.ts
```diff
@@ -19,20 +19,35 @@ import * as fs from "fs";
...
+        "semicolon": true,
+        "typedef-whitespace": [true, {
+            "call-signature": "nospace",
+            "index-signature": "nospace",
+            "parameter": "nospace",
+            "property-declaration": "nospace",
+            "variable-declaration": "nospace"
+        }],
```
adidahiya: let's add `triple-equals`; it's pretty uncontroversial

adidahiya: can you add `typedef-whitespace` to our own tslint.json as well?

src/tslint-cli.ts
```diff
@@ -42,6 +42,10 @@ let processed = optimist
             alias: "help",
             describe: "display detailed help"
         },
+        "i": {
+            alias: "init",
+            describe: "generate a tslint.json config file in the cwd"
```
adidahiya: I would expand cwd to 'current working directory'

src/tslint-cli.ts
```diff
@@ -81,6 +85,17 @@ if (argv.v != null) {
     process.exit(0);
 }
 
+if (argv.i != null) {
+    if (fs.existsSync(CONFIG_FILENAME)) {
+        console.error(`Cannot generate ${CONFIG_FILENAME}: file already exists`);
+        process.exit(1);
+    }
+
+    const tslintJSON = JSON.stringify(DEFAULT_CONFIG, undefined, "   ");
```
adidahiya: this is 3 spaces; it should be 2 or 4

adidahiya: rename to `tslintJson`

test/check-bin.sh
```diff
@@ -43,8 +42,27 @@ expectOut $? 0 "tslint with valid arguments did not exit correctly"
...
+./tslint --init
+if [ ! -f tslint.json ]; then
+  echo "--init failed, tslint.json not created"
+  num_failures=$(expr $num_failures + 1)
+fi
+expectOut $? 0 "tslint with --init flag did not exit correctly"
+
+# should fail since tslint.json already exists
+./tslint --init
+expectOut $? 1 "tslint with --init flag did not exit correctly when tslint.json already exists"
```
adidahiya: :+1:

src/configuration.ts
JKillian: Single quotes are scientifically proven to be easier for humans on standard english keyboards to type!

src/configuration.ts
JKillian: :+1:  with the `allow-null-check` option I think

src/tslint-cli.ts
JKillian: (JS stdlib treats it as JSON, not Json, more consistent to keep it capitalized)

src/configuration.ts
adidahiya: meh, but it doesn't make sense to have the default config be different from our own code like this. we should discuss moving to single quotes elsewhere... we choose double to make it easier to switch back-and-forth between languages like Java to TS. many others, including the compiler team, use double quotes for the same reason.

src/configuration.ts
adidahiya: yep

src/tslint-cli.ts
adidahiya: > JS stdlib treats it as JSON

what does that mean?

The "JSON" substring feels a bit too direct of a conflict with the builtin `JSON` value.

JKillian:
[Changes made]

adidahiya:
:+1:

PR #899, opened 2016-01-04
Only import types from ./lint to prevent circular dependencies

JKillian:


JKillian:
`tsc` will remove imports from the final compiled code if they only import interfaces/types. By not using `Lint.scanAllTokens`, we're no longer using anything but types from `./lint` and thus the import will be removed.

src/enableDisableRules.ts
```diff
@@ -17,6 +17,7 @@
 
 import * as ts from "typescript";
 import * as Lint from "./lint";
```
adidahiya: let's remove this line too and do an explicit import of `IEnableDisablePosition`

JKillian: :+1:  Fixed and amended to the commit

Issue #50, opened 2013-12-10
comment-format rule triggers on ///<reference>'s


Bartvds:
The `comment-format` rule gets triggered by TypeScript's typical `///<reference path="foo.ts">` comments. 

Is this intentional? Maybe a special-case is in order to let these pass?

PR #904, opened 2016-01-08
fix semicolon rule

jerryorta-dev:
I don't think this is a legitimate fix. It just illustrates the issue:
semicolonRule - TypeError: Cannot read property 'length' of undefined #50


jerryorta-dev:
I am no longer getting the error. Upgraded tslint to 3.2.1

JKillian:
:+1: Glad it's working correctly!

PR #905, opened 2016-01-08
Fix bug where await expressions were marked as unused

### LINKED ISSUE #887: `no-unused-expression` should permit `await`
The following code causes a linter error:

	async function f(foo: Promise<void>): Promise<number> {
		await foo
		return 0
	}

The output I get is:

	src/foo.ts[2, 2]: expected an assignment or function call

Here's my `tslint.json`:

	{
		"rules": {
			"no-unused-expression": true
		}
	}

JKillian:
Fixes #887

### DIFF
--- a/src/rules/noUnusedExpressionRule.ts
+++ b/src/rules/noUnusedExpressionRule.ts
@@ -38,17 +38,17 @@ class NoUnusedExpressionWalker extends Lint.RuleWalker {
         this.expressionIsUnused = true;
         super.visitExpressionStatement(node);
         if (this.expressionIsUnused) {
-            // ignore valid unused expressions
-            if (node.expression.kind === ts.SyntaxKind.StringLiteral) {
-                const expressionText = node.expression.getText();
-                if (expressionText === "\"use strict\"" || expressionText === "'use strict'") {
-                    return;
-                }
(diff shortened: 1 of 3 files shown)


src/rules/noUnusedExpressionRule.ts
```diff
@@ -38,17 +38,17 @@ class NoUnusedExpressionWalker extends Lint.RuleWalker {
...
-                const expressionText = node.expression.getText();
-                if (expressionText === "\"use strict\"" || expressionText === "'use strict'") {
-                    return;
-                }
-            } else if (node.expression.kind === ts.SyntaxKind.DeleteExpression || node.expression.kind === ts.SyntaxKind.YieldExpression) {
-                return;
-            }
+            const { expression } = node;
+            const { kind } = expression;
+            const isValidStandAloneExpressionKind = kind === ts.SyntaxKind.DeleteExpression
```
adidahiya: nit: "standalone" is one word, not 2, and I think `Kind` is useless in this identifier

i.e. `isValidStandaloneExpression`

adidahiya:
feel free to merge after small rename :+1:

src/rules/noUnusedExpressionRule.ts
JKillian: Haha good call, meant to look that one up actually

PR #912, opened 2016-01-14
Improve handling of paths to custom rules directories

### LINKED ISSUE #910: Report descriptive error when custom rules directories fail to load
Currently, if custom rules directories are provided but no rules are found, TSLint does not report any errors or warnings.

JKillian:
New behavior:

* Throw an error if the rules directory supplied does not exist
* If rules directory is supplied in a tslint.json or package.json file, look for directory relative to .json file.
* If rules directory is supplied via CLI, look for directory relative to CWD.

Fixes #910

### DIFF
--- a/src/configuration.ts
+++ b/src/configuration.ts
@@ -51,39 +51,56 @@ export const DEFAULT_CONFIG = {
         ],
     }
 };
-const moduleDirectory = path.dirname(module.filename);
 
 export function findConfiguration(configFile: string, inputFileLocation: string): any {
-    if (configFile == null) {
-        // first look for package.json from input file location
-        configFile = findup("package.json", { cwd: inputFileLocation, nocase: true });
+    const configPath = findConfigurationPath(configFile, inputFileLocation);
+    return loadConfigurationFromPath(configPath);
+}
 
-        if (configFile) {
-            const content = require(configFile);
+export function findConfigurationPath(suppliedConfigFilePath: string, inputFilePath: string) {
+    if (suppliedConfigFilePath != null) {
+        if (!fs.existsSync(suppliedConfigFilePath)) {
+            throw new Error(`Could not find config file at: ${path.resolve(suppliedConfigFilePath)}`);
+        } else {
+            return suppliedConfigFilePath;
+        }
+    } else {
+        // search for package.json with tslintConfig property
+        let configFilePath = findup("package.json", { cwd: inputFilePath, nocase: true });
+        if (configFilePath != null && require(configFilePath).tslintConfig != null) {
+            return configFilePath;
+        }
 
-            if (content.tslintConfig) {
-                return content.tslintConfig;
+        // search for tslint.json from input file location
+        configFilePath = findup(CONFIG_FILENAME, { cwd: inputFilePath, nocase: true });
+        if (configFilePath != null && fs.existsSync(configFilePath)) {
+            return configFilePath;
+        }
+
+        // search for tslint.json in home directory
+        const homeDir = getHomeDir();
+        if (homeDir != null) {
+            configFilePath = path.join(homeDir, CONFIG_FILENAME);
+            if (fs.existsSync(configFilePath)) {
+                return configFilePath;
             }
         }
 
-        // next look for tslint.json
-        const homeDir = getHomeDir();
-        if (!homeDir) {
-            return undefined;
-        }
-
-        const defaultPath = path.join(homeDir, CONFIG_FILENAME);
-
-        configFile = findup(CONFIG_FILENAME, { cwd: inputFileLocation, nocase: true }) || defaultPath;
+        // no path could be found
+        return undefined;
     }
+}
 
-    if (fs.existsSync(configFile)) {
-        let fileData = fs.readFileSync(configFile, "utf8");
+export function loadConfigurationFromPath(configFilePath: string) {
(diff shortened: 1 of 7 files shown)


test/tslint-custom-rules-with-dir.json
```diff
@@ -0,0 +1,6 @@
+{
```
adidahiya: can we move this into a new dir `test/config` please? I'd like to not clutter top-level directories.

test/tslint-custom-rules.json
```diff
@@ -0,0 +1,5 @@
+{
```
adidahiya: same with this file

test/ruleLoaderTests.ts
```diff
@@ -17,8 +17,7 @@
 import {loadRules} from "./lint";
 
 describe("Rule Loader", () => {
-    const path = require("path");
-    const rulesDirectory = path.join(global.process.cwd(), "build/rules");
+    const rulesDirectory = "build/src/rules";
```
adidahiya: why is the old code not necessary?
this should probably be `RULES_DIRECTORY` now

src/configuration.ts
```diff
@@ -103,20 +120,27 @@ function getHomeDir() {
...
+export function getRulesDirectories(directories: string | string[], relativeTo?: string): string[] {
     let rulesDirectories: string[] = [];
 
     if (directories != null) {
         if (typeof directories === "string") {
-            rulesDirectories = [getRelativePath(<string>directories)];
+            rulesDirectories = [getRelativePath(<string>directories, relativeTo)];
         } else {
-            rulesDirectories = (<string[]>directories).map((dir) => getRelativePath(dir));
+            rulesDirectories = (<string[]>directories).map((dir) => getRelativePath(dir, relativeTo));
```
adidahiya: is the type not narrowed automatically here by the compiler?

src/configuration.ts
```diff
@@ -103,20 +120,27 @@ function getHomeDir() {
...
+        }
+    }
+
+    for (const directory of rulesDirectories) {
+        if (!fs.existsSync(directory)) {
+            throw new Error(`Could not find custom rule directory: ${directory}`);
```
adidahiya: should this be `path.resolve(directory)` in the error message?

src/ruleLoader.ts
```diff
@@ -63,6 +63,7 @@ export function findRule(name: string, rulesDirectories?: string | string[]) {
 
     for (let rulesDirectory of directories) {
         // then check for rules within the first level of rulesDirectory
+        console.log(rulesDirectory);
```
adidahiya: remove this line

src/tslint-cli.ts
```diff
@@ -18,8 +18,10 @@
 import * as fs from "fs";
 import * as glob from "glob";
 import * as optimist from "optimist";
+import * as path from "path";
 import * as Linter from "./tslint";
-import {CONFIG_FILENAME, DEFAULT_CONFIG, getRulesDirectories} from "./configuration";
+import {CONFIG_FILENAME, DEFAULT_CONFIG, getRulesDirectories,
+        findConfigurationPath, loadConfigurationFromPath} from "./configuration";
```
adidahiya: an alternative style to consider that's a little more maintainable:

```ts
import {
    CONFIG_FILENAME,
    DEFAULT_CONFIG,
    ...
} from "./configuration";
```

src/tslint-cli.ts
```diff
@@ -167,14 +169,13 @@ const processFile = (file: string) => {
...
     const contents = fs.readFileSync(file, "utf8");
-    const configuration = Linter.findConfiguration(argv.c, file);
+    const configurationPath = findConfigurationPath(argv.c, file);
+    const configuration = loadConfigurationFromPath(configurationPath);
 
-    if (configuration == null) {
-        console.error("Unable to find tslint configuration");
-        process.exit(1);
-    }
+    // if configurationPath is null, this will be equal to ".", which is the current directory and is fine
```
adidahiya: slightly better wording would be "... this will be set to..."

test/ruleLoaderTests.ts
JKillian: The old code actually didn't even work, but we never noticed since TSLint just ignored bad paths to rule directories. A relative path is based off of the cwd by default, so no need to do that path.join

src/configuration.ts
JKillian: `getRelativePath` is already resolving to absolute paths, so this will show users an absolute path already

test/check-bin.sh
```diff
@@ -19,7 +19,10 @@ expectOut () {
   expect=$2
   msg=$3
 
-  if [ $expect != $actual ]; then
+  nodeV=`node -v`
+  
+  # if Node 0.10.*, node will sometimes exit with status 8 when an error is thrown
+  if [[ $expect != $actual || $nodeV == v0.10.* && $expect == 1 && $actual == 8 ]] ; then
```
adidahiya: is this how precedence works here? `[[ $expect != $actual || [[$nodeV == v0.10.* && $expect == 1 && $actual == 8]] ]]` can we make that explicit?

JKillian: Can you nest brackets like that? I don't know bash, but I did look up that & is higher precedence than ||

PR #917, opened 2016-01-15
Prepare release v3.2.2-dev.1

JKillian:


JKillian:
CI shows failures because of a bug in the most recent TS nightly, tests would be passing otherwise

JKillian:
Retriggered the CI since a new TS nightly is out, builds should be passing now and this should be good to merge

adidahiya:
:+1:

PR #944, opened 2016-01-26
'learn more' button goes to 404

mmv:
Update href to be coherent with the "the full usage guide" link.


JKillian:
Good catch, thanks @mmv!

Issue #856, opened 2015-12-06
Multi-line literals with a single entry are treated as single-line literals for trailing commas


prencher:
Given the following:

```
{
   foo
}
```

TSLint will treat this as a single line literal rather than a multi-line literal. Because of this, you can't find all these cases without also turning on single-line literals, which means also hitting e.g. import statements such as `import { Foo } from './bar'`

adidahiya:
Looks like a bug. ESLint's comma-dangle rule flags that example as a failure, so we should do the same.

JKillian:
Fixed in #952 by @pe8ter

adidahiya:
Fix is now available in `v3.4.0-dev.1`

PR #952, opened 2016-02-01
Identify multiline literals with single property correctly.

pe8ter:
This PR addresses https://github.com/palantir/tslint/issues/856. It's only a one line fix, but quite a few test cases were added to cover the various flavors of multiline literals.


JKillian:
Looks good to me @pe8ter! Thanks for going through and adding those test cases as well.

pe8ter:
Glad to help!

Issue #598, opened 2015-08-23
Log a message if a rule cannot be found


HamletDRC:
Assuming that you have written all your rule config in a file called "tslint.json" then I would like to see a console error if one of the referenced rules cannot be found. 

For example, if I define  "no-unused-whatever" and that rule cannot be found, then I want to see something on the console. 

It looks like you just need to add a console.log message in the `ruleLoader.ts::findRule()` function. Right above `return undefined`

JKillian:
This will be handled by #953. Right now it's going to throw an error - do you think just a log message would be better (see https://github.com/palantir/tslint/pull/953#discussion_r51509416)

adidahiya:
Fix is now available in `v3.4.0-dev.1`

PR #953, opened 2016-02-01
Throw error if rule specified in config cannot be found.

### LINKED ISSUE #911: Report error when a rule is configured in tslint.json but no corresponding rule implementation is found

JKillian:
Also, fix a bug in our tslint.json and a bug in our tests

Fixes #911 and #598


JKillian:
As a side note, we should also throw an error for rules that are configured incorrectly.

src/ruleLoader.ts
```diff
@@ -43,11 +45,18 @@ export function loadRules(ruleConfiguration: {[name: string]: any},
                 const ruleSpecificList = (ruleName in enableDisableRuleMap ? enableDisableRuleMap[ruleName] : []);
                 const disabledIntervals = buildDisabledIntervalsFromSwitches(ruleSpecificList, allList);
                 rules.push(new Rule(ruleName, ruleValue, disabledIntervals));
+            } else {
```
adidahiya: let's change this conditional block to read better (prefer the positive form of conditional expressions first):

```ts
if (Rule == null) {
  notFoundRules.push(ruleName);
} else {
  ...
}
```

src/ruleLoader.ts
```diff
@@ -43,11 +45,18 @@ export function loadRules(ruleConfiguration: {[name: string]: any},
...
+                notFoundRules.push(ruleName);
             }
         }
     }
 
-    return rules;
+    if (notFoundRules.length > 0) {
+        const errorMessage = `Could not find the following rules specified in the configuration:\n${notFoundRules.join("\n")}`;
+        throw new Error(errorMessage);
```
adidahiya: this is a breaking change for users who currently have invalid configurations... maybe we should just log the error?

JKillian: I like throwing an error because there's no reason you should have a misnamed rule in your config. Also, depending on the context that TSLint is being run in, users may not see the console / log messages. However, you're right, I guess this could cause some troubles for people (although easily fixable ones)

adidahiya:
ok looks good :+1: this is a small breaking change i'm ok with

mr270:
It would be nice for us if this could be added to the outputs so those of us processing the outputs could handle it in our post processing tasks (json in our case). This could be an optional argument.

Issue #955, opened 2016-02-03
typedef-whitespace call-signature false positive with multi-line args list and indentation


kourge:
With typedef-whitespace call-signature set to `nospace`, this is recognized as a warning:
```typescript
function f() : any {}
```
And this passes:
```typescript
function f(): any {}
```
When arguments are spread across multiple lines, indentation produces a false positive, e.g. this works:
```typescript
function f(
  a: any,
  b: any
): any {}
```
Whereas this is identified as violating `nospace`:
```typescript
  function f(
    a: any,
    b: any
  ): any {}
```
Notice that the only difference is in indentation.

adidahiya:
Looks like a bug, thanks for the report @kourge

pe8ter:
I found a related false negative. This passes linting but should actually fail under the `nospace` rule:

```
foo( ) : void {}
```

No multiple lines or tabs involved. What both these failure cases have in common is that the last character in the parameter list is not followed immediately by the closing parenthesis. The AST returned by the TypeScript module doesn't provide enough explicit information to locate the closing parenthesis with a quick one-line fix, but a helper function could fairly quickly be created to find it. I'll keep poking at it.

pe8ter:
To be clearer, the linter incorrectly assumes that a closing parenthesis follows the last character of the parameter list (see lines that contain `node.parameters.end + 1` in _typedefWhitespaceRule.ts_). The position of the closing parenthesis could be far beyond the last character in the parameter list, including after spaces and comments, but that information isn't available in the AST that TypeScript outputs. The only way to truly find the parenthesis is to use the end position of the parameter list `node.parameters.end` and the start of the return type `node.type.pos` combined with a scanner to extract the character run that contains a parenthesis somewhere in it and a RegEx to test for the existence of an offending space.

JKillian:
@pe8ter I'm thinking the AST might be able to help us here infact. I stuck the code below into Atom and used the Ast Full atom-typescript command. The key difference between ast/ast full in atom-typescript is that the first uses `ts.forEachChild` which only iterates over important child nodes, whereas the latter iterates over `node.getChildren` which contains all children. [As explained here](https://basarat.gitbooks.io/typescript/content/docs/compiler/ast-tip-children.html). My hypothesis is that we can just find the colon token node (highlighted in red below) and work off of it.
```
function f(
  a: any,
  b: any
): any {};
```


[...]

Issue #888, opened 2015-12-20
Feature request: typedef-whitespace should have options for both sides of colons


andy-hanson:
`"typedef-whitespace": [true, {"call-signature": "space"}]` permits a missing space on the right side:

	function f() :number {
		return 0
	}

Wheras "nospace" permits a space on the right side:

	function f(): number {
		return 0
	}

I think this rule should have options for ensuring space on both sides of a colon.
People selecting `space` likely want everything to look like `foo : bar` and not `foo :bar`.
People selecting `nospace` may vary; personally I want to mandate `foo: bar` and forbid `foo:bar` (this is what http://www.typescriptlang.org/Handbook uses), but many likely want the opposite.

JKillian:
Agreed, it's somewhat limited / counterintuitive as it is now.

b0r3as:
The first thing, that is still unclear to me, is, if the `"space"` option means exactly one space or at least one space. The latter is currently implemented. Should all three cases be configurable (nospace, space, spaces) or is either exactly one or at least one space preferable?

Secondly, how should we handle the configuration of whitespace after the colon? I propose something like this:

```json
"typedef-whitespace": [
  true,
  {
    "call-signature": {
      "left": "nospace/space/spaces",
      "right": "nospace/space/spaces"
    },
    "index-signature": "nospace/space/spaces"
  }
]
```

The second form, demonstrated with `"index-signature"`, could either be used as a shorthand, if both left and right are equal or could retain the current functionality for compatibility reasons.

Any thoughts on this? I would be willing to make a PR if we can agree on a specification.

PR #958, opened 2016-02-05
Identify spaces before function return type colons correctly.

pe8ter:
This PR addresses https://github.com/palantir/tslint/issues/955.

There are some bugs related to looking for spaces before and after the colon on a function's return type caused by assuming that the colon comes exactly one character after the end of the parameter list. Instead of assuming, search the AST for the actual colon token.

The only thing I don't like about this solution is that there might not be a colon after the parameter list in which case I return `0` as a sentinel. I could make sure there is a return type before calling my new helper function, but that logic is already happening inside the `checkSpace()` function which is overgeneralized to handle all "check space" logic even for non-function-like cases.

I also noticed that some function types like setter methods and regular instance methods didn't have any tests so I filled in those gaps.

### DIFF
--- a/src/rules/typedefWhitespaceRule.ts
+++ b/src/rules/typedefWhitespaceRule.ts
@@ -26,20 +26,17 @@ export class Rule extends Lint.Rules.AbstractRule {
 
 class TypedefWhitespaceWalker extends Lint.RuleWalker {
     public visitFunctionDeclaration(node: ts.FunctionDeclaration) {
-        const positionBeforeColon = this.positionBeforeColon(node);
-        this.checkSpace("call-signature", node, node.type, positionBeforeColon);
+        this.checkSpace("call-signature", node, node.type, positionBeforeColon(node));
         super.visitFunctionDeclaration(node);
     }
 
     public visitFunctionExpression(node: ts.FunctionExpression) {
-        const positionBeforeColon = this.positionBeforeColon(node);
-        this.checkSpace("call-signature", node, node.type, positionBeforeColon);
+        this.checkSpace("call-signature", node, node.type, positionBeforeColon(node));
         super.visitFunctionExpression(node);
     }
 
     public visitGetAccessor(node: ts.AccessorDeclaration) {
-        const positionBeforeColon = this.positionBeforeColon(node);
-        this.checkSpace("call-signature", node, node.type, positionBeforeColon);
+        this.checkSpace("call-signature", node, node.type, positionBeforeColon(node));
         super.visitGetAccessor(node);
     }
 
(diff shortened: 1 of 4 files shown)


pe8ter:
In cases where space is desired before the colon is the goal of the typedef whitespace rule to enforce that _exactly_ one space exists or that _at least_ one space exists? If it means that _exactly_ one space should exist then neither the current implementation nor this PR are correct yet.

JKillian:
Thanks for the PR @pe8ter! I'll give this a look-over in the next day or two. As far as your question, I'm not sure of the original intent. However, if the current implementation allows more than one space, it's fine by me if this PR keeps things the same in that regard. I also think the rule is a little too limited right now... I think #888 will be a nice future improvement! (But of course that definitely shouldn't be in this PR)

src/rules/typedefWhitespaceRule.ts
```diff
@@ -119,4 +125,15 @@ class TypedefWhitespaceWalker extends Lint.RuleWalker {
         const options = allOptions[0];
         return options[option];
     }
+
+    private positionBeforeColon(node: ts.FunctionDeclaration | ts.FunctionExpression | ts.AccessorDeclaration |
+                                        ts.MethodDeclaration | ts.SignatureDeclaration | ts.AccessorDeclaration) {
+        const children = node.getChildren();
```
JKillian: I think you might be able to simplify this a little bit, something like:

```
const colon = node.getChildren().filter((c) => c.kind === ts.SyntaxKind.ColonToken)[0];
return colon == null ? 0 : colon.pos;

File: src/rules/typedefWhitespaceRule.ts [hunk 2]
```diff
@@ -119,4 +125,15 @@ class TypedefWhitespaceWalker extends Lint.RuleWalker {
         const options = allOptions[0];
         return options[option];
     }
+
+    private positionBeforeColon(node: ts.FunctionDeclaration | ts.FunctionExpression | ts.AccessorDeclaration |
```
JKillian: Let's just simplify this to `node: ts.Node` since we're only using the `getChildren()` method which should be available on all nodes

JKillian: Since this `positionBeforeColon` method never accesses `this.` anything, you can declare it outside of the class. Then above, you can simplify things a little:

```
this.checkSpace("call-signature", node, node.type, positionBeforeColon(node))
```

JKillian:
Looks good functionality-wise @pe8ter! I left a few comments with some ideas to simplify/tweak a few things just to clean up the code / make it fit style-wise with the rest of the codebase better

adidahiya:
:+1:

PR #959, opened 2016-02-06
Labels can be used with for-of statements.

ChaseMoskal:
Added for-of statement to the label position rule whitelist.


JKillian:
Thanks for the PR @ChaseMoskal! Could you add a test to `test/rules/label-position/test.ts.lint`? In this case, you just need to add a labeled for-of statement to the file (this would cause tests to fail without your change, but with your change tests should pass).

ChaseMoskal:
I think...

...I think I did it right.

Right?

adidahiya:
lgtm :+1:

PR #962, opened 2016-02-09
Fix trailing-comma lint failures

adidahiya:


JKillian:
Let's merge these after we release v3.4.0 (as they shouldn't cause the build to fail then).

PR #969, opened 2016-02-12
Treat member variables initialized to a function as a function for ordering purposes

### LINKED ISSUE #226: Member ordering too strict
After using the member-ordering rule like so
```
"member-ordering": [true,
"public-before-private",
"static-before-instance",
"variables-before-functions"
],
```
It seems to inherently enact the rule that this applies to functions as well as to member variables.  Can we have it so that we can only enforce this on the member variables excluding the functions?

JKillian:
Fixes #226

### DIFF
--- a/README.md
+++ b/README.md
@@ -171,7 +171,8 @@ A sample configuration file with all options is available [here](https://github.
 * `member-ordering` enforces member ordering. Rule options:
     * `public-before-private` All public members must be declared before private members.
     * `static-before-instance` All static members must be declared before instance members.
-    * `variables-before-functions` All variables needs to be declared before functions.
+    * `variables-before-functions` All member variables need to be declared before member functions.
+       Member variables initialized to a function literal are treated as member functions.
 * `no-any` diallows usages of `any` as a type decoration.
 * `no-arg` disallows access to `arguments.callee`.
 * `no-bitwise` disallows bitwise operators.
--- a/src/rules/memberOrderingRule.ts
+++ b/src/rules/memberOrderingRule.ts
@@ -78,12 +78,16 @@ export class MemberOrderingWalker extends Lint.RuleWalker {
     }
 
     public visitPropertyDeclaration(node: ts.PropertyDeclaration) {
-        this.checkModifiersAndSetPrevious(node, getModifiers(false, node.modifiers));
+        const { initializer } = node;
+        const isFunction = initializer != null
+            && (initializer.kind === ts.SyntaxKind.ArrowFunction || initializer.kind === ts.SyntaxKind.FunctionExpression);
+        this.checkModifiersAndSetPrevious(node, getModifiers(isFunction, node.modifiers));
         super.visitPropertyDeclaration(node);
     }
 
-    public visitPropertySignature(node: ts.Node) {
-        this.checkModifiersAndSetPrevious(node, getModifiers(false, node.modifiers));
+    public visitPropertySignature(node: ts.PropertyDeclaration) {
+        const isFunction = node.type != null && node.type.kind === ts.SyntaxKind.FunctionType;
+        this.checkModifiersAndSetPrevious(node, getModifiers(isFunction, node.modifiers));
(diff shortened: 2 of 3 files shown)


src/rules/memberOrderingRule.ts
```diff
@@ -78,12 +78,16 @@ export class MemberOrderingWalker extends Lint.RuleWalker {
...
+        const { initializer } = node;
+        const isFunction = initializer != null
+            && (initializer.kind === ts.SyntaxKind.ArrowFunction || initializer.kind === ts.SyntaxKind.FunctionExpression);
+        this.checkModifiersAndSetPrevious(node, getModifiers(isFunction, node.modifiers));
         super.visitPropertyDeclaration(node);
     }
 
-    public visitPropertySignature(node: ts.Node) {
-        this.checkModifiersAndSetPrevious(node, getModifiers(false, node.modifiers));
+    public visitPropertySignature(node: ts.PropertyDeclaration) {
```
JKillian: It's debatable if `ts.PropertyDeclaration` is the right type here. However, this is what we use in other places for this method signature so I went with it here.

adidahiya:
Exciting! What a long standing issue. Will take a look today.

src/rules/memberOrderingRule.ts
adidahiya: Looks like it is discussed here: https://github.com/Microsoft/TypeScript/issues/3833

`PropertySignature` nodes will be visited when linting interfaces. But looking at that code snippet from the parser, `ts.PropertyDeclaration` is probably safe enough to use as the typedef here. I guess we'll rely on testing :)

README.md
```diff
@@ -171,7 +171,8 @@ A sample configuration file with all options is available [here](https://github.
 * `member-ordering` enforces member ordering. Rule options:
     * `public-before-private` All public members must be declared before private members.
     * `static-before-instance` All static members must be declared before instance members.
-    * `variables-before-functions` All variables needs to be declared before functions.
+    * `variables-before-functions` All member variables need to be declared before member functions.
+       Member variables initialized to a function literal are treated as member functions.
```
adidahiya: In retrospect, this doesn't seem like the right option name... shouldn't it be `properties-before-methods`? If you agree I'll file an issue as a breaking change for v4.x.

test/rules/member-ordering/method/test.ts.lint
```diff
@@ -19,4 +19,18 @@ class Foo {
...
+    b(): void;
+    c: () => void;
+}
+
+class ArrowMethods {
+    public a = 5;
+    public b = "foo";
+    public c() {}
+    public d = () => {};
+    public e = function() {};
```
adidahiya: I think we should add more tests. Check that it enforces:

- alph for properties declared as functions
- member variables cannot be declared after member functions that use this syntax

src/rules/memberOrderingRule.ts
JKillian: Oh, nice find. We really only use this to access common fields like `node.type` so shouldn't be a problem

test/rules/member-ordering/method/test.ts.lint
JKillian: I don't think this rule checks anything about alphabetization. Second point seems like a good idea though

README.md
JKillian: Eh, it's not perfect because functions assigned with `=` are properties really but we treat them as methods. (I don't think you'd call functions on the prototype "properties" but not sure.) Let me think about this, but I'm inclined just to leave it as it is because I don't see a big clarity benefit worth making a breaking change

test/rules/member-ordering/method/test.ts.lint
adidahiya: oops, you're right. filed #973 for that FR

JKillian:
Updated (via an amended commit)

adidahiya:
:+1:

PR #971, opened 2016-02-13
Extend options of typedef-whitespace rule to control whitespace on both sides of the typedef colon

b0r3as:
This feature was requested here: #888. 

The rule configuration now could look like this (slashes meaning one of the three, of course): 

```json
"typedef-whitespace": [
  true,
  {
    "call-signature": "nospace/space/spaces",
    "index-signature": {
      "left": "nospace/space/spaces",
      "right": "nospace/space/spaces"
    },
    "parameter": {
      "left": "nospace/space/spaces",
    },
    "property-declaration": {
      "right": "nospace/space/spaces"
    },
    "variable-declaration": {
      "left": "nospace",
      "right": "space"
    }
  }
]
```

The three possible values for an option are `"nospace"`, `"space"` and `"spaces"` meaning no space, exactly one and at least one space, respectively. If one of these values is placed directly on an option, it applies to both sides of the colon. This is a shorthand for setting `"left"` and `"right"` to the same value. 

You can omit either `"left"` or `"right"`, which will skip the check for that side of the colon altogether.

*This PR breaks compatibility*. If you had the options for this rule configured to `"nospace/space"` before, you would have to use

```json
{
  "left": "nospace/spaces"
}
```

on each option now to replicate the exact same behaviour.


JKillian:
Thanks for the PR @b0r3as! I'll give it a review in the next day or two

src/rules/typedefWhitespaceRule.ts
```diff
@@ -112,16 +105,138 @@ class TypedefWhitespaceWalker extends Lint.RuleWalker {
...
+        isFailure = isFailure || !hasWS &&
+            (optionValue === "space" || optionValue === "spaces");
+
+        if (isFailure) {
+            this.addFailure(this.createFailure(failurePos, 1, message));
+        }
+    }
+
+    private getColonPosition(node: ts.Node): number {
+        const colon: ts.Node = node.getChildren().filter((child: ts.Node) =>
```
JKillian: In general, we leave out type annotations where they can be figured out automatically by tsc, so here, it can just be `child` instead of `child: ts.Node` and `colon` instead of `colon: ts.Node`. (This may apply a few other places in your changes as well)

src/rules/typedefWhitespaceRule.ts
```diff
@@ -112,16 +105,138 @@ class TypedefWhitespaceWalker extends Lint.RuleWalker {
...
+        // has at no space but should have at least one
+        isFailure = isFailure || !hasWS &&
+            (optionValue === "space" || optionValue === "spaces");
+
+        if (isFailure) {
+            this.addFailure(this.createFailure(failurePos, 1, message));
+        }
+    }
+
+    private getColonPosition(node: ts.Node): number {
```
JKillian: Another general rule we have is that if a method doesn't make any references to `this`, it can be moved outside of the class as its own function.

JKillian: return type can be figured out by the compiler here, no need for `: number`.

src/rules/typedefWhitespaceRule.ts
```diff
@@ -112,16 +105,138 @@ class TypedefWhitespaceWalker extends Lint.RuleWalker {
...
+        const optionConfig = this.getOption(option);
+        if (!this.hasRightOption(option)) {
+            return null;
+        }
+
+        return typeof optionConfig === "string" ? optionConfig : optionConfig.right;
+    }
+
+    private checkLeft(option: string, node: ts.Node, scanner: ts.Scanner, colonPosition: number): void {
+        let checkLeadingWhitespace: boolean = this.hasLeftOption(option);
```
JKillian: Feel free to just inline this at line 149 since it's only used once there

src/rules/typedefWhitespaceRule.ts
```diff
@@ -112,16 +105,138 @@ class TypedefWhitespaceWalker extends Lint.RuleWalker {
...
+        }
+    }
+
+    private performFailureCheck(optionValue: string, hasWS: boolean, hasSeveralWS: boolean, failurePos: number, message: string): void {
+        // has several spaces but should have one or none
+        let isFailure: boolean = hasSeveralWS &&
+            (optionValue === "space" || optionValue === "nospace");
+        // has at least one space but should have none
+        isFailure = isFailure || hasWS && optionValue === "nospace";
+        // has at no space but should have at least one
```
JKillian: typo here: "has at no space"

test/rules/typedef-whitespace/both/nospace/test.ts.lint
```diff
@@ -0,0 +1,815 @@
...
+             ~          [expected nospace after colon in parameter]
+        b   :   number
+           ~            [expected nospace before colon in parameter]
+             ~          [expected nospace after colon in parameter]
+    )   :   void {}
+       ~            [expected nospace before colon in call-signature]
+         ~          [expected nospace after colon in call-signature]
+}
+
+// endregion
```
JKillian: Oh man, I admire the thoroughness of this, but maybe it could be shortened a little bit for readability purposes? (Perhaps < 200 lines?)

src/rules/typedefWhitespaceRule.ts
```diff
@@ -112,16 +105,138 @@ class TypedefWhitespaceWalker extends Lint.RuleWalker {
...
+                hasSeveralTrailingWhitespaces,
+                colonPosition + 1,
+                message
+            );
+        }
+    }
+
+    private performFailureCheck(optionValue: string, hasWS: boolean, hasSeveralWS: boolean, failurePos: number, message: string): void {
+        // has several spaces but should have one or none
+        let isFailure: boolean = hasSeveralWS &&
```
JKillian: type `: boolean` not needed here (this applies elsewhere also)

JKillian:
@b0r3as: Talked with @adidahiya who also manages this project and we both think it would be better to go with the option schema [as described here](https://github.com/palantir/tslint/issues/888#issuecomment-166159680). It's not that your way of configuring things is bad in and of itself, but we'd rather not have a breaking change for users, and the linked suggestion lets us maintain current behavior.

Also noted a couple stylistic issues you can tweak to make the code more consistent with existing TSLint code. Let me know if you have any questions!

b0r3as:
The code style issues should be easy enough to fix. Switching over to the other config scheme should also be doable.
Is it still OK, to add the `spaces` value to discern between exactly one and at least one space? This would still slightly change the rule's behaviour for existing configurations (`space` now meaning exactly one instead of at least one, like it was before). But IMO, the precision gained with this additional value outweighs the small break introduced.

src/rules/typedefWhitespaceRule.ts
b0r3as: I agree, that it does not need to be a class member, as it does not reference `this`. But moving it out of the class somehow feels wrong to me from an OO point of view. Wouldn't `private static` be the way to go here?

b0r3as:
Config now looks like this: 

```json
"typedef-whitespace": [
      true,
      {
        "call-signature": "nospace/onespace/space",
        "index-signature": "nospace/onespace/space",
        "parameter": "nospace/onespace/space",
        "property-declaration": "nospace/onespace/space",
        "variable-declaration": "nospace/onespace/space"
      },
      {
        "call-signature": "nospace/onespace/space",
        "index-signature": "nospace/onespace/space",
        "parameter": "nospace/onespace/space",
        "property-declaration": "nospace/onespace/space",
        "variable-declaration": "nospace/onespace/space"
      }
    ]
```
First object describes left hand side, second object describes right hand side of the colon. This is 100% compatible to previous configurations.

test/rules/typedef-whitespace/both/nospace/test.ts.lint
b0r3as: I wanted to test all permutations of the three possible values for all five options. I'm actually not quite sure, which cases to omit to shorten the tests.

adidahiya:
@b0r3as cool, I like the new config. I'm ready to merged this, can you just squash this into one commit and rebase on the latest master? Thanks!

test/rules/typedef-whitespace/both/nospace/test.ts.lint
JKillian: Cool, this is fine then. I have some ideas for possible tweaks to the test system for the future which might change things anyways, so I'm good with leaving this as is :+1:

src/rules/typedefWhitespaceRule.ts
JKillian: Honestly, not a big deal if it's a static member function, member function, or plain old function. In this sort of JS setup though, the module is the high-level encapsulating context so having plain old functions living in it is fine. Other languages prefer that everything is inside a class, but I feel like in JS there's no good reason to go down that road. But no big deal either way

b0r3as:
Rebase done, let me know, if I need to do anything else.

PR #974, opened 2016-02-17
Extend semicolon rule with a never option

### LINKED ISSUE #363: Add ASI (automatic semicolon insertion) support to semicolon rule
As explained in https://github.com/Microsoft/TypeScript/issues/2575

Some coders would like to forget about trailing semicolons, but in some cases it can be a source of errors that are undetected by both the TypeScript compiler and `tslint`.

Example 1:
```ts
function dangerous(): string {
    return
        "watch the semi-colons!"
}
dangerous()
```
Compiles into:
```js
function dangerous() {
    return;
    "watch the semi-colons!";
}
dangerous();
```

Example 2:
```ts
var foo = true
// some comment
['a', 'b'].forEach((l: string) => {/**/})
```
compiles into
```js
var foo = true['a', 'b'].forEach(function (l) {/**/})
```

That would be awesome to see some warnings in these cases, e.g. by turning on a new rule :heart_eyes:

Pajn:
This only fails when you add a semicolon that is unnecessary, it does not
warn for cases where the code might do thr wrong thing by leaving them out.

I don't know if that should be added but if so I guess a warning for return
statements should be added as well as that's more or less the same case.

fixes #363


src/rules/semicolonRule.ts
```diff
@@ -92,11 +96,28 @@ class SemicolonWalker extends Lint.RuleWalker {
...
+        const always = this.hasOption(OPTION_ALWAYS) || (this.getOptions() && this.getOptions().length === 0);
 
-        if (!hasSemicolon) {
+        if (always && !hasSemicolon) {
             const sourceFile = this.getSourceFile();
             const position = node.getStart(sourceFile) + node.getWidth(sourceFile);
-            this.addFailure(this.createFailure(Math.min(position, this.getLimit()), 0, Rule.FAILURE_STRING));
+            this.addFailure(this.createFailure(Math.min(position, this.getLimit()), 0, Rule.FAILURE_STRING_MISSING));
+        } else if (this.hasOption(OPTION_NEVER) && hasSemicolon) {
+            const sourceFile = this.getSourceFile();
```
JKillian: Let's extract this line to above the conditional

src/rules/semicolonRule.ts
```diff
@@ -92,11 +96,28 @@ class SemicolonWalker extends Lint.RuleWalker {
     private checkSemicolonAt(node: ts.Node) {
         const children = node.getChildren(this.getSourceFile());
         const hasSemicolon = children.some((child) => child.kind === ts.SyntaxKind.SemicolonToken);
+        // Backwards compatible with plain {"semicolon": true}
+        const always = this.hasOption(OPTION_ALWAYS) || (this.getOptions() && this.getOptions().length === 0);
 
-        if (!hasSemicolon) {
+        if (always && !hasSemicolon) {
             const sourceFile = this.getSourceFile();
             const position = node.getStart(sourceFile) + node.getWidth(sourceFile);
```
JKillian: Let's also extract this variable to above the conditional

src/rules/semicolonRule.ts
```diff
@@ -92,11 +96,28 @@ class SemicolonWalker extends Lint.RuleWalker {
...
-            this.addFailure(this.createFailure(Math.min(position, this.getLimit()), 0, Rule.FAILURE_STRING));
+            this.addFailure(this.createFailure(Math.min(position, this.getLimit()), 0, Rule.FAILURE_STRING_MISSING));
+        } else if (this.hasOption(OPTION_NEVER) && hasSemicolon) {
+            const sourceFile = this.getSourceFile();
+            const position = node.getStart(sourceFile) + node.getWidth(sourceFile) - 1;
+            const scanner = ts.createScanner(ts.ScriptTarget.ES5, false, ts.LanguageVariant.Standard, this.getSourceFile().text);
+            scanner.setTextPos(position + 1);
```
JKillian: Here we can just do `scanner.setTextPos(position);`

src/rules/semicolonRule.ts
```diff
@@ -92,11 +96,28 @@ class SemicolonWalker extends Lint.RuleWalker {
...
+
+            let tokenKind = scanner.scan();
+            while (tokenKind === ts.SyntaxKind.WhitespaceTrivia || tokenKind === ts.SyntaxKind.NewLineTrivia) {
+                tokenKind = scanner.scan();
+            }
+
+            if (tokenKind !== ts.SyntaxKind.OpenParenToken && tokenKind !== ts.SyntaxKind.OpenBracketToken
+                    && tokenKind !== ts.SyntaxKind.PlusToken && tokenKind !== ts.SyntaxKind.MinusToken) {
+                this.addFailure(this.createFailure(Math.min(position, this.getLimit()), 0, Rule.FAILURE_STRING_UNNECESSARY));
```
JKillian: And this should become `position-1`

test/rules/semicolon/never/test.ts.lint
```diff
@@ -0,0 +1,100 @@
+var x = 3;
+         ~nil [unnecessary semicolon]
```
JKillian: All of these will go from being `~nil` to just plain old `~`.

test/rules/semicolon/never/test.ts.lint
```diff
@@ -0,0 +1,100 @@
...
+               ~nil [unnecessary semicolon]
+    baz: boolean
+}
+
+import {Router} from 'aurelia-router'
+
+import {Controller} from 'my-lib';
+                                 ~nil [unnecessary semicolon]
+
+// Edge cases when omitting semicolon needs to be supported
```
JKillian: I think you mean "not omitting"

test/rules/semicolon/never/test.ts.lint
```diff
@@ -0,0 +1,100 @@
...
+;("1" + "2").length
+
+var a = 1
+;[].length
+
+var a = 1
+;+"a"
+
+var a = 1
+;-1
```
JKillian: Can we add a test case in here similar to this:

```
for(var i = 0; i < 10; ++i) {
  // do something
}
```

JKillian:
@Pajn This looks really good! Just left a few comments above on tweaks I think will be an improvement, let me know if anything's unclear

Pajn:
Great comments and good catch on the for loop.
Should be fixed now.

src/rules/semicolonRule.ts
```diff
@@ -92,11 +96,26 @@ class SemicolonWalker extends Lint.RuleWalker {
...
-            this.addFailure(this.createFailure(Math.min(position, this.getLimit()), 0, Rule.FAILURE_STRING));
+        const sourceFile = this.getSourceFile();
+        const position = node.getStart(sourceFile) + node.getWidth(sourceFile);
+        // Backwards compatible with plain {"semicolon": true}
+        const always = this.hasOption(OPTION_ALWAYS) || (this.getOptions() && this.getOptions().length === 0);
+
+        if (always && !hasSemicolon) {
+            this.addFailure(this.createFailure(Math.min(position, this.getLimit()), 0, Rule.FAILURE_STRING_MISSING));
+        } else if (this.hasOption(OPTION_NEVER) && hasSemicolon) {
+            const scanner = ts.createScanner(ts.ScriptTarget.ES5, false, ts.LanguageVariant.Standard, this.getSourceFile().text);
```
JKillian: Can use `sourceFile` variable here

src/rules/semicolonRule.ts
```diff
@@ -92,11 +96,26 @@ class SemicolonWalker extends Lint.RuleWalker {
     private checkSemicolonAt(node: ts.Node) {
         const children = node.getChildren(this.getSourceFile());
```
JKillian: Can use `sourceFile` variable here too

JKillian:
Nice, looking forward to merging this! Two pedantic comments above, but if you don't have time to make the tweak in the next couple hours, no matter, I'll just merge this and can do it myself later

Pajn:
Fixed :)

adidahiya:
:+1: thanks @Pajn!

Issue #522, opened 2015-07-22
Excludes for variable-name


Anahkiasen:
I have a variable in my app that comes from a third party and needs to be named a certain way. Would it be possible to add the ability to exclude a certain variable name? Like,

``` js
"variable-name": [true,
    "TheVarName",
],
```

Or something like that?

sundaysen:
You can insert rule exclusions in your code using comments
```
/* tslint:disable: name-of-rule-to-disable */
yourCodeHere
/* tslint:enable */
```

Anahkiasen:
Yes but in my case it's a service I import in multiple controllers and I don't want to litter my codes with comments like that, it would be way easier to ignore it globally. If the syntax is the issue then maybe an object (which would be clearer anyway than the array thing used currently):

```js
"variable-name": {
  enabled: true,
  ignore: ['TheVarName'],
},
```

adidahiya:
What's the exact code this rule is failing on? Are you re-declaring the variable? If it's provided by a third party library, shouldn't it be declared outside of the code you're linting?

Anahkiasen:
It's an Angular service I'm injecting so the variable name has to match the service name registered on the container by the third party

adidahiya:
ah, ok. This is similar to #457. Do you have use cases where the injected variable is used as something _other_ than a constructor? Meaning, if we resolve #457, would that satisfy your use case completely?

Anahkiasen:
Hm I think it might, not sure, it's a factory. Here's my code dumbed down:

```ts
class MyClass {
  constructor(public Flash: Flash) {}

  someMethod() {
    this.Flash.create('success', 'The message');
  }
```

adidahiya:
Even with #577 implemented, that still doesn't address variables from third party libs which are declared as interfaces instead of classes. So I agree we should implement excludes for the variable-name rule.

ajcrites:
Another example of when this comes up is using stateless functional components in react.

    export const Example: React.SFC<{}> = () => <div />;

With the current rule, this variable name doesn't pass the lint check, but React requires that custom components including SFCs be capitalized.

ChrisSG:
The SFC is what hits us at the moment. I think that maybe only classnames should be PascalCase and one should explicitely do it like this:

const example = ...

export {example as Example}

Any more thoughts on this?

joshghent:
[...]

PR #981, opened 2016-02-19
Add ignore-pattern option to no-unused-variable

### LINKED ISSUE #314: Selectively suppres unused variable warning
Haskell has a nice way to selectively suppress their equivalent of *unused variable* by prepending the variable with an underscore. I'm particularly interested in suppressing the warning for unused function arguments.

Pajn:
fixes #314

### DIFF
--- a/README.md
+++ b/README.md
@@ -200,6 +200,7 @@ A sample configuration file with all options is available [here](https://github.
     * `"check-parameters"` disallows unused function and constructor parameters.
         * NOTE: this option is experimental and does not work with classes that use abstract method declarations, among other things. Use at your own risk.
     * `"react"` relaxes the rule for a namespace import named `React` (from either the module `"react"` or `"react/addons"`). Any JSX expression in the file will be treated as a usage of `React` (because it expands to `React.createElement`).
+    * `{"ignore-pattern": "pattern"}` where pattern is a case-sensitive regexp. Variable names that match the pattern will be ignored.
 * `no-use-before-declare` disallows usage of variables before their declaration.
 * `no-var-keyword` disallows usage of the `var` keyword, use `let` or `const` instead.
 * `no-var-requires` disallows the use of require statements except in import statements, banning the use of forms such as `var module = require("module")`.
--- a/src/rules/noUnusedVariableRule.ts
+++ b/src/rules/noUnusedVariableRule.ts
@@ -80,7 +80,10 @@ class NoUnusedVariablesWalker extends Lint.RuleWalker {
                 && !this.isReactUsed
                 && !this.hasSeenJsxElement) {
             const nameText = this.reactImport.name.getText();
-            this.addFailure(this.createFailure(this.reactImport.name.getStart(), nameText.length, `${Rule.FAILURE_STRING}'${nameText}'`));
+            if (!this.isIgnored(nameText)) {
+                const start = this.reactImport.name.getStart();
+                this.addFailure(this.createFailure(start, nameText.length, `${Rule.FAILURE_STRING}'${nameText}'`));
+            }
         }
     }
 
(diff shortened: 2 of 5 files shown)


JKillian:
Thanks @Pajn! Will review in the next couple of days.

As a sidenote, we might want to think about a similar option for #522 as well.

README.md
```diff
@@ -200,6 +200,7 @@ A sample configuration file with all options is available [here](https://github.
     * `"check-parameters"` disallows unused function and constructor parameters.
         * NOTE: this option is experimental and does not work with classes that use abstract method declarations, among other things. Use at your own risk.
     * `"react"` relaxes the rule for a namespace import named `React` (from either the module `"react"` or `"react/addons"`). Any JSX expression in the file will be treated as a usage of `React` (because it expands to `React.createElement`).
+    * `{"ignore-pattern": "pattern"}` where pattern is a case sensitive regexp. Variable names that matches the pattern will be ignored.
```
JKillian: I wish we could accept a literal regex, but since we're using JSON, I suppose we have no good way to do this.

JKillian: Just a couple of grammar tweaks would be good here:

* "case sensitive" => "case-sensitive"
* "matches" => "match"

src/rules/noUnusedVariableRule.ts
```diff
@@ -279,8 +282,18 @@ class NoUnusedVariablesWalker extends Lint.RuleWalker {
...
-        if (highlights == null || highlights[0].highlightSpans.length <= 1) {
+        if ((highlights == null || highlights[0].highlightSpans.length <= 1) && !this.isIgnored(name)) {
             this.addFailure(this.createFailure(position, name.length, `${Rule.FAILURE_STRING}'${name}'`));
         }
     }
+
+    private isIgnored(name: string) {
+        const options = this.getOptions();
+        for (const option of options) {
+          if (typeof option === "object") {
```
JKillian: Indentation is off for this `if` block

src/rules/noUnusedVariableRule.ts
```diff
@@ -279,8 +282,18 @@ class NoUnusedVariablesWalker extends Lint.RuleWalker {
...
+            const {"ignore-pattern": ignorePattern} = option;
+            return ignorePattern && new RegExp(ignorePattern).test(name);
```
JKillian: We normally use an explicit check here, i.e. `ignorePattern != null`

JKillian:
Looks solid to me @Pajn, left a few notes for grammar/style tweaks

Pajn:
Fixed

JKillian:
Thanks! Merged

src/rules/noUnusedVariableRule.ts
```diff
@@ -279,8 +282,20 @@ class NoUnusedVariablesWalker extends Lint.RuleWalker {
...
         }
     }
+
+    private isIgnored(name: string) {
+        const options = this.getOptions();
+        for (const option of options) {
+            if (typeof option === "object") {
+                const {"ignore-pattern": ignorePattern} = option;
+                if (ignorePattern != null) {
+                    return new RegExp(ignorePattern).test(name);
```
wereHamster: What performance overhead does compiling the pattern on each check have? I'd expect this function to be called very frequently...

Pajn: It will be falled once per unused variable, to be noticeable you would need loads of unused variables.

JKillian: @wereHamster This is a fair point which I considered and we can change it if needed so the RegExp is just compiled once and stored as a class member. However, my hunch is that compared to `languageService.getDocumentHighlights` the impact is small. Have you been running into performance issues in general?

adidahiya: regardless, recompiling the regexp is unnecessary; let's move it to a class member.

PR #984, opened 2016-02-21
[WIP] Enforce tsfmt of tslint code base

myitcv:
_Recreating the PR because I managed to mess up #841_

[View diff ignoring whitespace changes](https://github.com/palantir/tslint/compare/master...myitcv:tsfmt_code?w=1)

Tasks as part of this PR:

* [x] Agree on settings for `tsfmt.json`
* [x] Define scope of which files should be formatted
  * `+(src|test)/**/*.@(ts|tsx)`
* [x] Format all code (excluding test files) - partially complete subject to previous points
* [ ] Write `grunt` task to enforce formatting of files in scope (fails if `tsfmt --verify` fails) (see #732)
* [ ] Write `grunt` task to format and replace files in scope
* [ ] Add docs for contributors explaining the setup:
  * Reference new `grunt` tasks
  * Suggest they setup `tsfmt` as part of 'save' hook in editor


test/typings/chai/chai.d.ts
```diff
@@ -51,7 +51,7 @@ declare module Chai {
         haveOwnProperty: OwnProperty;
```
JKillian: This file shouldn't be formatted with tsfmt

test/files/eofline.test.ts
```diff
@@ -1,4 +1,4 @@
 var test = {
```
JKillian: This file also shouldn't be formatted (will break a test!)

test/files/formatters/jsonFormatter.test.ts
```diff
@@ -1,6 +1,6 @@
 var x = 123;
```
JKillian: This file shouldn't be formatted either

test/files/enabledisable.test.ts
```diff
@@ -20,7 +20,7 @@ var AAAaA = 'test'
 var AAAaA = 'test'
```
JKillian: file shouldn't be formatted

src/tslint-cli.ts
```diff
@@ -21,11 +21,11 @@ import * as optimist from "optimist";
 import * as path from "path";
 import * as Linter from "./tslint";
 import {
-    CONFIG_FILENAME,
-    DEFAULT_CONFIG,
-    findConfigurationPath,
-    getRulesDirectories,
-    loadConfigurationFromPath,
+CONFIG_FILENAME,
```
JKillian: Hmm, this is a little strange, wonder what the normal way to do things is in this situation?

myitcv:
Thanks for the comments @JKillian - I'll look to address those when I can (particularly those about not formatting certain `test/**` files)

JKillian:
@myitcv Looking over my comments, I think the `test/typings` and the `test/files` directories are the ones that need to be excluded, other `.ts` files in `test` will be source and good to be formatted.

adidahiya:
Closing this PR as it's very out of date, but I'm happy to continue discussion around this in a new issue. We'll also want to think of this in context with the new auto-fix formatter built-in to TSLint.

myitcv:
@adidahiya @JKillian apologies, I dropped the ball on this. Quite happy with you closing...

PR #1009, opened 2016-03-02
Alias some config functions at top-level to make more accessible to 3rd-party tools

JKillian:


JKillian:
This should make it easier to fix:
* https://github.com/panuhorsmalahti/gulp-tslint/issues/55
* https://github.com/wbuchwalter/tslint-loader/issues/16

adidahiya:
Let's add JSDoc comments to the exported functions in `configuration.ts` now that they're part of the public API.

JKillian:
Updated. We also now search for tslint.json before searching for package.json - I think this is a sensible thing to switch

PR #1016, opened 2016-03-05
Add extends support

### LINKED ISSUE #997: Extend configuration files
ESLint has a nice feature that allows you to extend configuration files. We should add support for this as well.

(edit: updated link) http://eslint.org/docs/user-guide/configuring#extending-configuration-files

unional:
I want to add some test in but don't know how. I have added a config.json file. Can you give me some pointers?

Also, please comment on the name and location of the functions.

I can adjust accordingly.

---

edit from @adidahiya: closes #997

### DIFF
--- a/custom-typings/path-is-absolute.d.ts
+++ b/custom-typings/path-is-absolute.d.ts
@@ -1,0 +1,6 @@
+
+declare module "path-is-absolute" {
+    namespace pathIsAbsolute {}
+    function pathIsAbsolute(path: string): boolean;
+    export = pathIsAbsolute;
+}
--- a/src/configuration.ts
+++ b/src/configuration.ts
@@ -18,10 +18,12 @@
 import * as fs from "fs";
 import * as path from "path";
 import * as findup from "findup-sync";
+import * as pathIsAbsolute from "path-is-absolute";
 
 import {arrayify, objectify} from "./utils";
 
 export interface IConfigurationFile {
+    extends?: string | string[];
     rulesDirectory?: string | string[];
     rules?: any;
 }
@@ -60,7 +62,7 @@ export const DEFAULT_CONFIG = {
 };
 
 const PACKAGE_DEPRECATION_MSG = "Configuration of TSLint via package.json has been deprecated, "
-   + "please start using a tslint.json file instead (http://palantir.github.io/tslint/usage/tslint-json/).";
+    + "please start using a tslint.json file instead (http://palantir.github.io/tslint/usage/tslint-json/).";
 
 /**
  * Searches for a TSLint configuration and returns the data from the config.
@@ -128,13 +130,43 @@ export function loadConfigurationFromPath(configFilePath: string): IConfiguratio
         console.warn(PACKAGE_DEPRECATION_MSG);
         return require(configFilePath).tslintConfig;
     } else {
-        let fileData = fs.readFileSync(configFilePath, "utf8");
-        // remove BOM if present
-        fileData = fileData.replace(/^\uFEFF/, "");
-        const configFile: IConfigurationFile = JSON.parse(fileData);
-        configFile.rulesDirectory = getRulesDirectories(configFile.rulesDirectory, path.dirname(configFilePath));
+        const resolvedConfigFilePath = resolveConfigurationPath(configFilePath);
+        let configFile: IConfigurationFile = require(resolvedConfigFilePath);
+        const configFileDir = path.dirname(resolvedConfigFilePath);
+
+        configFile.rulesDirectory = getRulesDirectories(configFile.rulesDirectory, configFileDir);
+        configFile.extends = arrayify(configFile.extends);
+
+        for (const name of configFile.extends) {
+            const baseConfigFilePath = resolveConfigurationPath(name, configFileDir);
+            const baseConfigFile = loadConfigurationFromPath(baseConfigFilePath);
+            configFile = extendConfigurationFile(configFile, baseConfigFile);
+        }
         return configFile;
     }
+}
+
+/**
+ * Resolve configuration file path
+ * @var relativeFilePath Relative path or package name (tslint-config-X) or package short name (X)
+ */
+function resolveConfigurationPath(relativeFilePath: string, relativeTo?: string) {
+    let resolvedPath: string;
+    if (pathIsAbsolute(relativeFilePath)) {
+        resolvedPath = relativeFilePath;
+    } else if (relativeFilePath.indexOf(".") === 0) {
+        resolvedPath = getRelativePath(relativeFilePath, relativeTo);
+    } else {
+        try {
+            resolvedPath = require.resolve(relativeFilePath);
+        } catch (err) {
+            throw new Error(`Invalid "extends" configuration value - could not require "${relativeFilePath}". ` +
+                "Review the Node lookup algorithm (https://nodejs.org/api/modules.html#modules_all_together) " +
+                "for the approximate method TSLint uses to find the referenced configuration file.");
+        }
+    }
+
+    return resolvedPath;
 }
 
 export function extendConfigurationFile(config: IConfigurationFile, baseConfig: IConfigurationFile): IConfigurationFile {
--- a/test/configurationTests.ts
+++ b/test/configurationTests.ts
@@ -14,7 +14,8 @@
  * limitations under the License.
  */
 
-import {IConfigurationFile, extendConfigurationFile} from "../src/configuration";
+import * as fs from "fs";
+import {IConfigurationFile, extendConfigurationFile, loadConfigurationFromPath} from "../src/configuration";
 
 describe("Configuration", () => {
     it("extendConfigurationFile", () => {
@@ -49,6 +50,51 @@ describe("Configuration", () => {
                 "c": 3,
             },
             rulesDirectory: ["foo", "bar", "baz"],
+        });
+    });
+
+    describe("loadConfigurationFromPath", () => {
+        it("extends with relative path", () => {
+            let config = loadConfigurationFromPath("./test/config/tslint-extends-relative.json");
+
+            assert.isArray(config.rulesDirectory);
+            assert.isTrue(config.rules["no-fail"]);
+            assert.isFalse(config.rules["always-fail"]);
+        });
+
+        it("extends with package", () => {
+            let config = loadConfigurationFromPath("./test/config/tslint-extends-package.json");
+
+            assert.isArray(config.rulesDirectory);
+            assert.isFalse(config.rules["always-fail"]);
+            assert.isTrue(Object.keys(config.rules).length > 1);
+        });
+
+        it("extends with package without customization", () => {
+            let config = loadConfigurationFromPath("./test/config/tslint-extends-package-no-mod.json");
+
+            assert.isArray(config.rulesDirectory);
+            assert.isTrue(Object.keys(config.rules).length > 1);
+        });
+
+        it("extends with package two levels (and relative path in rulesDirectory)", () => {
+            let config = loadConfigurationFromPath("./test/config/tslint-extends-package-two-levels.json");
+
+            assert.isArray(config.rulesDirectory);
+            assert.lengthOf(config.rulesDirectory, 2);
+            assert.isTrue(fs.existsSync(config.rulesDirectory[0]));
+            assert.isTrue(fs.existsSync(config.rulesDirectory[1]));
+            assert.isFalse(config.rules["always-fail"]);
+            assert.isTrue(Object.keys(config.rules).length > 1);
+        });
+
+        it("extends with array", () => {
+            let config = loadConfigurationFromPath("./test/config/tslint-extends-package-array.json");
+
+            assert.isArray(config.rulesDirectory);
+            assert.isTrue(config.rules["no-fail"]);
+            assert.isFalse(config.rules["always-fail"]);
+            assert.isTrue(Object.keys(config.rules).length > 1);
         });
     });
 });
--- a/package.json
+++ b/package.json
@@ -25,6 +25,7 @@
     "findup-sync": "~0.2.1",
     "glob": "^6.0.1",
     "optimist": "~0.6.0",
+    "path-is-absolute": "^1.0.0",
     "underscore.string": "~3.1.1"
   },
   "devDependencies": {
@@ -39,6 +40,8 @@
     "grunt-tslint": "latest",
     "mocha": "^2.2.5",
     "tslint": "latest",
+    "tslint-config-typings": "0.1.6",
+    "tslint-config-unional": "0.1.5",
     "typescript": "latest"
   },
   "peerDependencies": {
--- a/src/tsconfig.json
+++ b/src/tsconfig.json
@@ -12,6 +12,7 @@
         "rewriteTsconfig": false
     },
     "filesGlob": [
+        "../custom-typings/**/*.d.ts",
         "../typings/**/*.d.ts",
         "./*.ts",
         "./formatters/**/*.ts",
@@ -20,6 +21,7 @@
         "./test/**/*.ts"
     ],
     "files": [
+        "../custom-typings/path-is-absolute.d.ts",
         "../typings/colors/colors.d.ts",
         "../typings/diff/diff.d.ts",
         "../typings/findup-sync/findup-sync.d.ts",
--- a/test/config/tslint-extends-package-array.json
+++ b/test/config/tslint-extends-package-array.json
@@ -1,0 +1,9 @@
+{
+  "extends": [
+    "tslint-config-unional",
+    "./tslint-custom-rules-with-two-dirs.json"
+  ],
+  "rules": {
+    "always-fail": false
+  }
+}
--- a/test/config/tslint-extends-package-no-mod.json
+++ b/test/config/tslint-extends-package-no-mod.json
@@ -1,0 +1,3 @@
+{
+  "extends": "tslint-config-unional"
+}
--- a/test/config/tslint-extends-package-two-levels.json
+++ b/test/config/tslint-extends-package-two-levels.json
@@ -1,0 +1,7 @@
+{
+  "extends": "tslint-config-typings",
+  "rulesDirectory": "./rules",
+  "rules": {
+    "always-fail": false
+  }
+}
--- a/test/config/tslint-extends-package.json
+++ b/test/config/tslint-extends-package.json
@@ -1,0 +1,6 @@
+{
+  "extends": "tslint-config-unional",
+  "rules": {
+    "always-fail": false
+  }
+}
--- a/test/config/tslint-extends-relative.json
+++ b/test/config/tslint-extends-relative.json
@@ -1,0 +1,6 @@
+{
+  "extends": "./tslint-custom-rules-with-two-dirs.json",
+  "rules": {
+    "always-fail": false
+  }
+}
--- a/test/tsconfig.json
+++ b/test/tsconfig.json
@@ -11,6 +11,7 @@
         "rewriteTsconfig": false
     },
     "filesGlob": [
+        "../custom-typings/**/*.d.ts",
         "../typings/**/*.d.ts",
         "./typings/**/*.d.ts",
         "../src/**/*.ts",
@@ -19,6 +20,7 @@
         "./rule-tester/*.ts"
     ],
     "files": [
+        "../custom-typings/path-is-absolute.d.ts",
         "../typings/colors/colors.d.ts",
         "../typings/diff/diff.d.ts",
         "../typings/findup-sync/findup-sync.d.ts",


unional:
Will update the PR. It should merge the whole config. Not just the `rules`.

Do you want it to be done directly in `loadConfigurationFromFile()`?

unional:
Move extends logic into `loadConfigurationFromPath`

Test is skipped right now because I don't know how to get the config copied and referenced.

unional:
Currently it should support relative path. Once this is okey from you then I will code for array and package support.

unional:
Updated to support package and array. Definitely need testing

JKillian:
Thanks for the PR @unional! I haven't reviewed the code, but I like the `extends` syntax, simple and usable

unional:
Please take a look at the PR when you have time. While the test passed, when I try to use it using `npm link` (with vscode) it is not working. I want to figure out why. 

Haven't test directly with `tslint` cli or `gulp-tslint` yet.

src/configuration.ts
```diff
@@ -120,8 +120,83 @@ export function loadConfigurationFromPath(configFilePath: string) {
...
+        } else if (typeof config.rulesDirectory === "string") {
+            (<string[]>parentRulesDirectory).unshift(<string>config.rulesDirectory);
+            config.rulesDirectory = parentRulesDirectory;
+        } else {
+            config.rulesDirectory = (<string[]>config.rulesDirectory).concat(<string[]>parentRulesDirectory);
+        }
+    }
+    let parentRules = parentConfig.rules;
+    if (parentRules) {
+        for (let name in parentRules) {
```
JKillian: We prefer `for (const name of Object.keys(parentRules)) {`

src/configuration.ts
```diff
@@ -120,8 +120,83 @@ export function loadConfigurationFromPath(configFilePath: string) {
...
+ * Resolve config file path.
+ * @var {string} relativeFilePath Relative path or
+ *   package name (tslint-config-X) or package short name (X)
+ */
+function resolveConfigFilePath(relativeFilePath: string, relativeTo: string) {
+    if (relativeFilePath.indexOf("/") === -1) {
+        relativeFilePath = `tslint-config-${relativeFilePath}`;
     }
+
+    if (relativeFilePath.indexOf("tslint-config-") === 0) {
```
JKillian: Instead of doing this manually here, we should just use `require` instead where appropriate to take care of this logic for us.

src/configuration.ts
```diff
@@ -120,8 +120,83 @@ export function loadConfigurationFromPath(configFilePath: string) {
...
+    rules: any;
+}
+
+/**
+ * Resolve config file path.
+ * @var {string} relativeFilePath Relative path or
+ *   package name (tslint-config-X) or package short name (X)
+ */
+function resolveConfigFilePath(relativeFilePath: string, relativeTo: string) {
+    if (relativeFilePath.indexOf("/") === -1) {
```
JKillian: Should we also check here that `relativeFilePath.indexOf("tslint-config-") === -1`? If they specify the path as something like "tslint-config-mypackage", we don't want to prefix it with "tslint-config-"

src/configuration.ts
```diff
@@ -120,8 +120,83 @@ export function loadConfigurationFromPath(configFilePath: string) {
...
         // remove BOM if present
         fileData = fileData.replace(/^\uFEFF/, "");
-        return JSON.parse(fileData);
+        let config = JSON.parse(fileData) as IConfigFile;
+
+        let parentConfigNames = config.extends;
+        if (parentConfigNames) {
+            const configFileDir = path.dirname(configFilePath);
+            if (Array.isArray(parentConfigNames)) {
+                (<string[]>parentConfigNames).forEach(name => {
```
JKillian: We prefer `for-of` loops over `forEach` normally

src/configuration.ts
```diff
@@ -120,8 +120,83 @@ export function loadConfigurationFromPath(configFilePath: string) {
         let fileData = fs.readFileSync(configFilePath, "utf8");
         // remove BOM if present
         fileData = fileData.replace(/^\uFEFF/, "");
-        return JSON.parse(fileData);
+        let config = JSON.parse(fileData) as IConfigFile;
+
+        let parentConfigNames = config.extends;
+        if (parentConfigNames) {
+            const configFileDir = path.dirname(configFilePath);
+            if (Array.isArray(parentConfigNames)) {
```
JKillian: Let's do

```
if (typeof parentConfigNames === "string") {
    parentConfigNames = [parentConfigNames];
}
```

And then below, we can just handle the array case and not have to duplicate logic

src/configuration.ts
```diff
@@ -120,8 +120,83 @@ export function loadConfigurationFromPath(configFilePath: string) {
         let fileData = fs.readFileSync(configFilePath, "utf8");
         // remove BOM if present
         fileData = fileData.replace(/^\uFEFF/, "");
-        return JSON.parse(fileData);
+        let config = JSON.parse(fileData) as IConfigFile;
+
+        let parentConfigNames = config.extends;
+        if (parentConfigNames) {
```
JKillian: In general in TSLint code, we use explicit checks for `null`/`undefined` instead of implicit ones. So instead of `if (parentConfigNames)` we use `if (parentConfigNames != null)`. Or instead of `if (!someVar)` we would use `if (someVar == null)`.  This removes the weirdness of empty strings, `false`, etc being handled incorrectly and is more explicit overall.

unional:
@JKillian , what would you suggest in adding test? You can see in https://github.com/palantir/tslint/pull/1016/files#diff-7adcebcb9ab1532bda827cc5be7bc94bR6, I'm routing back to the test source for the config file.

Any suggestion?

src/configuration.ts
unional: What do you mean by `require` here? Put `resolveConfigFilePath` to another file?

src/configuration.ts
unional: Yes

src/configuration.ts
unional: Yes

src/configuration.ts
unional: ok

src/configuration.ts
unional: Yup. Should have done that. Was trying to get the typing work and then forgot about this.

JKillian:
Well, the cwd for the tests should be the root of the project, so instead of doing `__dirname + "/../../test/config/tslint-extends-rules.json"` you should just be able to use `./test/config/tslint-extends-rules.json`.

As far as testing with a package, I think the easiest solution will be to make a super-simple npm package and list it as a devDependency.

unional:
> Well, the cwd for the tests should be the root of the project, so instead of doing __dirname + "/../../test/config/tslint-extends-rules.json" you should just be able to use ./test/config/tslint-extends-rules.json.

That doesn't work because the `config/*` is not copied to `build` folder.

unional:
> As far as testing with a package, I think the easiest solution will be to make a super-simple npm package and list it as a devDependency

I have one ready.

JKillian:
> That doesn't work because the config/* is not copied to build folder.

But `loadConfigurationFromPath` uses `fs.readFileSync`, which if supplied a relative path, will load that path relative to the process's current working directory. In this case, the cwd for running tests should be the project root, and you should be able to use a relative path like `./test/config/tslint-extends-rules.json`. Have you tried this and it doesn't work? Perhaps I'm not understanding something correctly here.

unional:
The `cwd` at that moment is `./build/test/`, thus  __dirname + "../../test/config/..." point to the right file.

unional:
What's your opinion on `delete config.extends`?

unional:
Don't have to review code at the moment. Found bugs. Since I can now run tests, I'll get them all work and do a rebase. Will ping you soon. :rose:

JKillian:
> What's your opinion on `delete config.extends`?

Probably better to not do this I'd say

unional:
Alrighty.

unional:
Tests in, passed. Code rebased. Ready to go! @JKillian

src/configuration.ts
```diff
@@ -120,8 +120,79 @@ export function loadConfigurationFromPath(configFilePath: string) {
...
+    if (relativeFilePath.indexOf("/") === -1 && relativeFilePath.indexOf("tslint-config") === -1) {
+        relativeFilePath = `tslint-config-${relativeFilePath}`;
     }
+
+    if (relativeFilePath.indexOf("tslint-config-") === 0) {
+        let fileData = fs.readFileSync(path.resolve(`node_modules/${relativeFilePath}/package.json`), "utf8");
+        let packageJson = JSON.parse(fileData);
+        return path.resolve(`node_modules/${relativeFilePath}/${packageJson.main}`);
+    } else {
+        return path.resolve(relativeTo, relativeFilePath);
```
unional: Do you want to change this to:
```ts
return getRelativePath(relativeFilePath, relativeTo);
```

unional:
With this in, custom rules would be published as `tslint-config-X` too. Rules directory will be merged.:
```js
// tslint-eslint-rules -> tslint-config-eslint
// main: tslint.json
{
  "rulesDirectory": "./<local rule relative path>",
  "rules": {
    // Can add default rules value here, and
    // disable conflicting core rules if needed.
  } 
}
```

unional:
Speaking of which, would it make sense to resolve package shorthands with `tslint-` along with `tslint-config-`?
I originally do `tslint-config-` to mirror `eslint-config-` behavior.

JKillian:
Let's keep it simple here, let's only use `tslint-config-`, no need for the other.

Your previous comment actually brought up an interesting point:
When we resolve rulesDirectory fields from extended `tslint.json` files, we need to resolve those paths relative to each `tslint.json` and not just relative to the root/starting `tslint.json`. I think this should be pretty easy to take care of in `applyExtends`

src/configuration.ts
JKillian: Sure, let's use that instead

src/configuration.ts
```diff
@@ -120,8 +120,80 @@ export function loadConfigurationFromPath(configFilePath: string) {
...
+ *   package name (tslint-config-X) or package short name (X)
+ */
+function resolveConfigFilePath(relativeFilePath: string, relativeTo: string) {
+    // `relativeFilePath.indexOf("/") === -1` will fail sub-rule use case when it is supported.
+    if (relativeFilePath.indexOf("/") === -1 && relativeFilePath.indexOf("tslint-config") === -1) {
+        relativeFilePath = `tslint-config-${relativeFilePath}`;
     }
+
+    if (relativeFilePath.indexOf("tslint-config-") === 0) {
+        let fileData = fs.readFileSync(path.resolve(`node_modules/${relativeFilePath}/package.json`), "utf8");
```
JKillian: To clarify on my previous comment, instead of guessing the package location, parsing the `package.json` file, and finding the main file, we can use `require` instead. If we just do `require('package-name')` it will do all of the above for us and we won't have to implement a half-baked version of this.

The tricky thing here then, is what this function should do. I think this function should just return the name of the package. Then `loadConfigurationFromPath` can take care of actually `require`ing the package

unional: I see. make sense. However, what should the package expose?
```ts
import * as tslintJson form 'somepackage';
import tslintJson from 'somepackage';
```
or something else? Is there a fixed convention?
Because from `tslint-eslint-rules`, it asks consumer to:

```js
{
  "ruleDirectory": "node_modules/tslint-eslint-rules/dist/rules"
}
```
So I'm not sure what the convention should be.

Do you have defined this convention?

src/configuration.ts
unional: ok

unional:
> When we resolve rulesDirectory fields from extended tslint.json files, we need to resolve those paths relative to each tslint.json and not just relative to the root/starting tslint.json

I'll resolve them to full path. Ok?

src/configuration.ts
unional: Also, the original implementation only `require('<path to package.json>')` and get the tslintConfig, and manually read the file `using fs.readFileSync()` (and remove BOM). I don't know would `require('package-name')` would work correctly in this case.

Of course, I can change and test it. Have tests in place.

unional:
I saw you are trying to get ride of the package.json/tsLintConfig:
https://github.com/palantir/tslint/issues/1020

So if you define what the `require()` would do, then maybe I can just do it here and code that too.

unional:
@JKillian , don't know why the build is failing. Test runs correctly in local.  Can you take a look?

JKillian:
So it looks like `extends with package two levels (and relative path in rulesDirectory)` is failing. The error message is: ` AssertionError: Count not find node_modules/tslint-eslint-rules/dist/rules`.

What's happening here is the danger of relying on npm to install things at a specific location. On your system, the directory structure is something like this and things work as expected (tree diagram is abbreviated):

```
.
└── node_modules
    └── tslint-config-unional
        ├── tslint.json
        └── node_modules
            └── tslint-eslint-rules
```

However, NPM can also perfectly legitimately install things like this, and NPM3 will:

```
.
└── node_modules
    ├── tslint-eslint-rules
    └── tslint-config-unional
        └── tslint.json
```

And with this directory structure, the relative path in `tslint-config-unional/tslint.json` won't work correctly :/

unional:
Um... I think that is not the case. I'm using npm 3.8. It is the same structure as the second example you have.

And note that the path is `node_modules/...`, and not `./node_modules/...` (well, that likely doesn't make a difference. :stuck_out_tongue: ). And `path.resolve()` is called without other params, i.e. it is using the local dir. i.e. it's the project root as you mentioed.

unional:
or.....unless you mean test failed on older node versions. Then I understand your point.

unional:
This means, it is actually a *usability bug* on the custom rule plugin side. It should have implemented as extends, so that:
```js
// tsling-config-unional/tslint.json
{
  "extends": "tslint-eslint-rules",
  ...
}

// tslint-eslint-rules/tslint.json
{
  "rulesDirectory": "./dist/rules",
  ...
}
```

unional:
cc/@buzinas

JKillian:
Indeed - it failed for npm 2. If you check out [this travis-ci build](https://travis-ci.org/palantir/tslint/builds/114714670) you can see it's only with npm 3 that it passed.

> And path.resolve() is called without other params, i.e. it is using the local dir. i.e. it's the project root as you mentioned

I think the paths should be resolved relative to the location of the `tslint.json` file containing them. (Although that makes it hard to handle a situation like this correctly.) We really don't want to resolve relative to the current working directory here - that would lead to all sort of strange bugs where your paths are affected by where you run the command from. For example, a user running `tslint -c proj/tslint.json proj/file.ts` ought to be able to expect the same results as `cd proj; tslint -c tslint.json file.ts`. But if we resolve relative to the current working directory, we could get different results here

buzinas:
@unional Sorry, I can't get what this "extends" is trying to solve. Can you briefly explain it to me? Is there anything I can help you guys with?

unional:
@buzinas , sure. It is nicely summarized by @JKillian here: https://github.com/palantir/tslint/pull/1016#issuecomment-194682347

What happen is that relying on `ruleDirectory` pointing to `node_modules/<package>/<path>` would lead to npm 2/3 tree vs flat issue.
Since in `tslint-config-unional` I try to use your rules, it is written as:
```js
// tslint-config-unional/tslint.json
{
  "rulesDirectory": "node_modules/tslint-eslint-rules/dist/rules",
  ...
}
```
just as described in your readme.

However since it is an extends, then the relative path approach will mess up when:
```sh
tslint-config-typings
+ --- tslint-config-unional
       + --- tslint-eslint-rules
```

unional:
Oh, and `extends` is the same feature as `extends` in `eslint`.

unional:
> I think the paths should be resolved relative to the location of the tslint.json file containing them.

Agree to your point 100% Let me make the change.

src/configuration.ts
```diff
@@ -34,13 +34,16 @@ export const DEFAULT_CONFIG = {
...
+        "typedef-whitespace": [
+            true,
+            {
+                "call-signature": "nospace",
+                "index-signature": "nospace",
+                "parameter": "nospace",
+                "property-declaration": "nospace",
+                "variable-declaration": "nospace",
+            },
+        ],
```
unional: This alignment is changed to avoid vscode constantly kill the space in the original line 43: `},<space>],` when execute `format code`, and causes build error.

src/configuration.ts
```diff
@@ -111,19 +114,98 @@ export function findConfigurationPath(suppliedConfigFilePath: string, inputFileP
...
+    return filePath;
+}
+
+function applyExtends(config: IConfigFile, parentConfig: IConfigFile): IConfigFile {
+    let parentRulesDirectory = parentConfig.rulesDirectory;
+    if (parentRulesDirectory != null) {
+        if (config.rulesDirectory == null) {
+            config.rulesDirectory = parentRulesDirectory;
+        } else {
+            config.rulesDirectory = (<string[]>config.rulesDirectory).concat(<string[]>parentRulesDirectory);
```
unional: `config.rulesDirectory` will already be `string[]` here because of https://github.com/palantir/tslint/pull/1016/files#diff-8b01d5598b81332eb62385d26f3ead22R138

I agree that it is a bit logic bleeding, so change suggestion welcome.

unional:
I have updated my package to not depends on `tslint-eslint-rules` at the moment. At such, the build succeeds.

I'll open another issue regarding the `rulesDirectory` usage.

unional:
@JKillian , please let me know do you want to add `up(...)` code to look up the tree for `node_modules/*` usage to support existing `rulesDirectory` usage.

I didn't put that in because I think it would complicate the code unnecessary, but if you think that is a good way to solve the problem, then I can make the change.

JKillian:
@unional I used some of your code/ideas in writing #1035, which is needed badly for a lot of 3rd-party tools. Once we merge that, you an rebase your PR and we'll have a few less lines of code to be dealing with here.

JKillian:
(Sorry, accidentally clicked the close button...)

JKillian:
Haven't forgotten about this! I'm planning on making sure we have a good design for dealing with node_modules and paths and configuration and all that this week

unional:
Ok :)

JKillian:
@unional, see #1049 for how I'm thinking about loading rules / extending files from node modules.

The gist of it is this:

We get rid of the magic behavior that happens if you have a path that starts with `tslint-config-`. Instead, any path that references a node module will have a `node_module:` prefix. For example, `"node_module:tslint-amazing-config/tslint.json"`. This path then gets parsed (see the `resolveNodeModulePath` function of #1049). Essentially, we use [`require.resolve`](https://nodejs.org/api/globals.html#globals_require_resolve) to find the path to the `tslint-amazing-rules` module and then we append the rest of the path, in this case "tslint.json" to that path.

The same thing will work with directories to custom rules - someone could provide a path: `"node_module:tslint-amazing-rules/path/to/rules"` for example.

And of course, people will still be able to use regular old file paths for both `extends` and `rulesDirectory` if they want to.

unional:
Updated. Please review. :smile:

unional:
The fail build it the original one. The relative path with node_modules.

Updating to a version that does not have that dependency

unional:
Just made one more checkin to make sure npm < 3 test case are properly covered.

unional:
Ready to go!

Hope can make it to 3.7

src/configuration.ts
```diff
@@ -128,15 +129,46 @@ export function loadConfigurationFromPath(configFilePath: string): IConfiguratio
...
+        }
         return configFile;
     }
 }
 
+/**
+ * Resolve configuration file path
+ * @var relativeFilePath Relative path or package name (tslint-config-X) or package short name (X)
+ */
+function resolveConfigurationPath(relativeFilePath: string, relativeTo: string) {
```
JKillian: I think this function should have three cases (pseduocode below):

```
if path.isAbsolute(filePath):  // it could be an absolute path
  return filePath
else if filePath.startsWith("."):
  return getRelativePath(filePath, relativeTo)  // ha, this should really be called resolveRelativePath, but too late now
else:
  return require.resolve(filePath);
```

So no matter what, it returns an absolute path to a file. Also, let's not do any magic prefixing by adding "tslint-config-": if they provide a package name let's just use it as is.

adidahiya: +1 for not magically handling the "tslint-config-" prefix. For example, I like how stylelint does `extends` https://github.com/stylelint/stylelint/blob/master/docs/user-guide/configuration.md#extends

JKillian: `relativeTo?: string` (can be made optional)

src/configuration.ts
```diff
@@ -128,15 +129,46 @@ export function loadConfigurationFromPath(configFilePath: string): IConfiguratio
         console.warn(PACKAGE_DEPRECATION_MSG);
         return require(configFilePath).tslintConfig;
     } else {
-        let fileData = fs.readFileSync(configFilePath, "utf8");
-        // remove BOM if present
-        fileData = fileData.replace(/^\uFEFF/, "");
-        const configFile: IConfigurationFile = JSON.parse(fileData);
-        configFile.rulesDirectory = getRulesDirectories(configFile.rulesDirectory, path.dirname(configFilePath));
+        let configFile: IConfigurationFile;
```
JKillian: Then with the changes noted below, we should be able to do something like:
```
const resolvedConfigFilePath = resolveConfigurationPath(configFilePath);
const configFile: IConfigurationFile = require(resolvedConfigFilePath);
const configFileDir = path.dirname(resolvedConfigFilePath);
```

We should be able to eliminate lines 132 through 143 then

src/configuration.ts
unional: Interesting. I looked at stylelint.  It is using `resolve-from` package. which uses `require('module')`. Do you want to use the `resolve-from` package, or I just re-use those code?

unional: > not magically handling the "tslint-config-" prefix.

ok


> relativeTo?: string (can be made optional)

Yes. Done

JKillian: > I can't use require('module') directly, seems like it is not typed in node.d.ts.

Can you elaborate on why this doesn't work? Looks like the typings are fine:

image

unional: Module is not typed.

image

unional:
Decided to just use `require.resolve()`. Any error message suggestion if it throws?

unional:
Build failing because `path.isAbsolute` is not available in older node. Any suggestion?

unional:
https://github.com/sindresorhus/path-is-absolute

JKillian:
> https://github.com/sindresorhus/path-is-absolute

Feel free to just add that as a dependency and use it. I'm not sure how the legal technicalities work of copying code form an MIT-licensed project into an Apache licensed one

JKillian:
Also, I like how simple & clean this code is ending up for such a powerful feature! This will be a great PR once it gets merged in

unional:
Ok

What you think on the exception message for resolve?

unional:
Also, for the typings for `path-is-abs`, I have added it to `typings`. But adding to DT might take a while. What's your approach? Custom?

JKillian:
> What you think on the exception message for resolve?

Maybe something like this:

_Error: Invalid "extends" configuration value - could not require "some-invalid-path". Review the Node lookup algorithm (https://nodejs.org/api/modules.html#modules_all_together) for the approximate method TSLint uses to find the referenced configuration file._

> Also, for the typings for path-is-abs, I have added it to typings. But adding to DT might take a while. What's your approach? Custom?

Seems like a good solution to me, just drop it in `typings/custom`.

unional:
Ok. I drop the file under `/custom-typings/` instead of `/typings/custom` to differentiate it from the typings installed by `tsd`.

Should be good to go V2!

src/configuration.ts
```diff
@@ -78,7 +79,7 @@ export function findConfiguration(configFile: string, inputFilePath: string): IC
 /**
  * Searches for a TSLint configuration and returns the path to it.
  * Could return undefined if not configuration is found.
- * @param suppliedConfigFilePath A path to an known config file supplied by a user. Pass null here if
+ * @param suppliedConfigFilePath A path to an know``n config file supplied by a user. Pass null here if
```
adidahiya: stray ``

unional: Ouch

custom-typings/path-is-absolute.d.ts
```diff
@@ -0,0 +1,5 @@
+
+declare module "path-is-absolute" {
+    function pathIsAbsolute(path: string): boolean;
+    export = pathIsAbsolute;
```
JKillian: can you add a line here: `namespace pathIsAbsolute {}`. This will let us import it with ES6 syntax

src/configuration.ts
```diff
@@ -18,10 +18,12 @@
 import * as fs from "fs";
 import * as path from "path";
 import * as findup from "findup-sync";
+import pathIsAbsolute = require("path-is-absolute");
```
JKillian: change to `import * as pathIsAbsolute from "path-is-absolute"`

src/configuration.ts
```diff
@@ -128,15 +130,46 @@ export function loadConfigurationFromPath(configFilePath: string): IConfiguratio
...
+            throw new Error(`Invalid "extends" configuration value - could not require "${relativeFilePath}". ` +
+                "Review the Node lookup algorithm (https://nodejs.org/api/modules.html#modules_all_together) " +
+                "for the approximate method TSLint uses to find the referenced configuration file.");
+        }
+    }
+
+    return resolvedPath;
+}
+
+
```
JKillian: remove extra newline here

test/configurationTests.ts
```diff
@@ -14,7 +14,9 @@
  * limitations under the License.
  */
 
-import {IConfigurationFile, extendConfigurationFile} from "../src/configuration";
+import fs = require("fs");
```
JKillian: import * as fs from "fs";

test/configurationTests.ts
```diff
@@ -51,4 +53,45 @@ describe("Configuration", () => {
...
+
+    describe("loadConfigurationFromPath", () => {
+        it("extends with relative path", () => {
+            let config = loadConfigurationFromPath("./test/config/tslint-extends-relative.json");
+
+            assert.isArray(config.rulesDirectory);
+            assert.isTrue(config.rules["no-fail"]);
+            assert.isFalse(config.rules["always-fail"]);
+        });
+        it("extends with package", () => {
```
JKillian: newlines between tests are nice

JKillian:
Sweet! Comments I left are just minor style things, I'll make the tweaks in a follow-up PR if you don't have a chance to do it yourself.

unional:
Sure

unional:
Regarding the `declare namespace ...` trick, it is neat and I haven't thought about that before.

I don't know if you aware of https://github.com/Microsoft/TypeScript/issues/7398. There may be problem in the future.

JKillian:
:balloon:

JKillian:
I wrote a little post about this: http://palantir.github.io/tslint/2016/03/31/sharable-configurations-rules.html

Does it all seem correct to you @unional?

unional:
Seems like you go all the way to the complex scenario. :smile:
Haven't try the `tslint-crazy-config.js` route myself. But given of what it is, it should work.

The simplest case would be:
```js
// package.json
{
  "main": "tslint.json"  // or any json file
}
```
```js
// tslint.json
{
  "rulesDirectory": "./rules",
  "rules": {
     // default values
  }
}
```

Then if you want to extend from someone else, just change your `tslint.json` to:
```js
// tslint.json
{
  "extends": "some-other-package",
  "rulesDirectory": "./rules",
  "rules": {
     // my rules default
     // overriding `some-other-package` defaults
  }
}
```

unional:
In your example you are suggesting the package authors to share multiple configs.  It is powerful, but it is also more advanced. :rose:

unional:
So you may organize the post into three sections:
- how do I use it as the end user
- how do I use it as a package author extending other packages (i.e. adding my own touch)
- how do I use it as a package author to expose multiple configuration for my user to use.

My two cents.

PR #1022, opened 2016-03-08
fix isBlockScopeBoundary, add test case for no-shadowed-variable.

### LINKED ISSUE #1021: no-shadowed-variable will make a mistake if there was an alone `{}` block
```javascript
/** case 1 */
if (true) {
    const a = 3
}
function test1() {
    const a = 5
}

/** case 2 */
{
    const b = 3
}
function test2() {
    const b = 5
}

```

Case 1 will pass, but 2 will fail. That is incorrect.

meowtec:
fix #1021

If node is a alone block, e.g.
```
{
   const a = 3
}
```
`isBlockScopeBoundary()` should return true


adidahiya:
:+1:

JKillian:
Thanks @meowtec!

PR #1026, opened 2016-03-10
Add no-type-assertion-expression rule.

### LINKED ISSUE #639: Enforce use of as operator instead of <cast>
TypeScript 1.6 introduced the [`as` operator][0] for typecasts. In `.tsx` files it is the only valid cast method, whereas in `.ts` files either cast method can be used. It makes sense to just enforce using `as` everywhere, for consistency and for ease of changing a `.ts` file to a `.tsx` file.

[0]: https://github.com/Microsoft/TypeScript/wiki/What%27s-new-in-TypeScript#new-tsx-file-extension-and-as-operator

Arnavion:
Checks for uses of the `<>` type assertion syntax and suggests changing it to the `as` syntax.

Fixes #639 

---

The name of the rule comes from TS - the `<>` syntax is called a TypeAssertionExpression and the `as` syntax is called an AsExpression. But I can imagine it'll be confusing to tslint users because functionally they're both type assertion expressions. I'm open to suggestions for the name. (`no-angle-bracket-type-assertions` ? I'm hesitant about calling it `old` or `legacy`.)

I also didn't make this a tri-state (disabled, enforce `as`, enforce `<>`) because I'm not sure anyone will care for the last one.

For the test case - I wrote it initially without the errors, then ran `grunt` and accepted its output that contained the errors. Is there anything else I need to do?

### DIFF
--- a/src/rules/noAngleBracketTypeAssertionRule.ts
+++ b/src/rules/noAngleBracketTypeAssertionRule.ts
@@ -1,0 +1,34 @@
+/**
+ * @license
+ * Copyright 2013 Palantir Technologies, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import * as ts from "typescript";
+import * as Lint from "../lint";
+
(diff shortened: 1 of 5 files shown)


JKillian:
Thanks @Arnavion, I'll look over this shortly! Your build is showing up as failing, but it's actually fine, the build system just had an unhappy moment when you pushed

JKillian:
Code looks great @Arnavion! The way you wrote the test is the same way I often do it.

As far as naming of the rule, I'm not sure either. Any thoughts @adidahiya?

adidahiya:
`no-angle-bracket-type-assertion` sounds appropriate.

looks like the appveyor failure is a flake.

JKillian:
Sounds good, agreed that `no-type-assertion-expression` is not immediately clear, lets use `no-angle-bracket-type-assertion` instead.

Arnavion:
Updated to `no-angle-bracket-type-assertion` and rebased onto latest master.

JKillian:
Thanks @Arnavion! This'll be available in the next release

Issue #1020, opened 2016-03-07
Drop support for loading configuration from `package.json`


JKillian:
Handling `package.json` files as a potential source of configuration information clutters our code and increases complexity. It seems like a somewhat unneeded feature to me as well - let's just standardize on `tslint.json` files.

adidahiya:
:+1:

PR #1027, opened 2016-03-10
Add deprecation warning.

hamhut1066:
This is a different approach to issue #1020.

### DIFF
--- a/src/configuration.ts
+++ b/src/configuration.ts
@@ -19,6 +19,8 @@ import * as fs from "fs";
 import * as path from "path";
 import * as findup from "findup-sync";
 
+const PACKAGE_DEPRECATION_MSG = "Configuration of TSLint via package.json has been deprecated,\
+please start using the standard tslint.json (http://palantir.github.io/tslint/usage/tslint-json/)";
 export const CONFIG_FILENAME = "tslint.json";
 export const DEFAULT_CONFIG = {
     "rules": {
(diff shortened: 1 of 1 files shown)


src/configuration.ts
```diff
@@ -91,6 +92,7 @@ export function findConfigurationPath(suppliedConfigFilePath: string, inputFileP
         // search for package.json with tslintConfig property
         configFilePath = findup("package.json", { cwd: inputFilePath, nocase: true });
         if (configFilePath != null && require(configFilePath).tslintConfig != null) {
+            console.warn(PACKAGE_DEPRICATION_MSG);
```
JKillian: Let's not add the warning message here - let's just keep it in Line 120 below

src/configuration.ts
```diff
@@ -19,6 +19,7 @@ import * as fs from "fs";
 import * as path from "path";
 import * as findup from "findup-sync";
 
+const PACKAGE_DEPRICATION_MSG = "configuration of tslint via package.json has been deprecated, please start using the standard tslint.json"
```
JKillian: * `DEPRECATION` instead of `DEPRICATION`
* `TSLint` instead of `tslint` (I know we use these sort of interchangeably)
* semicolon at the end of the line
* `Configuration` instead of `configuration`.
* Could we also include this URL? http://palantir.github.io/tslint/usage/tslint-json/

JKillian:
Thanks @hamhut1066! I think this is the right approach for now, and then we'll remove the feature in a later version. Could you make the few small changes I noted above? Just amend your current commit with them instead of making a whole new commit

JKillian:
Thanks @hamhut1066!

PR #1035, opened 2016-03-14
Add configurationPath option

### LINKED ISSUE #1019: `configurationPath` option for programmatic API.
With more features coming to `tslint.json` files, such as the `rulesDirectory` option and the [`extends` option](https://github.com/palantir/tslint/issues/997), it becomes harder and harder for third-party tools that wrap TSLint to properly load configurations themselves. (See https://github.com/palantir/tslint/issues/1007 for example.)

The easiest and cleanest solution (that I can think of) for this is to change the Linter options API.

**Current:**
```ts
export interface ILinterOptions {
    configuration: any;
    formatter: string;
    formattersDirectory: string;
    rulesDirectory: string | string[];
}
```

**Option 1**
```ts
export interface ILinterOptions {
    configuration: ITSLintConfig | string;
    formatter: string;
    formattersDirectory: string;
    rulesDirectory: string | string[];
}
```

**Option2**
```ts
export interface ILinterOptions {
    configuration?: ITSLintConfig;
    configurationPath?: string;
    formatter: string;
    formattersDirectory: string;
    rulesDirectory: string | string[];
}
```

If a path to a configuration is provided (instead of passing a configuration object directly), TSLint would handle all responsibilities such as resolving rules directories, resolving extends clauses, etc. If a configuration object is passed directly, the third-party tool would have to take care of these tasks.

JKillian:
This commit lets us simplify the codebase in a couple other ways:
- helper `arrayify` and `objectify` functions make it easier to handle T
  | T[] types.
- `loadConfigurationFromPath` now resolves rules directories to absolute paths. This means 3rd-party tools can do `config: findConfiguration("possible/user/supplied/config/path", "file/ts/lint.ts")` and everything should work correctly.
- Adds a `node_module:` prefix that can be used to specify rules directories inside a node module (see #1049)

Closes #1019


src/lint.ts
```diff
@@ -45,7 +45,8 @@ export interface LintResult {
 }
 
 export interface ILinterOptions {
-    configuration: any;
+    configuration?: any;
+    configurationPath?: string;
```
JKillian: Most of the changes in this PR are good, but I'm wondering if we actually need this `configurationPath` option. If 3rd-party tools use `findConfiguration/loadConfigurationFromPath`, those functions can resolve relative paths which the programatic API should then be able to handle.

src/tslint.ts
```diff
@@ -88,6 +91,21 @@ class Linter {
...
+        if (configuration == null && configurationPath == null) {
+            throw new Error("An ILinterOptions object must have either `configuration` or `configurationPath` specified");
+        }
+
+        const configFromPath = configurationPath != null ? loadConfigurationFromPath(configurationPath) : {};
+        const configProvided = configuration != null ? configuration : {};
+        const finalConfig = extendConfigFile(configProvided, configFromPath);
+
+        this.options.configuration = finalConfig;
+        this.options.rulesDirectory = arrayify(rulesDirectory).concat(arrayify(finalConfig.rulesDirectory));
```
JKillian: This will overwrite rules directories supplied in this option, best to tweak this

JKillian:
Build is fine, review still needed

JKillian:
@adidahiya This needs review, and after this merges we can add the extends feature

unional:
Can't wait for this to go in so I can complete the extend feature.

src/configuration.ts
```diff
@@ -19,6 +19,13 @@ import * as fs from "fs";
 import * as path from "path";
 import * as findup from "findup-sync";
 
+import {arrayify, objectify} from "./utils";
+
+export interface IConfigFile {
```
adidahiya: let's be consistent about the abbreviation -- I think it should be "Configuration" everywhere, not "Config"

src/configuration.ts
```diff
@@ -124,8 +131,28 @@ export function loadConfigurationFromPath(configFilePath: string) {
...
+
+    combinedConfig.rules = {};
+    for (const name of Object.keys(objectify(baseConfig.rules))) {
+        combinedConfig.rules[name] = baseConfig.rules[name];
+    }
+    for (const name of Object.keys(objectify(config.rules))) {
+        combinedConfig.rules[name] = config.rules[name];
     }
+
+    return combinedConfig;
```
unional: I actually would suggest to base the return on current config and add base config if not available. i.e. `_.defaults(config, baseConfig)`.
This is more fragile as you need to change this function every time a new field is added.

src/tslint.ts
```diff
@@ -18,11 +18,13 @@
 import {IFormatter} from "./language/formatter/formatter";
 import {RuleFailure} from "./language/rule/rule";
 import {getSourceFile} from "./language/utils";
-import {findConfiguration, findConfigurationPath, getRelativePath, getRulesDirectories, loadConfigurationFromPath} from "./configuration";
+import {DEFAULT_CONFIG, findConfiguration, findConfigurationPath, getRelativePath,
+        getRulesDirectories, loadConfigurationFromPath, } from "./configuration";
```
adidahiya: expand to one import name per line please

src/tslint.ts
```diff
@@ -88,6 +91,15 @@ class Linter {
     private containsRule(rules: RuleFailure[], rule: RuleFailure) {
         return rules.some((r) => r.equals(rule));
     }
+
+    private computeFullOptions() {
+        let {configuration, rulesDirectory} = this.options;
+        if (configuration == null ) {
```
adidahiya: extra space before )

src/configuration.ts
adidahiya: :+1: `_.defaults` is well-suited for this, let's use that

JKillian: Wait:
* We don't depend on lodash AFAIK, do we want to add it?
* `_.defaults` won't concat arrays or anything, which means that things like rules directories from extended configs may not get added, which is bad.

> This is more fragile as you need to change this function every time a new field is added.

This is true, but theoretically we're not going to be adding new fields to `tslint.json` all that often...

src/configuration.ts
JKillian: So just leave this as is, right? Exported things are a blurry line between private/public, but I assume we're only treating as public things in the top level package

src/configuration.ts
unional: Agree that `tslint` should not depends on `lodash` (`tslint` already take quite a while to install :worried:).
I meant to suggest creating a copy of `config` and apply/merge `baseConfig` using the `_.defaults` principle.

src/configuration.ts
unional: Sorry to chime in, but I think it should be renamed is `IConfigurationFile` as this interface is useful to extensions.

src/configuration.ts
unional: Sorry. take the `_.defaults` idea back. Just realize I fall into the generalization trap. (slapping my hand).

Each property: `rules`, `rulesDirectory` and `extend` should be handled differently. Ordering in `rulesDirectory` doesn't matter, but it does in `extend`.

unional: I would suggest this:
```ts
export function extendConfigFile(config: IConfigFile, baseConfig: IConfigFile): IConfigFile {
    let combinedConfig: IConfigFile = JSON.parse(JSON.stringify(config));

    const baseRulesDirectory = arrayify(baseConfig.rulesDirectory);
    const configRulesDirectory = arrayify(config.rulesDirectory);
    combinedConfig.rulesDirectory = configRulesDirectory.concat(baseRulesDirectory);
    
    let rules = combinedConfig.rules || {};
    const baseRuleKeys = Object.keys(objectify(baseConfig.rules));
    for (const name of baseRuleKeys) {
        if (!rules.hasOwnProperty(name)) {
            rules[name] = baseConfig.rules[name];
        }
    }

    combinedConfig.rules = rules;

    return combinedConfig;
}
```

adidahiya: ah, sorry about that re: _.defaults. you're right @JKillian 

@unional those look like minor stylistic preferences... I'm fine with the code layout in this function as it is now. in the future, please leave line-by-line comments rather than rewriting the whole implementation... otherwise it gets pretty hard to review your proposed changes

src/configuration.ts
adidahiya: anything exported _is_ part of the public API; let's just assume people will use anything they can find in our generated typings

adidahiya:
overall lgtm :+1: only blocking issue is the naming convention in public API surface

src/configuration.ts
JKillian: Agreed, the tricky thing here is we already have public APIs like
'DEFAULT_CONFIG' before that don't obey this convention.

JKillian:
[Updated]

adidahiya:
:+1:

PR #1047, opened 2016-03-19
Update tslint.json to use extended semicolon rule format

ChrisPearce:
Update tslint.json file in root to use extended semicolon rule format
that includes “always”/“never” option


JKillian:
Thanks @ChrisPearce!

PR #1049, opened 2016-03-21
Add `node_module:` prefix for loading rules from a node module

JKillian:
Allows one to provide directories to custom rules living in a node module using the "node_module:" prefix:

```
{
  rulesDirectory: "node_module:tslint-eslint-rules/dist/rules"
  rules: { ... }
}
```

This is better than relative paths to node modules, which can behave funky depending on where the node module gets installed. Internally, this uses `require.resolve` to find the location of the node module, which means that it's using node's package finding algorithm, which is good.

### DIFF
--- a/package.json
+++ b/package.json
@@ -39,6 +39,7 @@
     "grunt-tslint": "latest",
     "mocha": "^2.2.5",
     "tslint": "latest",
+    "tslint-eslint-rules": "1.1.1",
     "typescript": "latest"
   },
   "peerDependencies": {
--- a/src/configuration.ts
+++ b/src/configuration.ts
@@ -186,7 +186,13 @@ export function getRelativePath(directory: string, relativeTo?: string): string 
  * @return An array of absolute paths to directories potentially containing rules
  */
 export function getRulesDirectories(directories: string | string[], relativeTo?: string): string[] {
-    const rulesDirectories = arrayify(directories).map((dir) => getRelativePath(dir, relativeTo));
+    const rulesDirectories = arrayify(directories).map((dir) => {
+        if (isNodeModulePath(dir)) {
+            return resolveNodeModulePath(dir);
+        } else {
+            return getRelativePath(dir, relativeTo);
+        }
+    });
 
     for (const directory of rulesDirectories) {
         if (!fs.existsSync(directory)) {
@@ -195,4 +201,19 @@ export function getRulesDirectories(directories: string | string[], relativeTo?:
     }
 
     return rulesDirectories;
+}
+
+function isNodeModulePath(nodeModulePath: string) {
+    return nodeModulePath.indexOf("node_module:") === 0;
+}
+
+function resolveNodeModulePath(nodeModulePath: string) {
+    const packagePath = nodeModulePath.split(":")[1];
+    const [packageName, ...insidePackagePathParts] = packagePath.split("/");
+
+    // we don't want the path to the .js module, we just want the path to its directory
+    const pathToPackage = path.dirname(require.resolve(packageName));
+    const insidePackagePath = insidePackagePathParts.join("/");
+
+    return path.join(pathToPackage, insidePackagePath);
 }
--- a/test/configurationTests.ts
+++ b/test/configurationTests.ts
@@ -14,7 +14,8 @@
  * limitations under the License.
  */
 
-import {IConfigFile, extendConfigFile} from "../src/configuration";
+import {IConfigFile, extendConfigFile, getRulesDirectories} from "../src/configuration";
+import {loadRules} from "./lint";
 
 describe("Configuration", () => {
     it("extendConfigFile", () => {
@@ -50,5 +51,11 @@ describe("Configuration", () => {
             },
             rulesDirectory: ["foo", "bar", "baz"],
         });
+    });
+
+    it("resolves rules directories with a node_module: prefix correctly", () => {
+        const resolvedDirectories = getRulesDirectories("node_module:tslint-eslint-rules/dist/rules");
+        const rules = loadRules({"valid-typeof": true}, {}, resolvedDirectories);
+        assert.strictEqual(rules.length, 1);
     });
 });


JKillian:
Tests will be failing for now until https://github.com/buzinas/tslint-eslint-rules/pull/59 merges and there's a new release of tslint-eslint-rules

unional:
Input related to https://github.com/palantir/tslint/pull/1016#issuecomment-199413583:

It depends on how do you envision the user use it. My opinion is to mirror the convention of `eslint`, i.e.: `tslint-config-ABC` or `ABC` or `plugin:XYZ`.
http://eslint.org/docs/user-guide/configuring#extending-configuration-files

I think that provides a better experience and consistency between the tools.
In their example it has relative path `./node_modules/...`. That would subject to the same issue unless they search up for the `node_modules` folder.

unional:
> That would subject to the same issue unless they search up for the node_modules folder.

That means this PR would be not necessary. My two cents. :smirk:

JKillian:
I like having the more explicit `node_module:` prefix (or some other name would work as well). The problem with things like "tslint-config/whatever" is that it seems like it's a relative path when in reality it's being treated differently behind the scene.

unional:
IMO I can't think of a semantic meaning to `node_module:` (or a different name):
* If it stands for "plugin" or "config", that's what "extends" is for.

While it is explicit, it seems to be extra noise:
* In the context of `extends`, user would not assume you `extends` something in a relative path fashion. It meant to be a package/library.
* In the context of `rulesDirectory`, user expect it to be a relative path.

I think what mentioned here is very clean already:
https://github.com/palantir/tslint/pull/1016#issuecomment-194686772

unional:
btw, It's just IMO, not trying to be harsh or anything. :smile:

unional:
> IMO I can't think of a semantic meaning to node_module: (or a different name):

A name that I can think of making a little bit of sense is `npm:`, noting it is a package/path from the npm ecosystem.

However, with that name, it suggests that you are planning to support others such as `bower:`, or `jspm:`.
Which, I think would be unnecessary.

JKillian:
@unional I see what you're saying, although I'm not sure if I agree haha. @adidahiya, what do you think?

adidahiya:
I prefer being explicit about the resolution logic as @JKillian suggests in https://github.com/palantir/tslint/pull/1016#issuecomment-199413583. The `node_module:` prefix looks to be the most transparent, let's go with that.

JKillian:
:+1: I'm going to merge this into my config related PR which this depends upon. We can still review/tweak things there if necessary

unional:
> @unional I see what you're saying, although I'm not sure if I agree haha.

Interested in understanding your reason.

This seems like a extra burden to the user :disappointed: 
```js
{
  "extends": "node_modules:tslint-config-unional",
  "rulesDirectory": "node_modules:tslint-eslint-rules/dist/rules"
}
```
in compare to:
```js
{
  "extends": [
    "tslint-config-unional",
    "tslint-eslint-rules"
  ]
}
```

adidahiya:
@unional how do you distinguish `"tslint-config-unional"` from a relative path? If we support relative paths for `rulesDirectory`, then why not for `extends` as well?

unional:
Because:
1. `extends` are meant for package. Just like in `npm` you won't `npm install ../anotherModule`, you `npm install anotherModule` or `npm link`
2. If you really want to have relative path, then `"./<relative path>"`.

`rulesDirectory` make sense to be relative path.

To see it in action, it would look like this:
```js
// tslint-eslint-rules/tslint.json
{
  "rulesDirectory": "./dist/rules",
  ...
}

// tsling-config-unional/tslint.json
{
  "extends": "tslint-eslint-rules",
  "rulesDirectory": "./dist/moreRules",
  ...
}

// app1/tslint.json
{
  "extends": "tslint-config-unional",
  "rulesDirectory": "./myCustomRules"
}

// app2/tslint.json
{
  "extends": [
    "tslint-config-typings"
    "tslint-config-airbnb"
  ]
}
```

unional:
One more thing regarding `node_modules:` resolution: the user should not be aware, concern, or use certain syntax due to **a historic** reason. By exposing this mechanism, you are forcing new user (of npm 3+) to learn about this historic reason that they 100% don't need to know.

This is especially true since this is `tslint`, not `npm`. There is no need to carry that over.

Cheers :beers:

blakeembrey:
@JKillian @adidahiya I just got linked to this, why would you add a proprietary syntax for resolution? It seems really confusing... Why wouldn't it just use `require.resolve` semantics like the rest of the node ecosystem?

JKillian:
@blakeembrey Right now, the only way you can specify directories that additional rules are in is by a file path, either absolute or relative. This is obviously not at all ideal for rules that are contained in a node module, as file paths can get a little wonky with NPM as I'm sure you well know.

So we obviously need some sort of way to supply a path that references a directory in a node module. The choices discussed then are:

* What this PR suggested: use `require.resolve` (node module) resolution for paths prefixed with `node_module:` and path resolution otherwise. It works, it's explicit, but it does require an arbitrary syntax.
* What @unional suggested above: only support relative paths for `rulesDirectory` but support node module resolution for an `extends` feature. I think this is a little less flexible than we would prefer.
* What you implied: use path resolution for absolute paths or paths starting with a `./`, and use node module resolution otherwise. The issue with this is that you end up with strange situations like this: `tslint -r somePath/to/rules file/to/lint.ts` tries to look for a node module named "somePath" with a "to/rules" directory inside whereas `tslint -r ./somePath/to/rules file/to/lint.ts` does what's expected. (Although this behavior might be a little less unexpected in a `tslint.json` file.) This would also break things for people who haven't prefixed a relative path with a "." currently, although I think we could space this change out over a release or two and provide warnings / error messages that would make things easy enough to correct.
* A variation of the above two options - only use node module resolution if the path starts with a magic package prefix, like "tslint-rules-", otherwise, use path resolution.

I was leaning towards the first one obviously, but am definitely open to being convinced otherwise

unional:
> The issue with this is that you end up with strange situations like this:

My suggestion would be do not rely on module resolution for `rulesDirectory`. If someone use the cli, that means he/she is the end user. The end user should use relative path and knows relative path anyway.
e.g.: `ts-node node_modules/tape/bin/tape src/**/*.spec.ts`

By the way, the name actually suggests it. It is a "rules directory", not a "rules package".

It initial cause for this PR (I assume) is the nesting issue I mentioned in the `extends` feature, which only occurs when **the extension author** writes `"rulesDirectory": "node_module/tslint-eslint-rules/dist/rules"`. 
The end user can still write  `"rulesDirectory": "node_module/tslint-eslint-rules/dist/rules"` and things would be resolved correctly.

unional:
> This is obviously not at all ideal for rules that are contained in a node module, as file paths can get a little wonky with NPM as I'm sure you well know.

```js
{
  "extends": "that-package"
}
```

adidahiya:
@unional's overall suggestion / @JKillian's option 2 in https://github.com/palantir/tslint/pull/1049#issuecomment-200157349 is probably the cleanest. The only limitation here is that authors of custom rule modules will have to add a `tslint.json` at the root of their module for users to get `extends` working properly, but this isn't a breaking change -- users have no clean way to reference rules in node_modules currently. So I think we should take these suggestions, change the behavior to be more prescriptive, and remove the `node_module:` syntax.

Locking `extends` into node module resolution still allows for a good amount of flexibility. Let's say two users wanted to have different configs for the rules in [tslint-microsoft-contrib](https://github.com/Microsoft/tslint-microsoft-contrib/).

- `my-config-a` module would have `tslint.json`:

  ```
  {
    "extends": "tslint-microsoft-contrib",
    "rules": {
      // explicitly enable / disable the ones you want
    }
  }
  ```

- `my-config-b` module could have:

  ```
  {
    "extends": "tslint-microsoft-contrib",
    "rulesDirectory": "dist/rules/", // just local to this module
    "rules": {
      // explicitly enable / disable the ones you want
    }
  }
  ```

Thanks for the input @unional and @blakeembrey :rose: 

also tagging @HamletDRC for any thoughts

unional:
> The only limitation here is that authors of custom rule modules will have to add a tslint.json at the root of their module for users to get extends working properly

You don't have to. Package author can do: `package.json/main: "<path>"`. That's what I did.

unional:
https://github.com/typings/tslint-config-typings/blob/master/package.json#L5

unional:
Ar, but yes. Need to have `tslint.json` somewhere.

unional:
https://github.com/airbnb/javascript/blob/master/packages/eslint-config-airbnb/package.json#L5
https://github.com/airbnb/javascript/blob/master/packages/eslint-config-airbnb/index.js

JKillian:
Alright, let's do it this way then. As far as "extend", we can just have it use `require.resolve` / `require` directly and not have to deal with any weird resolution logic on our end (since luckily you can `require` JSON files). I don't think we need to look for any magic prefix or anything, let's go with what @blakeembrey suggested and use straight up require. 

As far as "rulesDirectory", since we're pointing to a directory and not a .js or .json file, requiring things is a bit less intuitive and makes less sense. So in this case, we'll simply treat whatever is passed as a path and handle it as such. The only weird part about this is that it's a little strange that the behavior is different here than with extends. Maybe it makes sense to handle them both using require semantics as @blakeembrey suggested?

adidahiya:
let's keep `rulesDirectory` as it is and assume it is a straight up relative path. we may reconsider later and introduce another option with different semantics. the current name "rules directory" is already a slight misnomer since you can pass an array of paths.

JKillian:
I essentially "unmerged" this from the feature branch I had merged it into

Issue #602, opened 2015-08-24
add allow-undefined-check option to triple-equals rule


vmmitev:
The official [TypeScript compiler coding guidelines](https://github.com/Microsoft/TypeScript/wiki/Coding-guidelines) state:

[Use **`undefined`**, do not use `null`](https://github.com/Microsoft/TypeScript/wiki/Coding-guidelines#null-and-undefined)

Wouldn't it be beneficial to have an additional `allow-undefined-check` option for the `triple-equals` rule for developers that want to adhere to the above guidelines?

adidahiya:
Sure, that sounds like a reasonable feature request. Just want to point out, though -- those coding guidelines are for the compiler source code, not all TypeScript projects in general.

vmmitev:
That's absolutely right, I edited the original comment to reflect the nature of the URL more clearly.

It's worth noting, though, that this is the article most likely to be the first result in search engines when looking up terms like `typescript` + `code style` or `coding standards/convention`. At least that was the case for me and I was set on that path ever since.

JKillian:
If you only are using `undefined` in your code, you could just use strict equality, right? Maybe I'm misunderstanding the use-case for this feature, could you elaborate just a little @vmmitev?

vmmitev:
The only feasible use-case would be to support code-style standards where using the `null` keyword is not allowed. JavaScript is a bit confusing when it comes to `null`-like values where you have 3 possible null states (one of which results in `TypeError`). Omitting one possibility may result in better to understand code. If you're heavily working with JSON you may go in favour of `null` since that would be the prevalent default value, but unless explicitly set, in all other cases, it's `undefined`.

Sadly I can't back up this use-case with any statistical data (of how often teams decide to drop either `null` or `undefined`) so I'll leave it up to you to decide if it makes sense to add this slight flexibility option. If you do see value, I can prep up a pull request with the necessary changes.

adidahiya:
Are any values besides `undefined` coerced to `== undefined` when you use double-equals like that? If not, then I don't think we should implement this rule option.

vmmitev:
As the person that brought up this change I also think that in the end it might be unnecessary. I hope this makes it easier for you to make the decision.

adidahiya:
Alright; i'm going to close this and leave it up to custom rules for users who want it.

joerixaop:
[...]

PR #1056, opened 2016-03-24
allow-undefined-check option for tripleEqualsRule

patsissons:
As discussed in #602, a basic PR to enable the undefined double equals checks.

Two notes about this PR:

1. I wasn't sure what the most appropriate way to check if an expression was undefined, this was the best I could come up with quickly. I'm open to suggestions on a better strategy (if one exists).
2. I updated the sample and default `tslint` configs. Not sure if that is desired or not. If not let me know and i'll purge that last commit from this PR.


test/rules/triple-equals/tslint.json
```diff
@@ -1,5 +1,5 @@
 {
   "rules": {
-    "triple-equals": [true, "allow-null-check"]
+    "triple-equals": [true, "allow-null-check", "allow-undefined-check"]
```
JKillian: Ideally in these tests, it would be better to make a separate test case for this option. See how the `semicolon` rule has its tests structured for example.

src/rules/tripleEqualsRule.ts
```diff
@@ -19,6 +19,11 @@ import * as ts from "typescript";
 import * as Lint from "../lint";
 
 const OPTION_ALLOW_NULL_CHECK = "allow-null-check";
+const OPTION_ALLOW_UNDEFINED_CHECK = "allow-undefined-check";
+
+function isUndefinedExpression(expression: ts.Expression) {
+    return expression.kind === ts.SyntaxKind.Identifier && expression.getText() === "undefined";
```
JKillian: Looks good to me, I don't know of any better way to do this

src/configuration.ts
```diff
@@ -33,7 +33,7 @@ export const DEFAULT_CONFIG = {
         "one-line": [true, "check-open-brace", "check-whitespace"],
         "quotemark": [true, "double"],
         "semicolon": [true, "always"],
-        "triple-equals": [true, "allow-null-check"],
+        "triple-equals": [true, "allow-null-check", "allow-undefined-check"],
```
JKillian: Same here, let's just keep "allow-null-check" for now

docs/sample.tslint.json
```diff
@@ -96,7 +96,8 @@
     ],
     "triple-equals": [
       true,
-      "allow-null-check"
+      "allow-null-check",
+      "allow-undefined-check"
```
JKillian: Let's just keep "allow-null-check" in here for now - better to be using one or the other in

src/rules/tripleEqualsRule.ts
```diff
@@ -55,7 +60,10 @@ class ComparisonWalker extends Lint.RuleWalker {
     private isExpressionAllowed(node: ts.BinaryExpression) {
         const nullKeyword = ts.SyntaxKind.NullKeyword;
 
-        return this.hasOption(OPTION_ALLOW_NULL_CHECK)
-            && (node.left.kind ===  nullKeyword || node.right.kind === nullKeyword);
+        return (
+            this.hasOption(OPTION_ALLOW_NULL_CHECK) && (node.left.kind ===  nullKeyword || node.right.kind === nullKeyword)
```
JKillian: Seems like there's an extra space here: `===  nullKeyword`. Looks like it was here before too, but if you're tweaking this file, feel free to remove it :)

test/rules/triple-equals/test.ts.lint
```diff
@@ -13,3 +13,6 @@ function testFunction() {
 
 // allow double-equality null checks
 var x = (y == null) ? 1 : 2;
+
+// allow double-equality undefined checks
+var x = (y == undefined) ? 1 : 2;
```
JKillian: New line at the end of these files

JKillian:
Looks solid to me @patsissons, thanks! I left a few minor tweaks for you to make and then we'll merge this

patsissons:
Thanks everyone, I'll try and get to these revisions this morning.

JKillian:
Thanks @patsissons - merged!

PR #1060, opened 2016-03-25
Implements "no-triple-equals-null" rule. Closes #923.

### LINKED ISSUE #923: New rule to warn against triple equals null
Almost always I want to compare against `null` and `undefined`, not just `null`. So `=== null` is probably an error.

This rule should suggest to the user, `Did you mean == null instead?`

martellaj:
_:baby: My first contribution to tslint!_

**Summary**
- Work item taken from issue #923. 
- Implements "no-triple-equals-null" rule.
- Adds tests for "no-triple-equals-null" rule.
- Documents "no-triple-equals-null" rule in README.

### DIFF
--- a/src/rules/noTripleEqualsNullRule.ts
+++ b/src/rules/noTripleEqualsNullRule.ts
@@ -1,0 +1,67 @@
+/**
+ * @license
+ * Copyright 2013 Palantir Technologies, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import * as ts from "typescript";
+import * as Lint from "../lint";
+
+const OPTION_ALLOW_NULL_CHECK = "no-triple-equals-null";
+
+export class Rule extends Lint.Rules.AbstractRule {
+    public static EQ_FAILURE_STRING = "Did you mean == null instead?";
+    public static NEQ_FAILURE_STRING = "Did you mean != null instead?";
+
+    public apply(sourceFile: ts.SourceFile): Lint.RuleFailure[] {
+        const noTripleEqualsNullWalker = new NoTripleEqualsNullWalker(sourceFile, this.getOptions());
+        return this.applyWithWalker(noTripleEqualsNullWalker);
+    }
+}
+
+class NoTripleEqualsNullWalker extends Lint.RuleWalker {
+    public visitBinaryExpression(node: ts.BinaryExpression) {
+        if (!this.isExpressionAllowed(node)) {
+            const position = node.getChildAt(1).getStart();
+            const expressionWidth = node.right.getFullWidth() + 3;
+            this.handleBinaryComparison(position, expressionWidth, node.operatorToken.kind, node.right.kind);
+        }
+
+        super.visitBinaryExpression(node);
+    }
+
+    private handleBinaryComparison(position: number, expressionWidth: number, operator: ts.SyntaxKind, right: ts.SyntaxKind) {
+        switch (operator) {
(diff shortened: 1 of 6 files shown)


src/rules/noTripleEqualsNullRule.ts
```diff
@@ -0,0 +1,67 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import * as ts from "typescript";
+import * as Lint from "../lint";
+
+const OPTION_NO_NULL_KEYWORD = "no-null-keyword";
```
JKillian: Hmm, this option is confusing, passing this option to the rule seems to make the rule not do anything? Perhaps we should have two options: "allow-null" and "allow-undefined". By default the rule would disallow `=== null`, `!== null`,  `=== undefined`, and `!== undefined`, but if you passed these options it would allow the corresponding constructs.

src/rules/noTripleEqualsNullRule.ts
```diff
@@ -0,0 +1,67 @@
...
+    public apply(sourceFile: ts.SourceFile): Lint.RuleFailure[] {
+        const noTripleEqualsNullWalker = new NoTripleEqualsNullWalker(sourceFile, this.getOptions());
+        return this.applyWithWalker(noTripleEqualsNullWalker);
+    }
+}
+
+class NoTripleEqualsNullWalker extends Lint.RuleWalker {
+    public visitBinaryExpression(node: ts.BinaryExpression) {
+        if (this.isExpressionAllowed(node)) {
+            const position = node.getChildAt(1).getStart();
```
JKillian: maybe we should use `node.operatorToken` here for clarity?

src/rules/noTripleEqualsNullRule.ts
```diff
@@ -0,0 +1,67 @@
...
+            const expressionWidth = node.right.getFullWidth() + 3;
```
JKillian: Probable better to do something like: `const expressionWidth = node.right.getEnd() - position;` in case there's extra whitespace lying around.

src/rules/noTripleEqualsNullRule.ts
```diff
@@ -0,0 +1,67 @@
...
+                if (right === ts.SyntaxKind.NullKeyword) {
+                    this.addFailure(this.createFailure(position, expressionWidth, Rule.NEQ_FAILURE_STRING));
+                }
+                break;
+            default:
+                break;
+        }
+    }
+
+    private isExpressionAllowed(node: ts.BinaryExpression) {
```
JKillian: it seems that this method is somewhat redundant with `handleBinaryComparison`, maybe there's a way to refactor this?

test/rules/triple-equals-null/test.ts.lint
```diff
@@ -0,0 +1,50 @@
...
+if (x !== null) {
+      ~~~~~~~~          [Did you mean != null instead?]
+}
+
+/**
+ * More complex tests
+ */
+ 
+ x === null ? true : false;
+   ~~~~~~~~             [Did you mean == null instead?]
```
JKillian: add a newline at the end of files, just helps keep all files the same

test/rules/triple-equals-null/tslint.json
```diff
@@ -0,0 +1,6 @@
+{
+  "rules": {
+    "no-triple-equals-null": true,
+    "no-null-keyword": false
```
JKillian: we can just enable `no-triple-equals-null` here

src/rules/noTripleEqualsNullRule.ts
JKillian: @adidahiya What are you thinking API-wise for this rule?

JKillian:
Thanks @martellaj! This is looking really good! Overall left a few comments/notes. The most important thing to get figured out here is my first comment regarding the rule and its options, once we have a good plan there, everything else should be really easy to tweak

src/rules/noTripleEqualsNullRule.ts
adidahiya: Let's split this up into two rules, `no-triple-equals-null` and `no-triple-equals-undefined`, with no options on either of them. Right now we have no way of communicating between rules so this seems like the most flexible path forward for users.

adidahiya:
ping @martellaj

ash211:
Ping @martellaj

PR #1070, opened 2016-03-28
Clarify the "unsorted key" failure message

chrismbarr:
Adding TSLint to our projects has been a huge benefit, but nearly every developer who has encountered the `unsorted key 'a'` message has come to me and asked "I understood how to fix the other errors I saw, but I have no idea what this one means."  

I think just saying "unsorted key" is not enough information to give someone who has never encountered this error before.  So i'm proposing just a small text change to help clarify this.

**Before** :-1: 
>unsorted key 'a'

**After** :+1: 
>The key 'a' is not sorted alphabetically


adidahiya:
Nice, thanks @chrismbarr! This is definitely an improvement.

PR #1062, opened 2016-03-25
Added --force (-n) option. Fixes #1059

sasidhar:
Fixes #1059

Issue #1059, opened 2016-03-25
--force option for CLI


adidahiya:
`tslint --force` should return a status code of 0 even if there are lint failures.

follow-up from https://github.com/palantir/tslint/issues/1053

sasidhar:
Ignore above pull request. It was made on some old forked repo.

PR #1073, opened 2016-04-01
No alias for --force

JKillian:
Related: #1059 and #1062


adidahiya:
nice, thx for adding a test. can we send a PR for the optimist typings?

Issue #1076, opened 2016-04-01
Feature Request: Don't require semi-colons after bound-method declarations.


AndyMoreland:
#### TypeScript code being linted

```ts
private handleFoo = (newState: boolean) => {
    this.setState({ banana: newState });
}
```
#### Actual behavior

When I have the "semicolon" rule on, tslint complains of a missing semi-colon after the closing `}` in this method definition.

#### Expected behavior

I want a way to make the semi-colon after the closing `}` optional. I view this construct as a way of defining a /method/, not a /property/, and methods don't close with `;`.

adidahiya:
We could change the `semicolon` rule to start being lenient here, or add a new `allow-member-functions` to enable leniency. My slight preference is for the former. Thoughts @JKillian?

AndyMoreland:
I've pushed a PR that addresses this, I think. If the semi-colon rule is set to "enabled", it'll be lenient. When it's "always", it'll require them.

PR #1081, opened 2016-04-04
Make semi-colon rule laxer with bound-method properties

AndyMoreland:
Methods declared like:

```ts
class Foo {
      bar = () => { }
}
```

shouldn't be marked as requiring a trailing semi-colon for the closing `}`

Addresses: #1076

### DIFF
--- a/src/rules/semicolonRule.ts
+++ b/src/rules/semicolonRule.ts
@@ -83,8 +83,8 @@ class SemicolonWalker extends Lint.RuleWalker {
 
     public visitPropertyDeclaration(node: ts.PropertyDeclaration) {
         const initializer = node.initializer;
-        if (this.hasOption(OPTION_ALWAYS) || this.hasOption(OPTION_NEVER) ||
-            !(initializer && initializer.kind === ts.SyntaxKind.ArrowFunction)) {
+        /* ALWAYS === "enabled" for this rule. */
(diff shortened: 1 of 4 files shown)


adidahiya:
Huh, I just now realized that the semicolon rule has three states, enabled / always / never. We should improve the docs around this, but this PR looks good to me :+1:

JKillian:
My hesitation here: this rule currently only has two states, always & never. (if you provide neither option, it behaves as if you said always). I'm hesitant to introduce a third state to the rule without more justification, but if you guys think it's worth it, I'll listen

adidahiya:
Uh, yeah, after posting my comment I changed my mind because of the same hesitation. Having `true` and `[true, "always"]` behave differently could be confusing. Do ESLint or stylelint have any rule configs that work like that?

Alternatives:

1. always be lenient about bound-method properties
2. introduce a new option to be lenient on them, as suggested in #1076

AndyMoreland:
I didn't intentionally add a third state. It seemed like it already existed. I thought I saw an issue somewhere where it was discussed but I can't find it anymore so it's possible I hallucinated it entirely. I'm happy to implement either (1) or (2).

AndyMoreland:
I believe I've addressed the comments here by going with option (1) from above.

adidahiya:
👍

PR #1084, opened 2016-04-05
Treat all -c option CLI values as a path

JKillian:


adidahiya:
so the idea here is that with the CLI, a user does not ever expect the `-c` argument to be resolved as a node module? makes sense, at least for now.

JKillian:
Yep, and in fact, it wouldn't even have worked if the node module did exist because there was a check to see if `fs.exists(argv.c)`

adidahiya:
:+1:

PR #1089, opened 2016-04-05
CircleCI > Travis

adidahiya:


adidahiya:
Travis failed because I killed .travis.yml and it can't infer the build config. And appveyor takes at least 10 minutes every time to queue a build...

PR #1094, opened 2016-04-06
JS directives are not unused expressions

### LINKED ISSUE #1050: no-unused-expression rule doesn't handle directives
### Bug Report

- __TSLint version__: `3.6.0`
- __TypeScript version__: `1.8.9`
- __Running TSLint via__: CLI , but doesn't actually matter

```ts
'ngInject';
console.log('abc');
'use strict';
```
with tslint.json
```json
{
  "rules": {
    "no-unused-expression": true
  }
}
```

#### Actual behavior

```
1.ts[1, 1]: expected an assignment or function call
```
1. Directive 'ngInject'; triggers no-unused-expression violation
2. "no-op" expression 'use strict'; does not trigger no-unused-expression violation

#### Expected behavior

The no-unused-expression rule does not handle directive prologues at all. It only has a hard coded detection of 'use strict' expression in any position.

1. Directive prologue should be handled and any directive in directive prologue position should not trigger the violation (i.e. there should be no error on 'ngInject' in the example code).
2. 'use strict' must not be hardcoded and allowed in just any position (i.e. there should be an error on 'use strict' in the example code).

'ngInject' directive is just an example. Any directive should work in directive prologue position.

This "reopens" #423 which was valid but maybe not explained well enough.

For reference, the directive prologue definition from [ECMAScript 5.1](http://ecma-international.org/ecma-262/5.1/#sec-14.1)
 
> 14.1 Directive Prologues and the Use Strict Directive
> 
> A Directive Prologue is the longest sequence of ExpressionStatement productions occurring as the initial SourceElement productions of a Program or FunctionBody and where each ExpressionStatement in the sequence consists entirely of a StringLiteral token followed a semicolon. The semicolon may appear explicitly or may be inserted by automatic semicolon insertion. A Directive Prologue may be an empty sequence.

JKillian:
Fixes #1050

### DIFF
--- a/src/rules/noUnusedExpressionRule.ts
+++ b/src/rules/noUnusedExpressionRule.ts
@@ -43,10 +43,9 @@ class NoUnusedExpressionWalker extends Lint.RuleWalker {
             const isValidStandaloneExpression = kind === ts.SyntaxKind.DeleteExpression
                 || kind === ts.SyntaxKind.YieldExpression
                 || kind === ts.SyntaxKind.AwaitExpression;
-            const isValidStringExpression = kind === ts.SyntaxKind.StringLiteral
-                && (expression.getText() === '"use strict"' || expression.getText() === "'use strict'");
+            const isValidDirective = isDirective(node);
 
-            if (!isValidStandaloneExpression && !isValidStringExpression) {
+            if (!isValidStandaloneExpression && !isValidDirective) {
                 this.addFailure(this.createFailure(node.getStart(), node.getWidth(), Rule.FAILURE_STRING));
             }
         }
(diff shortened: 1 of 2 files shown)


src/rules/noUnusedExpressionRule.ts
```diff
@@ -43,10 +43,9 @@ class NoUnusedExpressionWalker extends Lint.RuleWalker {
             const isValidStandaloneExpression = kind === ts.SyntaxKind.DeleteExpression
                 || kind === ts.SyntaxKind.YieldExpression
                 || kind === ts.SyntaxKind.AwaitExpression;
-            const isValidStringExpression = kind === ts.SyntaxKind.StringLiteral
-                && (expression.getText() === '"use strict"' || expression.getText() === "'use strict'");
+            const isValidDirective = isDirective(node);
 
-            if (!isValidStandaloneExpression && !isValidStringExpression) {
+            if (!isValidStandaloneExpression && !isValidDirective) {
```
adidahiya: nit: I would just inline the call to `isDirective(node)`

adidahiya:
:+1:

PR #1098, opened 2016-04-08
Fix false positive from no-internal-module on global augmentation

### LINKED ISSUE #1069: false negative: no-internal-module
https://github.com/typed-typings/npm-chai/blob/master/lib/Chai.d.ts#L28-L33
```ts
/* tslint:disable:no-internal-module */
declare global {
  interface Object {
    should: A.Assertion;
  }
}
```
Global augmentation is detected as internal module.

JKillian:
Fixes #1069


adidahiya:
👍

PR #1105, opened 2016-04-11
Add "no-invalid-this" rule

AndyMoreland:
This PR adds a rule that flags usage of the `this` keyword outside of class-body scopes.

Often I find myself refactoring code that looks like:

```ts
class Foo {
    private aMethod() {
        doWorkOn(this.someState.x);
    }
}
```

to pull `aMethod` out into a function that looks like:

```ts
function(someState) {
    doWorkOn(someState.x);
}
```

When I make this change by hand I find that I screw up a lot and accidentally write:

```ts
function(someState) {
    doWorkOn(this.someState.x); /* note `this` */
}
```

which does not cause a compile error because typescript unhelpfully infers `this` with type `any`.

I've found that `this` outside of a `class` in code that is using ES6-class syntax is almost always a mistake. I wouldn't turn this on by default for all users of tslint, but it'd be very useful in my case (and I imagine for others.)

### DIFF
--- a/src/language/walker/scopeAwareRuleWalker.ts
+++ b/src/language/walker/scopeAwareRuleWalker.ts
@@ -25,10 +25,10 @@ export abstract class ScopeAwareRuleWalker<T> extends RuleWalker {
         super(sourceFile, options);
 
         // initialize stack with global scope
-        this.scopeStack = [this.createScope()];
+        this.scopeStack = [this.createScope(sourceFile)];
     }
 
-    public abstract createScope(): T;
+    public abstract createScope(node: ts.Node): T;
 
     public getCurrentScope(): T {
         return this.scopeStack[this.scopeStack.length - 1];
@@ -57,7 +57,7 @@ export abstract class ScopeAwareRuleWalker<T> extends RuleWalker {
         const isNewScope = this.isScopeBoundary(node);
 
         if (isNewScope) {
-            this.scopeStack.push(this.createScope());
+            this.scopeStack.push(this.createScope(node));
         }
 
         this.onScopeStart();
--- a/src/language/walker/scopeAwareRuleWalker.ts
+++ b/src/language/walker/scopeAwareRuleWalker.ts
@@ -80,6 +80,7 @@ export abstract class ScopeAwareRuleWalker<T> extends RuleWalker {
             || node.kind === ts.SyntaxKind.ArrowFunction
             || node.kind === ts.SyntaxKind.ParenthesizedExpression
             || node.kind === ts.SyntaxKind.ClassDeclaration
+            || node.kind === ts.SyntaxKind.ClassExpression
             || node.kind === ts.SyntaxKind.InterfaceDeclaration
             || node.kind === ts.SyntaxKind.GetAccessor
             || node.kind === ts.SyntaxKind.SetAccessor;
--- a/src/rules/noInvalidThisRule.ts
+++ b/src/rules/noInvalidThisRule.ts
@@ -1,0 +1,51 @@
+/**
+ * @license
+ * Copyright 2016 Palantir Technologies, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import * as ts from "typescript";
+import * as Lint from "../lint";
+
+export class Rule extends Lint.Rules.AbstractRule {
+    public static FAILURE_STRING = "the \"this\" keyword is disallowed outside of a class body" ;
+
+    public apply(sourceFile: ts.SourceFile): Lint.RuleFailure[] {
+        return this.applyWithWalker(new NoInvalidThisWalker(sourceFile, this.getOptions()));
+    }
+}
+
+class NoInvalidThisWalker extends Lint.ScopeAwareRuleWalker<{inClass: boolean}> {
+    public createScope(node: ts.Node): { inClass: boolean } {
+        const isClassScope = node.kind === ts.SyntaxKind.ClassDeclaration || node.kind === ts.SyntaxKind.ClassExpression;
+
+        return {
+            inClass: isClassScope,
+        };
+    }
+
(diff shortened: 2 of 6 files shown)


src/language/walker/scopeAwareRuleWalker.ts
```diff
@@ -80,6 +80,7 @@ export abstract class ScopeAwareRuleWalker<T> extends RuleWalker {
             || node.kind === ts.SyntaxKind.ArrowFunction
             || node.kind === ts.SyntaxKind.ParenthesizedExpression
             || node.kind === ts.SyntaxKind.ClassDeclaration
+            || node.kind === ts.SyntaxKind.ClassExpression
```
AndyMoreland: I believe this is correct but would appreciate a second set of eyes here.

AndyMoreland: This is for things like:

```
const A = class {
...
}
```

JKillian: 👍 Looks right to me

src/rules/unboundFunctionThisRule.ts
```diff
@@ -0,0 +1,51 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import * as ts from "typescript";
+import * as Lint from "../lint";
+
+export class Rule extends Lint.Rules.AbstractRule {
+    public static FAILURE_STRING = "disallowed \"this\" outside of class body" ;
```
JKillian: Can we make this a little clearer? Maybe: `the "this" keyword is disallowed outside of a class body`

JKillian:
Looks good to me! Although I'm not sure how I feel about the rule name, any thoughts @adidahiya? Also,  tweaking the error message would be great

src/rules/unboundFunctionThisRule.ts
```diff
@@ -0,0 +1,51 @@
...
+    public createScope(node: ts.Node): { inClass: boolean } {
+        let isClassScope = node.kind === ts.SyntaxKind.ClassDeclaration || node.kind === ts.SyntaxKind.ClassExpression;
+
+        return {
+            inClass: isClassScope,
+        };
+    }
+
+    protected validateThisKeyword(node: ts.Node) {
+        if (!this.getAllScopes().some(scope => scope.inClass)) {
```
JKillian: Normally as a style guideline we surround arrow func arguments with parens:
`(scope) => scope.inClass`. Just makes them easier to distinguish visually. This should be a lint rule!

adidahiya:
I like `no-unbound-this` for the rule name

adidahiya:
oh, actually it's called `no-invalid-this` in ESLint... we should probably be consistent? http://eslint.org/docs/rules/no-invalid-this

JKillian:
Let's go with that: `no-invalid-this`

myitcv:
Just a quick question on what we're looking to achieve here: aren't we ultimately trying to catch situations where the type of `this` is `any`?

Therefore I'm not totally clear how scopes etc comes into it.... other than, in the absence of full type information (#680), we're trying to look at scopes and infer that the type of `this` would be `any`.

Does that make sense?

adidahiya:
@myitcv I'd consider this one of the stricter rules that I don't expect all TSLint users to enable. Yes, I would love to have full type information and check if `this` is `any`, but banning `this` outside `class` scopes is a decent analog that we can accomplish syntactically.

AndyMoreland:
@myitcv when/if tslint gives me type info I'd be happy to write that rule.

AndyMoreland:
I believe I've addressed the comments. I renamed to no-invalid-this and fixed the `scope =>` to be `(scope) =>`.

adidahiya:
yeah we can merge this / refine the error message afterwards

src/tsconfig.json
```diff
@@ -113,6 +113,7 @@
         "rules/tripleEqualsRule.ts",
         "rules/typedefRule.ts",
         "rules/typedefWhitespaceRule.ts",
+        "rules/unboundFunctionThisRule.ts",
```
JKillian: seems like the wrong file name here. If you run grunt locally it should automatically fix this

src/rules/noInvalidThisRule.ts
```diff
@@ -0,0 +1,51 @@
...
+
+export class Rule extends Lint.Rules.AbstractRule {
+    public static FAILURE_STRING = "the \"this\" keyword is disallowed outside of a class body" ;
+
+    public apply(sourceFile: ts.SourceFile): Lint.RuleFailure[] {
+        return this.applyWithWalker(new UnboundFunctionThisWalker(sourceFile, this.getOptions()));
+    }
+}
+
+class UnboundFunctionThisWalker extends Lint.ScopeAwareRuleWalker<{inClass: boolean}> {
```
JKillian: Class name is also not updated

test/tsconfig.json
```diff
@@ -141,4 +142,4 @@
         "rule-tester/testData.ts",
         "rule-tester/utilsTests.ts"
     ]
-}
+}
```
JKillian: no reason to remove this newline 😉

src/rules/noInvalidThisRule.ts
```diff
@@ -0,0 +1,51 @@
...
+    public createScope(node: ts.Node): { inClass: boolean } {
+        let isClassScope = node.kind === ts.SyntaxKind.ClassDeclaration || node.kind === ts.SyntaxKind.ClassExpression;
```
JKillian: can be a const

src/rules/noInvalidThisRule.ts
```diff
@@ -0,0 +1,51 @@
+/**
+ * @license
+ * Copyright 2013 Palantir Technologies, Inc.
```
JKillian: Can be 2016

JKillian:
Still needs a few tweaks, then it should be good to merge

Note to self: add documentation / tweak error message after this merges

AndyMoreland:
I've made the four changes you requested. Sorry for the sloppiness!

JKillian:
Sweet, merged! Release should be sometime this week, Thursday at the latest

AndyMoreland:
Thanks!

aaronresnick:
> @myitcv when/if tslint gives me type info I'd be happy to write that rule.

@AndyMoreland I believe writing a rule that checks for `this` usage only when its type is any is possible now, right? If so, that would be super helpful for me. Are you still interested in writing it?

dugagjinll:
```````

goktugyil:
In mocha, we need to use this.skip outside the scope of classes: http://prntscr.com/ns01dr
This gives me a warning. Is there a way to disable cases like these? https://mochajs.org/

PR #1107, opened 2016-04-11
Added "allow-pascal-case" to variable-name-rule

### LINKED ISSUE #1079: Allow PascalCase for "variable-name" rule
It would be nice to have this in addition to camelCase and UPPER_CASE. 

It would also be nice if you could assign me to implement it :)

JoshuaKGoldberg:
Fixes #1079

### DIFF
--- a/src/rules/variableNameRule.ts
+++ b/src/rules/variableNameRule.ts
@@ -24,6 +24,7 @@ const OPTION_LEADING_UNDERSCORE = "allow-leading-underscore";
 const OPTION_TRAILING_UNDERSCORE = "allow-trailing-underscore";
 const OPTION_BAN_KEYWORDS = "ban-keywords";
 const OPTION_CHECK_FORMAT = "check-format";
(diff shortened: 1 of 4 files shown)


src/rules/variableNameRule.ts
```diff
@@ -120,10 +121,21 @@ class VariableNameWalker extends Lint.RuleWalker {
         if (!this.hasOption(OPTION_TRAILING_UNDERSCORE) && lastCharacter === "_") {
             return false;
         }
-        return firstCharacter === firstCharacter.toLowerCase() && middle.indexOf("_") === -1;
+        if (this.hasOption(OPTION_ALLOW_PASCAL_CASE)) {
+            if (!isLowerCase(firstCharacter) && !isUpperCase(firstCharacter)) {
```
JKillian: What is being checked for here?

JoshuaKGoldberg: You know what, you're totally right. I'll remove it.

JKillian:
Looks good, thanks for the PR!

PR #1109, opened 2016-04-11
Removed no-duplicate-key rule

### LINKED ISSUE #885: Deprecate `no-duplicate-key` rule (covered by compiler)
The `no-duplicate-key` rule is covered by the compiler and thus unnecessary. For example, the following code will not compile:

```ts
const bar = {
    foo: 'baz',
    foo: 'baz'
}
```

JoshuaKGoldberg:
Fixes #885.


JKillian:
Thanks @JoshuaKGoldberg! Look good, although we probably won't merge this until we're ready for TSLint 4.0

JKillian:
We're working towards a 4.0 release now, so if you'd like to update this it'll be mergeable after

PR #1117, opened 2016-04-12
"extends" tests now use code inside this project

JKillian:
This helps avoid weird dependency bugs that cause builds to fail


adidahiya:
👍

PR #1119, opened 2016-04-12
Prepare release v3.7.0-dev.5

JKillian:


JKillian:
This should be the last version of `next` where the build fails. The issue here is that this build depends on the previous release of `next`, and that release still depends on `tslint-config-typings` which causes the build to fail because of invalid peer reasons. Once this is released, that should all be taken care of

PR #1124, opened 2016-04-13
Clearer message for `no-var-keyword` rule

chrismbarr:
Instead of just telling someone what they can't do, give suggestions about how to fix it and what they should do as an alternative.

Currently the rule `no-var-keyword` complains with 
>forbidden var keyword

But that really doesn't help, especially if someone is new to TypeScript and is not even aware that `let` and `const` are possibilities.

Now it will complain with this message

>forbidden 'var' keyword, use 'let' or 'const' instead

### DIFF
--- a/src/rules/noVarKeywordRule.ts
+++ b/src/rules/noVarKeywordRule.ts
@@ -19,7 +19,7 @@ import * as ts from "typescript";
 import * as Lint from "../lint";
 
 export class Rule extends Lint.Rules.AbstractRule {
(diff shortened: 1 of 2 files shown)


adidahiya:
lgtm 👍  thanks @chrismbarr!

chrismbarr:
Awesome!  Should this also be put in the [3.8 milestone](https://github.com/palantir/tslint/milestones/TSLint%203.8)?

PR #1125, opened 2016-04-13
Adding missing sample options

chrismbarr:
A few options were missing form the sample `tslint.json` file.  I've added them now.

----

:question: **Question**, should the `ban` option also have an example of how to ban a method call?  Perhaps this would be a good sample? Maybe just a subset of this?

```json
"ban": [
	true,
	[ "_", "forEach" ],
	[ "_", "each" ],
	[ "$", "each" ],
	[ "angular", "forEach" ]
]
```

*(We ban these library loops in our apps since in most cases when devs want to use this, they are just looping over a flat JS array and a standard `for` loop would perform faster)*

### DIFF
--- a/docs/sample.tslint.json
+++ b/docs/sample.tslint.json
@@ -20,7 +20,10 @@
       true,
       "spaces"
     ],
-    "interface-name": true,
+    "interface-name": [
+      true,
+      "always-prefix"
+    ],
     "jsdoc-format": true,
     "label-position": true,
     "label-undefined": true,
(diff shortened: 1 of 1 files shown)


adidahiya:
thanks @chrismbarr! yeah I like your idea of showing examples for the ban rule, can you add that in a new commit?

chrismbarr:
@adidahiya Sure, I can do that! Should I just add what I suggested up there, or should it maybe be something that's obviously more "fake" like
```json
"ban": [
    true,
    [ "someObject", "someProperty" ]
]
```

adidahiya:
I like what you suggested up there; it's something I try to enforce in my own code (prefer native iteration with `for-of`)

PR #1129, opened 2016-04-14
Fix config cache issue

### LINKED ISSUE #1128: Require.resolve() in extends cause config not loaded when changed (in vscode)
Referencing:
https://github.com/Microsoft/vscode-tslint/issues/45

This is caused by https://github.com/palantir/tslint/blob/master/src/configuration.ts#L139 using `require(...)` to load the file. `require()` would cache the file, preventing any change to file to be reloaded.

The proposed change is to load the file directly as before. Since now the path is properly resolve (including using require.resolve()) in `resolveConfigurationPath(configFilePath);`, we are safe to do that.

cc @egamma 

I can create a PR on that tomorrow.

unional:
And strip comments. i.e. now allows user to add comments on `tslint.json`

Fixes #1128

### DIFF
--- a/src/configuration.ts
+++ b/src/configuration.ts
@@ -49,7 +49,7 @@ export const DEFAULT_CONFIG = {
             "parameter": "nospace",
             "property-declaration": "nospace",
             "variable-declaration": "nospace",
-        },],
+        }, ],
         "variable-name": [true, "ban-keywords"],
         "whitespace": [true,
             "check-branch",
--- a/src/configuration.ts
+++ b/src/configuration.ts
@@ -49,7 +49,7 @@ export const DEFAULT_CONFIG = {
             "parameter": "nospace",
             "property-declaration": "nospace",
             "variable-declaration": "nospace",
-        }, ],
+        },],
         "variable-name": [true, "ban-keywords"],
         "whitespace": [true,
             "check-branch",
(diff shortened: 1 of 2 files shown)


src/configuration.ts
```diff
@@ -129,14 +129,44 @@ export function findConfigurationPath(suppliedConfigFilePath: string, inputFileP
  * @returns a configuration object for TSLint loaded from the file at configFilePath
  */
 export function loadConfigurationFromPath(configFilePath: string): IConfigurationFile {
+    function stripComments(content: string): string {
```
JKillian: Can you move this method to `src/utils.ts`? I'd just rather keep the implementation logic of removing comments separate from the  "business" logic of loading a configuration

src/configuration.ts
```diff
@@ -129,14 +129,44 @@ export function findConfigurationPath(suppliedConfigFilePath: string, inputFileP
...
+
     if (configFilePath == null) {
         return DEFAULT_CONFIG;
     } else if (path.basename(configFilePath) === "package.json") {
         console.warn(PACKAGE_DEPRECATION_MSG);
         return require(configFilePath).tslintConfig;
     } else {
         const resolvedConfigFilePath = resolveConfigurationPath(configFilePath);
-        let configFile: IConfigurationFile = require(resolvedConfigFilePath);
+        const fileContent = stripComments(fs.readFileSync(resolvedConfigFilePath).toString());
```
JKillian: Looks good for `.json` files, let's add a separate case for all other files (see https://github.com/Microsoft/vscode-tslint/issues/45#issuecomment-209997779)

src/configuration.ts
unional: Sure

src/configuration.ts
unional: Yup

unional:
Sorry for the fail build. I wanted to fix that for quite a while when I started working on the extends feature. :stuck_out_tongue_winking_eye:

src/configuration.ts
```diff
@@ -136,7 +138,15 @@ export function loadConfigurationFromPath(configFilePath: string): IConfiguratio
...
-        let configFile: IConfigurationFile = require(resolvedConfigFilePath);
+        let fileContent: string;
+        if (path.extname(resolvedConfigFilePath) === ".json") {
+            fileContent = stripComments(fs.readFileSync(resolvedConfigFilePath).toString());
+        } else {
+            fileContent = require(resolvedConfigFilePath);
+            delete require.cache[resolvedConfigFilePath];
+        }
+
+        let configFile: IConfigurationFile = JSON.parse(fileContent);
```
JKillian: `JSON.parse` only needs to be done if loading a JSON file. When we require a `.js` file, we'll get back an object automatically

unional: What was I doing........ :stuck_out_tongue_winking_eye:

PR #1133, opened 2016-04-15
Support disallowing namespaces (noNamespace).

mprobst:
This allows code bases to outlaw pre-ES6 modules and namespaces. It's often
still useful to declare namespaces when interacting with non-TypeScript code, so
this has an option to allow "declare namespace" style usage.

### DIFF
--- a/src/rules/noInternalModuleRule.ts
+++ b/src/rules/noInternalModuleRule.ts
@@ -38,17 +38,10 @@ class NoInternalModuleWalker extends Lint.RuleWalker {
         // an internal module declaration is not a namespace or a nested declaration
         // for external modules, node.name.kind will be a LiteralExpression instead of Identifier
         return !Lint.isNodeFlagSet(node, ts.NodeFlags.Namespace)
-            && !isNestedDeclaration(node)
+            && !Lint.isNestedModuleDeclaration(node)
             && node.name.kind === ts.SyntaxKind.Identifier
             && !isGlobalAugmentation(node);
     }
-}
-
-function isNestedDeclaration(node: ts.ModuleDeclaration) {
-    // in a declaration expression like 'module a.b.c' - 'a' is the top level module declaration node and 'b' and 'c'
-    // are nested therefore we can depend that a node's position will only match with its name's position for nested
-    // nodes
-    return node.name.pos === node.pos;
 }
 
 function isGlobalAugmentation(node: ts.ModuleDeclaration) {
--- a/src/rules/noNamespaceRule.ts
+++ b/src/rules/noNamespaceRule.ts
@@ -1,0 +1,38 @@
+/**
+ * @license
+ * Copyright 2016 Palantir Technologies, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import * as ts from "typescript";
+import * as Lint from "../lint";
+
+export class Rule extends Lint.Rules.AbstractRule {
+    public static FAILURE_STRING = "'namespace' and 'module' are disallowed";
+
+    public apply(sourceFile: ts.SourceFile): Lint.RuleFailure[] {
+        return this.applyWithWalker(new NoNamespaceWalker(sourceFile, this.getOptions()));
+    }
+}
+
+class NoNamespaceWalker extends Lint.RuleWalker {
+    public visitModuleDeclaration(decl: ts.ModuleDeclaration) {
+        super.visitModuleDeclaration(decl);
+        // declare module 'foo' {} is an external module, not a namespace.
+        if (decl.name.kind === ts.SyntaxKind.StringLiteral) { return; }
+        if (Lint.isNodeFlagSet(decl, ts.NodeFlags.Ambient) && this.hasOption("allow-declarations")) { return; }
+        if (Lint.isNestedModuleDeclaration(decl)) { return; }
+        this.addFailure(this.createFailure(decl.getStart(), decl.getWidth(), Rule.FAILURE_STRING));
+    }
+}
--- a/README.md
+++ b/README.md
@@ -241,6 +241,8 @@ A sample configuration file with all options is available [here](https://github.
 * `no-inferrable-types` disallows explicit type declarations for variables or parameters initialized to a number, string, or boolean.
 * `no-internal-module` disallows internal `module` (use `namespace` instead).
 * `no-invalid-this` disallows using the `this` keyword outside of classes.
+* `no-namespace` disallows both internal `module`s and `namespace`, but allows ES6-style external modules.
+    * `allow-declarations` Allow `declare module ... {}` to describe external APIs.
 * `no-null-keyword` disallows use of the `null` keyword literal.
 * `no-reference` disallows `/// <reference path=>` imports (use ES6-style imports instead).
 * `no-require-imports` disallows invocation of `require()` (use ES6-style imports instead).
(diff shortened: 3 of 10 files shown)


src/rules/noNamespaceRule.ts
```diff
@@ -0,0 +1,42 @@
...
+
+export class Rule extends Lint.Rules.AbstractRule {
+    public static FAILURE_STRING = "'namespace' and 'module' are disallowed";
+
+    public apply(sourceFile: ts.SourceFile): Lint.RuleFailure[] {
+        return this.applyWithWalker(new NullWalker(sourceFile, this.getOptions()));
+    }
+}
+
+class NullWalker extends Lint.RuleWalker {
```
JKillian: Can we give this class a more meaningful name, like `NoNamespaceWalker`?

src/rules/noNamespaceRule.ts
```diff
@@ -0,0 +1,42 @@
...
+    public visitNode(node: ts.Node) {
```
JKillian: Instead of overriding `visitNode`, we can override `visitModuleDeclaration` which lets us skip line 32 and 33 below

src/rules/noNamespaceRule.ts
```diff
@@ -0,0 +1,42 @@
...
+        super.visitNode(node);
+        if (node.kind !== ts.SyntaxKind.ModuleDeclaration) { return; }
+        let decl = <ts.ModuleDeclaration>node;
+        // declare module 'foo' {} is an external module, not a namespace.
+        if (decl.name.kind === ts.SyntaxKind.StringLiteral) { return; }
+        /* tslint:disable:no-bitwise */
+        if (node.flags & ts.NodeFlags.Ambient && this.hasOption("allow-declarations")) { return; }
```
JKillian: Can you wrap `node.flags & ts.NodeFlags.Ambient` in parens? This code is logically correct as is, I just always forget bitwise operator precedence, so it's nice to have it visually distinguished

test/rules/namespace/allow-declarations/test.ts.lint
```diff
@@ -0,0 +1,3 @@
+declare namespace Foo {
```
JKillian: This file and all the files added to `test/` should be in `test/no-namespace`

test/rules/namespace/disallow/test.ts.lint
```diff
@@ -0,0 +1,17 @@
+namespace Foo {
```
JKillian: This file should be in `test/rules/namespace/default/` to indicate that it's a test for the default configuration

test/rules/namespace/disallow/test.ts.lint
```diff
@@ -0,0 +1,17 @@
...
+~~~~~~~~~~~~~~~
+}
+~ ['namespace' and 'module' are disallowed]
+
+module Foo {
+~~~~~~~~~~~~~~~
```
JKillian: Can you fix this line to be the same width as the line above? (See #1137)

test/rules/namespace/disallow/test.ts.lint
```diff
@@ -0,0 +1,17 @@
...
+}
+~ ['namespace' and 'module' are disallowed]
+
+declare namespace Foo {
+~~~~~~~~~~~~~~~
```
JKillian: Same here

JKillian:
Thanks @mprobst, I like this rule! Makes sense to me to have this rule as more TS codebases will want to enforce ES6 module usage over namespaces.

Left you a few tweaks to make, and then we should be able to merge this. Looks like the rule also needs a rebase, a previously merged PR must've created a conflict

src/rules/noNamespaceRule.ts
mprobst: Done.

src/rules/noNamespaceRule.ts
mprobst: Done.

src/rules/noNamespaceRule.ts
mprobst: Done.

test/rules/namespace/allow-declarations/test.ts.lint
mprobst: Done.

test/rules/namespace/disallow/test.ts.lint
mprobst: Done.

test/rules/namespace/disallow/test.ts.lint
mprobst: Done.

test/rules/namespace/disallow/test.ts.lint
mprobst: Done.

mprobst:
Thanks for the review! Please take another look.

mprobst:
So, in the mean time I found the `no-internal-module` rule (no idea how I didn't notice that before). I think it might make sense to unify the two configuration settings - `no-namespace` is sort of a natural extension of `no-internal-module`, and the code should share the logic around global augmentations and nested declarations.

I've added that as a separate commit as it makes the configuration a bit harder to understand. Let me know what you think.

JKillian:
It's a tough call. In my mind:

 `no-internal-module` encourages the use of the `namespace` keyword instead of the outdated `module` keyword where appropriate

`no-namespace` encourages people to stop using namespaces all together and instead organize their code with ES6 modules.

They're definitely about similar syntaxic elements, but the goals of the two are somewhat different. I'm unsure here about what's best.

mprobst:
I agree, it's a bit tricky, and I presume it might also be nicer to have just one option going forward (`no-namespace`). I could also refactor the code so we only have one shared implementation, but still two different options.

How do we resolve? Get a third opinion?

JKillian:
@adidahiya any thoughts?

mprobst:
@alexeagle knows his way around static analysis tools and might venture an opinion?

alexeagle:
I think tslint lacks a curated default set of checks, in which case things that have been deprecated by the language (`no-internal-module`) should be enabled by default and users don't think about it (unless they need to disable it). If that were the case, the answer would clearly be to make a new, separate `no-namespace` rule.

In my opinion, the convenience of `no-namespace` implying `no-internal-module` is not worth the complexity of understanding the configuration. Are there any other options that currently flip other options? Users are expected to study a long list of rules and decide what to opt-in to, so follow that model.

JKillian:
Makes sense to me. We should be sure that `no-namespace` doesn't conflict with `no-internal-module` though: that is, it shouldn't be possible to get into a position where one of the two rules will always error. 

An example of this would be if you had a rule that required explicit typedefs and had a rule that required implicit typedefs. It would be impossible to declare a variable and satisfy both of these rules.

I think we'll be fine here, as `no-namespace` is just more strict that `no-internal-module` which shouldn't lead to conflicts. Let's keep it as a separate `no-namespace` rule if you're in agreement @mprobst

mprobst:
Please Take Another Look. I've "unmerged" the code, and now just share the implementation of `isNestedDeclaration` across the two linters.

mprobst:
Friendly ping :-)

src/rules/noInternalModuleRule.ts
```diff
@@ -44,7 +44,7 @@ class NoInternalModuleWalker extends Lint.RuleWalker {
     }
 }
 
-function isNestedDeclaration(node: ts.ModuleDeclaration) {
+export function isNestedDeclaration(node: ts.ModuleDeclaration) {
```
JKillian: Can you instead move this function to `src/language/utils` and rename it `isNestedModuleDeclaration`? Then we avoid any chance of circular dependencies and keep shared code in a separate location

src/rules/noNamespaceRule.ts
```diff
@@ -0,0 +1,40 @@
...
+    }
+}
+
+class NoNamespaceWalker extends Lint.RuleWalker {
+    public visitModuleDeclaration(decl: ts.ModuleDeclaration) {
+        super.visitModuleDeclaration(decl);
+        // declare module 'foo' {} is an external module, not a namespace.
+        if (decl.name.kind === ts.SyntaxKind.StringLiteral) { return; }
+        if (Lint.isNodeFlagSet(decl, ts.NodeFlags.Ambient) && this.hasOption("allow-declarations")) { return; }
+        if (isNestedDeclaration(decl)) { return; }
```
JKillian: And this can become `Lint.isNestedModuleDeclaration` with the change mentioned above

JKillian:
Sorry for the delay! Can you make the one change listed above and also rebase off the latest develop? (The rebase will help make sure that CI passes)

mprobst:
No worries! Refactored and rebased, PTAL.

JKillian:
Looks great, thank you!

PR #1139, opened 2016-04-19
Support outlawing `/// <reference>` style imports.

mprobst:
The reasoning is that these are superseded by ES6 style imports, and the syntax
overall isn't too pleasant.

### DIFF
--- a/src/rules/noReferenceRule.ts
+++ b/src/rules/noReferenceRule.ts
@@ -1,0 +1,35 @@
+/**
+ * @license
+ * Copyright 2016 Palantir Technologies, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import * as ts from "typescript";
+import * as Lint from "../lint";
+
+export class Rule extends Lint.Rules.AbstractRule {
+    public static FAILURE_STRING = "<reference> is not allowed, use imports";
+
+    public apply(sourceFile: ts.SourceFile): Lint.RuleFailure[] {
+        return this.applyWithWalker(new NoReferenceWalker(sourceFile, this.getOptions()));
+    }
+}
+
+class NoReferenceWalker extends Lint.RuleWalker {
+    public visitSourceFile(node: ts.SourceFile) {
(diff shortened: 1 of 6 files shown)


src/rules/noReferenceRule.ts
```diff
@@ -0,0 +1,35 @@
...
+    public static FAILURE_STRING = "<reference> is not allowed, use imports";
+
+    public apply(sourceFile: ts.SourceFile): Lint.RuleFailure[] {
+        return this.applyWithWalker(new NoReferenceWalker(sourceFile, this.getOptions()));
+    }
+}
+
+class NoReferenceWalker extends Lint.RuleWalker {
+    public visitSourceFile(node: ts.SourceFile) {
+        for (let ref of node.referencedFiles) {
```
JKillian: `node.referencedFiles` only refers to `<reference path=` statements, which is good

JKillian:
I like this rule too! This theme of rules that encourages using modern-TS is a good use-case for TSLint in my opinion. 

Couple notes: TS [is adding a `<reference types=...>`](https://github.com/Microsoft/TypeScript/issues/7125#issuecomment-209670094) directive. We might want to allow that - I haven't read enough into its use case yet to say for certain though. There's also [`<reference no-default-lib="true"/>`](http://www.typescriptlang.org/docs/handbook/triple-slash-directives.html) which we also probably want to allow (and your code does).

With all that in mind, I think we should change the name of this rule slightly, perhaps to `no-reference-path` or `no-path-reference` or something similar?

test/tsconfig.json
```diff
@@ -90,6 +90,7 @@
         "../src/rules/noInternalModuleRule.ts",
         "../src/rules/noInvalidThisRule.ts",
         "../src/rules/noNullKeywordRule.ts",
+        "../src/rules/noReferenceRule.ts",
```
JKillian: `src/tsconfig.json` should also have a similar change, running `grunt` should automatically update both files.

test/tsconfig.json
```diff
@@ -142,4 +143,4 @@
         "rule-tester/testData.ts",
         "rule-tester/utilsTests.ts"
     ]
-}
+}
```
JKillian: no big deal, but we do like our newlines at the end of files 😉

JKillian: Actually, ignore this, it seems that grunt always removes this newline, so lets just get it removed and keep it removed

mprobst:
What about making this `no-reference`, and then adding `disallow-types` if it becomes a thing? It makes sense to me to make anything that's `reference` related in one option, as much as that's possible.

So `"no-reference": true` for now, and then later if we need to support it `"no-reference": [true, "disallow-types"]`.

test/tsconfig.json
mprobst: Done.

JKillian:
:+1: Agreed, keeping it as `no-reference` seems fine, and various options (either negative or positive) can be added as needed later.

mprobst:
Cool. I think this is good to go then, unless I'm missing something.

JKillian:
Agreed - thanks for the PR!

PR #1148, opened 2016-04-21
Array padding rule

### LINKED ISSUE #1045: Require whitespace inside inline array literals
good

``` ts
const x = [ "one", "two" ];
```

bad

``` ts
const x = ["one", "two"];
```

leeavital:
@JKillian has generously offered to finish this 😄 

---

edit by @adidahiya: fixes #1045

### DIFF
--- a/test/rules/arrayPadding/test.ts.lint
+++ b/test/rules/arrayPadding/test.ts.lint
@@ -1,0 +1,7 @@
+["abc"]
+~~~~~~~ [array padding]
+
+["this", "is", "ok"]
+
+["so", "is",
+"this"]
--- a/src/rules/arrayPaddingRule.ts
+++ b/src/rules/arrayPaddingRule.ts
@@ -1,0 +1,47 @@
+/**
+ * @license
+ * Copyright 2013 Palantir Technologies, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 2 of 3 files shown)


test/rules/arrayPadding/test.ts.lint
```diff
@@ -0,0 +1,7 @@
+["abc"]
+~~~~~~~ [array padding]
+
+["this", "is", "ok"]
```
thasner: shouldn't this break the rule?

adidahiya:
I think this should be called `array-bracket-spacing` like ESLint: http://eslint.org/docs/rules/array-bracket-spacing.

Related: https://github.com/buzinas/tslint-eslint-rules/issues/57

thasner:
sgtm

JKillian:
I'm fine with `array-bracket-spacing`, although we could also make one rule that has options for object-brace-spacing, import-brace-spacing, jsx-attribute-brace-spacing and this. Or we can just keep them all as separate rules.

chrismbarr:
Reporting a failure message that just says
> array padding

Isn't very descriptive or helpful. Perhaps something that actually describes what the issue is and how to fix it, like:

>Arrays require whitespace between brackets

Also, does this rule consider initializing arrays: ` arr = [];`
or would you have to initialize like `arr = [ ];` ?

test/rules/arrayPadding/test.ts.lint
chrismbarr: Ageed.  I don't see the difference.  I would expect:
`[ "this", "is", "ok" ]`
not
`["this", "is", "ok"]`

adidahiya:
👍  for more descriptive error message.

And I think `arr = []` should still be allowed; enforcing `arr = [ ]` feels unnecessary.

JKillian:
Sorry, this is very much not complete. I had told @leeavital he could just throw whatever he had up here and I would finish it, but very clearly I haven't done that yet...

adidahiya:
this PR is old, open a new one with rebased commits if you'd like to take up this task

PR #1159, opened 2016-04-24
Recognize directives in class constructors and properties.

ArturDorochowicz:
This builds on work in #1094 (related to #1050) by improving no-unused-expression rule to recognize directives in class constructors and property getters and setters.


JKillian:
Thanks @ArturDorochowicz, good catch!

PR #1168, opened 2016-04-26
Restricted Linter options type

### LINKED ISSUE #1167: Linter should throw if passed a string configuration
#### Repro

I did a silly and forgot to call `JSON.parse` on the value being passed to `Linter`.

```typescript
// wrong
let linter = new Linter("file.ts", "...", "{ \"rules\": ... } ");
```

```typescript
// correct
let linter = new Linter('file.ts", "...", { rules: ... });
```

#### Actual behavior

It ran without error, but reported 0 lint failures.

#### Expected behavior

It should have blown up and told me I'm an idiot for passing an invalid tslint.json equivalent.

:(

*(I'll send a PR for this unless directed otherwise)*

JoshuaKGoldberg:
Fixes #1167

Wasn't sure where unit tests for this would go, if you do want to test constructor logic.


JKillian:
👍   We can probably get away without a test here for now, it would break some of the full CLI tests if it didn't accept valid input. Do you think we should also add a check for `== null` and throw an error then?

JoshuaKGoldberg:
Yeah, that sounds like a good thing to add.

Do you consider any option fields mandatory? If so, instead of checking for `typeof` it might make sense to check for their existence.

JKillian:
Strangely enough, it looks like `ILinterOptions.formatter` must be specified, while everything else can be left undefined. This seems more like a bug to me than anything, so let's not worry about it - I think `ILinterOptions.formatter` should be made optional in the long run.

On the other hand, if everything is optional, passing `null` as the current options isn't unreasonable. Maybe we should just modify `computeFullOptions` to handle the case where the whole options object is `null` and to provided a default value for `formatter` if not provided?

JoshuaKGoldberg:
Thoughts on this draft? Is there some source of truth for what `formatter`should default to?

src/tslint.ts
```diff
@@ -99,13 +98,19 @@ class Linter {
...
         }
-        this.options.rulesDirectory = arrayify(rulesDirectory).concat(arrayify(configuration.rulesDirectory));
-        this.options.configuration = configuration;
+
+        let { configuration, formatter, formattersDirectory, rulesDirectory } = options;
+
+        return {
+            configuration: configuration || DEFAULT_CONFIG,
+            formatter: formatter || "json",
+            formattersDirectory: formattersDirectory || "",
```
JKillian: let's not add the `|| ""` here, it's fine if this stays `null`. (An empty string would get treated as the current directory which could lead to unexpected behavior)

src/tslint.ts
```diff
@@ -99,13 +98,19 @@ class Linter {
...
+            options = {};
         }
-        this.options.rulesDirectory = arrayify(rulesDirectory).concat(arrayify(configuration.rulesDirectory));
-        this.options.configuration = configuration;
+
+        let { configuration, formatter, formattersDirectory, rulesDirectory } = options;
+
+        return {
+            configuration: configuration || DEFAULT_CONFIG,
+            formatter: formatter || "json",
```
JKillian: let's just have it default to "prose", since that's what the CLI does. The results are still accessible programmatically alongside of the prose description

src/tslint.ts
```diff
@@ -99,13 +98,19 @@ class Linter {
         return rules.some((r) => r.equals(rule));
     }
 
-    private computeFullOptions() {
-        let {configuration, rulesDirectory} = this.options;
-        if (configuration == null) {
-            configuration = DEFAULT_CONFIG;
+    private computeFullOptions(options: ILinterOptionsRaw): ILinterOptions {
+        if (!options || typeof options !== "object") {
```
JKillian: I actually liked what you had before where you threw an error if `typeof options !== "object"`. It's clearly a mistake if that happens, so let's throw an error! As far as if `options` is just null, then we can continue using all the defaults below

src/tslint.ts
```diff
@@ -44,11 +44,10 @@ class Linter {
     private source: string;
     private options: ILinterOptions;
 
-    constructor(fileName: string, source: string, options: ILinterOptions) {
+    constructor(fileName: string, source: string, options: ILinterOptionsRaw) {
         this.fileName = fileName;
         this.source = source;
-        this.options = options;
-        this.computeFullOptions();
+        this.options = this.computeFullOptions(options);
```
JKillian: I like this! Better to return a new object than mutate

src/lint.ts
```diff
@@ -44,6 +44,13 @@ export interface LintResult {
     output: string;
 }
 
+export interface ILinterOptionsRaw {
```
JKillian: Mixed feelings about if we really need a new interface here - I'm sort of inclined to just making the properties of `ILinterOptions` optional. What do you think?

JKillian:
Left my thoughts above @JoshuaKGoldberg, feel free to disagree on any of them!

src/tslint.ts
JoshuaKGoldberg: Agreed. I'll switch this out.

src/lint.ts
JoshuaKGoldberg: There's a TypeScript issue filed somewhere for letting us do something like `ILinterOptions?`, where every field in `ILinterOptions` would be marked optional (so exactly this).

My problem with re-using one interface for this is that after this initial parsing, the fields *aren't* optional. The semantic intent changes from input option settings to stored options data.

src/tslint.ts
JoshuaKGoldberg: Ok - would you consider it better to hardcode `"prose"` again or move the setting to configuration.ts?

src/lint.ts
JKillian: Makes sense to me, let's keep this then. It shouldn't cause any backwards compat issues for library users as fulfilling the previous `ILinterOptions` interface will still fulfill the `ILinterOptionsRaw` interface. Maybe we should make the `formattersDirectory` field optional on both though, since it really is optional

src/tslint.ts
JKillian: I'm good with keeping it here, but either way is fine by me

JKillian:
Beautiful! This turn out very nice

PR #1172, opened 2016-04-27
Use `resolve` instead of Node.JS require for resolving configuration files

### LINKED ISSUE #1171: global tslint throws error if `extends` config option references local package
### Bug Report
- __TSLint version__: `3.8.1`

`tslint.json`:

```json
{
  "extends": "shared-tslint-config"
}
```

(where `shared-tslint-config` is a locally installed NPM package providing a tslint config)

#### Actual behavior

(when using a globally installed `tslint`)

```bash
$ tslint src/index.ts --config tslint.json

/usr/local/lib/node_modules/tslint/lib/configuration.js:122
            throw new Error(("Invalid \"extends\" configuration value - could not require \"" + relativeFilePath + "\". ") +
            ^

Error: Invalid "extends" configuration value - could not require "shared-tslint-config". Review the Node lookup algorithm (https://nodejs.org/api/modules.html#modules_all_together) for the approximate method TSLint uses to find the referenced configuration file.
    at resolveConfigurationPath (/usr/local/lib/node_modules/tslint/lib/configuration.js:122:19)
    at loadConfigurationFromPath (/usr/local/lib/node_modules/tslint/lib/configuration.js:101:38)
    at Object.findConfiguration (/usr/local/lib/node_modules/tslint/lib/configuration.js:45:12)
    at processFile (/usr/local/lib/node_modules/tslint/lib/tslint-cli.js:111:41)
    at Array.forEach (native)
    at Object.<anonymous> (/usr/local/lib/node_modules/tslint/lib/tslint-cli.js:128:41)
    at Module._compile (module.js:409:26)
    at Object.Module._extensions..js (module.js:416:10)
    at Module.load (module.js:343:32)
    at Function.Module._load (module.js:300:12)
```
#### Expected behavior

tslint should resolve configuration file from locally installed package `shared-tslint-config`

janslow:
Fixes #1171 by using [resolve](https://www.npmjs.com/package/resolve) to resolve package-based extension configuration files from the directory of the current configuration file.

I've modified the tests for extends so that they have their own node_modules (created by the `run:installTestDeps` task) in `./test/config`, so that `tslint-test-config` and `tslint-test-custom-rules` can't be accessed by tslint directly with `require()`.

### DIFF
--- a/src/configuration.ts
+++ b/src/configuration.ts
@@ -18,7 +18,6 @@
 import * as fs from "fs";
 import * as path from "path";
 import * as findup from "findup-sync";
-import * as pathIsAbsolute from "path-is-absolute";
 import * as resolve from "resolve";
 
 import {arrayify, objectify, stripComments} from "./utils";
@@ -167,22 +166,14 @@ export function loadConfigurationFromPath(configFilePath: string): IConfiguratio
  * @var relativeFilePath Relative path or package name (tslint-config-X) or package short name (X)
  */
 function resolveConfigurationPath(relativeFilePath: string, relativeTo?: string) {
-    let resolvedPath: string;
-    if (pathIsAbsolute(relativeFilePath)) {
-        resolvedPath = relativeFilePath;
-    } else if (relativeFilePath.indexOf(".") === 0) {
-        resolvedPath = getRelativePath(relativeFilePath, relativeTo);
-    } else {
-        try {
-            resolvedPath = resolve.sync(relativeFilePath, { basedir: relativeTo });
-        } catch (err) {
-            throw new Error(`Invalid "extends" configuration value - could not require "${relativeFilePath}". ` +
-                "Review the Node lookup algorithm (https://nodejs.org/api/modules.html#modules_all_together) " +
-                "for the approximate method TSLint uses to find the referenced configuration file.");
-        }
+    const basedir = relativeTo || process.cwd();
+    try {
+        return resolve.sync(relativeFilePath, { basedir });
+    } catch (err) {
+        throw new Error(`Invalid "extends" configuration value - could not require "${relativeFilePath}". ` +
+            "Review the Node lookup algorithm (https://nodejs.org/api/modules.html#modules_all_together) " +
+            "for the approximate method TSLint uses to find the referenced configuration file.");
     }
-
-    return resolvedPath;
 }
 
 export function extendConfigurationFile(config: IConfigurationFile, baseConfig: IConfigurationFile): IConfigurationFile {
--- a/src/configuration.ts
+++ b/src/configuration.ts
@@ -170,9 +170,13 @@ function resolveConfigurationPath(relativeFilePath: string, relativeTo?: string)
     try {
         return resolve.sync(relativeFilePath, { basedir });
     } catch (err) {
-        throw new Error(`Invalid "extends" configuration value - could not require "${relativeFilePath}". ` +
-            "Review the Node lookup algorithm (https://nodejs.org/api/modules.html#modules_all_together) " +
-            "for the approximate method TSLint uses to find the referenced configuration file.");
+        try {
+            return require.resolve(relativeFilePath);
+        } catch (err) {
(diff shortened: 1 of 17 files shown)


src/configuration.ts
```diff
@@ -166,22 +166,14 @@ export function loadConfigurationFromPath(configFilePath: string): IConfiguratio
...
-        resolvedPath = getRelativePath(relativeFilePath, relativeTo);
-    } else {
-        try {
-            resolvedPath = require.resolve(relativeFilePath);
-        } catch (err) {
-            throw new Error(`Invalid "extends" configuration value - could not require "${relativeFilePath}". ` +
-                "Review the Node lookup algorithm (https://nodejs.org/api/modules.html#modules_all_together) " +
-                "for the approximate method TSLint uses to find the referenced configuration file.");
-        }
+    const basedir = relativeTo || process.cwd();
```
JKillian: Love how this could be simplified!

I'm wondering if we should also attempt to load with regular `require.resolve` as well. I.e. does it also make sense to look for the node module relative to the installed location of TSLint? I think it does, but only if the "path" specified is not  a path but instead a package reference

JKillian:
> I've modified the tests for extends so that they have their own node_modules (created by the run:installTestDeps task)

Clever, seems like a good way to test things out! Seems like it's causing [the Windows CI](https://ci.appveyor.com/project/ashwinr/tslint/build/1.0.1562) to fail though, any ideas?

image

src/configuration.ts
janslow: That would mean it's not the same as the [`require.resolve() algorithm](https://nodejs.org/dist/latest-v4.x/docs/api/modules.html#modules_all_together). Before this pull request `Y` (in the algorithm) was always the location of the tslint package, now it's the location of the parent `tslint.json` file (or the tslint package the first time).

It's similar to how you'd expect `node -e "require('foo')"` to fail if `foo` was installed globally but not locally.

However, it's arguably a breaking change, so I'm happy to add a fallback which resolves with `require.resolve` to avoid issues for other users.

janslow:
It may be as simple as adding `npm` as a dev dependency. If not, I'll try using a grunt plugin like [`grunt-npm-install`](https://www.npmjs.com/package/grunt-npm-install) or [`grunt-npm-command`](https://www.npmjs.com/package/grunt-npm-command)

src/configuration.ts
JKillian: Agree with you that ideally `extends` references in a `tslint.json` file should only reference packages installed where the file is, but I can see this being a situation where people are relying on other cases, so let's prevent the breaking change here

JKillian:
> It may be as simple as adding npm as a dev dependency. If not, I'll try using a grunt plugin like grunt-npm-install or grunt-npm-command.

Nice fix on this issue, looks like `grunt-npm-command` solves things well

janslow:
These commits should maintain backwards compatibility with packages installed relative to tslint, with some extra unit tests and CLI tests.

janslow:
Unfortunately, I can't reproduce the error on ci/circleci (including in a clean checkout).

test/check-bin.sh
```diff
@@ -70,6 +70,18 @@ expectOut $? 0 "tslint with with -r pointing to custom rules did not find lint f
 ./bin/tslint -c test/config/tslint-almost-empty.json src/tslint.ts
 expectOut $? 0 "-c relative path without ./ did not work"
 
+# make sure calling tslint with a config file which extends a package relative to the config file works
+./bin/tslint -c test/config/tslint-extends-package-no-mod.json src/tslint.ts
+expectOut $? 0 "tslint (with config file extending relative package) did not work"
+
+# check that a tslint file (outside of the resolution path of the tslint package) can resolve a package that is
+# installed as a dependency of tslint. See palantir/tslint#1172 for details.
+tmpDir=$(mktemp -d)
```
JKillian: On the version of `mktemp` installed on my Mac, this is unfortunately an invalid usage. I'm okay with getting rid of this test here though, as you have a similar test below. Plus your test below should work on Windows also, which is great.

JKillian:
This looks great to me! I'll make the one change I mentioned above in a follow-up commit

PR #1173, opened 2016-04-27
member-ordering: add a check of constructor

### LINKED ISSUE #1158: member-ordering: doesn't check constructor
### Bug Report 

- __TSLint version__: `3.8.0`
- __Running TSLint via__: CLI

with `tslint.json`:

```json
{
   "member-ordering": [
           true,
           "public-before-private",
           "static-before-instance",
           "variables-before-functions"
        ]
}
```

When checking this code:

```ts
class MyClass {
       foo() {}
       blah: number = 0;
}
```

I get the expected warning:

  > [30, 10]: Declaration of public instance member variable not allowed to appear after declaration of public instance member function

But when checking this code it passes. (renamed 'foo' to 'constructor')

```ts
class MyClass {
       constructor() {}
       blah: number = 0;
}
```

#### Expected behavior

I would expect the constructor case to have the same warning because there is a method before property definition.

YuichiNukiyama:
fix #1158

### DIFF
--- a/src/rules/memberOrderingRule.ts
+++ b/src/rules/memberOrderingRule.ts
@@ -77,6 +77,11 @@ export class MemberOrderingWalker extends Lint.RuleWalker {
         super.visitMethodSignature(node);
     }
 
+    public visitConstructorDeclaration(node: ts.ConstructorDeclaration) {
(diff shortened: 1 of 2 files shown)


JKillian:
Thanks @YuichiNukiyama!

PR #1175, opened 2016-04-28
Single line ignore switches

### LINKED ISSUE #144: Ignore a line with a single comment
Is there a way to ignore a single line, like in jshint?
I have tried // tslint ignore:line but that does not seem to work.

patsissons:
This PR supports using ESLint style single line ignore switches by using `tslint:disable-line` and `tslint:disable-next-line`.

This PR resolves #144

### DIFF
--- a/src/enableDisableRules.ts
+++ b/src/enableDisableRules.ts
@@ -26,6 +26,7 @@ export class EnableDisableRulesWalker extends SkippableTokenAwareRuleWalker {
     public visitSourceFile(node: ts.SourceFile) {
         super.visitSourceFile(node);
         const scan = ts.createScanner(ts.ScriptTarget.ES5, false, ts.LanguageVariant.Standard, node.text);
+        let lineStartPos = 0;
 
         scanAllTokens(scan, (scanner: ts.Scanner) => {
             const startPos = scanner.getStartPos();
@@ -36,16 +37,21 @@ export class EnableDisableRulesWalker extends SkippableTokenAwareRuleWalker {
                 return;
             }
 
+            // keep track of the beginning of the line so that we can use it for same line switches
+            if (scanner.getToken() === ts.SyntaxKind.NewLineTrivia) {
+                lineStartPos = startPos + 1;
+            }
+
             if (scanner.getToken() === ts.SyntaxKind.MultiLineCommentTrivia ||
                 scanner.getToken() === ts.SyntaxKind.SingleLineCommentTrivia) {
                 const commentText = scanner.getTokenText();
                 const startPosition = scanner.getTokenPos();
-                this.handlePossibleTslintSwitch(commentText, startPosition);
+                this.handlePossibleTslintSwitch(commentText, startPosition, lineStartPos, scanner);
             }
         });
     }
 
-    private handlePossibleTslintSwitch(commentText: string, startingPosition: number) {
+    private handlePossibleTslintSwitch(commentText: string, startingPosition: number, lineStartingPosition: number, scanner: ts.Scanner) {
         // regex is: start of string followed by "/*" or "//" followed by any amount of whitespace followed by "tslint:"
         if (commentText.match(/^(\/\*|\/\/)\s*tslint:/)) {
             const commentTextParts = commentText.split(":");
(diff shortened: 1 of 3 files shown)


src/enableDisableRules.ts
```diff
@@ -36,25 +37,33 @@ export class EnableDisableRulesWalker extends SkippableTokenAwareRuleWalker {
                 return;
             }
 
+            // keep track of the beginning of the line so that we can use it for same line switches
+            if (scanner.getToken() === ts.SyntaxKind.NewLineTrivia) {
```
JKillian: As opposed to tracking this, you could use:

`node.getLineAndCharacterOfPosition` in conjunction with `scanner.getTokenPos`

src/enableDisableRules.ts
```diff
@@ -63,10 +72,46 @@ export class EnableDisableRulesWalker extends SkippableTokenAwareRuleWalker {
...
+                            position: scanner.getTextPos() + 1,
+                        });
+                    } else {
+                        this.enableDisableRuleMap[ruleToAdd].push({
+                            isEnabled: isEnabled,
+                            position: startingPosition,
+                        });
+
+                        if (isNextLine) {
+                            const nextLineEndPos = scanner.lookAhead(() => {
```
JKillian: I think you should be able to simplify this by passing in the `SourceFile` node and using `node.getPositionOfLineAndCharacter` or maybe `node.getLineStarts`

patsissons:
I'll see what I can come up with tonight

JKillian:
Looking good to me @patsissons! Let's see if we can simplify things a little using the `SourceFile` APIs, but if it seems like a bad idea after trying it to you, we can stick with something more like it is now

patsissons:
I'm all for optimized clean code :+1: what you're suggesting sounds pretty reasonable.

src/enableDisableRules.ts
```diff
@@ -63,10 +72,31 @@ export class EnableDisableRulesWalker extends SkippableTokenAwareRuleWalker {
...
+                    if (isCurrentLine) {
+                        // start at the beginning of the current line
+                        this.enableDisableRuleMap[ruleToAdd].push({
+                            isEnabled: isEnabled,
+                            position: this.getStartOfLinePosition(node, startingPosition),
+                        });
+                        // end at the beginning of the next line
+                        this.enableDisableRuleMap[ruleToAdd].push({
+                            isEnabled: !isEnabled,
+                            position: scanner.getTextPos() + 1,
```
JKillian: Should this be:
`position: this.getStartOfLinePosition(node, startingPosition, 1),` ?

Possible that the current version is fine, I just don't understand what it's doing exactly

JKillian:
👍 Like this new version, much easier to read and understand! Just had one comment above, looks good to me otherwise

src/enableDisableRules.ts
patsissons: The third param is line offset, which defaults to 0, the current line. For current line switches we want to start at the beginning of the current line, so line offset isn't necessary.

patsissons:
Thoughts on tests to cover the start and stop interval creation? Or is that better captured in a new PR?

src/enableDisableRules.ts
JKillian: Sorry for the lack of clarity, I was referring to if `position: scanner.getTextPos() + 1,` should be changed

JKillian:
As far as tests, I think we should be able to move this over to the newer, more visual testing system. (I.e. the files in `test/rules`. Test case(s) could go in `test/rules/_integration/comments`. That can be a follow-up though

src/enableDisableRules.ts
patsissons: Oh sorry, i see what you were trying to say now. Either version should yield the same position. `scanner.getTextPos()` returns the position of the end of the text, the `+1` is just to bump the position to the beginning of the next line. I guess that the current expression may be a little terse, I could add an extra comment for clarity if you like?

JKillian: No worries, the comment above is clear enough! Just wanted to check that this was as intended. PR merged!

JKillian:
@patsissons Do you want to do the test PR? Or should I try to knock it out? Just want to make sure we don't duplicate work

patsissons:
I probably wouldn't get to it for a few days, so go ahead. Plus you'll definitely have a better understanding on how to improve things than I will.

PR #1176, opened 2016-04-28
Make member-variable-declaration check that for property assignment of an arrow function, either the property has a type def, or all parameters of the arrow function have type defs

tomduncalf:
The rationale behind this is that with the existing `"member-variable-declaration"` option, the type definition for a arrow function member variable needs to be on the variable itself:

```
myMember: (param: string) => void = (param) => {
```

which leads to duplication (`param` is written twice) - even more so if you are using the `"arrow-parameter"` rule, in which case you would need:

```
myMember: (param: string) => void = (param: string): void => {
```

A more natural syntax, which has the same type safety, is:

```
myMember = (param: string): void => {
```

To enable this syntax, we need to split out the typedef checking of arrow function member variables vs. all other member variables, hence the `"arrow-member-variable-declaration"` rule.

However, if you are not using `"arrow-parameter"`, then tslint won't enforce a type definition on the arrow function either, so with just `"arrow-member-variable-declaration"`, this would be allowed:

```
myMember = (param) => {
```

So in this case, we probably want to enforce that arrow functions *should* have parameter definitions, but only if they are a member variable, which is what `"arrow-member-variable-declaration-in-function"` achieves, so with that the following is valid:

```
myMember = (param: string): void => {
```

There might be a better way to achieve this (or a less verbose way of naming the options!), I kind of hacked it together so I'm happy to discuss further!

### DIFF
--- a/src/rules/typedefRule.ts
+++ b/src/rules/typedefRule.ts
@@ -40,7 +40,8 @@ class TypedefWalker extends Lint.RuleWalker {
 
     public visitArrowFunction(node: ts.FunctionLikeDeclaration) {
         const location = (node.parameters != null) ? node.parameters.end : null;
-        if (node.parent.kind !== ts.SyntaxKind.CallExpression) {
+
+        if (node.parent.kind !== ts.SyntaxKind.CallExpression && !isTypedPropertyDeclaration(node.parent)) {
             this.checkTypeAnnotation("arrow-call-signature", location, node.type, node.name);
         }
         super.visitArrowFunction(node);
@@ -85,8 +86,23 @@ class TypedefWalker extends Lint.RuleWalker {
     public visitParameterDeclaration(node: ts.ParameterDeclaration) {
         // a parameter's "type" could be a specific string value, for example `fn(option: "someOption", anotherOption: number)`
         if (node.type == null || node.type.kind !== ts.SyntaxKind.StringLiteral) {
-            const optionName = node.parent.kind === ts.SyntaxKind.ArrowFunction ? "arrow-parameter" : "parameter";
-            this.checkTypeAnnotation(optionName, node.getEnd(), <ts.TypeNode> node.type, node.name);
+            const isArrowFunction = node.parent.kind === ts.SyntaxKind.ArrowFunction;
+            let performCheck = true;
+
+            let optionName: string;
+            if (isArrowFunction && isTypedPropertyDeclaration(node.parent.parent)) {
+                performCheck = false;
+            } else if (isArrowFunction && isPropertyDeclaration(node.parent.parent)) {
+                optionName = "member-variable-declaration";
+            } else if (isArrowFunction) {
+                optionName = "arrow-parameter";
+            } else {
+                optionName = "parameter";
+            }
+
+            if (performCheck) {
+                this.checkTypeAnnotation(optionName, node.getEnd(), <ts.TypeNode> node.type, node.name);
+            }
         }
         super.visitParameterDeclaration(node);
     }
@@ -105,7 +121,14 @@ class TypedefWalker extends Lint.RuleWalker {
 
     public visitPropertyDeclaration(node: ts.PropertyDeclaration) {
         const optionName = "member-variable-declaration";
-        this.checkTypeAnnotation(optionName, node.name.getEnd(), node.type, node.name);
+
+        // If this is an arrow function, it doesn't need to have a typedef on the property declaration
+        // as the typedefs can be on the function's parameters instead
+        const performCheck = !(node.initializer.kind === ts.SyntaxKind.ArrowFunction && node.type == null);
+
+        if (performCheck) {
+            this.checkTypeAnnotation(optionName, node.name.getEnd(), node.type, node.name);
+        }
         super.visitPropertyDeclaration(node);
     }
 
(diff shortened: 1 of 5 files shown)


JKillian:
You explained things very well, but I'm afraid this many options (which have subtle differences) might end up being too confusing for users.

What if we didn't add any new options, but instead, `member-variable-declaration` just checks that either what you describe as `arrow-member-variable-declaration` or as `arrow-member-variable-declaration-in-function` is true?

The downside to this is that it wouldn't make one / or the other style mandatory in your codebase, so you could end up with some of both types of type declarations. On the other hand, you'd always have an explicit type declaration, which would probably fulfill the intent of the rule well enough.

tomduncalf:
I did think that might be the case, but couldn't think of a good way to reduce the number of options. However, I like your idea - the main thing I am concerned about is having type safety but being able to express it more succinctly so that would be fine. I'll look at updating the PR next week, thanks.

JKillian:
Hi @tomduncalf, still interested in updating this?

tomduncalf:
Sorry, yes I am, but haven't found time to look at it yet. I will try and have a look next week if that is okay with you?

JKillian:
Sounds great!

tomduncalf:
Going to have be honest, I'm not really sure how to implement this!

The logic I'm trying to implement is:

- In `visitParameterDeclaration`, if this is an arrow function and the parent is a parameter assignment, and the parent parameter assignment has a type, don't check the parameter type.

- In `visitPropertyAssignment`, if the value being assigned is an arrow function, don't check the property type if the arrow function has types for its parameters and return value.

It's the second part I am struggling with – do you have any good ideas for how I might check if the arrow function being assigned has types, in a non-hacky way? Or is there a much simpler way to do this that I'm missing out on? :)

tomduncalf:
Never mind, think I've got it!

tomduncalf:
Alright, had a go at implementing what was discussed above – feedback welcome :)

src/rules/typedefRule.ts
```diff
@@ -155,3 +181,11 @@ class TypedefWalker extends Lint.RuleWalker {
         }
     }
 }
+
+function isPropertyDeclaration(node: ts.Node): boolean {
```
JKillian: You can actually make this a typeguard:

```ts
function isPropertyDeclaration(node: ts.Node): node is ts.PropertyDeclaration {
```

src/rules/typedefRule.ts
```diff
@@ -155,3 +181,11 @@ class TypedefWalker extends Lint.RuleWalker {
...
+    return node.kind === ts.SyntaxKind.PropertyDeclaration;
+}
+
+function isTypedPropertyDeclaration(node: ts.Node): boolean {
+    return isPropertyDeclaration(node) && !!(node as ts.PropertyDeclaration).type;
```
JKillian: Then with the typeguard from above, you can simplify this!

```
return isPropertyDeclaration(node) && node.type != null;
```

(The TSLint code-style uses explicit comparisons to null instead of `!!` to avoid the weirdness that happens with JS implicit type conversion. For example, `!!"" !== !!"0"`)

src/rules/typedefRule.ts
```diff
@@ -105,7 +121,17 @@ class TypedefWalker extends Lint.RuleWalker {
...
     public visitPropertyDeclaration(node: ts.PropertyDeclaration) {
         const optionName = "member-variable-declaration";
-        this.checkTypeAnnotation(optionName, node.name.getEnd(), node.type, node.name);
+        let performCheck = true;
+
+        // If this is an arrow function, it doesn't need to have a typedef on the property declaration
+        // as the typedefs can be on the function's parameters instead
+        if (!node.type && node.initializer.kind === ts.SyntaxKind.ArrowFunction) {
+            performCheck = false;
+        }
```
JKillian: Could simplify the above code to:

```
const performCheck = node.type != null || node.initializer.kind !== ts.SyntaxKind.ArrowFunction;
```

src/rules/typedefRule.ts
```diff
@@ -85,8 +86,23 @@ class TypedefWalker extends Lint.RuleWalker {
     public visitParameterDeclaration(node: ts.ParameterDeclaration) {
         // a parameter's "type" could be a specific string value, for example `fn(option: "someOption", anotherOption: number)`
         if (node.type == null || node.type.kind !== ts.SyntaxKind.StringLiteral) {
-            const optionName = node.parent.kind === ts.SyntaxKind.ArrowFunction ? "arrow-parameter" : "parameter";
-            this.checkTypeAnnotation(optionName, node.getEnd(), <ts.TypeNode> node.type, node.name);
+            const isArrowFunction = node.parent.kind === ts.SyntaxKind.ArrowFunction;
+            let performCheck = true;
+
+            let optionName: string = null;
```
JKillian: Don't need to initialize to `null` here

JKillian:
Looking good to me! Left a few small style comments, but really only trivial things

src/rules/typedefRule.ts
tomduncalf: Ah of course, my mind was a bit melted after trying to figure out how to implement the feature so I missed some silly things 😃

src/rules/typedefRule.ts
tomduncalf: That implicit coercion from the typeguard is very cool :)

src/rules/typedefRule.ts
tomduncalf: I did this slightly differently as it read easier to me: `const performCheck = !(node.initializer.kind === ts.SyntaxKind.ArrowFunction && node.type == null);`

tomduncalf:
Comments addressed

JKillian:
Thanks @tomduncalf!

tomduncalf:
Thanks for the merge! Going to go and rip out all my old `this.x = this.x.bind(this)` from React component constructors and replace them with shiny new property assigned arrow functions now 👍

PR #1179, opened 2016-04-28
Extend no-invalid-this rule to disallow `this` in functions in methods

inthemill:
This PR extends the [no-invalid-this] with the option "in-function-in-method".
If it is enabled it complains about the usage of the `this` keyword inside of function expression within a method.
`this` in such function expressions is usually meant to refer the this outside of the functions scope.
 This rule will not complain about this in arrow functions.

In our project, it happened often when migrating JS to TS. With an angular service written as
```javascript
angular.module('a',[]).factory('Service', function(Helper, Enhancer){
return {
    getAllMapped: function(){
         return Helper.getAll()
             .map(function(item){
                  return Enhancer.enhance(item);
             })
    }
}
}); 
```
We often end up with
```javascript
class ServiceClass {
    constructor(private Helper, private Enhancer){
    }

    public getAllMapped() {
         return this.Helper.getAll()
             .map(function(item){
                  return this.Enhancer.enhance(item);
             })
    }
}
angular.module('a',[]).service('Service', ServiceClass);
```
which will not work as ```this.Enhancer``` is not defined.

### DIFF
--- a/README.md
+++ b/README.md
@@ -240,7 +240,9 @@ A sample configuration file with all options is available [here](https://github.
 * `no-eval` disallows `eval` function invocations.
 * `no-inferrable-types` disallows explicit type declarations for variables or parameters initialized to a number, string, or boolean.
 * `no-internal-module` disallows internal `module` (use `namespace` instead).
-* `no-invalid-this` disallows using the `this` keyword outside of classes.
+* `no-invalid-this` disallows using the `this` keyword in scopes where they are often misinterpreted
+    * `outside-of-class` disallows using the `this` keyword outside of classes.
+    * `in-function-in-method` disallows using the `this` keyword in functions within class methods.
 * `no-namespace` disallows both internal `module`s and `namespace`, but allows ES6-style external modules.
     * `allow-declarations` allows `declare namespace ... {}` to describe external APIs.
 * `no-null-keyword` disallows use of the `null` keyword literal.
--- a/README.md
+++ b/README.md
@@ -240,9 +240,8 @@ A sample configuration file with all options is available [here](https://github.
 * `no-eval` disallows `eval` function invocations.
 * `no-inferrable-types` disallows explicit type declarations for variables or parameters initialized to a number, string, or boolean.
 * `no-internal-module` disallows internal `module` (use `namespace` instead).
-* `no-invalid-this` disallows using the `this` keyword in scopes where they are often misinterpreted
-    * `outside-of-class` disallows using the `this` keyword outside of classes.
-    * `in-function-in-method` disallows using the `this` keyword in functions within class methods.
+* `no-invalid-this` disallows using the `this` keyword outside of classes.
+    * `no-this-in-function-in-method` disallows using the `this` keyword in functions within class methods.
 * `no-namespace` disallows both internal `module`s and `namespace`, but allows ES6-style external modules.
     * `allow-declarations` allows `declare namespace ... {}` to describe external APIs.
 * `no-null-keyword` disallows use of the `null` keyword literal.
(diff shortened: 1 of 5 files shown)


README.md
```diff
@@ -240,7 +240,9 @@ A sample configuration file with all options is available [here](https://github.
 * `no-eval` disallows `eval` function invocations.
 * `no-inferrable-types` disallows explicit type declarations for variables or parameters initialized to a number, string, or boolean.
 * `no-internal-module` disallows internal `module` (use `namespace` instead).
-* `no-invalid-this` disallows using the `this` keyword outside of classes.
+* `no-invalid-this` disallows using the `this` keyword in scopes where they are often misinterpreted
+    * `outside-of-class` disallows using the `this` keyword outside of classes.
```
JKillian: Can we keep this behavior on by default? For backwards-compat reasons, we don't want to have to make users add a new option to their rule to get the same behavior as before

src/rules/noInvalidThisRule.ts
```diff
@@ -18,26 +18,46 @@
...
 import * as Lint from "../lint";
 
-export class Rule extends Lint.Rules.AbstractRule {
-    public static FAILURE_STRING = "the \"this\" keyword is disallowed outside of a class body" ;
+interface Scope {
+    inClass: boolean;
+    inFunction: boolean;
+}
 
+const OPTION_OUTSIDE = "outside-of-class";
```
JKillian: We can get rid of this option as noted above

src/rules/noInvalidThisRule.ts
```diff
@@ -18,26 +18,46 @@
...
-        if (!this.getAllScopes().some((scope) => scope.inClass)) {
-            this.addFailure(this.createFailure(node.getStart(), node.getWidth(), Rule.FAILURE_STRING));
+        let inClass = 0;
+        let inFunction = 0;
+        this.getAllScopes().forEach((scope, index) => {
+            inClass = scope.inClass ? index + 1 : inClass;
+            inFunction = scope.inFunction ? index + 1 : inFunction;
+        });
+
+        if (this.hasOption(OPTION_OUTSIDE) && !inClass) {
```
JKillian: `inClass === 0` (we prefer not using implicit casts in conditionals)

src/rules/noInvalidThisRule.ts
```diff
@@ -18,26 +18,46 @@
...
+    inClass: boolean;
+    inFunction: boolean;
+}
 
+const OPTION_OUTSIDE = "outside-of-class";
+const OPTION_INSIDE = "in-function-in-method";
+
+export class Rule extends Lint.Rules.AbstractRule {
+    public static FAILURE_STRING_OUTSIDE = "the \"this\" keyword is disallowed outside of a class body" ;
+    public static FAILURE_STRING_INSIDE = "the \"this\" keyword is disallowed in function bodys inside class methods";
```
JKillian: "bodies" instead of "body"

test/rules/no-invalid-this/enabled/test.ts.lint
```diff
@@ -9,9 +9,46 @@ function foo(x: number) {
...
+    }
+
     public aMethod() {
         this.x = 5;
     }
+
+    public bMethod() {
+        [3,4].forEach(function(nr){
+            console.log(this.x === nr);
+                        ~~~~[the "this" keyword is disallowed in function bodys inside class methods]
```
JKillian: add a space between the "~~~~" and "[" for readability

JKillian:
Looks good to me logic-wise @inthemill! If you could just make the couple tweaks listed above about the rule's API and style things, we'll be able to merge this after.

JKillian:
Looks great, thanks @inthemill!

PR #1183, opened 2016-04-29
no-unreachable: allow type alias after function return

### LINKED ISSUE #564: Incorrect unreachable code error for type alias after function return
```ts
function foo() {
  return 1;
  function bar() {}
  type Bar = string;
}
```
Reports unreachable code on the type alias. 

Using tslint v2.5.0-dev.1

YuichiNukiyama:
fix #564

### DIFF
--- a/src/rules/noUnreachableRule.ts
+++ b/src/rules/noUnreachableRule.ts
@@ -36,8 +36,9 @@ class NoUnreachableWalker extends Lint.RuleWalker {
 
     public visitNode(node: ts.Node) {
         const previousReturned = this.hasReturned;
-        // function declarations can be hoisted -- so set hasReturned to false until we're done with the function
-        if (node.kind === ts.SyntaxKind.FunctionDeclaration) {
+        // function declarations and type alias declarations can be hoisted
+        // -- so set hasReturned to false until we're done with the function
+        if (node.kind === ts.SyntaxKind.FunctionDeclaration || node.kind === ts.SyntaxKind.TypeAliasDeclaration) {
             this.hasReturned = false;
         }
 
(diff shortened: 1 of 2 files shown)


adidahiya:
thanks for the PR!

lgtm 👍

src/rules/noUnreachableRule.ts
```diff
@@ -99,4 +100,8 @@ class NoUnreachableWalker extends Lint.RuleWalker {
         super.visitThrowStatement(node);
         this.hasReturned = true;
     }
+
+    public visitTypeAliasDeclaration(node: ts.TypeAliasDeclaration) {
```
JKillian: I don't think we actually need to add these lines - `visitNode` gets called on all nodes by default

src/rules/noUnreachableRule.ts
```diff
@@ -36,8 +36,9 @@ class NoUnreachableWalker extends Lint.RuleWalker {
 
     public visitNode(node: ts.Node) {
         const previousReturned = this.hasReturned;
-        // function declarations can be hoisted -- so set hasReturned to false until we're done with the function
-        if (node.kind === ts.SyntaxKind.FunctionDeclaration) {
+        // function declarations and type alias declarations can be hoisted
+        // -- so set hasReturned to false until we're done with the function
+        if (node.kind === ts.SyntaxKind.FunctionDeclaration || node.kind === ts.SyntaxKind.TypeAliasDeclaration) {
```
JKillian: Down in line 54 we should restore the state of `this.hasReturned` like we do for function declarations

JKillian:
Hi @YuichiNukiyama, thanks for the PR, just a couple tweaks and then we can merge this!

YuichiNukiyama:
Thank you for your review. I did it.

JKillian:
Looks great, thanks!

Issue #811, opened 2015-11-20
Take optional flag in whitespace rules into account


pmentz:
Hi,

I have set

``` JSON
"typedef-whitespace": [
  true,
  {
    "parameter": "space"
  }
]
```

to enforce a space in front of the colon in method signatures. Unfortunately, what it seems to do is to force me to add a space after the parameter name, which only allows the following syntax for optional parameters:

``` Javascript
public myMethod(name ?:string):void { /* Looks like an optional type */ }
public myAlternateMethod(name ? :string):void { /* Looks like a ternary operator */ }
```

which is not quite, what I want to have.
For me, it would be very nice, if I could define this syntax:

``` Javascript
public myMethod(name? :string):void { /* empty*/ }
```

Am I overseeing a whitespace-rule for the optional flag?

JKillian:
This seems like a bug to me. I can't imagine anyone actually wants a space between the name and the ? of parameters (or fields or anything similar). Unfortunately there isn't any workaround that I know of.

b0r3as:
Shouldn't this be fixed with #958? Before the PR, the rule would take the end of the parameter name, advance one character and check that one for being a space (which is obviously wrong for the optional flag). The PR fixed this by explicitly searching for the colon and checking the character in front of it.

adidahiya:
@b0r3as Good call, I think you're right. We should add some regression tests to verify the fix.

th0r:
Another lint error that I faced after updating tslint from `5.8.0` to `5.9.1`:

![image](https://user-images.githubusercontent.com/302213/38558554-2b309bf4-3cd9-11e8-9225-1108cf568c93.png)

I use `?` to specify optional class property (and seems like it's a valid Typescript).

liuxh0:
Same issue using tslint 5.11.0

JoshuaKGoldberg:
Per #4534, this is an old issue with a formatting concern that has aged away.

ataravati:
Why is this issue closed without being resolved?

JoshuaKGoldberg:
Hi @ataravati! See #4534 (as linked in the comment just above yours). TSLint is deprecated and we're no longer accepting PRs for non-critical bugs such as this one.

ataravati:
Thank you, @JoshuaKGoldberg!

JoshuaKGoldberg:
[...]

PR #1184, opened 2016-04-29
Rule - Typedef-Whitespace - allow optional

mattslocum:
* allow QuestionToken before the colon

### DIFF
--- a/test/rules/typedef-whitespace/both/onespace/test.ts.lint
+++ b/test/rules/typedef-whitespace/both/onespace/test.ts.lint
@@ -140,13 +140,13 @@ interface WithPreceedingSpacesWithSubsequentSpacesInterface {
 
 // region Function
 
-function noPreceedingSpaceNoSubsequentSpaceFn(a:number, b:number):number {
+function noPreceedingSpaceNoSubsequentSpaceFn(a:number, b?:number):number {
                                               ~                              [expected onespace before colon in parameter]
                                                 ~                            [expected onespace after colon in parameter]
                                                         ~                    [expected onespace before colon in parameter]
-                                                          ~                  [expected onespace after colon in parameter]
-                                                                ~            [expected onespace before colon in call-signature]
-                                                                  ~          [expected onespace after colon in call-signature]
+                                                           ~                 [expected onespace after colon in parameter]
+                                                                 ~           [expected onespace before colon in call-signature]
+                                                                   ~         [expected onespace after colon in call-signature]
     var c:number = a + b;
         ~            [expected onespace before colon in variable-declaration]
           ~          [expected onespace after colon in variable-declaration]
@@ -163,10 +163,10 @@ function noPreceedingSpaceNoSubsequentSpaceFn(a:number, b:number):number {
     }
 }
 
-function noPreceedingSpaceWithSubsequentSpaceFn(a: number, b: number): number {
+function noPreceedingSpaceWithSubsequentSpaceFn(a: number, b?: number): number {
                                                 ~                              [expected onespace before colon in parameter]
                                                            ~                   [expected onespace before colon in parameter]
-                                                                    ~          [expected onespace before colon in call-signature]
+                                                                     ~         [expected onespace before colon in call-signature]
     var c: number = a + b;
         ~          [expected onespace before colon in variable-declaration]
     var d: number = a - b;
@@ -180,13 +180,13 @@ function noPreceedingSpaceWithSubsequentSpaceFn(a: number, b: number): number {
     }
 }
 
-function noPreceedingSpaceWithSubsequentSpacesFn(a:   number, b:   number):   number {
+function noPreceedingSpaceWithSubsequentSpacesFn(a:   number, b?:   number):   number {
                                                  ~                                    [expected onespace before colon in parameter]
                                                    ~                                  [expected onespace after colon in parameter]
                                                               ~                       [expected onespace before colon in parameter]
-                                                                ~                     [expected onespace after colon in parameter]
-                                                                         ~            [expected onespace before colon in call-signature]
-                                                                           ~          [expected onespace after colon in call-signature]
+                                                                 ~                    [expected onespace after colon in parameter]
+                                                                          ~           [expected onespace before colon in call-signature]
+                                                                            ~         [expected onespace after colon in call-signature]
     var c:   number = a + b;
         ~            [expected onespace before colon in variable-declaration]
           ~          [expected onespace after colon in variable-declaration]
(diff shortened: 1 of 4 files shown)


mattslocum:
This fixes https://github.com/palantir/tslint/issues/811

mattslocum:
I just realized that I think I broke multi-space rule for a case with 2 spaces. I'll fix it, and I'll write unit tests for my change and the bug.

JKillian:
Hi @mattslocum, thanks for the PR! According to [this comment](https://github.com/palantir/tslint/issues/811#issuecomment-182607753), #811 might already be fixed? Have you been seeing the problem described in #811 in your usage of TSLint?

mattslocum:
@JKillian I am still having the problem with 3.5.0. I have a code change that I also wanted to submit with unit tests, but I've never seen your style of tests. There are tildes and square brackets under function definitions. I can kids see what is going on, but not totally. Can you point me to some documentation on this style?

JKillian:
Here's a little write-up about the tests: http://palantir.github.io/tslint/develop/testing-rules/

Long story short, those test files you're looking at are just normal TypeScript files, except for the tildes and square brackets. The tildes mark where the linter should produce a lint failure, and the message in brackets next to it marks what the lint failure message should be

mattslocum:
Sorry. that took so long. I'm at ng-conf now. Hope this helps all the new typescript users that want spaces before the ':'

test/rules/typedef-whitespace/both/onespace/test.ts.lint
```diff
@@ -220,7 +220,7 @@ function withPreceedingSpaceNoSubsequentSpaceFn(a :number, b :number) :number {
     }
 }
 
-function withPreceedingSpaceWithSubsequentSpaceFn(a : number, b : number) : number {
+function withPreceedingSpaceWithSubsequentSpaceFn(a : number, b ?: number) : number {
```
JKillian: Do we want to allow this? Ideally I'd think it should be `b? : number`? Or is it customary to put the space before the question mark?

JKillian:
No worries on any delays @mattslocum! I'm still a little confused what this is tweaking though. Can you give a quick example of code that generated an error before that is now allowed (or vice versa)?

test/rules/typedef-whitespace/both/onespace/test.ts.lint
mattslocum: That is the coding standard that we use at my work. Both the ? and : are typescript characters. In my mind the ? is a modifier of the :, so we require the ? and : to be together, but a space before and after them.

I was actually just thinking if we should allow
`a ? : number`
but I was hoping to leave that to someone else, if that is okay.

mattslocum:
Oops forgot interface tests. Incoming...

test/rules/typedef-whitespace/both/onespace/test.ts.lint
JKillian: Ah, I see what you're getting at. So with your tweaks, the rule would allow either `foo? : number` or `foo ?: number`? (Assuming the settings are for one space on both sides of the colon.)

mattslocum: That is correct. It would allow both styles. I personally don't like typescript characters up against my JS variables. So I was going for the 2nd example you gave.

mattslocum:
Here's an example of the existing problem:
```
export interface IUser {
    name : string;
    age ?: number;
}

class Ctrl {
    private setOptions(
        user : IUser,
        favoriteColor ?: string
    ) : void {
    }
}
```
Without my pull request, I get the following errors:
```
[16:03:12] [gulp-tslint] error (typedef-whitespace) example.ts[3, 9]: expected space before colon in property-declaration
[16:03:12] [gulp-tslint] error (typedef-whitespace) example.ts[9, 23]: expected space before colon in parameter
```

test/rules/typedef-whitespace/both/onespace/test.ts.lint
adidahiya: most TS code I've seen, including ambient typings on DefinitelyTyped, uses the `b?: number` convention with no space before the `?`. so I'm inclined to disagree with this for the core TS rule. if you wish to enforce `b ?: number` in your codebase, then I think you are best served by a custom rule.

mattslocum: @adidahiya I don't mind if you want to do `b?: number`, but I should also be able to do `b ?: number`. There is truly a bug in that if I have it configured to require a space that it should allow the question mark next to the colon. You don't have to use this feature if you don't want. 

This has been a bug since November #811. This used to work back with tslint 3.3.0, and now I'm stuck on this version because no-one is fixing this. So I fixed it myself. Now it works like it used to. It isn't a new feature.

JKillian: I'll ask around and see if `varName ?: type` is used elsewhere in the TS community and considered good practice. If it's decently used, we can proceed with this PR. If not, you can always compile this rule yourself and use it as a custom rule.

Worth noting though that the original commenter in #811 wanted the space to come after the `?`, not before.

JKillian: I took a survey of random TS developers, and found that this syntax wasn't really used elsewhere. (Overwhelmingly people used `varName?: type`.) So unfortunately that makes this PR a change that's not appropriate for core TSLint. However, I would definitely recommend making a custom rule and using it at your company! Obviously you're welcome to use the current rule in TSLint as your starting point

mattslocum:
So just to clarify. There is already a configuration for whitespace required before the colon.

possible existing config:
```
"typedef-whitespace": [
      true,
      {
          "call-signature": "space",
          "index-signature": "space",
          "parameter": "space",
          "property-declaration": "space",
          "variable-declaration": "space"
      },
      {
          "call-signature": "space",
          "index-signature": "space",
          "parameter": "space",
          "property-declaration": "space",
          "variable-declaration": "space"
      }
],
```

This config will cause the following code to fail:
```
public myFunction(name ?: string, value ?: number) { }
```
But this is the only way to make it pass:
```
public myFunction(name? : string, value? : number) { }
```
So someone at one time thought it was acceptable to make a linter rule to require a space before the colon, but not before the question mark. I would think that if someone wanted to require a space (as the rule implies), that they should be able to put the `?` **optionally** against the `:`, as both the `?` and the `:` are typescript syntax and not JS syntax. 

So should we completely remove the rule to have a space before the 'colon' because the "survey of random TS developers" doesn't like the original rule?

ataravati:
@mattslocum This issue has not been resolved yet?

adidahiya:
TSLint is deprecated, see #4534

Issue #1054, opened 2016-03-24
New rule: Equals signs can't be used to check if variable is NaN (use isNaN() instead)


atsu85:
Equals signs can't be used to check if variable is `NaN`, `isNaN()` function should be used instead.

#### Currently following incorrect usage doesn't raise an error

Example of **incorrect** usage:
```ts
const maybeNumber = parseInt("a");
if( maybeNumber === NaN || maybeNumber === NaN ) {
  // this line is never reached <- UNEXPECTED
}
```
#### Example of correct usage
```ts
const maybeNumber = parseInt("a");
if( isNaN(maybeNumber) ) {
  // this line is reached, as expected
}
```

Avol-V:
You can use this: https://github.com/Microsoft/tslint-microsoft-contrib/blob/master/src/useIsnanRule.ts

myitcv:
@HamletDRC - has there been any thought to contributing these Microsoft rules back to the main `tslint` repo?

atsu85:
@HamletDRC, that would be nice :)

HamletDRC:
I am happy to move any of the rules to tslint. It is up to someone from the tslint team to go through the tslint-microsoft-contrib rulesset and tell me which ones they want. I am not going to create 60 new PR requests for tslint just to have 55 of them closed as "rule is too specific for teslint". So the ball is in Palentir's court. I am happy to do the work of porting the rules, but tslint must state which rules they want and which rules they do not want.

adidahiya:
@HamletDRC fair enough :) thanks for the offer. We'll take a look at the tslint-microsoft-contrib rule set. In the meantime, the new `extends` feature should make sharing configs easier (once the bugs are ironed out this week).

adidahiya:
Feel free to send a PR for this one in particular (and any others which have the "Status: Accepting PRs" label)

PR #1185, opened 2016-04-29
[Issue #1054] use-isnan fixed to ignore assignments to NaN

HamletDRC:
Fixes bug on issue #1054

### DIFF
--- a/src/rules/useIsnanRule.ts
+++ b/src/rules/useIsnanRule.ts
@@ -32,7 +32,8 @@ export class Rule extends Lint.Rules.AbstractRule {
 class UseIsnanRuleWalker extends Lint.RuleWalker {
 
     protected visitBinaryExpression(node: ts.BinaryExpression): void {
-        if (this.isExpressionNaN(node.left) || this.isExpressionNaN(node.right)) {
+        if (this.isExpressionNaN(node.left) || this.isExpressionNaN(node.right)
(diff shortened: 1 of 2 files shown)


src/rules/useIsnanRule.ts
```diff
@@ -32,7 +32,8 @@ export class Rule extends Lint.Rules.AbstractRule {
 class UseIsnanRuleWalker extends Lint.RuleWalker {
 
     protected visitBinaryExpression(node: ts.BinaryExpression): void {
-        if (this.isExpressionNaN(node.left) || this.isExpressionNaN(node.right)) {
+        if (this.isExpressionNaN(node.left) || this.isExpressionNaN(node.right)
```
JKillian: So considering operator precedence, this actually evaluates as:

```
this.isExpressionNaN(node.left) || 
(this.isExpressionNaN(node.right) && node.operatoreToken.kind !== ts.SyntaxKind.EqualsToken)
```

The way it's written it kind of implies the opposite. However, I think the behavior is actually correct... you'd never want an `NaN` on the left side, would you?

HamletDRC: ah well, I did make a mistake in the parens. But it does work either way.

JKillian: Right, no worries, PR merged!

PR #1190, opened 2016-04-30
no-inferrable-types: Add ignore-params flag

abierbaum:
This change allows using no-inferrable types at times when you
may want to still require types for method/function parameter lists
(for example when using the typedef rule)

It helps to workaround the issues in #711


src/rules/noInferrableTypesRule.ts
```diff
@@ -27,13 +27,23 @@ export class Rule extends Lint.Rules.AbstractRule {
 }
 
 class NoInferrableTypesWalker extends Lint.RuleWalker {
+    public ignoreParameters: boolean = false;
```
JKillian: Define a `const OPTION_IGNORE_PARAMS = "ignore-params";` above the `class Rule` declaration.

src/rules/noInferrableTypesRule.ts
```diff
@@ -27,13 +27,23 @@ export class Rule extends Lint.Rules.AbstractRule {
...
+
+    constructor(sourceFile: ts.SourceFile, options: Lint.IOptions) {
```
JKillian: And you should be able to delete this constructor all together

src/rules/noInferrableTypesRule.ts
```diff
@@ -27,13 +27,23 @@ export class Rule extends Lint.Rules.AbstractRule {
...
+    }
+
     public visitVariableDeclaration(node: ts.VariableDeclaration) {
         this.checkDeclaration(node);
         super.visitVariableDeclaration(node);
     }
 
     public visitParameterDeclaration(node: ts.ParameterDeclaration) {
-        this.checkDeclaration(node);
+        if (!this.ignoreParameters) {
```
JKillian: then here you can just do `if (!this.hasOption(OPTION_IGNORE_PARAMS))`

JKillian:
Looks good! Can you just make the change noted above to how you handle the option provided? This will make it more consistent with other rules

src/rules/noInferrableTypesRule.ts
abierbaum: done.

src/rules/noInferrableTypesRule.ts
abierbaum: done

src/rules/noInferrableTypesRule.ts
abierbaum: done

abierbaum:
@JKillian I made the changes suggested and pushed up a new commit with everything fixed up.  Let me know if there is anything more to change.

JKillian:
All good, thanks @abierbaum!

Issue #525, opened 2015-07-23
Rule to limit variable statements to a single declaration


DanielRosenwasser:
It would be nice if we could disallow the following in our codebase:

```TypeScript
var x: string, y: number;
```

ashwinr:
to clairfy: do you mean disallow multiple variable definitions in the same statement?

DanielRosenwasser:
Yup, sorry about that. It was fairly late.

adidahiya:
Looks like the corresponding eslint rule for this is called `one-var`: http://eslint.org/docs/rules/one-var

ashwinr:
:+1:

YuichiNukiyama:
related PR: #1194

JKillian:
Decided not to go with `one-var` as it's a somewhat brief name, instead we'll use `no-multiple-variable-declaration`, which fits TSLint's style better hopefully. Thanks to @YuichiNukiyama for implementing this! It'll be around in the next release. (Still up for renaming things if people feel strongly about it one way or the other)

JKillian:
Actually, I think `no-multiple-variable-declaration` is ambiguous in which style it's actually disallowing. Now thinking about `one-variable-per-declaration` instead

PR #1191, opened 2016-05-01
Add no-multiple-var rule

YuichiNukiyama:
related issue #525.

- for another linter compatibility, I define target options (`var`,
`let` and `const`)
- I check `for statement`, but don't check `for...in statement` and
`for...of statement`. becase, TS disallow multiple variable definitions
in the same statement by default


JKillian:
Thanks for the PR @YuichiNukiyama!

> for another linter compatibility, I define target options (var, let and const)
I'm not sure I understand why we need these options. I think this rule could maybe be simplified and not have any options? What do you think?

YuichiNukiyama:
@JKillian If you prefer simplicity to compatibility, I remove options (use only `true` option) and validate all of them (`var`, `let` and `const`) together.

JKillian:
@YuichiNukiyama No options seems best to me in this case. In what scenarios do you think people might want the different options?

YuichiNukiyama:
Speaking honestly, I don't want these options. So, I follow your idea.

JKillian:
Looks a lot better, thanks @YuichiNukiyama! One more idea, we might want an "ignore-for-loop" option to make the rule not check for loop declarations?

README.md
```diff
@@ -242,6 +242,8 @@ A sample configuration file with all options is available [here](https://github.
 * `no-internal-module` disallows internal `module` (use `namespace` instead).
 * `no-invalid-this` disallows using the `this` keyword outside of classes.
     * `no-this-in-function-in-method` disallows using the `this` keyword in functions within class methods.
+* `no-multiple-var` disallows multiple variable definitions in the same statement.
+* `no-use-before-declare` disallows usage of variables before their declaration.
```
JKillian: Also, can you delete this line?

YuichiNukiyama:
@JKillian `ignore-for-loop` option is good idea.  I' ll send new issue and PR.

PR #1193, opened 2016-05-02
Migrate enabledisable test to new testing system

JKillian:
@patsissons Just if you're curious, this is what the test looks like migrated to the new system! Exact same functionality, so much less code 😃


adidahiya:
👍

PR #1209, opened 2016-05-04
adding no-default-export rule

### LINKED ISSUE #1182: New rule: no-default-export
In my experience, `default` exports are not a great practice, especially anonymous ones. I think you should always name the things you are exporting. Also, default exports allow for subtle mistakes after refactoring. With default exports, you could change the semantics of a component but it might be imported elsewhere with a name suggesting its old semantic meaning.

```ts
// app.ts
import ActionPopover from "./action";
```

```ts
// action.ts
export default class ActionMenu {
    ...
}
// oops, we just renamed this from popover -> menu but our consumer had no idea!
```

patsissons:
I believe this is all that is required for this rule. open to any suggestions for improvements.

p.s. love the new testing framework!

resolves #1182

### DIFF
--- a/test/rules/no-default-export/test.ts.lint
+++ b/test/rules/no-default-export/test.ts.lint
@@ -1,3 +1,17 @@
+export { SingleItem };
+
+export { FirstItem, SecondItem };
+
+export { Item as aliasForItem };
+
+export { ItemFromImport } from 'module';
+
+export * from './relativeModule';
+
+export namespace SomeNamespace { }
+
+export import ItemAlias = someNamespace.Item;
+
 export const default = 'VALID';
 
 export default 'INVALID';
--- a/test/rules/no-default-export/test.ts.lint
+++ b/test/rules/no-default-export/test.ts.lint
@@ -1,0 +1,4 @@
+export const default = 'VALID';
+
+export default 'INVALID';
+       ~~~~~~~ [use of default exports is disallowed]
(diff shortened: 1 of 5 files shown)


test/rules/no-default-export/test.ts.lint
```diff
@@ -0,0 +1,4 @@
+export const default = 'VALID';
+
+export default 'INVALID';
+       ~~~~~~~ [use of default exports is disallowed]
```
JKillian: can you add a few more tests so we make sure we're not getting false positives?

```
export { Stream, writeToStream as write };
export { read, write, standardOutput as stdout } from "./inout";
export * from "./mod1";
```

test/rules/no-default-export/test.ts.lint
```diff
@@ -0,0 +1,4 @@
+export const default = 'VALID';
```
JKillian: do you think this should be valid? Under ES6 standards, this is actually the same as using a default export! (see https://github.com/Microsoft/TypeScript/issues/2242#issue-60179656). I'm leaning towards allowing this though

JKillian:
Glad you're liking the new testing framework - makes testing so much less painful! PR looks good to me other than the one note above

test/rules/no-default-export/test.ts.lint
patsissons: it is technically the same, but expressive instead of implicit. I feel like it should be allowed because it is not using the default export keyword.

test/rules/no-default-export/test.ts.lint
patsissons: I really didn't know what else to write down as a test, but tbh i didn't spend much time thinking about it. I'll add the above in there for good measure.

JKillian:
Merged, thanks!

Issue #1208, opened 2016-05-04
Make member-ordering rule more flexible


JKillian:
**New Options Syntax**

All of the below examples would be supported:

Full customization if needed:

```json
"member-ordering": [true, { "order": [
  "public-static-field",
  "protected-static-field",
  "private-static-field",
  "public-instance-field",
  "protected-instance-field",
  "private-instance-field",
  "constructor",
  "public-static-method",
  "protected-static-method",
  "private-static-method",
  "public-instance-method",
  "protected-instance-method",
  "private-instance-method",
]}]
```

Preset orderings (what most people would use):
```
// would do something similar to above
"member-ordering": [true, { "order": "smart" }]
```

Middle of the road orderings (although I'm debating if this is worth implementing):

```
"member-ordering": [true, { "order": [
  "static-field",        // public before private assumed
  "instance-field",
  "constructor",
  "static-method",
  "instance-method",
]}]
```

and

```
"member-ordering": [true, { "order": [
  "constructor",
  "static",       // field before method assumed and public before private assumed
  "instance",
]}]
```

lowkay:
Yupp this looks like a much neater way to do it. I think your approach to have a good default is a good choice.

I think that middle road presets would add unnecessary complexity to tslint. Start simple and if there's lots of noise think about middle ground.

Looking forward to seeing this implemented :)

JKillian:
👍  on avoiding the middle of the road options. These would add implementation and documentation complexity,  and for the most part the first two options should cover what people need.

Also interested in what presets people would like to see, and what they should be named, if anyone has ideas.

JKillian:
Closed by #1210

Issue #1186, opened 2016-04-29
`public-before-private` and `static-before-instance` are mutually exclusive


thasner:
### Bug Report

- __TSLint version__: `3.7.1`
- __TypeScript version__: `1.8.7`
- __Running TSLint via__: CLI

#### TypeScript code being linted

```ts
public static foo = 2;
private static bar = 3;
public x: string;
private y: string;
```

with `tslint.json`:

```json
{
    "member-ordering": [true,
      "public-before-private",
      "static-before-instance",
      "variables-before-functions"
    ],
}
```

#### Actual behavior

> Declaration of private static member variable not allowed to appear after declaration of public instance member variable

#### Expected behavior

no error

adidahiya:
Better support for member ordering will be available in TSLint 3.10. These options, `public-before-private` and `static-before-instance`, are now deprecated and will be removed in v4 (correct me if I'm wrong here @JKillian).

thasner:
this is awesome, thanks everyone!

i'd love to use this asap. could you release a minor version?

JKillian:
Yep, coming today!

PR #1210, opened 2016-05-04
Next-gen member-ordering rule

### LINKED ISSUE #1205: Member-ordering issue with constuctors
### Bug Report

- __TSLint version__: `3.9.0`
- __TypeScript version__: `1.8.9`
- __Running TSLint via__: Gulp

```ts
export class SkeletonDirective {
	/*@ngInject*/
	constructor(
	) {
		// todo: constructor injection and dependency initialization here
	}

	static factory(): IDirectiveFactory {
		/*@ngInject*/
		const directive = (
		) => {
			return new SkeletonDirective();
		};
		return directive;
	}
};
```

with `tslint.json`:

```json
		"member-ordering": [
			true,
			"public-before-private",
			"static-before-instance",
			"variables-before-functions"
		],
```

#### Actual behavior

Declaration of public static member function not allowed to appear after declaration of public instance member function

#### Expected behavior

No warning as it is a constuctor

JKillian:
Adds an optional new syntax for more explicit control of the `member-ordering` rule (see #1208).

This should resolve #1205 and #1186 by giving a user more power to define their own orderings. This keeps backwards-compatibility with the old rule options (which will be removed in TSLint 4.0). This also allows for the possibility of further fields, like an `alphabetical: true` option.

### DIFF
--- a/src/rules/memberOrderingRule.ts
+++ b/src/rules/memberOrderingRule.ts
@@ -17,9 +17,62 @@
 import * as Lint from "../lint";
 import * as ts from "typescript";
 
+/* start old options */
 const OPTION_VARIABLES_BEFORE_FUNCTIONS = "variables-before-functions";
 const OPTION_STATIC_BEFORE_INSTANCE = "static-before-instance";
 const OPTION_PUBLIC_BEFORE_PRIVATE = "public-before-private";
+/* end old options */
+
+/* start new options */
+const OPTION_ORDER = "order";
+const PRESET_ORDERS: { [preset: string]: string[] } = {
+    "fields-first": [
+        "public-static-field",
+        "protected-static-field",
+        "private-static-field",
+        "public-instance-field",
+        "protected-instance-field",
+        "private-instance-field",
+        "constructor",
+        "public-static-method",
+        "protected-static-method",
+        "private-static-method",
+        "public-instance-method",
+        "protected-instance-method",
+        "private-instance-method",
+    ],
+    "statics-first": [
+        "public-static-field",
+        "public-static-method",
+        "protected-static-field",
+        "protected-static-method",
+        "private-static-field",
+        "private-static-method",
+        "public-instance-field",
+        "protected-instance-field",
+        "private-instance-field",
+        "constructor",
+        "public-instance-method",
+        "protected-instance-method",
+        "private-instance-method",
+    ],
+    "instance-sandwich": [
+        "public-static-field",
+        "protected-static-field",
+        "private-static-field",
+        "public-instance-field",
+        "protected-instance-field",
+        "private-instance-field",
+        "constructor",
+        "public-instance-method",
+        "protected-instance-method",
+        "private-instance-method",
+        "public-static-method",
+        "protected-static-method",
+        "private-static-method",
+    ],
+};
+/* end new options */
 
 export class Rule extends Lint.Rules.AbstractRule {
     public apply(sourceFile: ts.SourceFile): Lint.RuleFailure[] {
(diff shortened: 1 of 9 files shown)


src/rules/memberOrderingRule.ts
```diff
@@ -49,36 +88,97 @@ function toString(modifiers: IModifiers): string {
...
-export class MemberOrderingWalker extends Lint.RuleWalker {
-    private previousMember: IModifiers;
+/* start new code */
+enum AccessLevel {
+    PRIVATE,
+    PROTECTED,
+    PUBLIC,
+}
+
+enum Bound {
```
JKillian: Is this the best name? Any ideas on better names?

src/rules/memberOrderingRule.ts
```diff
@@ -17,16 +17,55 @@
...
 
+/* start old options */
 const OPTION_VARIABLES_BEFORE_FUNCTIONS = "variables-before-functions";
 const OPTION_STATIC_BEFORE_INSTANCE = "static-before-instance";
 const OPTION_PUBLIC_BEFORE_PRIVATE = "public-before-private";
+/* end old options */
+
+/* start new options */
+const OPTION_ORDER = "order";
+const PRESET_ORDERS: { [preset: string]: string[] } = {
```
JKillian: @lowkay @thasner @alan-agius4 (and others) would you use one of these presets? If not, what preset would you like?

lowkay: @jkillian Yupp, fields-first would be my preference.

alan-agius4: Most likely yes and I'd go for fields-first.

JKillian: @marines also

src/rules/memberOrderingRule.ts
marines: How about?
```java
enum Membership {
    CLASS,
    INSTANCE
}
```

src/rules/memberOrderingRule.ts
```diff
@@ -17,16 +17,70 @@
...
+        "private-static-method",
+        "public-instance-field",
+        "protected-instance-field",
+        "private-instance-field",
+        "constructor",
+        "public-instance-method",
+        "protected-instance-method",
+        "private-instance-method",
+    ],
+    "instance-sandwich": [
```
JKillian: added this for you @thasner

src/rules/memberOrderingRule.ts
marines: Looks like the instance-sandwich will suit me the best but as long as I can order the rules by myself I'm totally fine with that.

src/rules/memberOrderingRule.ts
JKillian: I like that better, thanks!

src/rules/memberOrderingRule.ts
thasner: 👍

adidahiya:
code lgtm, but needs tests for the other presets

JKillian:
> code lgtm, but needs tests for the other presets
Think so? I was thinking that it wasn't really necessary to have tests for every preset

test/rules/member-ordering/order/fields-first/test.ts.lint
```diff
@@ -0,0 +1,39 @@
...
+    constructor() {}
+    public static g() {}
+    protected static h() {}
+    private static i = () => {};
+    j = () => {};
+    protected k() {};
+    private l = () => {};
+}
+    
+Interface IGood {
```
adidahiya: typo, `interface`

this is likely not being linted

adidahiya:
hm, ok, you're right that we don't need to test all the presets, but I'd like some more tests that check that access level modifiers are enforced

adidahiya:
👍

PR #1211, opened 2016-05-05
Fix path for tslintConfig in package.json

lijunle:
Do we want some test?


JKillian:
This is fine if `tslintConfig` is a `string`, bit what if it's a configuration object?

In all reality though, nobody should be using `tslintConfig` anymore, everyone should use a `tslint.json` file

lijunle:
The [changelog](https://github.com/palantir/tslint/blob/master/CHANGELOG.md#v046) shows it is a feature as of 0.4.6.

> [feature] package.json now contains a tslintConfig paramater to allow users to specify **the location of the configuration file** there

JKillian:
Interesting, although the way the code works now is that it loads an object specified there and not a path. I'd rather not change this code - although hopefully nobody is using it, if they are might as well not change things. Sorry!

Issue #1217, opened 2016-05-06
Single-line disabling flags also disable max-line-length for the next line


vitaminj:
### Bug Report
- **TSLint version**: `3.9.0`
- **TypeScript version**: `1.8.10`
- **Running TSLint via**: gulp-tslint / Visual Studio Code
#### TypeScript code being linted

Two long lines with the max-line-length rule.
First line either has disable-line on it, or disable-next-line before it.
#### Actual behavior

Linting passes.
#### Expected behavior

Linting should fail on second line.

PR with fix and testcases ready to be filed...

ajafff:
This should be fixed by #2062

PR #1218, opened 2016-05-06
Fixes line disabling flags from disabling max-line-length also on the following line (#1217)

vitaminj:
Fixes https://github.com/palantir/tslint/issues/1217

### DIFF
--- a/src/enableDisableRules.ts
+++ b/src/enableDisableRules.ts
@@ -81,7 +81,7 @@ export class EnableDisableRulesWalker extends SkippableTokenAwareRuleWalker {
                         // end at the beginning of the next line
                         this.enableDisableRuleMap[ruleToAdd].push({
                             isEnabled: !isEnabled,
-                            position: scanner.getTextPos() + 1,
+                            position: scanner.getTextPos(),
                         });
                     } else {
                         // start at the current position
@@ -93,7 +93,7 @@ export class EnableDisableRulesWalker extends SkippableTokenAwareRuleWalker {
                         if (isNextLine) {
                             this.enableDisableRuleMap[ruleToAdd].push({
                                 isEnabled: !isEnabled,
-                                position: this.getStartOfLinePosition(node, startingPosition, 2),
+                                position: this.getStartOfLinePosition(node, startingPosition, 2) - 1,
                             });
                         }
                     }
--- a/test/rules/_integration/enable-disable/test.ts.lint
+++ b/test/rules/_integration/enable-disable/test.ts.lint
@@ -38,8 +38,14 @@ var AAAaA = 'test' /* tslint:disable-line */
 
 // tslint:disable-next-line:quotemark variable-name
 var AAAaA = 'test'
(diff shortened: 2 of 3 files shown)


JKillian:
@patsissons, since you worked on this recently, do these changes look good to you?

patsissons:
My only suggestion would be to make sure that nested rules are not affected, I believe that's why I initially added the `+1` in there. But i had added that early on in development and so it may not be relevant any longer.

vitaminj:
Yes if you have suggestions for other tests to add? I added an end-of-line one to vaguely test off-by-one errors at that end, but couldn't immediately think of anything to test start-of-line?

src/enableDisableRules.ts
```diff
@@ -81,7 +81,7 @@ export class EnableDisableRulesWalker extends SkippableTokenAwareRuleWalker {
                         // end at the beginning of the next line
                         this.enableDisableRuleMap[ruleToAdd].push({
                             isEnabled: !isEnabled,
-                            position: scanner.getTextPos() + 1,
+                            position: scanner.getTextPos(),
```
JKillian: Just thinking, maybe this would be a better way to get the beginning of the next line?
`position: this.getStartOfLinePosition(node, startingPosition, 1) - 1,`

patsissons:
This is the best example test i can come up with off the top of my head, should validate both my original case and your new case:

```
// tslint:disable
// tslint:enable-next-line:max-line-length
var a = "testtesttesttesttesttesttesttesttesttesttesttesttest";
~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~ [exceeds maximum line length of 60]
var a = "testtesttesttesttesttesttesttesttesttesttesttesttest";
// tslint:enable
var a = "testtesttesttesttesttesttesttesttesttesttesttesttest";
~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~ [exceeds maximum line length of 60]

// tslint:disable-next-line:max-line-length
var a = "testtesttesttesttesttesttesttesttesttesttesttesttest";
var a = "testtesttesttesttesttesttesttesttesttesttesttesttest";
~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~ [exceeds maximum line length of 60]
```

JKillian:
@vitaminj Let me know what you think of the above comments. If everything seems as good as possible to you, we can merge this as well

patsissons:
i believe the test I added above will fail for one case with the current code, and fail for the other case in the proposed code (at least it did when i checked a few days ago).

JKillian:
Gotchya, thanks for the heads-up @patsissons. Let's make sure that @patsissons's test case is handled correctly with the new code before we merge then

adidahiya:
ping @vitaminj?

PR #1221, opened 2016-05-07
Bump NPM dependencies and switch to ESLint

### LINKED ISSUE #1220: Use ESLint instead of JSCS in our own build

adidahiya:
fixes #1220


package.json
```diff
@@ -22,21 +22,21 @@
   "dependencies": {
     "colors": "^1.1.2",
     "diff": "^2.2.1",
-    "findup-sync": "~0.3.0",
-    "glob": "^6.0.1",
+    "findup-sync": "~0.4.0",
```
JKillian: noooo undo!  (see https://github.com/palantir/tslint/pull/1108)

JKillian:
Also, build errors because of invalid peer dependencies :/

image

adidahiya:
yep, noticed the invalid peer deps... will try to investigate tomorrow

JKillian:
Related: https://github.com/unindented/grunt-npm-command/issues/7
I think that might solve the `peerDependency` issues

adidahiya:
I'd actually prefer to remove the grunt-npm-command dependency. It is a small package with very few users.

JKillian:
Sounds good, as long as things work correctly across platforms/CIs. Note https://github.com/palantir/tslint/pull/1172 which is the relevant discussion when it was added

JKillian:
`grunt-npm-command` released a new version to fix the peerDep error, so the build is passing now. I'm inclined to just let it stay for now, it's working well in our build.

adidahiya:
meh, ok. filed follow-up #1235

PR #1223, opened 2016-05-09
Add an option to ignore constructor in member ordering

### LINKED ISSUE #1222: Ignoring constructor in member ordering
Recently #1158 has been fixed however I'd like to opose to that since the constructor is a special kind of method so it shouldn't be treated as any other public method.

Let's take a look at the example:

```javascript
export interface IFoo {
    bar(): void;
}

class Foo implements IFoo {
    static $inject: string[] = ['$rootScope',
                                '$state',
                                '$location'];

    constructor(private $rootScope: ns.RootScope,
                private $state: angular.ui.IStateService,
                private $location: angular.ILocationService) {}

    private baz1: string;
    private baz2: number;

    bar(): void {
        // ...
    }

    private fooBar(): void {
        // ...
    }
}
```

I'm using all three `"public-before-private", "static-before-instance", "variables-before-functions"` options for "member-ordering" rule which force me to move the private properties above the constructor. Keeping them as they are makes the static $inject and the constructor look easy to read as they are related to each other. Having multiple private properties in between them would move them too far away.

marines:
Introduces "ignore-constructor" option for member-ordering rule which ignores constructor's order.

Fixes #1222

### DIFF
--- a/src/rules/memberOrderingRule.ts
+++ b/src/rules/memberOrderingRule.ts
@@ -20,6 +20,7 @@ import * as ts from "typescript";
 const OPTION_VARIABLES_BEFORE_FUNCTIONS = "variables-before-functions";
 const OPTION_STATIC_BEFORE_INSTANCE = "static-before-instance";
 const OPTION_PUBLIC_BEFORE_PRIVATE = "public-before-private";
+const OPTION_IGNORE_CONSTRUCTOR = "ignore-constructor";
 
 export class Rule extends Lint.Rules.AbstractRule {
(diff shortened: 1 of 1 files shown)


JKillian:
Hi @marines, thanks for the PR! I see why you want this option, but I believe #1210 will provide a more comprehensive and flexible solution to this issue. Feel free to check it out and make sure it'll meet your needs - if not, we can discuss ideas more.

JKillian:
Closed this as a duplicate, not because it's an identical duplicate, but because #1210 should provide a more complete solution

marines:
Agreed. It looks awesome! :)

PR #1224, opened 2016-05-09
README.md - no-switch-case-fall-through can be replaced with --noFall…

HamletDRC:
…throughCasesInSwitch

Updated the README.md file to explain that the no-switch-case-fall-through rule can be replaced by the TypeScript compiler flag named --noFallthroughCasesInSwitch


JKillian:
Thanks again!

PR #1225, opened 2016-05-09
README.md - no-unreachable is a tsc compiler flag

HamletDRC:
Updated README.md to explain that no-unreachable rule is a compiler flag on the tsc compiler that is enabled by default.


JKillian:
Thanks!

Issue #960, opened 2016-02-06
Options to force trailing comma on interfaces declaration


Strate:
I want to force trailing comma on interface declaration:
```ts
interface A {
  prop1: string,
  prop2: string,
  prop3: string,
}
```
Is there any option to do this?

JKillian:
The `trailing-comma` rule only checks array literals, object literals, named import lists, and destructuring constructs right now, so unfortunately there's no way to have object literal _types_ checked, whether in an interface declaration like this or as a declared type somewhere else.

Seems like a reasonable feature request though - I've tagged it as such.

adidahiya:
aren't fields in an interface supposed to end with a trailing semicolon?

http://www.typescriptlang.org/Handbook#interfaces

JKillian:
I had thought so also, but [the Interface grammar](https://github.com/Microsoft/TypeScript/blob/master/doc/spec.md#71-interface-declarations) allows for [semicolons or commas](https://github.com/Microsoft/TypeScript/blob/master/doc/spec.md#383-object-type-literals).

JKillian:
See #1233 for discussion on this issue. The "ignore-interfaces" option for the `semicolon` rule will let users make a custom rule for this request if desired.

schickling:
Since version 4.0 this seems to be always enforced when using `multiline: "always"`. Is there a way to allow to omit trailing commas for interfaces? I imagine the following: 

```json
  "rules": {
    "trailing-comma": [
      true,
      {
        "multiline": "always",
        "singleline": "never"
      },
      "ignore-interfaces"
    ]
  }
```

Unfortunately this change seemed to have broke all of our build once we upgraded to v4.

schickling:
Maybe @JKillian knows something about this?

adidahiya:
@schickling the `"ignore-interfaces"` is supported in the semicolon rule, not the trailing-comma rule (see the linked PR #1233). What exact format are you looking to write interfaces with? I'm a little hesitant to add more options to this rule, especially since TSLint v4 will automatically fix all violations of the trailing-comma rule for you.

schickling:
This is the way how we write interfaces:

```ts
interface Post {
  title: string
  age: number
}
```

Zjaaspoer:
We're writing our interfaces the exact same way as @schickling (basically interfaces without semicolons). Right now Tslint is adding trailing commas to the last row like so:

```
interface Post {
  title: string
  age: number,
}
```

[...]

PR #1233, opened 2016-05-10
Option to ignore interface declarations in the Semicolon rule

schmuli:
As discussed in palantir/tslint#960, this adds a new rule to check for a semicolon or comma for property declarations in Interfaces or Object Type literals.

In order for this PR not to overlap with the existing semicolon rule, I removed support for Interfaces from that rule.

Currently, the name of the rule is 'object-type-eol', however that doesn't sound so great, so any other names are welcome. The name is based on the fact that the declarations in an Interface are described as Object Type in the language specification.

The rule works similar to the semicolon rule: if enabled it defaults to 'semicolons', but if configured with 'comma' it will use that instead.

I would have liked to update the documentation for the site, but that doesn't appear to be in the repository.


JKillian:
Thanks @schmuli!

I'm a little worried though about removing the semicolon checking for interfaces from the semicolon rule. This seems like something people might not want.

But I understand why people use commas in situations like these:

```
interface Foo {
  field1: string,
  field2: string,
}
```

has a certain symmetry with

```
type Foo = {
  field1: string,
  field2: string,
}
```

for example.

Any thoughts @adidahiya?

schmuli:
I was thinking of that too, however since each rule is not aware of other enabled rules and their configuration (rightly so), using both rules would either be checking the same thing twice or would produce contradicting failures.

Perhaps, instead of removing Interfaces from 'semicolon', it could be another configuration option, something as simple as `'check-interfaces'`?

JKillian:
Yes, I think an "ignore-interfaces" option for `semicolon` would be a good way to go here. Then it would play nicely with `object-type-eol`. I also like that this is fully backwards-compatible, behavior for people doesn't change unless they decide to change their `tslint.json` configuration

src/rules/objectTypeEolRule.ts
```diff
@@ -0,0 +1,71 @@
...
+
+    protected visitInterfaceDeclaration(node: ts.InterfaceDeclaration) {
+        node.members.forEach(member => {
+            this.checkSemicolonAt(member);
+        });
+
+        super.visitInterfaceDeclaration(node);
+    }
+
+    private checkSemicolonAt(node: ts.Node) {
```
JKillian: `checkSemicolonAt` seems like a slightly confusing name, actually, could probably just role this into `checkLastToken` and shorten the code a little

src/rules/objectTypeEolRule.ts
```diff
@@ -0,0 +1,71 @@
...
+
+        const last = children[children.length - 1];
+        if (last.kind !== expectedKind) {
+            let position = 0;
+            let width = 0;
+            if (last.kind === unwantedKind) {
+                position = last.getStart(sourceFile);
+                width = 1;
+            } else {
+                position = node.getStart(sourceFile) + node.getWidth(sourceFile);
```
JKillian: not a big fan of ending up with zero-width failures. Let's just make it have a width of 1 and be on the character before?

src/rules/semicolonRule.ts
```diff
@@ -91,11 +92,15 @@ class SemicolonWalker extends Lint.RuleWalker {
...
+        if (this.hasOption(OPTION_IGNORE_INTERFACES)) {
+            return;
+        }
+
         for (let member of node.members) {
             this.checkSemicolonAt(member);
         }
         super.visitInterfaceDeclaration(node);
-    }
+     }
```
JKillian: extra space got added here by accident

test/rules/semicolon/always/test.ts.lint
```diff
@@ -71,7 +71,7 @@ interface ITest {
                ~nil [missing semicolon]
     baz: boolean;
 }
-
+
```
JKillian: extra space here

test/rules/semicolon/never/test.ts.lint
```diff
@@ -65,11 +65,11 @@ class MyClass {
 }
 
 interface ITest {
-    foo?: string;
-                ~ [unnecessary semicolon]
-    bar: number;
-               ~ [unnecessary semicolon]
-    baz: boolean
+    foo?: string;
```
JKillian: Looks like extra whitespace got added here as well

JKillian:
One big issue remains I think - how does this rule interact with single line interfaces and object literal types? We'll need some test cases for that I think. And that also brings up the naming question, as wer're not only dealing with eol's now

adidahiya:
Sorry for the late feedback. I'm actually not sure if I'm on board with the whole idea. It feels weird to have multiple rules in TSLint _core_ which enforce different conventions for terminating lines in interfaces (semicolon and this one). Especially now that we have shareable configs with `extends`, I think we can be a little more opinionated about what we bring into core. This rule should exist outside of the `tslint` package, and if you would like to share it @schmuli, you can publish an NPM package that allows others to use it (we list such custom rules packages in our README). And if users _really_ want it in core, they can vote on https://github.com/palantir/tslint/issues/960 with +1s and we might reconsider.

JKillian:
That seems reasonable, as this rule does end up semantically conflicting with the `semicolon` rule (although the "ignore-interfaces" option above prevents actual conflicts of course).

On the other hand, even if this were a custom rule, it would still have the same issue with the built-in `semicolon` rule which would mean that the package of custom rules would have to provide an almost-duplicate semicolon-like rule?

And finally, this brings up the action point for us that we should make sure to not put the "Accepting PRs" tag on issues that we might aren't accept PRs for. 😳

schmuli:
The whole reason this discussion (i.e. #960) came up, was because of limitations of the Semicolon rule as it is today:
1. Currently, the Semicolon rule will add a failure if an interface property doesn't terminate with a semicolon, even if it terminates with a comma, which is valid TS
2. Currently, the Semicolon rule ignores Object Type Literals entirely

In order for this new rule to work, it depends on the addition of an `'ingore-interfaces'` flag in the Semicolon rule, which will be required even if the new Rule is not accepted into TSLint core.

Regarding single-line interfaces and object type literals, I was actually thinking of ignoring any instances with just one property, even when not on one line, for example for delegate interfaces:

```javascript
interface Action<T> {
    (value: T): void
}
```

src/rules/objectTypeEolRule.ts
schmuli: This is actually based on the logic in the semicolon rule: https://github.com/palantir/tslint/blob/master/src/rules/semicolonRule.ts#L114

src/rules/objectTypeEolRule.ts
schmuli: `checkSemicolonAt` is the same name in the Semicolon rule. The functions are split because the `checkLastToken` code is reused between semicolons and commas, depending on the configured options.

adidahiya:
Ok you're right about the ignore-interfaces option. We will take a PR for that.

schmuli:
So, just to be clear on what is agreed:

TSLint will add the `'ignore-interfaces'` option for the Semicolon rule (PR pending), but the new `object-type-eol` rule will not be merged into core.

If this is correct, I am perfectly happy with this decision.

adidahiya:
@schmuli yes, that's correct. You may revise this PR or open up a new one, whichever suits you.

schmuli:
I will revise this PR soon, thank you

JKillian:
👍 This seems like a good resolution, makes sense to me

JKillian:
Thanks @schmuli! Sorry for the delay in merging this. I plan to do a release sometime in the next week, so this will be available then

Issue #1241, opened 2016-05-10
`no-default-export` missing cases


JKillian:
### Bug Report

- __TSLint version__: `3.10.0`
- __TypeScript version__: `1.8.10`
- __Running TSLint via__: CLI

#### TypeScript code being linted

```ts
export default class Test {}
```

with `tslint.json`:

```json
    "no-default-export": true
```

#### Actual behavior
No lint failure reported.

#### Expected behavior
Should return a lint failure.

cc @patsissons

patsissons:
I'm surprised that does not fail.. I can take a peek right now.

JKillian:
Me too! I think it's because it's classified as a class declaration, which makes things trickier. By the way, do you have a good TS AST viewer (I use atom-typescript's AST feature)? Super helpful for TSLint code-writing!

patsissons:
that sounds super handy. I'm using vscode as my daily IDE these days, but that AST feature sounds pretty handy for this scenario. Doesn't look like a similar feature exists in vscode land.

patsissons:
any other suggestions for additional tests?

I came up with these:
```ts
export default { }

export default class Test { }

export default function() { }
```

I don't use `export default`'s so it's hard to remember what other strange things you can do with `export default`.

JKillian:
I use vim and a TS plugin as my daily IDE, but always have atom-typescript on hand for AST viewing 😄 .   The AST feature is described here: https://github.com/TypeStrong/atom-typescript#ast-visualizer

Couple notes on the atom-typescript AST:
* the regular AST view shows the AST as if using `ts.forEachChild` (which only shows "important" children).
* the full AST view shows the AST as if using `someNode.getChildren()` which shows all children.
* the file has to be referenced in a `tsconfig.json` file for the AST to show up at all

JKillian:
Looks pretty good test-wise from your comment above, although you could also do non-named classes and named functions:

```
export default class {}
export default function foo() {}
```

patsissons:
interesting side note. My plan was to just scan for the `default` keyword and check if it is preceded by an `export` keyword. This actually doesn't work because `defaultKeywordNode.parent.getChildren()` doesn't contain `defaultKeywordNode`.

My new plan is to check the `parent.modifiers` array, and confirm `length === 2` and that elements are `export` followed by `default`. Is this appropriate? or should I box this scan to just class and function declarations?

PR #1242, opened 2016-05-10
improved no-default-export rule

patsissons:
more tests for less common export statements. new code to handle inline class and function declarations (and possibly other inline declarations if any exist).


patsissons:
fixes #1241

JKillian:
Looks good to me! I know your code should handle this correctly, but mind adding a code that tests a switch statement with a default case? Figure we might as well ensure that that doesn't trigger a failure ever in the future by accident.

patsissons:
sure no prob.

JKillian:
👍  Merged!

Issue #1250, opened 2016-05-12
Mistaken "Declaration of public static member variable" error


levibotelho:
### Bug Report

- __TSLint version__: `3.10.1`
- __TypeScript version__: `1.8`
- __Running TSLint via__: gulp-tslint & VS Code tslint extension

#### TypeScript code being linted

```ts
    static contextTypes = new class extends ValidationMap {
        router = React.PropTypes.object.isRequired;
    };

    static propTypes = new class extends ValidationMap {
        message = React.PropTypes.object.isRequired;
    };
```

with `tslint.json`:

#### Actual behavior

I get a "Declaration of public static member variable not allowed to appear after declaration of public instance member variable" error on the second instance. This also occurs if I switch the two declarations around.

#### Expected behavior

There should be no error.

adidahiya:
Looks like we probably don't handle class expressions properly. The fix is to add this method to `MemberOrderingWalker`:

```ts
    public visitClassExpression(node: ts.ClassExpression) {
        this.resetPreviousModifiers();

        this.newMemberList();
        super.visitClassExpression(node);
        this.checkMemberOrder();
    }
```

and make sure the appropriate `SyntaxWalker` super method is defined, etc.

PR #1252, opened 2016-05-13
Fix member ordering bug

### LINKED ISSUE #1251: Member Ordering with constructor
### Bug Report

- __TSLint version__: `3.10.1`
- __Running TSLint via__: (please choose one) Webpack

This is just a watered down example.

```ts
	export function FactoryGet(meh) {
		return class Res {
			constructor(res: string) {
				return res;
			}
		};
	}
```

with `tslint.json`:

```json
    "member-ordering": [true,
        "static-before-instance"
    ],
```

#### Actual behavior

I'm getting this error:

......./node_modules/tslint/lib/rules/memberOrderingRule.js:212
        this.memberStack[this.memberStack.length - 1].push(node);
TypeError: Cannot read property 'push' of undefined

#### Expected behavior

Not sure why i am getting this error... I'm fairly new to Typescript and Tslint.

I think it has something to do with the contractor call...

Any help would be appreciated. Thanks.

JKillian:
Should fix #1251 and #1250 

Also, only run code dealing with `order` option if it's specified. If there are any further bugs, this should prevent them for people using the old options

### DIFF
--- a/src/language/walker/syntaxWalker.ts
+++ b/src/language/walker/syntaxWalker.ts
@@ -70,6 +70,10 @@ export class SyntaxWalker {
         this.walkChildren(node);
     }
 
+    protected visitClassExpression(node: ts.ClassExpression) {
+        this.walkChildren(node);
+    }
+
     protected visitCatchClause(node: ts.CatchClause) {
         this.walkChildren(node);
     }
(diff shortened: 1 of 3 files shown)


JKillian:
@adidahiya I probably won't have time to make a release until Sunday, so if you have time to merge this and make a patch release that would be great. (Feel free to amend this PR if needed)

adidahiya:
👍

Issue #843, opened 2015-12-01
New Rule: no-mergable-namespace


adidahiya:
Similar to scss-lint's [MergeableSelector rule](https://github.com/brigade/scss-lint/blob/master/lib/scss_lint/linter/README.md#mergeableselector), we should lint against mergeable namespaces within the same file:

```ts
namespace Lint {
    ...
}

...

namespace Lint { // failure
    ...
}
```

PR #1253, opened 2016-05-14
Add new mergeable-namespace Rule

YuichiNukiyama:
related issue: #843


README.md
```diff
@@ -221,6 +221,7 @@ A sample configuration file with all options is available [here](https://github.
     * `"check-constructor"` enforces explicit visibility on constructors (can only be public)
 * `member-ordering` enforces member ordering. The first option should be an object with an `order` key.
    Values for `order` can be `fields-first`, `statics-first`, `instance-sandwich`, or a custom order.
+* `mergeable-namespace` disallows mergeable namespaces in the same file.
```
adidahiya: sorry, I should've clarified on the original issue -- this rule should be called __`no-mergeable-namespace`__ because it lints against the existence of certain syntax.

src/rules/mergeableNamespaceRule.ts
```diff
@@ -0,0 +1,54 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import * as ts from "typescript";
+import * as Lint from "../lint";
+
+export class Rule extends Lint.Rules.AbstractRule {
+    public static FAILURE_STRING = "mergeable namespace: ";
```
adidahiya: Going forward, for rules that have slightly customized failure strings like this, let's create a `failureStringFactory` method instead of a constant like this. In this case, we can actually make the failure much more helpful, something like this:

```ts
public static failureStringFactory(identifier: string, locationA: ts.LineAndCharacter, locationB: ts.LineAndCharacter): string {
    return `Mergeable namespace ${identifier} found. Merge its contents with the namespace on line ${locationB.line}.`;
}
```

src/rules/mergeableNamespaceRule.ts
```diff
@@ -0,0 +1,54 @@
...
+
+    public apply(sourceFile: ts.SourceFile): Lint.RuleFailure[] {
+        const languageService = Lint.createLanguageService(sourceFile.fileName, sourceFile.getFullText());
+        const mergeableNamespaceWalker = new MergeableNamespaceWalker(sourceFile, this.getOptions(), languageService);
+        return this.applyWithWalker(mergeableNamespaceWalker);
+    }
+}
+
+class MergeableNamespaceWalker extends Lint.RuleWalker {
+   private languageService: ts.LanguageService;
```
adidahiya: indentation looks slightly off here -- can you make sure every line uses 4 spaces?

src/rules/mergeableNamespaceRule.ts
```diff
@@ -0,0 +1,54 @@
...
+            && node.name.kind === ts.SyntaxKind.Identifier) {
+            this.validateReferencesForNamespace((<ts.Identifier> node.name).text, node.name.getStart());
+        }
+        super.visitModuleDeclaration(node);
+    }
+
+    private validateReferencesForNamespace(name: string, position: number) {
+        const fileName = this.getSourceFile().fileName;
+        const highlights = this.languageService.getDocumentHighlights(fileName, position, [fileName]);
+        if ((highlights == null || highlights[0].highlightSpans.length > 1)) {
```
adidahiya: excessive parentheses, please remove one set of `()` here

src/rules/mergeableNamespaceRule.ts
```diff
@@ -0,0 +1,54 @@
...
+   public visitModuleDeclaration(node: ts.ModuleDeclaration) {
+        if (Lint.isNodeFlagSet(node, ts.NodeFlags.Namespace)
+            && node.name.kind === ts.SyntaxKind.Identifier) {
+            this.validateReferencesForNamespace((<ts.Identifier> node.name).text, node.name.getStart());
+        }
+        super.visitModuleDeclaration(node);
+    }
+
+    private validateReferencesForNamespace(name: string, position: number) {
+        const fileName = this.getSourceFile().fileName;
```
adidahiya: small tweak to save some typing: `const { fileName } = this.getSourceFile();`

src/rules/mergeableNamespaceRule.ts
```diff
@@ -0,0 +1,54 @@
...
+        const highlights = this.languageService.getDocumentHighlights(fileName, position, [fileName]);
+        if ((highlights == null || highlights[0].highlightSpans.length > 1)) {
+            this.addFailure(this.createFailure(position, name.length, `${Rule.FAILURE_STRING}'${name}'`));
```
adidahiya: with the failure string factory I suggested above, this would become:

```ts
const failureString = Rule.failureStringFactory(name, this.findLocationToMerge(position, highlights));
this.addFailure(this.createFailure(position, name.length, failureString));
```

with a private helper method

```ts
private findLocationToMerge(currentPosition: number, highlightSpans: ts.HighlightSpan[]): ts.LineAndCharacter {
    const { line, character } = ts.getLineAndCharacterOfPosition(currentPosition);
    for (const span of highlightSpans) {
        const lineAndCharacter = ts.getLineAndCharacterOfPosition(span.textSpan.start);
        if (lineAndCharacter.line !== line && lineAndCharacter.character !== character) {
            return lineAndCharacter;
        }
    }
}
```

YuichiNukiyama: Good idea!!
I omit `&& lineAndCharacter.character !== character`. Because If the name of the namespaces is the same, value of character is always the same value.

YuichiNukiyama:
Thank you kindly advice.　I did it.

adidahiya:
nice, thanks @YuichiNukiyama!

PR #1254, opened 2016-05-15
adding linebreak-style rule

### LINKED ISSUE #123: Feature request: Linebreak style
A rule or setting to enforce (consistent) linebreaks would be good:

So either CRLF or LF (or CR), and no mixing.

patsissons:
initial attempt at the linebreak style rule. I used a manually created scanner because i need to tokenize the `NewLineTrivia` tokens, which aren't exposed in the standard rule framework. basically just walking the tokens until i see the new line token and checking its text content if it matches the configured linebreak type.

`CR` is not supported due to `CR` not really working properly at all. I left the `CR` code stubbed in and disabled for (potential) future use.

Tests are a bit funky because I can't create a mixed line endings file very easily (does git even allow such a thing via `.gitattributes`?). Right now I'm just testing that line endings are detected properly and success and failure scenarios work. But I cannot test a real world scenario where a paste with conflicting line endings occurs. That said it should not be any different than what the tests currently cover.

I would like to eventually add an `AUTO` option that would detect the dominant line endings in a file and then mark any differing line endings as failures, but I feel that it may be important to get the base functionality in first before I (or someone else) begin that type of work.

As always, open to any suggestions and improvements on this approach.

fixes #123

### DIFF
--- a/src/rules/linebreakStyleRule.ts
+++ b/src/rules/linebreakStyleRule.ts
@@ -6,8 +6,8 @@ const OPTION_LINEBREAK_STYLE_LF = "LF";
 
 export class Rule extends Lint.Rules.AbstractRule {
     public static FAILURE_STRINGS = {
-        CRLF: "Expected linebreak to be 'CRLF'",
-        LF: "Expected linebreak to be 'LF'",
+        CRLF: `Expected linebreak to be '${OPTION_LINEBREAK_STYLE_CRLF}'`,
+        LF: `Expected linebreak to be '${OPTION_LINEBREAK_STYLE_LF}'`,
     };
 
     public apply(sourceFile: ts.SourceFile): Lint.RuleFailure[] {
@@ -19,27 +19,16 @@ export class Rule extends Lint.Rules.AbstractRule {
             sourceFile.getFullText()
         );
 
-        let expectedEOL: string;
-        let failureString: string;
+        const linebreakStyle = this.getOptions().ruleArguments[0] || OPTION_LINEBREAK_STYLE_LF;
+        const expectLF = linebreakStyle === OPTION_LINEBREAK_STYLE_CRLF;
+        const expectedEOL = expectLF ? "\r\n" : "\n";
+        const failureString = expectLF ? Rule.FAILURE_STRINGS.CRLF : Rule.FAILURE_STRINGS.LF;
 
-        switch (this.getOptions().ruleArguments[0]) {
-            case OPTION_LINEBREAK_STYLE_CRLF:
-                expectedEOL = "\r\n";
-                failureString = Rule.FAILURE_STRINGS.CRLF;
-                break;
-            case OPTION_LINEBREAK_STYLE_LF:
-                expectedEOL = "\n";
-                failureString = Rule.FAILURE_STRINGS.LF;
-                break;
-        }
-
-        if (expectedEOL != null) {
-            for (let token = scanner.scan(); token !== ts.SyntaxKind.EndOfFileToken; token = scanner.scan()) {
-                if (token === ts.SyntaxKind.NewLineTrivia) {
-                    const text = scanner.getTokenText();
-                    if (text !== expectedEOL) {
-                        failures.push(this.createFailure(sourceFile, scanner, failureString));
-                    }
+        for (let token = scanner.scan(); token !== ts.SyntaxKind.EndOfFileToken; token = scanner.scan()) {
+            if (token === ts.SyntaxKind.NewLineTrivia) {
+                const text = scanner.getTokenText();
+                if (text !== expectedEOL) {
+                    failures.push(this.createFailure(sourceFile, scanner, failureString));
                 }
             }
         }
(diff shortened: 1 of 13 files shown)


JKillian:
I think we should be fine only supporting `LF` and `CR+LF`, given what Wikipedia says on the systems that use them:  😀 

image

test/rules/linebreak-style/failure/CRLF/test.ts.lint
```diff
@@ -0,0 +1,8 @@
+// this line uses CRLF
+~~~~~~~~~~~~~~~~~~~~~~ [Expected linebreak to be 'LF']
```
JKillian: marking the whole line as a failure seems good to me 👍

src/tsconfig.json
```diff
@@ -84,8 +85,8 @@
         "rules/noConditionalAssignmentRule.ts",
         "rules/noConsecutiveBlankLinesRule.ts",
         "rules/noConsoleRule.ts",
-        "rules/noConstructRule.ts",
         "rules/noConstructorVarsRule.ts",
+        "rules/noConstructRule.ts",
```
JKillian: not sure why this diff got in here, can you remove it?

.vscode/launch.json
```diff
@@ -0,0 +1,37 @@
+{
```
JKillian: What's this file for? If it's useful for developing from VSCode, can you make a separate PR for it?

src/rules/linebreakStyleRule.ts
```diff
@@ -0,0 +1,61 @@
+import * as ts from "typescript";
```
JKillian: Let's just remove the `CR` case from this rule all together for clarity. Shouldn't be too hard to add in later if need be (but that's unlikely anyways)

patsissons:
oops, i didn't realize i checked it a bunch of other files by accident. i'll fix that up in a sec.

src/rules/linebreakStyleRule.ts
```diff
@@ -0,0 +1,61 @@
...
+
+        let expectedEOL: string;
+        let failureString: string;
+
+        switch (this.getOptions().ruleArguments[0]) {
+            case "CR":
+                // expectedEOL = "\r";
+                failureString = Rule.FAILURE_STRINGS.CR;
+                break;
+            case "CRLF":
```
JKillian: Let's make constants for these options, [like here](https://github.com/palantir/tslint/blob/bbbd3b15ec75ab53809ed1580c4693615271295f/src/rules/tripleEqualsRule.ts#L21). I use them when creating the documentation to make sure I get the strings right 😃

src/rules/linebreakStyleRule.ts
```diff
@@ -0,0 +1,61 @@
...
+                expectedEOL = "\r\n";
+                failureString = Rule.FAILURE_STRINGS.CRLF;
+                break;
+            case "LF":
```
JKillian: Let's also make this the `default` case? What do you think?

src/rules/linebreakStyleRule.ts
```diff
@@ -0,0 +1,61 @@
...
+                expectedEOL = "\n";
+                failureString = Rule.FAILURE_STRINGS.LF;
+                break;
+        }
+
+        if (expectedEOL != null) {
```
JKillian: Then we can remove this, since we'll always have an `expectedEOL`

src/tsconfig.json
patsissons: i think that was just vscode ordering the tsconfig file alphabetically.

src/rules/linebreakStyleRule.ts
patsissons: probably worthy of discussion, i'll see what the eslint guys do and report back.

patsissons: they do indeed [use LF as the default](http://eslint.org/docs/rules/linebreak-style#options)

src/rules/linebreakStyleRule.ts
```diff
@@ -0,0 +1,49 @@
...
+        const failures = <Lint.RuleFailure[]>[];
+        const scanner = ts.createScanner(
+            sourceFile.languageVersion,
+            false,
+            sourceFile.languageVariant,
+            sourceFile.getFullText()
+        );
+
+        // do we need to allow a way to override this option with a disabled state?
+        const linebreakStyle = this.getOptions().ruleArguments[0] || OPTION_LINEBREAK_STYLE_LF;
```
patsissons: do we want to maybe allow something like `disable` or _empty string_ to allow overriding this rule with a _don't care_ behaviour?

JKillian: nah, the rule can just be disabled/turned off then

patsissons: i guess my use-case is for `extends` usage, where you want to stop a the rule from being used after loading it from a shared config.

patsissons: haha nevermind, i just realized this is completely irrelevant. I am spending too much time in different linters to remember which linters do what these days...

JKillian: You should still be able to do `"linebreak-style": false` in your own config to disable it

JKillian:
Thanks @patsissons!

PR #1261, opened 2016-05-20
Set up code for handling built-in configs

JKillian:
Support built-in configs. This allows a user to have a `tslint.json` file like this:

```
{
}
```

This does not allow `"tslint:recommended"` as an "entry point" configuration. That is, it can only be used by extending it, not doing something like `tslint --config tslint:recommended files/**/*.ts`. Should it allow this?

TODO:
- [x] Write tests for built-in config handling
- [x] Write built-in config handing code
- [x] Determine what rules built-in configs should have


src/configuration.ts
```diff
@@ -162,18 +164,29 @@ export function loadConfigurationFromPath(configFilePath: string): IConfiguratio
 }
 
 /**
- * Resolve configuration file path
- * @var relativeFilePath Relative path or package name (tslint-config-X) or package short name (X)
+ * Resolve configuration file path or node_module reference
+ * @var filePath Relative ("./path"), absolute ("/path"), node module ("path"), or built-in ("tslint:path")
```
adidahiya: missed this earlier -- `@var` should be `@param` instead

JKillian:
Added a staring point recommended config. I think we should recommend single quotes instead of double quotes though, less keystrokes!

We could also add a "basic" config that is less opinionated.

adidahiya:
nah, I'd rather stick with double quote for consistency with the TypeScript repo and our own internal projects. it's pretty easy for users to override....

src/configs/recommended.ts
```diff
@@ -0,0 +1,127 @@
...
+    ],
+    "class-name": true,
+    "comment-format": [true,
+        "check-space",
+    ],
+    "curly": true,
+    "eofline": true,
+    "forin": true,
+    "indent": [true, "spaces"],
+    "interface-name": true,
```
adidahiya: let's be more explicit and make the config `[true, "always-prefix"]`

src/configs/recommended.ts
```diff
@@ -0,0 +1,127 @@
...
+    "jsdoc-format": true,
+    "label-position": true,
+    "label-undefined": true,
+    "max-line-length": [true, 120],
+    "member-access": true,
+    "member-ordering": [true, {
+        "order": "statics-first",
+    }],
+    "new-parens": true,
+    "no-any": false,
```
adidahiya: how about adding `no-angle-bracket-type-assertion`?

src/configs/recommended.ts
```diff
@@ -0,0 +1,127 @@
...
+        "trace",
+    ],
+    "no-construct": true,
+    "no-constructor-vars": true,
+    "no-debugger": true,
+    "no-duplicate-key": true,
+    "no-duplicate-variable": true,
+    "no-empty": true,
+    "no-eval": true,
+    "no-internal-module": true,
```
adidahiya: how about `"no-inferrable-types": [true, "ignore-params"]`?

src/configs/recommended.ts
```diff
@@ -0,0 +1,127 @@
...
+        "check-finally",
+        "check-open-brace",
+        "check-whitespace",
+    ],
+    "one-variable-per-declaration": [true,
+        "ignore-for-loop",
+    ],
+    "quotemark": [true, "double", "avoid-escape"],
+    "radix": true,
+    "semicolon": true,
```
adidahiya: let's be explicit and say `[true, "always"]`?

adidahiya:
lgtm aside from the configuration suggestions I noted

src/configs/recommended.ts
JKillian: that seemed a little bit like overkill, given that many people aren't using .tsx at all

src/configs/recommended.ts
JKillian: worried about it conflicting with `typedef` which a lot of people like

src/configs/recommended.ts
```diff
@@ -0,0 +1,127 @@
...
+    "no-console": [true,
+        "debug",
+        "info",
+        "log",
+        "time",
+        "timeEnd",
+        "trace",
+    ],
+    "no-construct": true,
+    "no-constructor-vars": true,
```
JKillian: I think we should disable this one, it's a legit TS feature for removing redundancy in code.

Also, they're officially called "parameter property declarations", so this rule has a sort of unclear name as well

src/configs/recommended.ts
JKillian: Actually, thinking this is the wrong setting for the rule overall. Having an `I` prefix is like naming variable `length_number` and `name_string` and is redundant info

adidahiya: no, I actually think it's useful. By knowing which TS imports are interfaces vs those which are actual JS values at a glance, you can more easily tell which import statements will be elided at runtime.

```ts
import { IFoo, IBar } from "./a"; // elided
import { IBaz, Baz } from "./b"; // not elided
```

src/configs/recommended.ts
adidahiya: ok, i'm fine skipping it since baseline-web-config doesn't have it either

src/configs/recommended.ts
adidahiya: ok cool

src/configs/recommended.ts
adidahiya: ok sounds good

adidahiya: filed #1296

PR #1268, opened 2016-05-24
member-access: modify docs

### LINKED ISSUE #921: The docs for member-access (check-accessor) may need to be updated
The docs (README) for member-access (check-accessor) say "can only be public", but I don't see any compile errors when using "private" or "protected".

This is with a 1.8.x version of the TypeScript compiler – I haven't checked with earlier versions.

YuichiNukiyama:
fix #921


JKillian:
Thanks! I didn't even realize [that private constructors were a feature now](https://github.com/Microsoft/TypeScript/pull/6885) 😄

PR #1271, opened 2016-05-26
Better error message when rules from the config are not found

adidahiya:


src/ruleLoader.ts
```diff
@@ -52,8 +52,12 @@ export function loadRules(ruleConfiguration: {[name: string]: any},
     }
 
     if (notFoundRules.length > 0) {
-        const errorMessage = `Could not find the following rules specified in the configuration:\n${notFoundRules.join("\n")}`;
-        throw new Error(errorMessage);
+        const ERROR_MESSAGE = `
+            Could not find implementations for the following rules specified in the configuration:\n
+            ${notFoundRules.join("\n")}\n
+            Try upgrading TSLint and/or ensuring that you have all necessary custom rules installed.
```
JKillian: This error often also occurs when people upgrade to a new version of TSLint and still have rules left over from older versions. Maybe we could change the description to take that into account somehow?

src/ruleLoader.ts
```diff
@@ -52,8 +52,12 @@ export function loadRules(ruleConfiguration: {[name: string]: any},
     }
 
     if (notFoundRules.length > 0) {
-        const errorMessage = `Could not find the following rules specified in the configuration:\n${notFoundRules.join("\n")}`;
-        throw new Error(errorMessage);
+        const ERROR_MESSAGE = `
```
JKillian: since this is a template string, newlines and whitespace will be taken into account, so I think this message might end up with too much whitespace as is

PR #1277, opened 2016-05-27
Allow null keyword in a type position

### LINKED ISSUE #1265: Should not report an error on "null" when using in type position
We are trying to add linting to our library source files and run into issue below.

### Bug Report (delete this for feature requests)

- __TSLint version__: `@next`
- __TypeScript version__: `latest source-code built in master branch`
- __Running TSLint via__: Node.js API 

```ts
match(regexp: string): RegExpMatchArray | null;
```

with `tslint.json`: 

```json
     ....
     "no-null-keyword": true,
```

#### Actual behavior
Get an error `Use 'undefined' instead of 'null'`

#### Expected behavior

There shouldn't be an error when using `null` in type position

JKillian:
Fixes #1265

### DIFF
--- a/src/rules/noNullKeywordRule.ts
+++ b/src/rules/noNullKeywordRule.ts
@@ -31,8 +31,18 @@ export class Rule extends Lint.Rules.AbstractRule {
 class NullWalker extends Lint.RuleWalker {
     public visitNode(node: ts.Node) {
         super.visitNode(node);
-        if (node.kind === ts.SyntaxKind.NullKeyword) {
+        if (node.kind === ts.SyntaxKind.NullKeyword && !isPartOfType(node)) {
             this.addFailure(this.createFailure(node.getStart(), node.getWidth(), Rule.FAILURE_STRING));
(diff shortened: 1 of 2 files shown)


src/rules/noNullKeywordRule.ts
```diff
@@ -31,8 +31,18 @@ export class Rule extends Lint.Rules.AbstractRule {
...
     public visitNode(node: ts.Node) {
         super.visitNode(node);
-        if (node.kind === ts.SyntaxKind.NullKeyword) {
+        if (node.kind === ts.SyntaxKind.NullKeyword && !isPartOfType(node)) {
             this.addFailure(this.createFailure(node.getStart(), node.getWidth(), Rule.FAILURE_STRING));
         }
     }
 }
+
+function isPartOfType({ parent }: ts.Node) {
```
adidahiya: maybe `isPartOfTypeLiteral`?

adidahiya:
lgtm

PR #1285, opened 2016-05-31
README.md - explained that Formatter class must be named Formatter

HamletDRC:
README.md - explained that Formatter class must be named Formatter

I discovered this myself after working about 30 minutes to figure out why it was not found.


adidahiya:
thanks @HamletDRC!

PR #1289, opened 2016-06-03
Mention Emacs support

paldepind:
Flycheck is the most used Emacs package for on-the-fly syntax checking. And it supports [tslint](http://www.flycheck.org/en/latest/languages.html#typescript).


adidahiya:
Thanks @paldepind!

PR #1292, opened 2016-06-04
Visit JsxExpression node.

lijunle:
Visit JsxExpression node.

I am going to port [jsx-curly-spacing](https://github.com/yannickcr/eslint-plugin-react/blob/master/lib/rules/jsx-curly-spacing.js) rule and find no such node visitor function.


adidahiya:
Can you please put the method and case clause in alphabetical order? It should appear right before `JsxSelfClosingElement`.

lijunle:
Updated! :smile_cat:

adidahiya:
thanks @lijunle!

PR #1294, opened 2016-06-06
#325 now fixed in typescript

### LINKED ISSUE #325: Incorrect unused variable error
Very similar, if not the same as #237

Using the following contrived typescript file

```ts
    /// <reference path="typings/node/node.d.ts" />
    import fs = require('fs');
    import createWriteStream = fs.createWriteStream;
    console.log(createWriteStream);
```

With these rules

```json
    {
      "rules": {
        "no-unused-variable": true
      }
    }
```

We get this error about fs being unused:
`test.ts[1, 8]: unused variable: 'fs`

This has been the behavior since 1.0

JKillian:
Fixes #325

### DIFF
--- a/test/rules/no-unused-variable/default/import.ts.lint
+++ b/test/rules/no-unused-variable/default/import.ts.lint
@@ -29,3 +29,7 @@ baz();
 namedExport();
 
 import "jquery";
+
+import abc = require('abc');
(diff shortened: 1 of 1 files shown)


adidahiya:
nice!

PR #1303, opened 2016-06-09
Improving failure messages for the member-access rule to be more specific about what failed

chrismbarr:
:-1:  **Before**:
>default access modifier on member/method not allowed

:+1: **After**:
>The class method 'a' is missing an access modifier

or

>The class property 'i' is missing an access modifier

I always though a message about "default access modifier" was never very clear, especially for those new to TypeScript.  It doesn't **have** a default modifier, the issue is that it **does not have** a prefixed keyword that the rule requires.

Hopefully these changes will make this message easier to understand.

---

I considered making the message something like:
>The class method 'a' must be marked as either 'private', 'public', or 'protected'

I do like that it's more specific about what needs to be added, but I was unsure if it was too much.  What do you think, should I change it to be more like that example?

### DIFF
--- a/src/rules/memberAccessRule.ts
+++ b/src/rules/memberAccessRule.ts
@@ -21,7 +21,7 @@ import * as Lint from "../lint";
 export class Rule extends Lint.Rules.AbstractRule {
     public static FAILURE_STRING_FACTORY = (memberType: string, memberName: string, publicOnly: boolean) => {
         memberName = memberName == null ? "" : ` '${memberName}'`;
-        if(publicOnly){
+        if (publicOnly) {
             return `The ${memberType}${memberName} must be marked as 'public'`;
         }
         return `The ${memberType}${memberName} must be marked either 'private', 'public', or 'protected'`;
--- a/src/rules/memberAccessRule.ts
+++ b/src/rules/memberAccessRule.ts
@@ -19,7 +19,10 @@ import * as ts from "typescript";
 import * as Lint from "../lint";
 
 export class Rule extends Lint.Rules.AbstractRule {
-    public static FAILURE_STRING = "default access modifier on member/method not allowed";
+    public static FAILURE_STRING_FACTORY = (memberType: string, memberName: string) => {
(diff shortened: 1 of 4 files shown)


adidahiya:
Thanks for the suggestion @chrismbarr, I love descriptive error messages! Yes, I would go as far as suggesting the exact keywords that a user must add, like `private`, `public`, etc. I think along with the rule name (which is usually displayed along with failures), it gives users a clear idea of what TS syntax is expected.

src/rules/memberAccessRule.ts
```diff
@@ -19,7 +19,10 @@ import * as ts from "typescript";
 import * as Lint from "../lint";
 
 export class Rule extends Lint.Rules.AbstractRule {
-    public static FAILURE_STRING = "default access modifier on member/method not allowed";
+    public static FAILURE_STRING_FACTORY = (memberType: string, memberName: string) => {
+        memberName = memberName ? " '" + memberName + "'" : "";
```
adidahiya: can we make this a template string and avoid implicit type coercion please?

```ts
memberName = memberName == null ? "" : ` '${memberName}'`;
```

chrismbarr:
Good suggestions @adidahiya, thanks!  I've made those changes as well as adding a slightly different error message for `constructor` since those can only be `public`.  Otherwise the failure message would be giving you false information.

chrismbarr:
It looks like appveyor failure is a fluke, lots of `npm ERR! network read ECONNRESET` errors in there.  Can it be run again?  I'm very confident those errors are unrelated from my changes.

adidahiya:
lgtm 👍

glen-84:
@chrismbarr Constructors can also use `private` and `protected` now. ([link](https://github.com/Microsoft/TypeScript/wiki/What%27s-new-in-TypeScript#private-and-protected-constructors))

chrismbarr:
@glen-84 Neat, but not until TS 2.0.  Is there a way to detect the TS version from TSLint so the output is changed?

glen-84:
I have no idea. :stuck_out_tongue:

Issue #1249, opened 2016-05-11
no-consecutive-blank-lines rule doesn't catch lines with whitespace


chrismbarr:
### Bug Report

- __TSLint version__: `1.10.1`
- __TypeScript version__: `1.8.10`
- __Running TSLint via__: `gulp-tslint` & Visual Studio

__Config:__ `"no-consecutive-blank-lines": true,`

Consider `~~~>` to be a tab and `-` to be a space

#### TypeScript code being linted

```ts
----var x = "foo";
~~~>
~~~>
----
----
~~~>var y = "bar";
```

----

#### Actual behavior

No errors are thrown for consecutive lines that contain space or tab characters, even those for all practical purposes these lines should probably still be considered "blank"

#### Expected behavior

Throw error when there are consecutive blank lines, even if those "blank" lines only contain whitespace.

chrismbarr:
I am attempting to make a PR for this, but the [tests file for this one is really confusing me](https://github.com/palantir/tslint/blob/master/test/rules/no-consecutive-blank-lines/test.ts.lint).  I added some whitespace to some of the blank lines, and now the tests fail as expected, I'm attempting to rewrite the rule in a different way but I'm a little lost.

Where it has `this.createFailure(scanner.getStartPos(), 1, Rule.FAILURE_STRING);` what does `scanner.getStartPos()` return?  I had always assumed it was the line number, but it seems to return the character position from the start of the file.  Is this correct?
[See  :page_facing_up:  In context](https://github.com/palantir/tslint/blob/master/src/rules/noConsecutiveBlankLinesRule.ts#L48)

PR #1304, opened 2016-06-09
Fix for no-consecutive-blank-lines to also see whitespace

chrismbarr:
Related: #1249

Now this rule will consider lines that only contain whitespace as "blank" lines.

NOTE: Currently the tests for this rule fail, but it's because I'm not fully understanding the test file syntax.  All the failures appear to be reported correctly, except for that it reports some extra failures that I can't figure out.  It seems to report an incorrect failure at the beginning for every failure that exists afterwards. Any help here would be appreciated!  Below is a screenshot of the console output.  Note the debug output at the beginning.

image

### DIFF
--- a/src/rules/noConsecutiveBlankLinesRule.ts
+++ b/src/rules/noConsecutiveBlankLinesRule.ts
@@ -30,21 +30,29 @@ class NoConsecutiveBlankLinesWalker extends Lint.SkippableTokenAwareRuleWalker {
     public visitSourceFile(node: ts.SourceFile) {
         super.visitSourceFile(node);
 
+        const sourceFileText = node.getFullText();
+        const soureFileLines = sourceFileText.split(/\n/);
+
         // find all the lines that are blank or only contain whitespace
         let blankLineIndexes: number[] = [];
-        node.getFullText().split(/\n/).forEach(function(txt, i){
+        soureFileLines.forEach(function(txt, i){
             if (txt.trim() === "") {
                 blankLineIndexes.push(i);
             }
         });
 
         // now only throw failures for the fisrt number from groups of consecutive blank line indexes
-        for (let i = 0; i < blankLineIndexes.length; i++) {
-            let diff = Math.abs(blankLineIndexes[i + 1] - blankLineIndexes[i]);
-            let prevDiff = Math.abs(blankLineIndexes[i] - blankLineIndexes[i - 1]);
-            if (diff === 1 && prevDiff !== 1) {
-                this.addFailure(this.createFailure(blankLineIndexes[i], 1, Rule.FAILURE_STRING));
-            }
+        const sequences: number[][] = [];
+        let lastVal = -2;
+        for (const line of blankLineIndexes) {
+            line > lastVal + 1 ? sequences.push([line]) : sequences[sequences.length - 1].push(line);
+            lastVal = line;
         }
+        sequences
+            .filter((arr) => arr.length > 1).map((arr) => arr[0])
+            .forEach((startLineNum: number) => {
+                let startCharPos = node.getPositionOfLineAndCharacter(startLineNum + 1, 0);
+                this.addFailure(this.createFailure(startCharPos, 1, Rule.FAILURE_STRING));
+            });
     }
 }
--- a/src/rules/noConsecutiveBlankLinesRule.ts
+++ b/src/rules/noConsecutiveBlankLinesRule.ts
@@ -29,28 +29,52 @@ export class Rule extends Lint.Rules.AbstractRule {
 class NoConsecutiveBlankLinesWalker extends Lint.SkippableTokenAwareRuleWalker {
     public visitSourceFile(node: ts.SourceFile) {
         super.visitSourceFile(node);
-
-        // starting with 1 to cover the case where the file starts with two blank lines
-        let newLinesInARowSeenSoFar = 1;
-        Lint.scanAllTokens(ts.createScanner(ts.ScriptTarget.ES5, false, ts.LanguageVariant.Standard, node.text), (scanner: ts.Scanner) => {
-            const startPos = scanner.getStartPos();
-            if (this.tokensToSkipStartEndMap[startPos] != null) {
-                // tokens to skip are places where the scanner gets confused about what the token is, without the proper context
-                // (specifically, regex, identifiers, and templates). So skip those tokens.
-                scanner.setTextPos(this.tokensToSkipStartEndMap[startPos]);
-                newLinesInARowSeenSoFar = 0;
-                return;
-            }
-
-            if (scanner.getToken() === ts.SyntaxKind.NewLineTrivia) {
-                newLinesInARowSeenSoFar += 1;
-                if (newLinesInARowSeenSoFar >= 3) {
-                    const failure = this.createFailure(scanner.getStartPos(), 1, Rule.FAILURE_STRING);
-                    this.addFailure(failure);
-                }
-            } else {
-                newLinesInARowSeenSoFar = 0;
+        
+        //Find all the lines that are blank or only contain whitespace
+        let blankLineIndexes:number[] = [];
+        node.getFullText().split(/\n/).forEach(function(txt, i){
+            if(txt.trim()===""){
+                blankLineIndexes.push(i);
             }
         });
+
+        //console.log("==================")
+        //console.log("Blank or whitespace line numbers ", blankLineIndexes.map((n)=>n+1))
+        
+        //Now only keep the found blank lines that are consecutive
+        let consecutiveBlankLineStarts:number[] = [];
+        for(let i=0; i < blankLineIndexes.length; i++) {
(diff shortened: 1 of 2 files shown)


adidahiya:
Thanks @chrismbarr; assigning @JKillian to help out with the test syntax.

src/rules/noConsecutiveBlankLinesRule.ts
```diff
@@ -29,28 +29,52 @@ export class Rule extends Lint.Rules.AbstractRule {
...
+        //console.log("Blank or whitespace line numbers ", blankLineIndexes.map((n)=>n+1))
+        
+        //Now only keep the found blank lines that are consecutive
+        let consecutiveBlankLineStarts:number[] = [];
+        for(let i=0; i < blankLineIndexes.length; i++) {
+            let diff = blankLineIndexes[i+1] - blankLineIndexes[i];
+            if(Math.abs(diff)==1) {
+                consecutiveBlankLineStarts.push(blankLineIndexes[i]);
+            }
+        }
```
JKillian: I think you might be able to do all your logic in this one loop: keep track of something like `currentSequenceStart` and `currentSequenceLength`, and then push the value of `currentSequenceStart` into an array if the length is greater than one. Just a suggestion because the loop below feels a little complex

JKillian:
The test syntax is admittedly a little weird here, it uses the `~nil` mark, which means "there should be a failure reported here with a width of 0 characters." Ideally TSLint wouldn't report any errors like that, but it does, so that's why I built it into the test system.

In this case, there's not really a great location to report a failure for a completely empty line, so that's why it reports a error width error. For lines with actual length to them, I would report the error width to be the length of the line

JKillian:
I'd recommend just adding some existing test code below the current test code and leaving the current test code as is just so we can see the differences from your changes eariler

JKillian:
Also looks like there are some lint errors you'll want to get fixed so the build will pass!

chrismbarr:
ok, test file reverted back to what it was but now with an added "whitespace test" at the bottom.  I've also fixed the lint errors in the rule file, sorry about that!

I'm not totally sure that the 2 loops can be combined.  What happens is that lines 33-39 looks for the line indexes that are empty or whitespace only, that would return something like:
`[ 4, 5, 7, 9, 11, 12, 14, 16, 17, 19, 22, 23, 24, 25 ]`

Then the first loop goes through and finds the consecutive numbers from that array, so it would find
`[ 4, 11, 16, 22, 23, 24 ]`

Then the second loop only keeps the first number from each group, so now we only care about these lines
`[ 4, 11, 16, 22 ]`

I didn't have a lot of time today to think about how I might combine these loops.  It does seem possible, but I just wasn't able to today.

chrismbarr:
@JKillian ok, please check it out now.  I think I've simplified the code quite a bit, thanks for the suggestions!  The tests still fail due to me not totally understanding the odd markup in the test file, but perhaps you can fix it now that the rule works just like before, but with the added ability to ignore whitespace.

src/rules/noConsecutiveBlankLinesRule.ts
```diff
@@ -30,27 +30,21 @@ class NoConsecutiveBlankLinesWalker extends Lint.SkippableTokenAwareRuleWalker {
...
-                if (newLinesInARowSeenSoFar >= 3) {
-                    const failure = this.createFailure(scanner.getStartPos(), 1, Rule.FAILURE_STRING);
-                    this.addFailure(failure);
-                }
-            } else {
-                newLinesInARowSeenSoFar = 0;
+        // now only throw failures for the fisrt number from groups of consecutive blank line indexes
+        for (let i = 0; i < blankLineIndexes.length; i++) {
+            let diff = Math.abs(blankLineIndexes[i + 1] - blankLineIndexes[i]);
+            let prevDiff = Math.abs(blankLineIndexes[i] - blankLineIndexes[i - 1]);
```
JKillian: Nice! This code is a lot more friendly, I like the simplifications. I don't feel like accessing `blankLineIndexes[-1]` and `blankLineIndexes[length]` is great though, maybe we should handle those boundary cases specially?

JKillian:
Yeah, something weird is going on with the tests here, TSLint seems to be reporting multiple blank line errors at the beginning of the test file:

image

chrismbarr:
Yeah I got the same thing. But log out what `blankLineIndexes[i]` where it creates the failures. The line numbers there are correct and there aren't 4 extras there like the test reports. It's a bit of a mystery to me and it seems like the rule is written correctly!

I'll take a look at adding some boundary conditions tomorrow. However I believe that due to `Math.abs` those would produce `NaN` which technically *is* a number that we can compare to `1`. IMO this is fine to do, but if you'd prefer I can certainly add boundary conditions. 

**Side note**: `typeof NaN` is my favorite piece of "WTF JS" ever!

JKillian:
Ah, I know what the issue is here. You're passing in the line number to the `createFailure` method, but it expects the _character_ number. There are multiple ways you could get this, but easiest is the `node.getPositionOfLineAndCharacter` method (since `node` is the `SourceFile` in our context).

And yes, I like how `Not A Number` is actually a number! 😀

JKillian:
> IMO this is fine to do, but if you'd prefer I can certainly add boundary conditions.

It's a tough call, on one hand, if it works it works, on the other, propagating `undefined` and `NaN` in interesting ways always seems a bit unclear, nicer to stay in the world of non-nullness. However, I'm not sure the current code would work right for an array like `[5,6]`?

This is actually a pretty fun little bit of code to try to right neatly! Here's what I came up with, but it's untested. Feel free to use or not use as you wish:

```
const sequences = [];
const lastVal = -2;
for (const line of blankLineIndexes) {
    line > lastVal + 1 ? sequences.push([line]) : sequences[sequences.length -1].push(line);
    lastVal = line;
}
const errorLines = sequences.filter((arr) => arr.length > 1).map((arr) => arr[0]);
// convert lines to character position and add failures
```

chrismbarr:
Thanks for the tips, Your suggested code seems much simpler than what I had.  That is now in place, and it now uses the `getPositionOfLineAndCharacter` method to correctly get the char position instead of the line number.  I was not aware of this method!

So... now I'm in the opposite situation where my code works & all the tests pass, but I don't understand why!  I'll comment on lines to explain why.

test/rules/no-consecutive-blank-lines/test.ts.lint
```diff
@@ -14,20 +14,20 @@ class Clazz { // comment
...
 }
-~nil  [consecutive blank lines are disallowed]
+~nil [consecutive blank lines are disallowed]
 
 //Begin whitespace
 // The next two lines of "code" contain only tabs or spaces, they are also considered "blank" lines
-~nil                         [consecutive blank lines are disallowed]
         
         
+~ [consecutive blank lines are disallowed]
```
chrismbarr: Why does the test fail here on line 30?  I would expect it to fail on line 28, at the beginning of the blank line sequences.

JKillian: What I think is happening here is that the error position is ending up as the first character of the following line or something strange like that. The test system as is struggles a little bit with representing errors nicely visually for whitespace cases like this, but that's a test system quirk not an issue with your PR!

JKillian:
Merged, thanks!

test/rules/no-consecutive-blank-lines/test.ts.lint
chrismbarr: Alright, as long as it works I guess.  I've worked a bit with the microsoft-tslint-conrib repo and I find their test files _much_ easier to read and understand.  They can either write TS code as string inputs or refer to real pass/fail files.  Then they just compare the test result objects which contain the rule failure, with a line/char position.

[Here's an example of that with string input](https://github.com/Microsoft/tslint-microsoft-contrib/blob/master/tests/NoBannedTermsTests.ts#L16-L27)
And [here's an example test](https://github.com/Microsoft/tslint-microsoft-contrib/blob/master/tests/NoOctalLiteralTests.ts#L37-L147) that refers to [a real TS file](https://github.com/Microsoft/tslint-microsoft-contrib/blob/master/test-data/NoOctalLiteralTestInput.ts)

JKillian: Oh believe me, we used to do things that way, and we had comments from people (such as TS team members) who said they wanted to contribute to TSLint but that writing tests was too onerous.

The downsides of that approach are pretty painful: you have to manually figure out the position and length of each lint failure. Then, if you change your file at all, you have to go through and update the positions for every expected failure.

It's true that the current system doesn't work so well for this rule. But for other rules [it works nicely](https://github.com/palantir/tslint/blob/master/test/rules/triple-equals/allow-null-check/test.ts.lint).

Our testing system is inspired by the way the TS team does tests. I'd recommend [reading this issue](https://github.com/Microsoft/TypeScript/issues/6528) for some great pointers from them on testing, it's an interesting read

PR #1305, opened 2016-06-10
correct example tslint.json

allannienhuis:
The existing sample in the README is invalid json.
The example used array braces for rules definition, but this should be an object.


adidahiya:
haha good catch, thanks @allannienhuis :)

allannienhuis:
the only reason I caught it is because I'm lazy and copy/pasted from the example :)

PR #1308, opened 2016-06-13
Use latest tslint config and fix some lint errors

adidahiya:


adidahiya:
This is fairly uncontroversial, so I'm going to merge after a green build.

PR #1309, opened 2016-06-13
member-access: fix to correctly report the name for static methods/properties

chrismbarr:
I recently made PR #1303 (which was _just_ merged in!), but I realized a few moments ago that it did not properly get the member name in instances where the `static` keyword was used.  The tests never made use of this keyword, so I added some usages and updated the rule to properly get the member name regardless if the `static` keyword is used or not.

I like this way better anyway since it no longer assumes the hard-coded index of the identifier


chrismbarr:
Ugh appveyor :angry:  - `npm ERR! network read ECONNRESET`

adidahiya:
cool, this works, thanks for the additional tests 👍

Issue #1181, opened 2016-04-29
Stop testing with node v0.10 and v0.12, add v6.0 instead


adidahiya:
https://nodejs.org/en/blog/release/v6.0.0/#what-about-node-js-v0-10-and-v0-12

Related Tasks:
- [ ] Use Object.assign in `buildDocs.ts` and other locations if necessary

adidahiya:
At the same time, we should add testing for Node v6.

JKillian:
It seems that AppVeyor sometimes breaks on Node 0.10 now for who knows what reason, so all the better to do this

PR #1311, opened 2016-06-13
Metadata for each rule

JKillian:
Pushing up my work on metadata for rules. This metadata is currently used to generate large parts of http://palantir.github.io/tslint/.

Metadata contains:
* A short description of the rule
* A more detailed description of the rule
* A rationale behind the rule
* A JSON Schema for the options for the rule
* Examples of sensible rule option configurations
* The category of the rule

This PR:
* Adds metadata for each rule
* Adds a `gulp docs` command that runs a script which reads the metadata from each rule and outputs file into the Jekyll docs site
* Removes `docs/sample.tslint.json` because it's redundant documentation and also to make room for the docs build script.

TODO:
- [x] Switch to [JSON Schema](https://spacetelescope.github.io/understanding-json-schema/index.html) instead of own schema format for rule options


alexeagle:
cc @ScottSWu

src/language/rule/rule.ts
```diff
@@ -18,6 +18,53 @@
...
+    */
+    optionExamples?: string[];
+
+   /**
+    * An explanation of why the rule is useful.
+    */
+    rationale?: string;
+}
+
+export type RuleType = "functionality" | "maintainability" | "style" | "typescript";
```
JKillian: I'm not sure we use string literal union types anywhere else, and this might break things for 1.7 consumers. I feel okay with that at this point because 1.8 has been out for awhile, but we could use a different tool instead of string literal union types if we need to

docs/sample.tslint.json
```diff
@@ -1,168 +0,0 @@
-{
```
JKillian: This file is essentially redundant documentation at this point and an unnecessary burden to maintain, so I deleted it.

docs/tsconfig.json
```diff
@@ -0,0 +1,31 @@
+{
```
JKillian: Instead of putting the docs build script in `docs/` and compiling it on its own, I could put it in `src/` and let it get compiled with everything else. However, it seemed nicer to have it separate. I could also rename this directory `scripts/`

JKillian:
Ready for review!

docs/buildDocs.ts
```diff
@@ -0,0 +1,77 @@
...
+// write overall data file, this is used to generate the index page for the rules
+const fileData = JSON.stringify(rulesJson, undefined, "  ");
+fs.writeFileSync(path.join(DOCS_DIR, "_data", "rules.json"), fileData);
+
+function generateRuleFile(metadata: IRuleMetadata) {
+    const yamlData: any = {};
+    for (const key of Object.keys(metadata)) {
+        yamlData[key] = (<any> metadata)[key];
+    }
+    yamlData.optionsJSON = JSON.stringify(metadata.options, undefined, "  ");
```
adidahiya: minor: the last arg could be the number literal `2`; I think that's slightly more standard (?)

docs/buildDocs.ts
```diff
@@ -0,0 +1,77 @@
...
+}
+
+// write overall data file, this is used to generate the index page for the rules
+const fileData = JSON.stringify(rulesJson, undefined, "  ");
+fs.writeFileSync(path.join(DOCS_DIR, "_data", "rules.json"), fileData);
+
+function generateRuleFile(metadata: IRuleMetadata) {
+    const yamlData: any = {};
+    for (const key of Object.keys(metadata)) {
+        yamlData[key] = (<any> metadata)[key];
```
adidahiya: are you trying to do an object copy? let's use `Object.assign`

docs/buildDocs.ts
```diff
@@ -0,0 +1,77 @@
...
+    }
+    yamlData.optionsJSON = JSON.stringify(metadata.options, undefined, "  ");
+    yamlData.layout = "rule";
+    yamlData.title = `Rule: ${metadata.ruleName}`;
+    return `---\n${yaml.safeDump(yamlData, <any>{lineWidth: 140})}---`;
```
adidahiya: :( sad we have to do this `<any>` type assertion

docs/buildDocs.ts
```diff
@@ -0,0 +1,77 @@
...
+const DOCS_DIR = "../../tslint-gh-pages";
+const DOCS_RULE_DIR = path.join(DOCS_DIR, "rules");
+
+const rulePaths = glob.sync("../lib/rules/*Rule.js");
+const rulesJson: IRuleMetadata[] = [];
+for (const rulePath of rulePaths) {
+    const ruleModule = require(rulePath);
+    const Rule = ruleModule.Rule as typeof AbstractRule;
+    if (Rule != null && Rule.metadata != null) {
+        const metadata = Rule.metadata;
```
adidahiya: minor: use destructuring

docs/buildDocs.ts
```diff
@@ -0,0 +1,77 @@
...
+
+        rulesJson.push(metadata);
+    }
+}
+
+// write overall data file, this is used to generate the index page for the rules
+const fileData = JSON.stringify(rulesJson, undefined, "  ");
+fs.writeFileSync(path.join(DOCS_DIR, "_data", "rules.json"), fileData);
+
+function generateRuleFile(metadata: IRuleMetadata) {
```
adidahiya: What kind of file does this produce? yaml or html? could use a bit of documentation

docs/tsconfig.json
adidahiya: sgtm

src/language/rule/rule.ts
```diff
@@ -18,6 +18,53 @@
 import * as ts from "typescript";
 import {RuleWalker} from "../walker/ruleWalker";
 
+export interface IRuleMetadata {
+   /**
```
adidahiya: misaligned JSDoc comments (off by 1 space)

src/language/rule/rule.ts
adidahiya: yeah this seems fine

src/language/rule/rule.ts
```diff
@@ -18,6 +18,53 @@
...
+    * An explanation of the available options for the rule.
+    */
+    optionsDescription?: string;
+
+   /**
+    * Schema of the options the rule accepts.
+    * The first boolean for whether the rule is enabled or not is already implied.
+    * This field describes the options after that boolean.
+    * If null, this rule has no options and is not configurable.
```
adidahiya: "If undefined" ?

also, does that mean it should be an optional field?

src/rules/alignRule.ts
```diff
@@ -19,6 +19,31 @@ import * as Lint from "../lint";
...
 export class Rule extends Lint.Rules.AbstractRule {
+    /* tslint:disable:object-literal-sort-keys */
+    public static metadata: Lint.IRuleMetadata = {
+        ruleName: "align",
+        description: "Enforces vertical alignment.",
+        rationale: "Helps maintain a readable, consistent style in your codebase.",
+        optionsDescription: Lint.Utils.dedent`
+            Three arguments may be optionally provided:
+
+            * \`"parameters"\` checks alignment of function parameters.
```
adidahiya: why do we need to escape here? because of how js-yaml works?

Gruntfile.js
```diff
@@ -99,6 +111,12 @@ module.exports = function (grunt) {
         "mochaTest",
         "run:testRules",
     ].concat(checkBinTest));
+    grunt.registerTask("docs", [
```
adidahiya: this should probably be run as part of the CI build.

adidahiya:
mostly minor comments; 👎  for not running the new task as part of CI build

docs/buildDocs.ts
JKillian: Good call, much easier than guessing how many spaces are in a string

docs/buildDocs.ts
JKillian: Node 0.12 doesn't support Object.assign, we'll have to drop support for it if we do this. (although maybe some module we're depending on is already polyfilling it haha

src/language/rule/rule.ts
JKillian: jsyaml can handle `null`, but not `undefined`, so I was actually meant `null` specifically. Maybe it would be better to just make it optional and handle things in the docs build script, but I also don't mind requiring an explicit null

docs/buildDocs.ts
adidahiya: Dropping 0.12 is on the roadmap soon: https://github.com/palantir/tslint/issues/1181

If you prefer, you can leave this implementation, comment on that issue ^ about the future `Object.assign` refactor, and leave a code comment with a TODO linking to the GH issue.

src/rules/alignRule.ts
JKillian: Just to avoid having the backtick close the template string.

src/language/rule/rule.ts
adidahiya: meh I'd prefer trying to remove `null` from our APIs if possible, so I have a slight preference for handling it in our own code `if (foo = undefined) { /* do the null things */ }`

src/rules/alignRule.ts
adidahiya: oh derp I thought this was escaping quotemarks... nvm

Gruntfile.js
JKillian: Agreed, it can run when a tag is created for a non-dev release. People will still be able to do this manually though if they update other parts of the docs site that aren't generated via this task.

Can we automatically push to our repo from CircleCI?

adidahiya: It should be run as part of our verification step on every commit -- we want to make sure the docs TS code doesn't get broken

JKillian: I understand what you're getting at: we want to make sure that `docs.ts` builds and passes the linter. However, we _don't_ want the part of the task to run that actually updates the docs site, because the docs site would advertise unreleased TSLint rules and options.

For simplicity for now, I'll adjust things so that the build part of the current docs task is run on every build, including CircleCI builds. The actual updating of the files in the tslint-gh-pages, that is, the execution of `buildDocs.js`, will still have to be done manually for the time being

src/language/rule/rule.ts
JKillian: I'm going to leave this be for now, and when I get around to doing the validation code I'll see what ends up being the best design given the constrains validation presents.

adidahiya:
lgtm 👍

JKillian:
Sweet, let's merge this

PR #1315, opened 2016-06-16
Add a test for non-arrow members

JKillian:


adidahiya:
does this address any open issue?

JKillian:
Nope, just a test I thought should be there when I was looking through the file

PR #1316, opened 2016-06-16
Allow constructor calls for side effects with no-unused-expression, move checking to new "no-new" rule (#347)

### LINKED ISSUE #347: Allow constructor calls for side effects with no-unused-expression, move checking to new "no-new" rule
```ts
new SyncManager(fakeWebSocket, "session");
```

triggers the `no-unused-expression` rule, even though constructors are also function calls. This is difficult in testing when we're testing to make sure our class is calling the expected methods on `fakeWebSocket`.

bencoveney:
Fixes #347

I have assumed no-new should be in the default config files only because it replaces part of a rule which they were using before


src/rules/noNewRule.ts
```diff
@@ -0,0 +1,78 @@
...
+        return this.applyWithWalker(new NoNewWalker(sourceFile, this.getOptions()));
+    }
+}
+
+class NoNewWalker extends NoUnusedExpression.NoUnusedExpressionWalker {
+    private expressionContainsNew: boolean;
+
+    constructor(sourceFile: ts.SourceFile, options: Lint.IOptions) {
+        super(sourceFile, options);
+        this.expressionContainsNew = false;
```
JKillian: You should be able to initialize this in its declaration and delete this constructor entirely

src/rules/noNewRule.ts
```diff
@@ -0,0 +1,78 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import * as Lint from "../lint";
+import * as ts from "typescript";
+import * as NoUnusedExpression from "./noUnusedExpressionRule";
```
JKillian: `import { NoUnusedExpressionWalker } from "./noUnusedExpressionRule";` might be a nicer choice for this file

JKillian:
This looks good @bencoveney! I know the `no-new` name came from ESLint, but I'm not a big fan because it sounds like it disallows using `new` entirely. `no-unused-new` seems like a better name to me. What do you think? 
cc @adidahiya

adidahiya:
agreed @JKillian; I would even go further and call it `no-unused-new-expression` but `no-unused-new` is fine too

bencoveney:
Thanks for the help so far. This should address the comments.

JKillian:
Thanks @bencoveney!

PR #1317, opened 2016-06-19
Indent rule option to allow spaces after tabs

### LINKED ISSUE #1306: More lenient "indent": [true, "tabs"]
Consider the following formatting where `~~~>` is a tab and `-` is a space:

```typescript
~~~>if-(longArgumentlongArgumentlongArgument
~~~>---&& anotherLongArgumentanotherLongArgument) {
~~~>~~~>// ...
~~~>}
```

This gets flagged by the indent rule.

Would you be up for me adding an option to allow <4 spaces at the end of the tabs?

JoshuaKGoldberg:
Allows a configurable maximum number of spaces after tabs.

Fixes #1306.

### DIFF
--- a/README.md
+++ b/README.md
@@ -236,6 +236,7 @@ Core rules are included in the `tslint` package.
 * `forin` enforces a `for ... in` statement to be filtered with an `if` statement.
 * `indent` enforces indentation with tabs or spaces. Rule options (one is required):
     * `"tabs"` enforces consistent tabs.
+        * additionally pass an object with an `"allow-spaces"` member for a maximum number of spaces allowed after the tabs
     * `"spaces"` enforces consistent spaces.
 * `interface-name` enforces consistent interface names. Rule options:
     * `"always-prefix"` enforces interface names must have an 'I' prefix
--- a/src/rules/indentRule.ts
+++ b/src/rules/indentRule.ts
@@ -20,6 +20,7 @@ import * as Lint from "../lint";
 
 const OPTION_USE_TABS = "tabs";
 const OPTION_USE_SPACES = "spaces";
+const OPTION_ALLOW_SPACES = "allow-spaces";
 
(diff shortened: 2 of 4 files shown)


JKillian:
(Feeling a little hesitant)[https://github.com/palantir/tslint/issues/1306#issuecomment-227623464] on this, let's see what others think before going further. Worst case you can always make this a custom rule if needed in some of your projects

olegstepura:
IMHO one should make a choice and never mix spaces with tabs.

chrismbarr:
Yeah I agree, I can't see a lot of people needing a rule to enforce tabs only, but also sometime allow a maximum of 3 spaces after that.  Perhaps that makes sense for your team, but I cannot think of a good use case for that option.

Typically the choices are "tabs", "spaces", or "I don't care to enable this rule"

JoshuaKGoldberg:
@chrismbarr I've put a defense + use case in the other thread, https://github.com/palantir/tslint/issues/1306#issuecomment-227795888

JKillian:
I understand where you're coming form @JoshuaKGoldberg, but mixing tabs and spaces doesn't seem like an idea we want to encourage, so I'm going to close this. I never enjoy closing PRs that people have put time and effort into, but in this case it seems like the best course of action.

JoshuaKGoldberg:
Makes sense, no worries! I appreciate you & the other participants taking the time to think it over and give feedback.

vhakulinen:
I'll leave my (and probably my colleague's) two cents here and vouch for this opt-in feature. I can really relate to this comment: https://github.com/palantir/tslint/issues/1306#issuecomment-227795888 (already mentioned above by @JoshuaKGoldberg). _Really_ useful when you're using tabs for indention but still want to line up arguments or conditions in if statement with the line above (after indention tabs).

PR #1318, opened 2016-06-19
Added no-non-arrow-functions rule

### LINKED ISSUE #165: new rule: ban anonymous (non-arrow) functions
Advantages as advertised here:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/arrow_functions

JoshuaKGoldberg:
Disallows traditional `function () { ... }` declarations, preferring `() => { ... }` arrow lambdas.

Fixes #165.


src/rules/noNonArrowFunctionsRule.ts
```diff
@@ -0,0 +1,51 @@
...
+ */
+
+import * as ts from "typescript";
+import * as Lint from "../lint";
+
+export class Rule extends Lint.Rules.AbstractRule {
+    /* tslint:disable:object-literal-sort-keys */
+    public static metadata: Lint.IRuleMetadata = {
+        ruleName: "no-non-arrow-functions",
+        description: "Diallows traditional (non-arrow) function expressions.",
```
JKillian: typo, "Disallows"

test/rules/no-non-arrow-functions/test.ts.lint
```diff
@@ -0,0 +1,23 @@
+function foo(a: any) : any { // error
```
JKillian: Don't need the `// error` comments in this file since the markup indicates that itself

JKillian:
I wonder if `only-arrow-functions` would be a clearer name than `no-non-arrow-functions`? The double-negative in the latter worries me.

JoshuaKGoldberg:
> only-arrow-functions

Agreed, way better.

glen-84:
Also, isn't this rule supposed to only apply to anonymous functions? It's flagging stuff like:

```ts
function configureContainer(container: Container) {
    // ...
}
```

adidahiya:
@glen-84 I agree, I expect this rule to allow named functions using the `function` keyword. would you mind filing a separate issue?

glen-84:
Sure.

PR #1325, opened 2016-06-21
ordered-imports rule

### LINKED ISSUE #771: New rule: import-ordering
<sup>_Update July 1, 2016_</sup>

This rule was released as "ordered-imports" in TSLint v3.13.0.

<sup>_Update Jun 14, 2016_</sup>

**Import ordering**

Option: `check-groups` enforces formatting imports into groups (separated by single newlines) in a certain order:
1. third-party imports (standard NPM dependencies)
2. NPM-scoped imports
3. relative path imports

Option: `alphabetize` enforces alphabetic ordering of `from` clause
- if `check-groups` is enabled, then order is enforced _within_ each group
- if `check-groups` is not enabled, then enforce overall ordering

danvk:
Fixes #771

See test & rule metadata for how this works. No options for now, but I could imagine many different ways to order & group imports.

### DIFF
--- a/src/rules/orderedImportsRule.ts
+++ b/src/rules/orderedImportsRule.ts
@@ -1,0 +1,138 @@
+import * as Lint from "../lint";
+import * as ts from "typescript";
+
+export class Rule extends Lint.Rules.AbstractRule {
+    /* tslint:disable:object-literal-sort-keys */
+    public static metadata: Lint.IRuleMetadata = {
+        ruleName: "ordered-imports",
+        description: "Requires that import statements be alphabetized.",
+        descriptionDetails: Lint.Utils.dedent`
+            Enforce a consistent ordering for ES6 imports:
+            - Named imports must be alphabetized (i.e. "import {A, B, C} from "foo";")
+                - The exact ordering can be controled by the named-imports-order option.
+                - "longName as name" imports are ordered by "longName".
+            - Import sources must be alphabetized within groups, i.e.:
+                    import * as foo from "a";
+                    import * as bar from "b";
+            - Groups of imports are delineated by blank lines. You can use these to group imports
+                however you like, e.g. by first- vs. third-party or thematically.`,
+        optionsDescription: Lint.Utils.dedent`
+            You may set the \`"named-imports-order"\` option to control the ordering of named
+            imports (the \`{A, B, C}\` in \'import {A, B, C} from "foo"\`.)
+
+            Possible values for \`"named-imports-order"\` are:
+
+            * \`"case-insensitive'\`: Correct order is \`{A, b, C}\`. (This is the default.)
+            * \`"lowercase-first"\`: Correct order is \`{b, A, C}\`.
+            * \`"lowercase-last"\`: Correct order is \`{A, C, b}\`.
+        `,
+        options: {
+            type: "object",
+            properties: {
+                "named-imports-order": {
+                    type: "string",
+                    enum: ["case-insensitive", "lowercase-first", "lowercase-last"],
+                },
+            },
+            additionalProperties: false,
+        },
+        optionExamples: ["true", '[true, {"named-imports-order": "lowercase-first"}]'],
+        type: "style",
+    };
+    /* tslint:enable:object-literal-sort-keys */
+
+    public static NAMED_IMPORTS_UNORDERED = "Named imports must be alphabetized.";
+    public static IMPORT_SOURCES_UNORDERED = "Import sources within a group must be alphabetized.";
+
+    public apply(sourceFile: ts.SourceFile): Lint.RuleFailure[] {
+        const orderedImportsWalker = new OrderedImportsWalker(sourceFile, this.getOptions());
+        return this.applyWithWalker(orderedImportsWalker);
+    }
+}
+
+// Convert aBcD --> AbCd
+function flipCase(x: string): string {
+    return x.split("").map(char => {
+        if (char >= "a" && char <= "z") {
+            return char.toUpperCase();
+        } else if (char >= "A" && char <= "Z") {
+            return char.toLowerCase();
+        }
+        return char;
+    }).join("");
+}
+
+// After applying a transformation, are the nodes sorted according to the text they contain?
+// If not, return the pair of nodes which are out of order.
+function findUnsortedPair(xs: ts.Node[], transform: (x: string) => string): [ts.Node, ts.Node] {
+    for (let i = 1; i < xs.length; i++) {
+        if (transform(xs[i].getText()) < transform(xs[i - 1].getText())) {
+            return [xs[i - 1], xs[i]];
+        }
+    }
+    return null;
+}
+
+// Transformations to apply to produce the desired ordering of named imports.
+// The imports must be lexicographically sorted after applying the transform.
+const TRANFORMS: {[ordering: string]: (x: string) => string} = {
+    "case-insensitive": (x: string) => x.toLowerCase(),
+    "lowercase-first": flipCase,
+    "lowercase-last": (x: string) => x,
+};
+
+class OrderedImportsWalker extends Lint.RuleWalker {
+    // This gets reset after every blank line.
+    private lastImportSource: string = null;
+    private namedImportsOrder: string = null;
+
+    constructor(sourceFile: ts.SourceFile, options: Lint.IOptions) {
+        super(sourceFile, options);
+
+        const optionSet = this.getOptions()[0] || {};
+        this.namedImportsOrder = optionSet["named-imports-order"] || "case-insensitive";
+    }
+
+    // e.g. "import Foo from "./foo";"
+    public visitImportDeclaration(node: ts.ImportDeclaration) {
+        const source = node.moduleSpecifier.getText();
+
+        if (this.lastImportSource && source < this.lastImportSource) {
+            this.addFailure(this.createFailure(node.getStart(), node.getWidth(),
+                 Rule.IMPORT_SOURCES_UNORDERED));
+        }
+        this.lastImportSource = source;
+
+        super.visitImportDeclaration(node);
+    }
+
+    // This is the "{A, B, C}" of "import {A, B, C} from "./foo";".
+    // We need to make sure they're alphabetized.
+    public visitNamedImports(node: ts.NamedImports) {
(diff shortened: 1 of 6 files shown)


test/rules/ordered-imports/test.ts.lint
```diff
@@ -0,0 +1,21 @@
...
+        ~~~~                            [ordered-imports]
+
+
+// Import sources should be alphabetized.
+import * as bar from 'bar';
+import * as foo from 'foo';
+
+import * as foo from 'foo';
+import * as bar from 'bar'; // failure
+~~~~~~~~~~~~~~~~~~~~~~~~~~~            [ordered-sources]
```
thasner: consider adding a test for blank-lines resetting the `lastImportSource`

thasner:
👍

danvk:
I added a line to the test that would break if `lastImportSource` weren't reset.

danvk:
@adidahiya mind taking a look at this? Thanks!

src/rules/orderedImportsRule.ts
```diff
@@ -0,0 +1,125 @@
...
+    for (let i = 1; i < xs.length; i++) {
+        if (flipCase(xs[i].getText()) < flipCase(xs[i - 1].getText())) {
+            return [xs[i - 1], xs[i]];
+        }
+    }
+    return null;
+}
+
+// Returns "./utils" given the tree for, e.g. "import * as utils from "./utils";"
+function findFromSource(node: ts.Node): string {
```
JKillian: Instead of doing all this (which seems correct to me btw) you should be able to do something like `node.moduleSpecifier.getText()`. (Assuming `node` is a `ts.ImportDeclaration`.)

src/rules/orderedImportsRule.ts
```diff
@@ -0,0 +1,125 @@
...
+    const num = node.getChildCount();
+    for (let i = 0; i < num - 1; i++) {
+        const child = node.getChildAt(i);
+        if (child.kind !== ts.SyntaxKind.FromKeyword) {
+            continue;
+        }
+        const next = node.getChildAt(i + 1);
+        if (next.kind !== ts.SyntaxKind.StringLiteral) {
```
JKillian: I appreciate this explicit handling of errors, but it turns out its normally better to silently continue if something malformed is found. Reason being that when people use TSLint in their editor, their code gets linted as they type and the resulting AST isn't always completely well-formed. We don't really want TSLint to start throwing errors here though, it's better for it to just move on.

src/rules/orderedImportsRule.ts
```diff
@@ -0,0 +1,125 @@
...
+        }
+        this.lastImportSource = source;
+
+        super.visitImportDeclaration(node);
+    }
+
+    // This is the "{A, B, C}" of "import {A, B, C} from "./foo";".
+    // We need to make sure they're alphabetized.
+    public visitNamedImports(node: ts.NamedImports) {
+        const names = node.getChildAt(1);    // Three children: "{", "A, B, C" and "}"
```
JKillian: `node` here will have a field called `elements`, which is an array of `ts.ImportSpecifier`

test/rules/ordered-imports/test.ts.lint
```diff
@@ -0,0 +1,22 @@
...
+
+// Import sources should be alphabetized.
+import * as bar from 'bar';
+import * as foo from 'foo';
+
+import * as abc from 'abc';
+import * as foo from 'foo';
+import * as bar from 'bar'; // failure
+~~~~~~~~~~~~~~~~~~~~~~~~~~~            [ordered-sources]
+
```
JKillian: Could we add a test case with this kind of import:
```
import someDefault from "module";
```

as well as the combined form:

```
import someDefault, {nameA, nameB: differentIdentifier} from "./wherever";
```

Just is a nice sanity check that the rule doesn't throw errors or struggle with all the different forms of imports

src/rules/orderedImportsRule.ts
```diff
@@ -0,0 +1,125 @@
...
+        const imports: ts.Node[] = [];
+        for (const child of names.getChildren()) {
+            if (child.kind === ts.SyntaxKind.ImportSpecifier) {
+                imports.push(child);
+            }
+        }
+
+        const pair = findUnsortedPair(imports);
+        if (pair) {
```
JKillian: Small stylistic note: we like to explicitly do `pair != null` since it avoids the unexpected quirks of JS type conversions.

src/rules/orderedImportsRule.ts
```diff
@@ -0,0 +1,125 @@
...
+
+        super.visitNamedImports(node);
+    }
+
+    // Check for a blank line, in which case we should reset the import ordering.
+    public visitNode(node: ts.Node) {
+        const prefixLength = node.getStart() - node.getFullStart();
+        const prefix = node.getFullText().slice(0, prefixLength);
+
+        if (prefix.indexOf("\n\n") >= 0) {
```
JKillian: Will this work on files with `\r\n` linebreaks?

JKillian:
This PR looks very good, thanks @danvk! I left a few comments above for tweaks/improvements.

My first thoughts were that `import {SomeClass, someFunc} from "a"` should be enforced instead of `import {someFunc, SomeClass} from "a"`. Perhaps this is one of the cases though where it doesn't really matter which is the default, as long as users can choose which they want (eventually). There do seem to be many possible options for this rule, but perhaps we're better off starting simple as you've implemented it here.

danvk:
@JKillian Thanks for the educational review! I've taken all your suggestions and added an option to control the order of named declarations. Valid options are:

  * `"case-insensitive"`: Correct order is `{A, b, C}`. (This is the default.)
  * `"lowercase-first"`: Correct order is `{b, A, C}`.
  * `"lowercase-last"`: Correct order is `{A, C, b}`.

For named imports like:

```
import {longName as name} from 'foo';
import {name: longName} from 'foo';
```

I'm sorting by the full text, i.e. `longName as name` and `name: longName`.

Let me know if you have any more suggestions.

src/rules/orderedImportsRule.ts
```diff
@@ -0,0 +1,137 @@
...
+
+            * \`"case-insensitive'\`: Correct order is \`{A, b, C}\`. (This is the default.)
+            * \`"lowercase-first"\`: Correct order is \`{b, A, C}\`.
+            * \`"lowercase-last"\`: Correct order is \`{A, C, b}\`.
+        `,
+        options: {
+            type: "object",
+            properties: {
+                "named-imports-order": {
+                    oneOf: ["case-insensitive", "lowercase-first", "lowercase-last"],
```
JKillian: I don't believe this is valid JSO Schema. I think you'll want this:

```
"named-imports-order": {
    type: "string",
    enum: ["case-insensitive", "lowercase-first", "lowercase-last"]
}
```

JKillian:
My apologies for my one comment, I don't believe `import { symbol: localName } from "./wherever";` is valid ES6 or valid TS! Using `as` as you mention above is correct, and I think the way you handle it currently is good.

(As as aside, I wonder why the syntax differs between importing and destructuring?)

Those options you provide seem good to me. It also keeps the door open for adding ordering of groups in the future.

src/rules/orderedImportsRule.ts
```diff
@@ -0,0 +1,137 @@
...
+            },
+            additionalProperties: false,
+        },
+        optionExamples: ["true", '[true, {"named-imports-order": "lowercase-first"}]'],
+        type: "style",
+    };
+    /* tslint:enable:object-literal-sort-keys */
+
+    public static NAMED_IMPORTS_UNORDERED = "Named imports must be alphabetized.";
+    public static IMPORT_SOURCES_UNORDERED = "Import sources within a block must be alphabetized.";
```
JKillian: Since we use the word "group" in the documentation above, lets also use it here. I'm afraid "block" will imply the `{  }` type of block

src/rules/orderedImportsRule.ts
```diff
@@ -0,0 +1,137 @@
...
+        description: "Requires that import statements be alphabetized.",
+        descriptionDetails: Lint.Utils.dedent`
+            Enforce a consistent ordering for ES6 imports:
+            - Named imports must be alphabetized (i.e. "import {A, B, C} from "foo";")
+                - The exact ordering can be controled by the named-imports-order option.
+                - "longName as name" imports are ordered by "longName".
+            - Import sources must be alphabetized within groups, i.e.:
+                    import * as a from "a";
+                    import * as b from "b";
+            - groups of imports are delineated by blank lines. You can use these to group imports
```
JKillian: Let's capitalize "Groups" to be consistent with the above bullet points

src/rules/orderedImportsRule.ts
```diff
@@ -0,0 +1,137 @@
...
+    public static metadata: Lint.IRuleMetadata = {
+        ruleName: "ordered-imports",
+        description: "Requires that import statements be alphabetized.",
+        descriptionDetails: Lint.Utils.dedent`
+            Enforce a consistent ordering for ES6 imports:
+            - Named imports must be alphabetized (i.e. "import {A, B, C} from "foo";")
+                - The exact ordering can be controled by the named-imports-order option.
+                - "longName as name" imports are ordered by "longName".
+            - Import sources must be alphabetized within groups, i.e.:
+                    import * as a from "a";
```
JKillian: Can we change this to:

```
import * as foo from "a";
import * as bar from "b";
```

so users can more easily distinguish what they should sort on?

JKillian:
Looking good! Just left a few comments on some minor docs changes

danvk:
Addressed all comments.

I'm not sure why the syntaxes for imports and destructuring are different. I certainly find the `{name as localName}` syntax to be much clearer. I never use `const {a: b} = o;` because I can never remember whether that means `const a = o.b;` or `const b = o.a;`.

JKillian:
Sweet, merged!

glen-84:
I order my imports based on the full text, and not the "source". Two benefits are:

1. I can use something like "sort lines" in my editor (VS Code).
2. It's easier to look up a named import (for example) in this alphabetical order. The "source" is possibly less important.

I'm not 100% sure about this though. I may change my mind. :stuck_out_tongue:

PR #1326, opened 2016-06-22
Updated Grunt build now fails if a rule is *not* documented in README.md

HamletDRC:
Updated Grunt build now fails if a rule is *not* documented in README.md


JKillian:
I'm thinking about removing the list of rules from the README actually and simply linking to the docs site that has all the info. Two reasons for this:

* Maintaining duplicate documentation in two places is a pain
* The README is really long right now
* Sometimes when a new rule gets added, but before a release, people get confused when they see it in the README. It's easier to only update the docs site when doing a release.

JKillian:
Rule list removed from README so this is no longer needed, thanks though!

Issue #1335, opened 2016-06-24
Banning for loops when a TS for(...of...) loop might be better


chrismbarr:
Basically a standard `for` loop should only be allowed if the incrementer variable is used in the block of the loop for anything other than accessing the current array item.

 👍 **This is Fine** since the array index is needed for something
```ts
for (var j = 0; j <= arr.length; j++) {
    doMath(j);
}
```

👎 **This could be replaced with a `for(...of...)` loop** since the index is only used to access the array
```ts
for (var i = 0; i <= arr.length; i++) {
    doThing(arr[i]);
}
```
**Here's the preferred way to write the above loop**
```ts
for (var item of arr) {
    doThing(item);
}
```

----

Is this a feasible rule idea?  I've [already begun some work on it](https://github.com/chrismbarr/tslint/commit/78fa93f101eb622feab6e2f1dc5655140a9e0a12), but I'll work on it more later.  Right now it's only partially working since I'm not sure how to look for variable usages within a certain scope/Block/SyntaxList.  Any tips would be appreciated!

adidahiya:
I'm a fan of this rule; it could be called `prefer-for-of-loop`.

> I'm not sure how to look for variable usages within a certain scope/Block/SyntaxList

One idea is to use the regular `languageService.getDocumentHighlights` method, iterate through the highlights and verify that the source code positions of those spans are within the Block you are searching within.

scameron:
If you are adding elements to the end of the array being iterated inside the loop, don’t you need to use a standard for?

beenotung:
I agree for-of is more readable and easier to maintain, however, it may hurt the performance.

For example, in Firefox: for loop with index has 219,705 ops/sec, while for-of has only 35,827 ops/sec. Which is 84% slower

demo:
https://jsperf.com/foreach/19

wortwart:
This rule should not be default because in the DOM there are array-like objects which you can't iterate with for..of or .forEach() - at least not without extra hassle. I've been struggling repeatedly with this rule in node collections like you get with document.querySelectorAll(). 
Classic for loops are highly performant, simple, and readable and should IMHO be considered good practice.

PR #1340, opened 2016-06-25
New Rule: prefer-for-of

chrismbarr:
Based on discussions in #1335, i've made a new rule: `prefer-for-of`! 

**Reasoning**
Basically a standard `for` loop should only be allowed if the incrementer variable is used in the block of the loop for **anything other than accessing the current array item**.

 👍 **This is Fine** since the array index is needed for something
```ts
for (var j = 0; j <= arr.length; j++) {
    doMath(j);
}
```

👎 **This could be replaced with a `for(...of...)` loop** since the index is only used to access the array
```ts
for (var i = 0; i <= arr.length; i++) {
    doThing(arr[i]);
}
```

It will fail with the text:
>This standard for loop could be replaced with a for(... of ...) loop

👍 **Here's the preferred way to write the above loop**
```ts
for (var item of arr) {
    doThing(item);
}
```

This will also catch for loops that use syntax like: `for(var t=0, len=arr.length; t < len; t++)`


JKillian:
Sorry for the delay getting to this and other PRs! Not sure when I will, but promise I will eventually!

README.md
```diff
@@ -306,6 +306,7 @@ Core rules are included in the `tslint` package.
 * `one-variable-per-declaration` disallows multiple variable definitions in the same statement.
   * `"ignore-for-loop"` allows multiple variable definitions in for loop statement.
 * `only-arrow-functions` disallows traditional `function () { ... }` declarations, preferring `() => { ... }` arrow lambdas.
+* `prefer-for-of` recommends a [TypeScript `for(... of ...)` loop](https://www.typescriptlang.org/docs/handbook/iterators-and-generators.html) over a standard `for` loop if the index is only used to access the array being looped over.
```
JoshuaKGoldberg: This is an ES2015 construct, no?

chrismbarr: Well so it is!  I was not aware, I thought it was just TypeScript syntactic sugar.

JKillian:
Can you merge develop into your branch so the tests will pass? Will review this PR this week!

chrismbarr:
I was on vacation last week - back now!  I'll try and get my PR's fixed.

ugh.... this master merge really got screwed up.  Right now it might be easier for me to just create a new PR with a single commit...

chrismbarr:
closing this since there are so many commits and merge issues with master due to the age of this branch.  See #1435 instead.

Issue #1097, opened 2016-04-08
Rule request: forbid return in finally


louy:
Adding a `return` statement in a `finally` block is a very confusing thing. Thrown errors get ignored, `return` statements in `try` blocks get ignored, and your mind explodes. 

I'd love to have a rule to forbid this. Especially now that we're having `async`/`await`.

examples:

``` js
try {
return 1;
} catch(err) {
return 2;
} finally {
return 3;
}
// > 3

try {
throw new Error;
return 1;
} catch(err) {
return 2;
} finally {
return 3;
}
// > 3
```

adidahiya:
sounds reasonable. is there any analogous rule for ESLint? if not, that's cool too

louy:
not yet. https://github.com/eslint/eslint/issues/5808

andy-hanson:
Should be fixed by #1349.

PR #1349, opened 2016-06-26
Implement rule to forbid return statements in finally blocks #1097

lowkay:


lowkay:
Hmm, looking at this again and while I believe the code is easily understandable and simple, it does require 2 passes over finally blocks, which is not ideal.

I am thinking of an alternative approach to do it in a single pass by tracking current scope and flagging if the current scope is in a finally by taking advantage of scopeAwareRuleWalker.

JKillian:
Good idea for a rule! I wasn't aware of the weird semantics until I looked it up:

image

JKillian:
Could you add this rule to `src/configs/latest.ts`?

src/rules/noReturnInFinallyRule.ts
```diff
@@ -0,0 +1,87 @@
...
+import * as ts from "typescript";
+import * as Lint from "../lint";
+
+export class Rule extends Lint.Rules.AbstractRule {
+    /* tslint:disable:object-literal-sort-keys */
+    public static metadata: Lint.IRuleMetadata = {
+        ruleName: "no-return-in-finally",
+        description: "Disallows return statements in finally blocks.",
+        descriptionDetails: "",
+        rationale: "Return statements inside finally blocks have confusing semantics.",
```
JKillian: You could add a link ([example](https://github.com/palantir/tslint/blob/ae4ef37f0cd03dfe2b08650447fde7565d9cbdd4/src/rules/noArgRule.ts#L29)) or some more details about what these confusing semantics are. Not a blocking issue though

src/rules/noReturnInFinallyRule.ts
```diff
@@ -0,0 +1,87 @@
...
+        optionsDescription: "Not configurable.",
+        options: null,
+        optionExamples: ["true"],
+        type: "functionality",
+    };
+    /* tslint:enable:object-literal-sort-keys */
+
+    public static FAILURE_STRING = "finally block contains return statement";
```
JKillian: To make this message more consistent with other failure messages, I'd go with: "Return statements in finally blocks are forbidden"

JKillian:
Should we broaden this rule to watch for other statements beyond just `return`? [ESLint watches for a couple others](http://eslint.org/docs/rules/no-unsafe-finally) as well.

src/rules/noReturnInFinallyRule.ts
```diff
@@ -0,0 +1,87 @@
...
+        return {
+            isFinallyBlock: this.isFinallyBlock(node),
+        };
+    }
+
+    protected isScopeBoundary(node: ts.Node): boolean {
+        return super.isScopeBoundary(node) || this.isFinallyBlock(node);
+    }
+
+    private isFinallyBlock(node: ts.Node): boolean {
```
JKillian: You can move `isFinallyBlock` and `isTryStatement` outside of the class as they don't reference any class members. Very minor though, not a blocking issue for merging this

test/rules/no-return-in-finally/test.ts.lint
```diff
@@ -0,0 +1,132 @@
...
+            bar() {
+                return 1;
+            }
+        }
+    }
+
+    return 1;
+}
+
+[no-return-in-finally]: finally block contains return statement
```
JKillian: newlines preferred at the end of files (also below)

JKillian:
Code looks really good @lowkay! Only major thing here is to decide if the rule should handle more than just `return` statements as mentioned above. That'll affect the name of the rule, so we need to get it figured out before merging

lowkay:
I agree with you that it should handle the same as eslint, take the rule name from theirs "no-unsafe-finally"?

src/rules/noReturnInFinallyRule.ts
lowkay: Would it be acceptable to link to eslint.org (seems a bit weird)? Or maybe: "Using control-flow statements such as return, continue, break and throws inside finally blocks overrides any control-flow statements in the same try ... catch scope. This is considered confusing and unexpected behavior."

JKillian:
I like that it's consistent with ESLint (easier for users who use both), but I feel like "unsafe" implies the wrong idea. It's not so much unsafe as unexpected behavior. I'll ponder the name some more, feel free to make changes under the current name for now

lowkay:
Cool... Looking at this further, the new semantics mean that handling continue/break need additional scopes :(

So for example if we see a continue/break inside a switch inside a finally that's fine, but see a return anywhere outside a new scope boundary and we need to fail.

Then how about handling breaking out of labels? We need to track where the label is with respect to the finally block.

Essentially handling the different scoping of continue/break over return makes this rule much more complex to implement.

lowkay:
I *think* I know how to handle it though - each scope indicates it's scoping impact (full boundary for return/throws vs partial boundary for continue/break) then we walk up the scopes:
1. for return/throws that means walking up to the first full boundary and if that's a finally then error,
2. for continue/break that means walking up to the first applicable scope (tracking labels) , if we reach a full boundary and that is a finally then error.

lowkay:
Naming-wise I thought perhaps no-control-flow-in-finally but that's awkward...

JKillian:
@lowkay That implementation strategy seems good! If some weird edge case is overly complex (for example, tracking labels) we can opt for code simplicity for this initial PR and enhance later.

@adidahiya any thoughts on rule names?

src/rules/noReturnInFinallyRule.ts
adidahiya: @lowkay looks good, here are a few small edit suggestions:

> When used inside `finally` blocks, control flow statements such as `return`, `continue`, `break` and `throws` override any other control flow statements in the same try/catch scope. This is confusing and unexpected behavior.

adidahiya:
using the same rule name as eslint ("no-unsafe-finally") seems fine here since it has the same semantics

JKillian:
Seems good to me, let's go with `no-unsafe-finally` then

lowkay:
@JKillian @adidahiya I believe I have made all the changes mentioned, additionally I managed to get label support in too, so `continue label;` and `break label;` will cause an error if they reference a label outside of the finally scope.

JKillian:
Thanks @lowkay! Solid work on this which turned out to be trickier than expected

PR #1352, opened 2016-06-27
Fix #1269 - Too easy to disable all rules accidentally

### LINKED ISSUE #1269: Too easy to disable all rules accidentally
It's very easy to accidentally disable all lint rules by forgetting a `:` after `tslint:disable`. Can we also allow the following syntax?

```ts
/* tslint:disable max-line-length */
(code)
/* tslint:enable max-line-length */
```

#### Actual behavior

All lint rules are disabled for `(code)`.

#### Expected behavior

Only `max-line-length` is disabled for `(code)`.

lowkay:
Allow for tslint enable/disable rules to be specified without requiring a colon (':') delimiter before the list.


chrismbarr:
Will this also address the constant typo I make?

I always type this:
`/* tslint: disable: rule-name */`
instead of
`/* tslint:disable: rule-name */`

adidahiya:
lgtm, thanks @lowkay!

sorry @chrismbarr, this doesn't address that typo.

PR #1355, opened 2016-06-28
Update peer dependency for TypeScript nightly version of 2.0.0-dev

yuit:
TypeScript Compiler updates it version for release-2.0.0 [here](https://github.com/Microsoft/TypeScript/blob/master/package.json#L5) Update the peer dependency, so older node can install peer dependency correctly as this is currently broken TypeScript build

### DIFF
--- a/package.json
+++ b/package.json
@@ -46,7 +46,7 @@
     "typescript": "next"
   },
   "peerDependencies": {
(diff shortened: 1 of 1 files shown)


JKillian:
@yuit Do you need a new release with this change right away?

JKillian:
Published as `tslint@3.12.0-dev.2`

yuit:
@JKillian thanks! just try it out look good

PR #1357, opened 2016-06-28
Ignore MPEG transport streams in tslint

mprobst:
MPEG transport streams also use the `.ts` file extension. Linting causes odd
effects, either by reporting spurious errors, or by timing out because the files
are commonly too large.

### DIFF
--- a/src/tslint-cli.ts
+++ b/src/tslint-cli.ts
@@ -207,6 +207,22 @@ const processFile = (file: string) => {
         process.exit(1);
     }
 
+    const buffer = new Buffer(256);
+    buffer.fill(0);
+    const fd = fs.openSync(file, "r");
+    try {
+        fs.readSync(fd, buffer, 0, 256, null);
+        if (buffer.readInt8(0) === 0x47 && buffer.readInt8(188) === 0x47) {
+            // MPEG transport streams use the '.ts' file extension. They use 0x47 as the frame
+            // separator, repeating every 188 bytes. It is unlikely to find that pattern in
+            // TypeScript source, so tslint ignores files with the specific pattern.
+            console.warn(`${file}: ignoring MPEG transport stream`);
+            return;
+        }
+    } finally {
+        fs.closeSync(fd);
+    }
+
     const contents = fs.readFileSync(file, "utf8");
     const configuration = findConfiguration(possibleConfigAbsolutePath, file);
 


JKillian:
Ha, this PR made me laugh, love to see the Lint results on a MPEG transport stream file...

It seems unlikely to me that users have TypeScript and MPEG transport stream files mixed together when linting, but I know Google has tooling set up in some unexpected ways. It sounds like this problem came up in an actual scenario?

src/formatters/checkstyleFormatter.ts
```diff
@@ -3,7 +3,7 @@ import {RuleFailure} from "../language/rule/rule";
 
 export class Formatter extends AbstractFormatter {
     public format(failures: RuleFailure[]): string {
-        let output = '<?xml version="1.0" encoding="utf-8"?><checkstyle version="4.3">';
+        let output = "<?xml version=\"1.0\" encoding=\"utf-8\"?><checkstyle version=\"4.3\">";
```
JKillian: All these string changes shouldn't be necessary - possibly you were compiling with an old version of TSLint that didn't allow this?

mprobst: Not sure – I just run `grunt`, and get errors in this file. Wouldn't that use the just compiled version of TSLint?

src/tslint-cli.ts
```diff
@@ -207,6 +207,21 @@ const processFile = (file: string) => {
         process.exit(1);
     }
 
+    const buffer = new Buffer(256);
```
JKillian: the `Buffer` constructor doesn't initialize its content. As the file we're reading could be less than 256 bytes, we could end up reading junk data in the conditional. (Although I'm not entirely sure of the behavior of `readSync`.) I think we should zero the memory with `Buffer.fill` here

mprobst:
@JKillian our internal linting tools trigger just by file extension. So in our monolithic repository, if a change includes any `.ts` file, tslint runs and – in the case of videos – complains, mostly about missing semicolons. Surprisingly, developers are reluctant to add the required semicolons to their videos ;-)

src/formatters/checkstyleFormatter.ts
JKillian: We actually use the previous version of TSLint to lint the code, so it's likely you just had an old version in `node_modules`.

src/tslint-cli.ts
mprobst: Done. `readSync` iiuc returns the number of bytes read, but wouldn't overwrite sections of the buffer it doesn't touch.

src/formatters/checkstyleFormatter.ts
mprobst: Makes sense, and works out. Reverted this commit.

src/tslint-cli.ts
```diff
@@ -207,6 +207,22 @@ const processFile = (file: string) => {
...
+    buffer.fill(0);
+    const fd = fs.openSync(file, "r");
+    try {
+        fs.readSync(fd, buffer, 0, 256, null);
+        if (buffer.readInt8(0) === 0x47 && buffer.readInt8(188) === 0x47) {
+            // MPEG transport streams use the '.ts' file extension. They use 0x47 as the frame
+            // separator, repeating every 188 bytes. It is unlikely to find that pattern in
```
JKillian: One last concern, sorry I didn't notice this before. `0x47` is "G", which means a TS file where the first and the 189th character are "G" won't be linted. Admittedly, it's not highly likely that this condition is met, but it seems like a non-trivial likelihood, no?

JKillian: Just for fun, I grabbed about the top 1000 TypeScript repos off Github which combined had about 65,000 `.ts` files. About 700, or 1%, had a "G" as the first and the 189th character.

This seemed surprisingly high, given that "/" and "i" are by far (and expectedly) the most common first characters.

Turns out every one of the files with two "G"s was an MPEG transport stream file. 😉

mprobst:
Nice analysis!

src/tslint-cli.ts
```diff
@@ -207,6 +207,22 @@ const processFile = (file: string) => {
...
+            // TypeScript source, so tslint ignores files with the specific pattern.
+            console.warn(`${file}: ignoring MPEG transport stream`);
+            return;
+        }
+    } finally {
+        fs.closeSync(fd);
+    }
+
     const contents = fs.readFileSync(file, "utf8");
```
jwbay: rather than opening every file twice, would it make more sense to read the full contents into a buffer once, check the the bytes, then toString if it's okay and carry on? or maybe re-use the descriptor?

mprobst: I considered that, but another problem with videos is that they are often huge, hundreds of megabytes, which can be enough to kill the NodeJS process. So this code on purpose only peeks 256 bytes to make that decision.

Overall the biggest cost in reading the file is the hard disk, so after peeking it will be cached by the OS, and then parsing, which is unconcerned.

We could change this to keep reading from the file descriptor and appending into buffer, then converting that, but I don't think it'll be worth it.

Issue #1336, opened 2016-06-24
no-unused-variable: private ctor params shouldn't require `this.` when used in the ctor


chrismbarr:
### Bug Report

- __TSLint version__: `3.11.0`
- __TypeScript version__: `1.8.10`
- __Running TSLint via__: `gulp-tslint` via Visual Studio 2015

#### TypeScript code being linted

```ts
class Example {
  constructor(private thing1: string, private thing2: string){
    console.log(thing1);
    console.log(this.thing2);
  }
}
```

with `tslint.json` configuration:

```json
"no-unused-variable": [ true, "check-parameters" ],
```

#### Actual behavior

The private `thing1` variable is reported as unused, even though it is used without the `this.` prefix, which is allowed in a constructor.

#### Expected behavior

Used variables in the constructor should not require the `this.` prefix to be considered "used"

JKillian:
`this.thing1` vs. `thing1` have different meanings - consider [the case below][0]:

```
class Foo {
	constructor(public foo = 20) {
		this.foo = 10;
		foo = 5;
	}
}

const f = new Foo();
alert(f.foo);  // will display 10
```

If a private parameter property of a constructor is never accessed accessed anywhere in the class as `this.myProperty`, it actually still is unused as a class member and I think the lint rule is succeeding. Does this line up with what you're seeing from TSLint?

[0]: http://www.typescriptlang.org/play/#src=class%20Foo%20%7B%0A%09constructor(public%20foo%20%3D%2020)%20%7B%0A%09%09this.foo%20%3D%2010%3B%0A%09%09foo%20%3D%205%3B%0A%09%7D%0A%7D%0A%0Aconst%20f%20%3D%20new%20Foo()%3B%0Aalert(f.foo)%3B%0A%0A

chrismbarr:
I was not aware of that, I assumed they meant the same thing.  Thanks for the clarification!

chrismbarr:
Now that I'm thinking about this more... should there be a rule to ban the improper use like this?
```ts
constructor(public foo: number) {
        this.foo = 10;
        foo = 5; //<-- violation
}
```

PR #1358, opened 2016-06-29
New Rule: param-property-in-constructor

chrismbarr:
New rule created based on discussions in #1336

The rule is named `no-improper-constructor-param-usage` and has no configuration options.

Here’s a quick code sample of what will trigger a violation
```ts
constructor(public foo: number) {
    this.foo = 10;
    foo = 5; //<-- violation
}
```


chrismbarr:
Sorry about the merge commits.  I thought github desktop was better than that!  I'll do it from the command line in the future.

JKillian:
How do you feel about the name `param-property-in-constructor`?  "no-improper" feels like an uneeded double-negative to me, and "usage" seems unnecessary. "Parameter properties" is the official name the TS docs use for this sort of a construct, so I wanted to get that in the rule name

src/rules/noImproperConstructorParamUsageRule.ts
```diff
@@ -0,0 +1,96 @@
...
+        const languageService = Lint.createLanguageService(sourceFile.fileName, sourceFile.getFullText());
+        return this.applyWithWalker(new NoImproperCtorParamUsageWalker(sourceFile, this.getOptions(), languageService));
+    }
+}
+
+export class NoImproperCtorParamUsageWalker extends Lint.RuleWalker {
+
+    private languageService: ts.LanguageService;
+
+    constructor(sourceFile: ts.SourceFile, options: Lint.IOptions, languageService: ts.LanguageService) {
```
JKillian: You could just make this `private languageService: ts.LanguageService` if you want (funny considering what this rule does 😄 )

test/rules/no-improper-constructor-param-usage/test.ts.lint
```diff
@@ -0,0 +1,48 @@
...
+        this.num = 10;
+        num = 5;
+        ~~~ [The constructor parameter 'num' should only be accessed with the 'this' keyword: 'this.num']
+
+        animal = "aardvark";
+        ~~~~~~ [The constructor parameter 'animal' should only be accessed with the 'this' keyword: 'this.animal']
+        this.animal = "skunk";
+
+        fruit = "banana";
+        ~~~~~ [The constructor parameter 'fruit' should only be accessed with the 'this' keyword: 'this.fruit']
```
JKillian: This seems like a bug to me, `fruit` is just a regular old parameter, so this should be allowed, right? (Also occurs below)

src/rules/noImproperConstructorParamUsageRule.ts
```diff
@@ -0,0 +1,96 @@
...
+            const fileName = this.getSourceFile().fileName;
+            node.parameters.forEach((param: ts.ParameterDeclaration) => {
+                this.validateParam(param, fileName);
+            });
+        }
+        super.visitConstructorDeclaration(node);
+    }
+
+    private validateParam(param: ts.ParameterDeclaration, fileName: string) {
+        const highlights = this.languageService.getDocumentHighlights(fileName, param.name.getStart(), [fileName]);
```
JKillian: We only need to validate if the parameter is actually a parameter property, right? We check for that in another rule like this: https://github.com/palantir/tslint/blob/81c538c2cdc03bf319f5953bd573271420ff978a/src/rules/noConstructorVarsRule.ts#L48

src/rules/noImproperConstructorParamUsageRule.ts
```diff
@@ -0,0 +1,96 @@
...
+
+    constructor(sourceFile: ts.SourceFile, options: Lint.IOptions, languageService: ts.LanguageService) {
+        super(sourceFile, options);
+        this.languageService = languageService;
+    }
+
+    public visitConstructorDeclaration(node: ts.ConstructorDeclaration) {
+        if (node.parameters && node.parameters.length > 0) {
+            const fileName = this.getSourceFile().fileName;
+            node.parameters.forEach((param: ts.ParameterDeclaration) => {
```
JKillian: `for-of` usage is preferred when you don't need an index argument

adidahiya: Performance and consistency (one iteration style)

src/rules/noImproperConstructorParamUsageRule.ts
chrismbarr: lol, ugh why did I not notice this! :laughing:

test/rules/no-improper-constructor-param-usage/test.ts.lint
chrismbarr: Ah good point.  i suppose this rule should only check params that have a `public` or `private` modifier on it, correct?

chrismbarr:
The test failures seem to be related to changes that were made before this branch was started.  I think it would be fine to merge into master, but please double check me.

JKillian:
Can you merge master here as well, so we can see the CI passing before merging this PR?

test/rules/param-property-in-constructor/test.ts.lint
```diff
@@ -0,0 +1,65 @@
...
+
+        mineral = "Gypsum";
+        ~~~~~~~ [The constructor parameter 'mineral' should only be accessed with the 'this' keyword: 'this.mineral']
+        this.mineral = "Fluorite";
+    }
+}
+
+class FailingExample2 {
+    constructor(public thing: Object) {
+        const wow=thing.doTheThing()
```
JKillian: Hmm, I wasn't thinking about this use case. When accessing a value, it doesn't matter if you use `this.` or not. Do you think we should still disallow this?

JKillian: On the other hand, using `this.` doesn't hurt

chrismbarr: yeah that was my thinking with the way it was written initially.  I was thinking that if we just require `this.` on everything, then there is consistency for how all constructor params are used, regardless if they are used as properties or not.

JKillian:
Hmm, actually looks like TS has some sort of similar functionality build in: https://github.com/Microsoft/TypeScript/pull/10062

YuichiNukiyama:
@JKillian You're right. We can validate unused parameter property  with `--noUnusedParameters ` option.
This feature already has been implimented in TS2.0 beta. And my PR (#10062) only intend to change error message. see [Flag unused declarations with --noUnusedParameters and --noUnusedLocals ](https://github.com/Microsoft/TypeScript/wiki/What%27s-new-in-TypeScript#flag-unused-declarations-with---nounusedparameters-and---nounusedlocals).

adidahiya:
related: #1481

JKillian:
I don't think the benefits of this rule are worth it given the features in TS 2.0. Sorry @chrismbarr! Let me know if you strongly disagree

chrismbarr:
@JKillian :Some code I saw today in a project lead me back here to see why this PR was never accepted.  Reading over the above comments, I'm not entirely sure now.

The intent of this PR was **not** to find unused constructor parameters, it was meant to **enforce the correct usage** of those parameters.

```ts
constructor(public foo: number) {
    this.foo = 10;
    foo = 5; //<-- violation
}
```

I can re-submit this PR, but I don't want to put the work in for it unless it's desired.

PR #1360, opened 2016-06-30
New Rule: max-file-line-count

chrismbarr:
Adding a new rule that limits the number of lines that a file can contain.  When files go over this limit it's usually a sign that the code is too complex, or it has too many responsibilities and it needs to be broken up into smaller pieces. 

The goal here is to count the number of lines, and it should match the line number display in your IDE _(meaning things like blank lines, comments, etc. are included in this count)_

If the rule disabling comment `/* tslint:disable:max-file-line-count */` is added anywhere in the file, this rule is ignored for the file.


JKillian:
I think `file-max-line-count` would fall more in line with how other rules are named

src/rules/maxFileLineCountRule.ts
```diff
@@ -0,0 +1,68 @@
...
+
+    public apply(sourceFile: ts.SourceFile): Lint.RuleFailure[] {
+        const ruleFailures: Lint.RuleFailure[] = [];
+        const lineLimit: number = this.getOptions().ruleArguments[0];
+        const lineCount: number = sourceFile.getLineStarts().length;
+        const disabledIntervals = this.getOptions().disabledIntervals;
+
+        if (lineCount > lineLimit && disabledIntervals.length === 0) {
+            const errorString = Rule.FAILURE_STRING_FACTORY(lineCount, lineLimit);
+            ruleFailures.push(new Lint.RuleFailure(sourceFile, 0, 0, errorString, this.getOptions().ruleName));
```
JKillian: Instead of making this a zero-length error, can we make it either one character long or the length of the first line?

src/tsconfig.json
```diff
@@ -121,6 +122,7 @@
         "rules/oneLineRule.ts",
         "rules/oneVariablePerDeclarationRule.ts",
         "rules/onlyArrowFunctionsRule.ts",
+        "rules/orderedImportsRule.ts",
```
JKillian: Not sure why this change is in this diff?

test/rules/max-file-line-count/default/test.ts.lint
```diff
@@ -0,0 +1,23 @@
...
+        /*
+        Here is another comment
+        It's very helpful.
+        */
+
+        public giveNumber = () =>{  
+            alert(this.num);
+        };
+    }
+}
```
JKillian: newlines preferred at the end of files

JKillian:
Looking good, thanks @chrismbarr! Can you make the small changes noted above and merge `master` into this?

JKillian:
Let's just stay with `max-file-line-count` as the name, no need to change it

src/tsconfig.json
chrismbarr: This was added when I ran the `grunt` task.  Should I just not check in this file and ignore it manually?

JKillian: All good, if Grunt did it it's probably right

JKillian:
Thanks @chrismbarr, merged!

Issue #550, opened 2015-08-05
Unused variable failure string for unused import should be more specific


adidahiya:
The following code reports `unused variable`, when it _could_ be more helpful and say `unused import`:

```ts
import * as csp from "js-csp";
```

PR #1362, opened 2016-06-30
Update no-unused-variable rule to have better failure messages

chrismbarr:
Now when something is unused it fails with a specific message about what it is, instead of just calling everything a "variable".

The types that can be reported are:
* function :new:
* import :new:
* parameter :new:
* property :new:
* variable

### DIFF
--- a/src/rules/noUnusedVariableRule.ts
+++ b/src/rules/noUnusedVariableRule.ts
@@ -65,7 +65,7 @@ export class Rule extends Lint.Rules.AbstractRule {
     };
     /* tslint:enable:object-literal-sort-keys */
 
-    public static FAILURE_STRING_FACTORY = (name: string) => `Unused variable: '${name}'`;
+    public static FAILURE_STRING_FACTORY = (type: string, name: string) => `Unused ${type}: '${name}'`;
 
     public apply(sourceFile: ts.SourceFile): Lint.RuleFailure[] {
         const languageService = Lint.createLanguageService(sourceFile.fileName, sourceFile.getFullText());
@@ -128,7 +128,7 @@ class NoUnusedVariablesWalker extends Lint.RuleWalker {
             const nameText = this.reactImport.name.getText();
             if (!this.isIgnored(nameText)) {
                 const start = this.reactImport.name.getStart();
-                this.addFailure(this.createFailure(start, nameText.length, Rule.FAILURE_STRING_FACTORY(nameText)));
+                this.addFailure(this.createFailure(start, nameText.length, Rule.FAILURE_STRING_FACTORY_VAR(nameText)));
             }
         }
     }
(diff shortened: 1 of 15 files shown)


src/rules/noUnusedVariableRule.ts
```diff
@@ -220,7 +228,7 @@ class NoUnusedVariablesWalker extends Lint.RuleWalker {
             const variableName = (<ts.Identifier> node.name).text;
 
             if (Lint.hasModifier(modifiers, ts.SyntaxKind.PrivateKeyword)) {
-                this.validateReferencesForVariable(variableName, node.name.getStart());
+                this.validateReferencesForVariable("method", variableName, node.name.getStart());
```
adidahiya: Would be nice to also make "method" a static constant for consistency

src/rules/noUnusedVariableRule.ts
```diff
@@ -65,7 +65,14 @@ export class Rule extends Lint.Rules.AbstractRule {
     };
     /* tslint:enable:object-literal-sort-keys */
 
-    public static FAILURE_STRING_FACTORY = (name: string) => `Unused variable: '${name}'`;
+    public static FAILURE_TYPE_BIND_ELEM = "binding element";
```
adidahiya: Not so sure about this one -- it might be confusing for beginners, so I'm inclined to continue referencing these values as simply "variables".

src/rules/noUnusedVariableRule.ts
```diff
@@ -65,7 +65,14 @@ export class Rule extends Lint.Rules.AbstractRule {
...
+    public static FAILURE_TYPE_FUNC = "function";
+    public static FAILURE_TYPE_IMPORT = "import";
+    public static FAILURE_TYPE_PARAM = "parameter";
+    public static FAILURE_TYPE_PROP = "property";
+    public static FAILURE_TYPE_VAR = "variable";
```
adidahiya: I think these can all be private

adidahiya:
Nice PR! Left a few comments but generally this looks good.

src/rules/noUnusedVariableRule.ts
chrismbarr: Oops!  I missed this one!  These all started out as strings before they got moved to static strings on the class

src/rules/noUnusedVariableRule.ts
chrismbarr: I can get behind that, I'm not entirely familiar with the TS syntax this is referring to anyway

src/rules/noUnusedVariableRule.ts
chrismbarr: I don't think they can since they are in the `Rule` class, and not the `NoUnusedVariablesWalker` class.  I could move them to the walker class if desired, but I figured that all the failure message related text should stay together.

adidahiya:
sweet, thanks @chrismbarr. this PR fixes #550.

Issue #1323, opened 2016-06-20
Add optional type information to rules


ScottSWu:
We would like to add type information to nodes and give rules access to the type checker, thus resolving #680. To maintain compatibility with current rules, a tentative design is to accept a ts.Program object in the Linter, and pass that to rules that implement an "applyWithProgram" function. A PR is in the works for this feature.

From the cli, constructing the program would require tsconfig for compiler options. Would it be possible to add an argument in the TSLint cli for the config location and/or whether or not to enable type checking?

adidahiya:
Sure, adding a CLI option to point to a `tsconfig.json` file to pick up compiler options makes sense. See https://github.com/palantir/tslint/issues/858.

A lot of people are very interested in #680. Would you mind writing up a slightly longer proposal of how you plan to approach this? How will you construct a `ts.Program` to pass to the `Linter`?

Re: backcompat -- a new Rule method makes sense, but its API probably needs to be more clear than simply "applyWithProgram" (because a _walker_ is also involved). One alternative might be to make a new base class that extends `Lint.RuleWalker` and takes a `ts.Program` in its constructor -- something like `Lint.ProgramAwareRuleWalker`.

ScottSWu:
Sure, creating the `ts.Program` object would look something like:
- Obtain (or find) path to `tsconfig.json`
- Pass the file contents and directory path to `ts.parseJsonConfigFileContent`
- Call `ts.createProgram` with the parsed `fileNames` and `options`, and a standard compiler host (for `test.ts`, the compiler host needs to be modified a bit because TypeScript will not compile `.lint` files)
- Run the type checker by calling `ts.getPreEmitDiagnostics`

Some other changes tried in [this fork](https://github.com/ScottSWu/tslint/commit/bcba112f161f19e3988320ca4322591dd8c3f998):
- Extend `AbstractRule` to `TypedRule`, which has the `applyWithProgram` method
- Added the `ts.Program` and `ts.TypeChecker` objects to RuleWalker (now that you mention it, extending RuleWalker would be better)
- Walkers access the type checker through `getTypeChecker()`
- `Linter` receives program as an optional argument
- `Linter.lint` calls `applyWithProgram` for `TypedRule`s if a program is available, and `apply` otherwise

JKillian:
cc @myitcv and @alexeagle who have been involved in this discussion previously

alexeagle:
thanks @JKillian - actually @ScottSWu and I are working together on this proposal.
[...]

PR #1363, opened 2016-06-30
Add optional type information to rules (#1323)

ScottSWu:
* Changed `Linter` to accept an optional program object in the constructor
  * Added helper functions for creating a program using a tsconfig and getting relevant files
* Created `TypedRule` class which receives the program object
  * Rules extending this class must implement `applyWithProgram`
  * `Linter` passes program to `TypedRule`s if available
  * Calling `apply` on a `TypedRule` throws an error
* Added `requiresTypeInfo` boolean to metadata
* Created `ProgramAwareRuleWalker` which walks with the program / typechecker
* Added cli options `--project` and `--type-check`
  * `--project` takes a `tsconfig.json` file
  * `--type-check` enables type checking and `TypedRule`s, requires `--project`
* Added an example `restrictPlusOperands` rule and tests that uses types


src/language/rule/typedRule.ts
```diff
@@ -0,0 +1,28 @@
...
+ * limitations under the License.
+ */
+
+import * as ts from "typescript";
+import {RuleFailure} from "./rule";
+import {AbstractRule} from "./abstractRule";
+
+export abstract class TypedRule extends AbstractRule {
+    public apply(sourceFile: ts.SourceFile): RuleFailure[] {
+        return [];
```
alexeagle: Add a comment about the behavior here, like "Silently ignore this rule if tslint is executed without a program, as the rule requires type-check information"

src/rules/restrictPlusOperandsRule.ts
```diff
@@ -0,0 +1,68 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import * as ts from "typescript";
+import * as Lint from "../lint";
+
+export class Rule extends Lint.Rules.TypedRule {
+    /* tslint:disable:object-literal-sort-keys */
+    public static metadata: Lint.IRuleMetadata = {
```
alexeagle: should have a way to note in the metadata that the rule requires type info?
This would make it possible to check once at the beginning of execution for "you selected some rules x,y,z which require type information, but tslint was called without access to type info, see http://wiki/call-with-program"

src/language/walker/programAwareRuleWalker.ts
```diff
@@ -0,0 +1,40 @@
...
+    protected typeChecker: ts.TypeChecker;
+
+    constructor(sourceFile: ts.SourceFile, options: IOptions, program: ts.Program) {
+        super(sourceFile, options);
+
+        this.program = program;
+        this.typeChecker = program.getTypeChecker();
+    }
+
+    public getProgram(): ts.Program {
```
alexeagle: why have a protected field and also a public getter? seems like the getters are unneeded

src/rules/restrictPlusOperandsRule.ts
```diff
@@ -0,0 +1,68 @@
...
+}
+
+class RestrictPlusOperandsWalker extends Lint.ProgramAwareRuleWalker {
+    constructor(sourceFile: ts.SourceFile, options: Lint.IOptions, program: ts.Program) {
+        super(sourceFile, options, program);
+    }
+
+    public visitBinaryExpression(node: ts.BinaryExpression) {
+        if (node.operatorToken.kind === ts.SyntaxKind.PlusToken) {
+            const tc = this.getTypeChecker();
```
alexeagle: just this.typeChecker?

src/test.ts
```diff
@@ -46,17 +48,41 @@ export function runTest(testDirectory: string, rulesDirectory?: string | string[
...
+                    return ts.createSourceFile(fileBasename, fileTextWithoutMarkup, compilerOptions.target, true);
+                }
+            },
+            readFile: () => null,
+            useCaseSensitiveFileNames: () => true,
+            writeFile: () => null,
+        };
+
+        const program = ts.createProgram([fileCompileName], compilerOptions, compilerHost);
+        ts.getPreEmitDiagnostics(program);
```
alexeagle: comment that this call actually causes the type-checking to occur, and updates the `ts.SourceFile` objects to have symbol table refs

src/tslint.ts
```diff
@@ -42,17 +44,46 @@ class Linter {
 
     private fileName: string;
     private source: string;
+    private program: ts.Program;
     private options: ILinterOptions;
 
-    constructor(fileName: string, source: string, options: ILinterOptionsRaw) {
+    public static getProgram(configFile: string, projectDirectory?: string): ts.Program {
```
alexeagle: maybe `createProgram`? it should be clear that this is an expensive operation, not just a getter of a field

src/tslint.ts
```diff
@@ -42,17 +44,46 @@ class Linter {
...
+                projectDirectory = ".";
+            } else {
+                projectDirectory = configFile.substring(0, lastSeparator + 1);
+            }
+        }
+
+        const {config} = ts.readConfigFile(configFile, ts.sys.readFile);
+        const parsed = ts.parseJsonConfigFileContent(config, {readDirectory: ts.sys.readDirectory}, projectDirectory);
+        const host = ts.createCompilerHost(parsed.options, true);
+        const program = ts.createProgram(parsed.fileNames, parsed.options, host);
```
alexeagle: we need the program to be type-checked before it is handed to `applyWithProgram`. We could do the type-checking here, but then we have to report type-check errors which is awkward.
Alternatively we can leave the API user to do it, but we should verify that it was done, see https://github.com/angular/tsickle/commit/5d9281420b02121de788a41ecb8524d981744eb5 for an example

src/language/walker/programAwareRuleWalker.ts
ScottSWu: Oops, left over from from another branch. Changing back to private fields and leaving the getters.

src/rules/restrictPlusOperandsRule.ts
ScottSWu: Possibly adding `requiresTypeInfo?: boolean` to `IRuleMetadata`

src/language/walker/programAwareRuleWalker.ts
```diff
@@ -20,8 +20,8 @@ import {IOptions} from "../../lint";
 import {RuleWalker} from "./ruleWalker";
 
 export class ProgramAwareRuleWalker extends RuleWalker {
-    protected program: ts.Program;
-    protected typeChecker: ts.TypeChecker;
+    private program: ts.Program;
```
alexeagle: @JKillian is there a reason your style preference is for field declaration rather than a `private

src/language/rule/typedRule.ts
```diff
@@ -0,0 +1,31 @@
...
+ */
+
+import * as ts from "typescript";
+
+import {AbstractRule} from "./abstractRule";
+import {RuleFailure} from "./rule";
+
+export abstract class TypedRule extends AbstractRule {
+    public apply(sourceFile: ts.SourceFile): RuleFailure[] {
+        // if no program is given to the linter, this rule will be silently
```
mprobst: Is it a bug to call this method on a `TypedRule`? Should you maybe fail loudly instead of silently, e.g. by returning a `RuleFailure`, or even throwing?

src/language/walker/programAwareRuleWalker.ts
```diff
@@ -0,0 +1,41 @@
+/**
+ * @license
+ * Copyright 2013 Palantir Technologies, Inc.
```
mprobst: Fix copyright date, here and everywhere?

src/language/walker/programAwareRuleWalker.ts
```diff
@@ -0,0 +1,41 @@
...
+import {RuleWalker} from "./ruleWalker";
+
+export class ProgramAwareRuleWalker extends RuleWalker {
+    private program: ts.Program;
+    private typeChecker: ts.TypeChecker;
+
+    constructor(sourceFile: ts.SourceFile, options: IOptions, program: ts.Program) {
+        super(sourceFile, options);
+
+        this.program = program;
```
mprobst: use `private program` in the parameter declaration?

src/language/walker/programAwareRuleWalker.ts
mprobst: Maybe make the getters `protected`?

src/rules/restrictPlusOperandsRule.ts
```diff
@@ -0,0 +1,70 @@
...
+            const rightType = tc.typeToString(tc.getTypeAtLocation(node.right));
+
+            const width = node.getWidth();
+            const position = node.getStart();
+
+            if (leftType !== rightType) {
+                // mismatched types
+                const message = Rule.MISMATCHED_TYPES_FAILURE;
+                this.addFailure(this.createFailure(position, width, message));
+            } else if (leftType !== "number" && leftType !== "string") {
```
mprobst: What if I have `type MyNumber = number;` somewhere?

src/test.ts
```diff
@@ -46,17 +48,42 @@ export function runTest(testDirectory: string, rulesDirectory?: string | string[
...
+                    return ts.createSourceFile(fileBasename, fileTextWithoutMarkup, compilerOptions.target, true);
+                }
+            },
+            readFile: () => null,
+            useCaseSensitiveFileNames: () => true,
+            writeFile: () => null,
+        };
+
+        const program = ts.createProgram([fileCompileName], compilerOptions, compilerHost);
+        // perform type checking on the program, updating nodes with symbol table references
```
mprobst: Are you sure the result of `getPreEmitDiagnostics` is cached? I thought not.

src/tslint-cli.ts
```diff
@@ -81,6 +82,12 @@ let processed = optimist
         "test": {
             describe: "test that tslint produces the correct output for the specified directory",
         },
+        "project": {
+            describe: "tsconfig file",
```
mprobst: nit: tsconfig.json ?

src/tslint-cli.ts
```diff
@@ -243,8 +258,42 @@ const processFile = (file: string) => {
...
+if (argv.project != null) {
+    if (!fs.existsSync(argv.project)) {
+        console.error("Invalid option for project: " + argv.project);
+        process.exit(1);
+    }
+    program = Linter.createProgram(argv.project, path.dirname(argv.project));
+    if (files.length === 0) {
+        files = Linter.getFileNames(program);
+    }
+    if (argv["type-check"]) {
```
mprobst: I think it'd be nicer if we could detect whether we have any type based linter rules configured, and then throw an error here if type checking is not enabled. Otherwise we risk that users expect their type based linter rules run, but they are actually silently ignored.

src/language/walker/programAwareRuleWalker.ts
adidahiya: historically we've tended towards that code style so it's easy to locate all the fields in one place. we've loosened up the guidelines a bit; I don't really mind using constructor param properties.

src/rules/restrictPlusOperandsRule.ts
ScottSWu: At first glance, it seems `getTypeAtLocation` still returns "number". I'll add these tests for this rule.

src/language/rule/typedRule.ts
adidahiya: makes sense, we could throw an error here 👍

src/rules/restrictPlusOperandsRule.ts
```diff
@@ -0,0 +1,70 @@
...
+    public static UNSUPPORTED_TYPE_FAILURE = "cannot add types ";
+
+    public applyWithProgram(sourceFile: ts.SourceFile, program: ts.Program): Lint.RuleFailure[] {
+        return this.applyWithWalker(new RestrictPlusOperandsWalker(sourceFile, this.getOptions(), program));
+    }
+}
+
+class RestrictPlusOperandsWalker extends Lint.ProgramAwareRuleWalker {
+    constructor(sourceFile: ts.SourceFile, options: Lint.IOptions, program: ts.Program) {
+        super(sourceFile, options, program);
```
adidahiya: looks like this constructor can be omitted since it just calls `super`

test/files/tsconfig-test/badTypes.test.ts
```diff
@@ -0,0 +1 @@
+let x: string = 1 + "1";
```
adidahiya: nit: EOF newline here, and anywhere else indicated by the Github diff

src/tslint.ts
```diff
@@ -42,17 +45,50 @@ class Linter {
 
     private fileName: string;
     private source: string;
+    private program: ts.Program;
     private options: ILinterOptions;
 
-    constructor(fileName: string, source: string, options: ILinterOptionsRaw) {
+    public static createProgram(configFile: string, projectDirectory?: string): ts.Program {
```
adidahiya: please include JSDoc comments for public API methods like this

src/tslint.ts
```diff
@@ -42,17 +45,50 @@ class Linter {
...
+
+        const {config} = ts.readConfigFile(configFile, ts.sys.readFile);
+        const parsed = ts.parseJsonConfigFileContent(config, {readDirectory: ts.sys.readDirectory}, projectDirectory);
+        const host = ts.createCompilerHost(parsed.options, true);
+        const program = ts.createProgram(parsed.fileNames, parsed.options, host);
+
+        return program;
+    }
+
+    public static getFileNames(program: ts.Program): string[] {
```
adidahiya: same as above, please include a brief JSDoc description of this method

src/test.ts
ScottSWu: Not sure, but in this case, the tests just have to be parsed and type checked as much as possible, ignoring compiler errors.

src/tslint-cli.ts
```diff
@@ -243,8 +258,42 @@ const processFile = (file: string) => {
     }
 };
 
-const files = argv._;
+// if both files and tsconfig are present, use files
+let files = argv._;
+let program: ts.Program = undefined;
```
adidahiya: nit: don't need to initialize to `undefined`; variables start out that way

src/tslint-cli.ts
```diff
@@ -243,8 +258,42 @@ const processFile = (file: string) => {
...
+    }
+    if (argv["type-check"]) {
+        // if type checking, run the type checker
+        const diagnostics = ts.getPreEmitDiagnostics(program);
+        if (diagnostics.length > 0) {
+            const messages = diagnostics.map((diag) => {
+                // emit any error messages
+                let message = ts.DiagnosticCategory[diag.category];
+                if (diag.file) {
+                    message += " at " + diag.file.fileName + ":";
```
adidahiya: please use a template string here and on L284

src/tslint-cli.ts
adidahiya: Agreed, that would be nice, but I think we can do without it for the first version of this feature and file a follow-up issue to make it a later enhancement.

src/rules/restrictPlusOperandsRule.ts
```diff
@@ -0,0 +1,70 @@
...
+        description: "When adding two variables, operands must both be of type number or of type string.",
+        optionsDescription: "Not configurable.",
+        options: null,
+        optionExamples: ["true"],
+        type: "functionality",
+        requiresTypeInfo: true,
+    };
+    /* tslint:enable:object-literal-sort-keys */
+
+    public static MISMATCHED_TYPES_FAILURE = "types must match";
```
adidahiya: Let's make this message more specific: `"Types of values used in '+' operation must match"`

src/rules/restrictPlusOperandsRule.ts
```diff
@@ -0,0 +1,70 @@
...
+    public static UNSUPPORTED_TYPE_FAILURE = "cannot add types ";
```
adidahiya: Minor: make this a `UNSUPPORTED_TYPE_FAILURE_FACTORY = (leftType: string) => ...` for consistency with other rule failure messages.

src/language/walker/programAwareRuleWalker.ts
JKillian: Let's go ahead and use parameter properties where possible 👍 
In fact, just merged https://github.com/palantir/tslint/pull/1373 which does this in a few locations

ScottSWu:
Hi, sorry for the large PR. Could I get another review of the changes, and should I begin squashing commits?

adidahiya:
Sweet, this looks pretty good @ScottSWu! I think I'll just squash + merge at the end into one commit and take your PR description as the commit message body.

One last thing -- can you add documentation to the README about how to use these new APIs? This includes the new CLI options and static methods on the `Linter` class.

README.md
```diff
@@ -115,6 +115,8 @@ Options:
 -e, --exclude         exclude globs from path expansion
 -t, --format          output format (prose, json, verbose, pmd, msbuild, checkstyle)  [default: "prose"]
 --test                test that tslint produces the correct output for the specified directory
+--project             tsconfig.json file
```
alexeagle: path to tsconfig.json file

README.md
```diff
@@ -115,6 +115,8 @@ Options:
...
+--type-check          enable type checking when linting a project
```
alexeagle: note that it requires the `--project` flag be set?

README.md
```diff
@@ -214,6 +224,21 @@ const linter = new Linter(fileName, fileContents, options);
...
+#### Type Checking
+
+To enable rules that work with the type checker, a TypeScript program object may be passed to the linter. Helper functions are provided to create a program from a `tsconfig.json` file. A project directory can be specified if project files do not lie in the same directory as the `tsconfig.json` file.
+
+```javascript
+const program = Linter.createProgram("tsconfig.json", "projectDir/");
+const files = Linter.getFileNames(program);
+const results = files.map(file => {
+    const fileContents = program.getSourceFile(file).getFullText();
+    const linter = new Linter(file, fileContents, options, program());
```
alexeagle: parens after program look misplaced

README.md
```diff
@@ -313,6 +338,7 @@ Core rules are included in the `tslint` package.
     * `"jsx-double"` enforces double quotes for JSX attributes.
     * `"avoid-escape"` allows you to use the "other" quotemark in cases where escaping would normally be required. For example, `[true, "double", "avoid-escape"]` would not report a failure on the string literal `'Hello "World"'`.
 * `radix` enforces the radix parameter of `parseInt`.
+* `restrict-plus-operands` enforces the type of addition operands to be both `string` or both `number`
```
alexeagle: should we mention here which ones require type-checking?

adidahiya: yep, you can throw that on the end as "(requires type checking)." Also include that trailing period for consistency.

README.md
```diff
@@ -214,6 +224,21 @@ const linter = new Linter(fileName, fileContents, options);
 const result = linter.lint();
 ```
 
+#### Type Checking
+
+To enable rules that work with the type checker, a TypeScript program object may be passed to the linter. Helper functions are provided to create a program from a `tsconfig.json` file. A project directory can be specified if project files do not lie in the same directory as the `tsconfig.json` file.
```
adidahiya: This should probably be edited to "... a TypesScript program [must] be passed to the linter [when using the programmatic API]." (my additions in brackets)

Also it would be good to clarify that a program is automatically created for you when using the appropriate CLI flags.

README.md
ScottSWu: Fixed

README.md
ScottSWu: Fixed, and added a line below the code sample regarding the CLI.

adidahiya:
🎉

HamletDRC:
Will the grunt-tslint plugin need to be updated before grunt users can use this?

myitcv:
@ScottSWu @alexeagle et al - this is great work, thank you! Look forward to trying it out in the `@next` release

Fank:
@adidahiya Possible to release a new version, which contains these change, soon?

adidahiya:
@Fank @ScottSWu I'm attempting to cut a release here (https://github.com/palantir/tslint/pull/1395) but running into some compilation issues. I don't have time to dig into these right now but I'll try to get a release out for you guys ASAP (hopefully tonight or tomorrow)!

adidahiya:
Released v3.14.0-dev.0

glen-84:
Should it not silently fail if you don't enable type checking? This would allow you to see all the other errors/warnings in an editor, even when you haven't enabled type checking (which won't always be possible or easy to do).

Issue #1260, opened 2016-05-18
Support eslint quote-props rule


danvk:
I'd like a rule equivalent to eslint's [quote-props][1], so that I can enforce that object keys are quoted only when necessary.

i.e. I'd like a setting to make this illegal:

```ts
const o = { 'name': 123 };
```

in favor of:

```ts
const o = { name: 123 };
```

[1]: http://eslint.org/docs/rules/quote-props

adidahiya:
sounds like a good rule to me. open to PRs.

PR #1364, opened 2016-06-30
Core object-literal-key-quotes rule

danvk:
This adds a tslint equivalent to a subset of the eslint `quote-props` rule (see #1260).

The eslint rule is fairly elaborate. I've implemented the "as-needed" and "always" behaviors: http://eslint.org/docs/rules/quote-props

I verified that my tests produce the same results as eslint for the same code and settings.

There may be some incorrect behavior around things like unicode property names, but this is a good start.


danvk:
@JKillian @adidahiya This is ready for review.

JKillian:
Sorry for the delay @danvk, was away from the office. I'll try to give this a review by the end of the week at the latest

src/rules/quotePropsRule.ts
```diff
@@ -0,0 +1,103 @@
+import * as Lint from "../lint";
+import * as ts from "typescript";
+
+export class Rule extends Lint.Rules.AbstractRule {
+    /* tslint:disable:object-literal-sort-keys */
+    public static metadata: Lint.IRuleMetadata = {
+        ruleName: "quote-props",
+        description: "Quoting Style for Property Names",
```
JKillian: "Enforces consistent object literal property quote style."

I think the above would be more consistent with [the current descriptions](http://palantir.github.io/tslint/rules/).

src/rules/quotePropsRule.ts
```diff
@@ -0,0 +1,103 @@
...
+            var object2 = {
+                "property": true
+            };
+
+            In many cases, it doesn’t matter if you choose to use an identifier instead of a string
+            or vice-versa. Even so, you might decide to enforce a consistent style in your code.
+
+            This rules lets you enforce consistent quoting of property names. Either they should always
+            be quoted (default behavior) or quoted only as needed ("as-needed").
+        `,
```
JKillian: I'm not sure it matters, but I normally put this closing "`," up one line in order to prevent the extra newline.

src/rules/quotePropsRule.ts
```diff
@@ -0,0 +1,103 @@
...
+        optionsDescription: Lint.Utils.dedent`
+            Possible settings are:
+
+            * \`"always"\`: Property names should always be quoted. (This is the default.)
+            * \`"as-needed"\`: Only property names which require quotes may be quoted (e.g. those with spaces in them).
+
+            For ES6, computed property names (\`{[name]: value}\`) and methods (\`{foo() {}}\`) never need
+            to be quoted.
+        `,
+        options: {
```
JKillian: If the rule only takes one argument like this, I've been documenting it without an array as such: https://github.com/palantir/tslint/blob/7fbd31b01ecc6c9c080d30d81bb06c084b5354ac/src/rules/interfaceNameRule.ts#L36

I haven't implemented validation of rule options yet, but I found that the above is easier for reading in the documentation

src/rules/quotePropsRule.ts
```diff
@@ -0,0 +1,103 @@
...
+                {
+                    enum: ["always", "as-needed"],
+                    // TODO: eslint also supports "consistent", "consistent-as-needed" modes.
+                    // TODO: eslint supports "keywords", "unnecessary" and "numbers" options.
+                },
+            ],
+            minItems: 0,
+            maxItems: 1,
+        },
+        optionExamples: ["true", "[true, \"as-needed\"]", "[true, \"always\"]"],
```
JKillian: Once we start validating rule options, we'll probably require specifying the mode for the rule. I'd rather that in the documentation we don't mention a default value

test/rules/quote-props/always/tslint.json
```diff
@@ -0,0 +1,5 @@
+{
+  "rules": {
+    "quote-props": true
```
JKillian: Can we make this explicit, i.e.:

```
"quote-props": [true, "always"]
```

JKillian:
Looking good! Just found a few documentation tweaks to make if you don't mind

danvk:
Addressed all comments, thanks for the review!

src/rules/quotePropsRule.ts
```diff
@@ -0,0 +1,95 @@
...
+    public static UNNEEDED_QUOTES = (name: string) => `Unnecessarily quoted property '${name}' found.`;
+    public static UNQUOTED_PROPERTY = (name: string) => `Unquoted property '${name}' found.`;
+
+    public apply(sourceFile: ts.SourceFile): Lint.RuleFailure[] {
+        const quotePropsWalker = new QuotePropsWalker(sourceFile, this.getOptions());
+        return this.applyWithWalker(quotePropsWalker);
+    }
+}
+
+// This is simplistic. See https://mothereff.in/js-properties for the gorey details.
```
adidahiya: lol this url domain

danvk: the whole URL works pretty well for this topic! :)

adidahiya:
@danvk small note for future PRs: it makes it easier for us to review if you address CR comments in a follow-up commit (Github's squash+merge feature makes it easy to squash revision commits at the end). Thanks again for the contribution!

danvk:
👍 @adidahiya 

(I often use tools like Reviewable, which have a notion of revisions/snapshots independent from the current commits.)

danvk:
(split into two commits so you guys can see the diff more easily.)

JKillian:
Thanks! Last change, could we rename the rule to `object-literal-key-quotes`? This will keep it parallel to an existing rule `object-literal-sort-keys`. Sorry for not mentioning this last time

danvk:
@JKillian do you guys care about maintaining parity with eslint? That's where the `quote-props` name comes from: http://eslint.org/docs/rules/quote-props

JKillian:
@danvk That's definitely something that's taken into consideration. However, I feel like ESLint rule names can be a little bit hit or miss, so I'm leaning away from worrying about parity too much. I'd like to get better naming guidelines formalized (#1371), so if you have overall input, would be glad to hear your thoughts.

danvk:
`object-literal-key-quotes` it is!

danvk:
@JKillian this should be good to go

JKillian:
Thanks @danvk, merged!

PR #1370, opened 2016-07-01
Order imports so build passes

JKillian:


adidahiya:
👍 glad to see the new rule works well

PR #1373, opened 2016-07-02
Use parameter properties to simplify constructors across codebase

### LINKED ISSUE #619: Use parameter properties to simplify constructors across codebase
Instead of 
```typescript
class MyClass {
    private field1: boolean;
    constructor(field1: boolean) {
        this.field1 = field1;
    }
}
```
use instead:
```typescript
class MyClass {
    constructor(private field1: boolean) { }
}
```

bencoveney:
Small refactoring of constructors across code base to use parameter properties.

There was also a couple of places where constructors on derived classes were just passing arguments through to a base class with no additional logic, so i have removed those where I could see them.

Fixes #619


JKillian:
Thanks @bencoveney!

PR #1385, opened 2016-07-08
Better ban rule failure messages & allows for additional custom messages

chrismbarr:
This PR does 2 things.

---

### :one: Better Wording
The wording of the failure message for the `ban` rule (and the `no-console` rule since it is based on the `ban` rule) has been updated.  

:-1: **Before**
> function invocation disallowed: object.someFunction

_(The word "invocation" always reminded me of a wizard casting a spell... :crystal_ball: )_

:+1: **After**
> Calls to 'object.someFunction' are not allowed.

---
### :two: Optional Explanation Messages
Most TSLint rules are basically either turned on or off and they offer explanation text or it's generally understood **why** the rule is enabled in a project.  One thing that's always bothered my and my teammates is when a function is banned it basically just says "this is banned" and it doesn't tell you why. Since I'm the one who turned on the linting rules I eventually get someone asking me "why can't I use `_.filter`?  I can use in in this other project just fine..."

Well, now an optional 3rd parameter for each banned function call can be added.  This text could offer an explanation as to why it's banned and/or offer an alternative that could be used instead of the banned call.  Here's an example of how it would be configured

```json
"ban": [
      true, 
      ["myApp", "someFunc", "'myApp.someFunc' has been deprecated, use 'myApp.otherFunc' instead"],
      ["_", "filter", "Use the native JavaScript 'myArray.filter' instead."],
      ["_", "forEach"] //no additional custom message here
]
```

And here's what the failure message would look like with the additional text appended :clap: 
> Calls to '_.filter' are not allowed. Use the native JavaScript 'myArray.filter' instead.


src/rules/banRule.ts
```diff
@@ -25,23 +25,26 @@ export class Rule extends Lint.Rules.AbstractRule {
...
                 type: "array",
                 arrayMembers: [
                     { type: "string" },
                     { type: "string" },
+                    { type: "string" },
                 ],
             },
         },
-        optionExamples: [`[true, ["console", "log"], ["someObject", "someFunction"]]`],
+        optionExamples: [`[true, ["someObject", "someFunction"], ["someObject", "someFunction", "Optional explanation"]]`],
```
JKillian: We don't want both banned functions to be `someObject` and `someFunction`, can we change those terms for one of them?

README.md
```diff
@@ -224,7 +224,7 @@ Core rules are included in the `tslint` package.
   * `"parameters"` checks alignment of function parameters.
   * `"arguments"` checks alignment of function call arguments.
   * `"statements"` checks alignment of statements.
-* `ban` bans the use of specific functions. Options are ["object", "function"] pairs that ban the use of object.function().
+* `ban` bans the use of specific functions. Options are `["object", "function"]` pairs that ban the use of `object.function()`.  An optional 3rd parameter may be passed in (`["object", "function", "Use 'object.otherFunc' instead."]`) to offer an explanation as to why the function has been banned or to offer an alternative.
```
JKillian: super small change, I prefer "provided" instead of "passed in"

JKillian:
Looking good @chrismbarr! Two tiny changes above to make if you can, and can you merge `master` into this branch so it's mergeable?

JKillian:
Looking good, can you merge master here? I can't merge this PR as there are merge conflicts right now

PR #1388, opened 2016-07-12
Revert changes made to findup-sync during refactoring

weswigham:
Now that TS propertly supports module targets.

@alexeagle needs this for angular licensing reasons.


alexeagle:
LGTM

adidahiya:
👍

PR #1389, opened 2016-07-12
Revert local modification of findup typings

alexeagle:
findup typings in the `typings/` directory don't match the SHA from `tsd.json` which makes the build hard to repro externally.
This PR in place of https://github.com/DefinitelyTyped/DefinitelyTyped/pull/10061

### DIFF
--- a/src/configuration.ts
+++ b/src/configuration.ts
@@ -15,7 +15,7 @@
  * limitations under the License.
  */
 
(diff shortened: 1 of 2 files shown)


weswigham:
Duplicate of #1388? Except you used `const` rather than `import`! A critical mistake!

PR #1394, opened 2016-07-12
no-for-in-array rule

### LINKED ISSUE #1380: Add a rule to disallow "for … in" on array types
This is almost certainly an error:

```typescript
function foo(values: number[]) {
  for (const value in values) {
    ...
  }
}
```

It should be "for … of". If the intention is to iterate over the array indices, that's better done with `values.forEach` or a traditional `for (let i = 0; i < values.length; i++)` loop. Both of these alternatives give you numeric indices, rather than the strings that "for ... in" does.

Some people advocate disallowing for-in entirely, but I still find it useful for objects.

In theory tslint should know whether the object being iterated over is really an array and it could warn specifically on this mistake.

danvk:
Fixes #1380 

I would be interested in feedback on whether this is accurate:

```ts
const isArrayType = (type.flags & ts.TypeFlags.Reference) !== 0;
```

The `typescript.d.ts` file is quite terse about what a "Generic type reference" is, but in my testing it seems to correspond precisely to array types.

### DIFF
--- a/src/rules/noForInArrayRule.ts
+++ b/src/rules/noForInArrayRule.ts
@@ -1,0 +1,74 @@
+/**
+ * @license
+ * Copyright 2016 Palantir Technologies, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import * as ts from "typescript";
+
+import * as Lint from "../lint";
+
+export class Rule extends Lint.Rules.TypedRule {
+    /* tslint:disable:object-literal-sort-keys */
+    public static metadata: Lint.IRuleMetadata = {
+        ruleName: "no-for-in-array",
+        description: "Disallows for-in loops over array types.",
+        descriptionDetails: Lint.Utils.dedent`
+            A for-in loop (\`for (var k in o)\`) iterates over the properties of an Object.
+
+            While it is legal to use for-in loops with array types, it is not common.
+            for-in will iterate over the indices of the array as strings, omitting any "holes" in
+            the array.
+
+            More common is to use for-of, which iterates over the values of an array.
+            If you want to iterate over the indices, alternatives include:
+
+            array.forEach((value, index) => { ... });
+            for (const [index, value] of array.entries()) { ... }
+            for (let i = 0; i < array.length; i++) { ... }
+            `,
+        optionsDescription: "Not configurable.",
+        options: null,
+        optionExamples: ["true"],
+        requiresTypeInfo: true,
+        type: "functionality",
+    };
+    /* tslint:enable:object-literal-sort-keys */
+
+    public static FAILURE_STRING = "for-in loops over arrays are forbidden. Use for-of or array.forEach instead.";
+
+    public applyWithProgram(sourceFile: ts.SourceFile, program: ts.Program): Lint.RuleFailure[] {
+        const noForInArrayWalker = new NoForInArrayWalker(sourceFile, this.getOptions(), program);
+        return this.applyWithWalker(noForInArrayWalker);
+    }
+}
+
+class NoForInArrayWalker extends Lint.ProgramAwareRuleWalker {
+    public visitForInStatement(node: ts.ForInStatement) {
+        const iteratee = node.expression;
+        const tc = this.getTypeChecker();
+        const type = tc.getTypeAtLocation(iteratee);
+
+        /* tslint:disable:no-bitwise */
+        const isArrayType = (type.flags & ts.TypeFlags.Reference) !== 0;
+        const isStringType = (type.flags & ts.TypeFlags.StringLike) !== 0;
+        /* tslint:enable:no-bitwise */
+
+        if (isArrayType || isStringType) {
+            this.addFailure(this.createFailure(node.getStart(), node.getWidth(), Rule.FAILURE_STRING));
+        }
+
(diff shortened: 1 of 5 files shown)


JKillian:
Just a thought, what do you think about banning `for-in` all together? My thinking is that in the vast majority of cases, users only want the object's own properties and thus:

```
for(const key of Object.keys(someObj)) {
  // do stuff
}
```

is preferred. A `for-in` loop is really only beneficial if you want inherited properties, which is unlikely.

The other advantage is that this wouldn't need the type-checker, so it could be used in more situations and would be faster.

JKillian:
On the other hand, a general ban of `for-in` wouldn't address the issue you described, as `Object.keys(someArray)` will have the same problem. That would lead me to think that a `no-array-key-iteration` (naming?) rule might be a better solution than just a `no-for-in-array` rule.

JKillian:
> I would be interested in feedback on whether this is accurate:
> 
> ```ts
> const isArrayType = (type.flags & ts.TypeFlags.Reference) !== 0;
> ```


Unfortunately I don't think so, [check out the description of `TypeFlags.Reference` here](https://github.com/Microsoft/TypeScript/blob/master/src/compiler/types.ts#L2319). I think what's happening is that the Array type is inherently generic and thus always has that flag set? I think that potentially an instance of any generic class would have this flag set, but I haven't tested that out.

Maybe @weswigham could shed some insight on the best way to check if something is an Array type?

danvk:
So is that saying that an object of a generic type (say `Promise<T>`) would also have the `TypeFlags.Reference` bit set? I tried adding a test:

```ts
function foo<T>(promise: Promise<T>) {
  for (var k in promise);
}
```

But `promise` in that for-in loop does not have the `TypeFlags.reference` flag set. An example of something which has this flag set but is not an array would be very helpful!

I typically use for-in for iterating over object literals that I've created myself, so inherited properties aren't a big concern. for-in is compact and seems to be faster than for-of-Object.keys in modern browsers and node.

I can see the argument for disallowing it entirely, though. Maybe there should be two options for a "for-in" rule: `"for-in": [true, "never"]` and `"for-in": [true, "for-objects"]`?

JKillian:
When testing the type of `b` in this code:

```ts
class A<T> { }
let b: A<number>;
```

I get this (also included are the other variable declarations from your test case):

```
Reference: b:A<number>
Reference: array = [1, 2, 3, 4]
Reference: strArray = ['a', 'b', 'c']
Reference: objArray = [{a: 1}, {a: 2, b: 10}, {a: 3}]
Not Reference: o = {}
Not Reference: numArray = {
    1: "a",
    2: "b"
}
Not Reference: d = new Date()
Not Reference: str = "abc"
Not Reference: objectMap = {} as {[key: string]: number}
Not Reference: map = new Map([['a', 1], ['b', 2]])
Not Reference: set = new Set([1, 2, 3])
```

In case it helps, here's the method I added to your rule to get this output:

```ts
    public visitVariableDeclaration(node: ts.VariableDeclaration) {
        const tc = this.getTypeChecker();
        const type = tc.getTypeAtLocation(node);
        // tslint:disable
        const isReference = (type.flags & ts.TypeFlags.Reference) > 0;
        console.log(`${isReference ? "Reference" : "Not Reference"}: ${node.getText()}`);
        // tslint:enable
    }
```

JKillian:
If we go your route for this rule, we could actually just enhance [the current `forin` rule](http://palantir.github.io/tslint/rules/forin/). (It really should be hyphenated, but it's not.) It could then have three options: `"never", "guarded", "only-objects"` where `"guarded"` and `"only-objects"` could be used together. I think this might be the best way to go. With no options provided, the rule would default to `"guarded"` since that is its current behavior.

I know I proposed it above in the thread, but after more thought, I don't think a rule that warns about `for (const key of Object.keys(someArray))` is necessary. Although it's "wrong" in the same way that `for (const key of someArray)` is, it's not a mistake that would be accidentally made.

danvk:
Sure, I could add an `only-objects` setting to the existing rule.

I pushed a new version of the array check. The logic is now:

```ts
const isArrayType = type.symbol && type.symbol.name === "Array";
```

This handles the `A<number>` case correctly.

danvk:
(If this check seems correct then I'll add it to the existing rule.)

JKillian:
> I pushed a new version of the array check. The logic is now:
> 
> ```ts
> const isArrayType = type.symbol && type.symbol.name === "Array";
> ```

Your guess is as good as mine here, but this makes sense to me and can't think of any edge cases which would mess things up.

weswigham:
@JKillian 
```ts
function foo() {
  type Array = MyArray<T>;
  let x = new Array();
  // ...
}
```
It doesn't handle block scoped type declarations whose name is also `Array`.

JKillian:
> It doesn't handle block scoped type declarations whose name is also Array.

Let local type declarations named `Array` be anathema... Ha, are there any better alternatives?

danvk:
If you redefine `Array` locally and get a lint error, then I don't feel bad for you :)

weswigham:
@JKillian You want to be able to lookup the global array type, then check if the type you find at that location is an instantiation of it. Inside the compiler, this wouldn't be too difficult, but we don't publicly expose the functions you need to do this check on the type checker instance right now. Heck, we don't even expose a way to lookup a global type!

I've known about this shortcoming in our public API since I started working on compiler extensions and have had it at the back of my mind. I've opened #9879 to track the issue.

danvk:
@JKillian I pushed a new version which merges my `no-for-in-array` rule with the existing `forin` rule.

This introduces a new problem, though (and it's the reason that the tests will fail): the rule now requires type checking, whereas it didn't before. So existing configs will break.

What do you think? Is it still a good idea to merge the rules?

JKillian:
@weswigham Thanks for the info, sounds like the current method we're using is best for now then. Excited for the Type Relationship API! (And the other ones, like the Type Builder API).

@danvk Shoot, wasn't thinking about that. We definitely don't want to break existing users. Let's just keep it as `no-for-in-array` as you suggested. (We can add an option to ban all `for-in` loops to the main non-typechecked `forin` rule down the road if we want to.) Apologies for the needless churn

danvk:
@JKillian I've rolled this back to the version with a separate no-for-in-array rule. I believe this is ready to go.

JKillian:
Thanks @danvk!

PR #1398, opened 2016-07-13
no-mergable-namespace -> no-mergeable-namespace

vslinko:


JKillian:
This is already fixed in the source via https://github.com/palantir/tslint/pull/1361, but this is a good quick fix until the docs site is regenerated completely.

PR #1403, opened 2016-07-14
Add linterOptions to tslint.json

### LINKED ISSUE #1402: Running tests is much slower now that type checking is always enabled
As a result of #1363.

adidahiya:
- only enable type checking during tests when `typeCheck: true` is configured
- fixes #1402 

I haven't documented this as a public API yet; we're only using it internally for tests. If we did want to make it public, we'd have to refactor tslint-cli.ts to read the configuration file before it tries to run the type checker on a program.

### DIFF
--- a/src/configuration.ts
+++ b/src/configuration.ts
@@ -23,6 +23,9 @@ import * as resolve from "resolve";
 import {arrayify, objectify, stripComments} from "./utils";
 
 export interface IConfigurationFile {
+    linterOptions?: {
+        typeCheck?: boolean,
+    };
     extends?: string | string[];
     rulesDirectory?: string | string[];
     rules?: any;
--- a/src/configuration.ts
+++ b/src/configuration.ts
@@ -23,10 +23,10 @@ import * as resolve from "resolve";
 import {arrayify, objectify, stripComments} from "./utils";
 
 export interface IConfigurationFile {
+    extends?: string | string[];
     linterOptions?: {
         typeCheck?: boolean,
     };
-    extends?: string | string[];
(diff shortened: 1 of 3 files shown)


src/configuration.ts
```diff
@@ -24,6 +24,9 @@ import {arrayify, objectify, stripComments} from "./utils";
 
 export interface IConfigurationFile {
     extends?: string | string[];
+    linterOptions?: {
```
JKillian: Thinking we don't need this `linterOptions` property, `typeCheck` can reside at the top level

src/configuration.ts
```diff
@@ -24,6 +24,9 @@ import {arrayify, objectify, stripComments} from "./utils";
...
+        typeCheck?: boolean,
+    };
     rulesDirectory?: string | string[];
     rules?: any;
 }
```
JKillian: `extendConfigurationFile` further down in this file should probably support this field

JKillian:
I think if would be good to add support for this field when used via the CLI as well. It would only apply is the CLI used with the `--project` flag.

Sidenote: Should you be able to specify `--project` in tslint.json as well? If TSLint was run in a directory and not supplied any files, it could look for a `tslint.json` file and then use the project option passed in it.

JKillian:
Changed my mind, the comments above weren't that important / not fully thought out yet. We'll need to think how we want to handle the `typeCheck` option publicly, but for now let's merge this to get the tests faster.

JKillian:
For now, this is will remain a private/undocumented API as you mentioned until we have the end-user API figured out

PR #1404, opened 2016-07-14
Allow gh-pages to build on CI

adidahiya:
This also allows me to build using this new Jekyll CMS I want to try out called [Forestry](https://forestry.io/).

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -1,3 +1,3 @@
-ruby '2.2'
+ruby '2.2.3'
 source 'https://rubygems.org'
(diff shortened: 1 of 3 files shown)


JKillian:
How did you like Forestry? Just noticed now that you had made a few changes with it, recommend it?

adidahiya:
It's ok, not bad for quick edits, but only marginally better than editing in Github directly. I think they're going to start charging for Forestry, so it probably won't be worth it at that point (we don't have many non-technical contributors working on our docs).

PR #1406, opened 2016-07-15
task(formatter): added stylish formatter

gjuchault:
Hi!

This is a small pull request for adding a "styilsh" formatter (pretty much equivalent to eslint' stylish formatter).

I might have missed one or two thing (tests/readme/cli/etc.), so please tell me if anything is missing!


gjuchault:
The appveyor build is not passing, strange thing that colors did not pass with this one when passing circleci tests

JKillian:
Here's what the failure from AppVeyor looks like:

image

gjuchault:
image
This is from circleci

Is there any difference between appveyor and circleci commands ?

adidahiya:
@gjuchault appveyor runs on a windows VM

gjuchault:
@adidahiya Which makes the colors package outputs nothing in colors; when in Linux/compatible terminal it will work (cmder and all that stuff).

Do you have an idea how to handle the fact that there is no colors ? I could go something like `if colors.green('foo') === 'foo') {` in the tests

JKillian:
Hmm, not sure if we have any other formatters that use colors right now. A work around like you proposed seems fine for now @gjuchault. The colors library also [has a color support detection module](https://github.com/Marak/colors.js/blob/master/lib/system/supports-colors.js) that you may be able to use

gjuchault:
@JKillian error is fixed now, but I still have a doubt on the character not showing up. Is that normal (I'm using `getLineAndCharacter` [here](https://github.com/gjuchault/tslint/blob/master/src/formatters/stylishFormatter.ts#L46))

test/formatters/stylishFormatterTests.ts
```diff
@@ -0,0 +1,62 @@
...
+    it("formats failures", () => {
+        const maxPosition = sourceFile.getFullWidth();
+
+        const failures = [
+            new RuleFailure(sourceFile, 0, 1, "first failure", "first-name"),
+            new RuleFailure(sourceFile, 2, 3, "&<>'\" should be escaped", "escape"),
+            new RuleFailure(sourceFile, maxPosition - 1, maxPosition, "last failure", "last-name"),
+            new RuleFailure(sourceFile, 0, maxPosition, "full failure", "full-name"),
+        ];
+        const expectedResult = (require("colors").supportsColor) ?
```
JKillian: Can we import colors with an ES6 `import` statement at the top of the file instead of here?

src/formatters/stylishFormatter.ts
```diff
@@ -0,0 +1,96 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import {AbstractFormatter} from "../language/formatter/abstractFormatter";
+import {RuleFailure} from "../language/rule/rule";
+
+import * as colors from "colors";
+
+export class Formatter extends AbstractFormatter {
```
JKillian: 4-space indentation in this file please!

JKillian:
> @JKillian error is fixed now, but I still have a doubt on the character not showing up. Is that normal 

I'm sorry @gjuchault, I'm not quite sure what you're asking. Are you saying `lineAndCharacter.character` is undefined sometimes?

JKillian:
Ah, I see what you're saying now. No, the character should definitely show up! I'm not sure why it isn't.

src/formatters/stylishFormatter.ts
```diff
@@ -0,0 +1,96 @@
...
+      const output = `${positionTuple}  ${ruleName}  ${failureString}`;
+
+      outputLines.push(output);
+    }
+
+    return outputLines.join("\n") + "\n\n";
+  }
+
+  private pad(str: string, len: number): string {
+    const padder = Array(len).join(" ");
```
JKillian: This will make a string that is one character _shorter_ than `len`. I believe you may have meant `new Array(len + 1).join(" ");`  This should fix the problem where the character number doesn't show up!

test/formatters/stylishFormatterTests.ts
gjuchault: Well that's gonna be a problem, the [definition](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/colors/colors.d.ts) does not contain the `supportsColor` boolean

src/formatters/stylishFormatter.ts
gjuchault: Done!

gjuchault:
@JKillian Finally ! :)

JKillian:
Thanks @gjuchault!

PR #1422, opened 2016-07-21
Adding new rule to disallow optional parameters in callback positions

rakatyal:


src/rules/noOptionalParameterInCallbackRule.ts
```diff
@@ -0,0 +1,107 @@
...
+ * limitations under the License.
+ */
+
+import * as ts from "typescript";
+
+import * as Lint from "../lint";
+
+export class Rule extends Lint.Rules.TypedRule {
+
+    /* tslint:disable:object-literal-sort-keys */
```
weswigham: Why disable this rule around the metadata description? Why not just sort the keys?

rakatyal: It was a common pattern in the other rules. I thought I will stick to it.

src/rules/noOptionalParameterInCallbackRule.ts
```diff
@@ -0,0 +1,107 @@
...
+    private validateSignature(node: ts.SignatureDeclaration) {
+        const tc = this.getTypeChecker();
+        const signatureParametersType: ts.Type[] = node.parameters.filter(p => !!p.type)
+            .map(function (parameter: ts.ParameterDeclaration) { return tc.getTypeAtLocation(parameter); });
+        for (const type of signatureParametersType) {
+            if (this.isFunctionType(type)) {
+                const signatures = tc.getSignaturesOfType(type, ts.SignatureKind.Call);
+                for (const signature of signatures) {
+                    for (const param of signature.parameters) {
+                        const declarations = param.getDeclarations().map(function (declaration: ts.Declaration) {
```
weswigham: All this map does is typecast to `ts.ParameterDeclaration[]`, if I'm reading it correctly. If so, remove the map and just cast the array you get back from `param.getDeclarations()`.

src/rules/noOptionalParameterInCallbackRule.ts
```diff
@@ -0,0 +1,107 @@
...
+        description: "Disallows optional parameters in callback positions.",
+        optionsDescription: "Not configurable.",
+        options: null,
+        optionExamples: ["true"],
+        type: "functionality",
+        requiresTypeInfo: true,
+    };
+    /* tslint:enable:object-literal-sort-keys */
+
+    public static FAILURE_STRING = `No optional parameters in callback positions`;
```
weswigham:
```
public static FAILURE_STRING = `No optional parameters in callback positions - consider writing \`T | undefined\` or something similar instead.`
```

I like my lint failures to suggest an easy or common way to fix the failure. This is mostly just my preference, but I find it useful.

src/rules/noOptionalParameterInCallbackRule.ts
```diff
@@ -0,0 +1,107 @@
...
+    private isFunctionType(type: ts.Type): boolean {
+        const tc = this.getTypeChecker();
+        return type.flags && ts.TypeFlags.ObjectType && tc.getSignaturesOfType(type, ts.SignatureKind.Call).length > 0;
+    }
+
+    private validateSignature(node: ts.SignatureDeclaration) {
+        const tc = this.getTypeChecker();
+        const signatureParametersType: ts.Type[] = node.parameters.filter(p => !!p.type)
+            .map(function (parameter: ts.ParameterDeclaration) { return tc.getTypeAtLocation(parameter); });
+        for (const type of signatureParametersType) {
```
weswigham: In addition to parameters which are directly function types, you should probably recur into any object type parameters and check all their members for callbacks, too.

src/rules/noOptionalParameterInCallbackRule.ts
JKillian: I did this in all the rules when I added their initial metadata because I found the order of metadata with sorted keys was harder to read / quickly glance at. Feel free to debate whether it was a good decision haha

test/rules/no-optional-parameter-in-callback/test.ts.lint
```diff
@@ -0,0 +1,86 @@
...
+
+function f3() {
+    function f4(x?: number, y?: (k?: string) => void): number {
+                                 ~~~~~~~~~~  [error]
+        return 1;
+    }
+}
+
+var v1 = {
+    foo(x: (y?: number): void {
```
weswigham: This line has a syntax error on it. It should be `    foo(x: (y?: number) => void): void {`

src/rules/noOptionalParameterInCallbackRule.ts
```diff
@@ -0,0 +1,107 @@
...
+            if (this.isFunctionType(type)) {
+                const signatures = tc.getSignaturesOfType(type, ts.SignatureKind.Call);
+                for (const signature of signatures) {
+                    for (const param of signature.parameters) {
+                        const declarations = param.getDeclarations().map(function (declaration: ts.Declaration) {
+                            return <ts.ParameterDeclaration> declaration;
+                        });
+                        for (const declaration of declarations) {
+                            if (tc.isOptionalParameter(declaration)) {
+                                this.addFailure(this.createFailure(declaration.getStart(), declaration.getWidth(), Rule.FAILURE_STRING));
```
weswigham: Through the magic of type aliases, I _think_ this can error on this same thing multiple times. Example:

```ts
type CB = (things?: any) => void;
function do(a: CB, b: CB) {
  // ...
}
```

If this is the case, you _probably_ only want to error on a given declaration once, so you may want to collect all of your errors before reporting them, then make them per-declaration. Alternatively (and this is probably the better way to do it), keep a list of `declarations` you've already checked, and skip ones you've already seen.

src/rules/noOptionalParameterInCallbackRule.ts
weswigham: Then again, how does one differentiate methods on an object from callbacks in a map? There may not be a _good_ way to do this. Maybe make it an optional setting and only look into anonymous types?

src/rules/noOptionalParameterInCallbackRule.ts
rakatyal: Wouldn't you like to give multiple failures in this scenario? In the same context:

`function do(a: (b?: string) => void, c: (d?: number) => void) { }`

should give you 2 errors for 'b' and 'd'

weswigham: In that case, you're erroring on two different parameter declarations. In mine, both of them would be
```ts
type CB = (things?: any) => void;
           ^^^^^^^ [error] here
function do(a: CB, b: CB) {
  // ...
}
```
And seeing two errors get stacked on the same parameter declaration is a pain, since it gets fixed at the same time, in the same place.

rakatyal: You are right. But I don't see this happening on any other rules and we tend to give multiple errors. And on discussing this offline, keeping a visited array of declarations sounds like a stretch to solve a tiny issue.

JKillian:
All good from your side @weswigham? If so, I'm happy to go ahead and merge it

src/rules/noOptionalParameterInCallbackRule.ts
```diff
@@ -0,0 +1,126 @@
...
+                    if (this.checker.isOptionalParameter(declaration)) {
+                        this.addFailure(this.createFailure(declaration.getStart(), declaration.getWidth(), Rule.FAILURE_STRING));
+                    }
+                }
+            }
+        }
+    }
+
+    private validateSignature(node: ts.SignatureDeclaration) {
+        const that = this;
```
weswigham: This is unnecessary. Just use an arrow function below.

weswigham:
In addition to my other comment, I'd like to see the type alias case I mentioned added to the test case, so at least in any scenario the current behavior is recorded.

Additionally, the following test case would be nice to see, too:
```ts
interface Callback {
  (a: string, b?: number): Callback;
}

function callbackable(cb: Callback) {
                          ^^^^^^^^ <---- Error should probably be here (at use), not on declaration. Though it would be useful to point back to the type's declaration, since there's multiple ways to fix the issue
  cb("");
}
```

And on an interface with overloads:
```ts
interface Multicall {
  (a: string, b: number, c: number, d?: Multicall): Multicall;
  (a: string, b: number, c?: Multicall): Multicall;
  (a: string): Multicall;
  (): void;
}

function multicaller(x: Multicall) {
  x();
}
```
(Actually you should probably add a bunch of tests with overloads with a varying set of them)

rakatyal:
@weswigham: I believe the error shown would be better on declaration instead of use so that they are shown just once in case of multiple uses. I have added few more test cases to support this, please have a look.

rakatyal:
@JKillian: I am unable to figure out what's causing the build to break since it doesn't seem to be coming from my changes. Could you help me out?

weswigham:
@rakatyal But the issue isn't with the type definition (which is possibly shared with both parameters, variables, and object members) - it is the usage of that type within the context of a specific parameter. It can be fixed either by changing the declaration of the parameter's type _or_ by changing the declared type of the parameter. IDK, nonlocal errors seem like a really bad choice. (Along the lines of:: "I added a function with an argument of this type and then got a lint error in the type's declaration, what gives - the type didn't change?")

rakatyal:
@weswigham: Point well taken. I modified the code to make it work. Please have a look again.

nchen63:
@rakatyal Can you rebase or merge with `master`? I can take a look at the errors afterwards

adidahiya:
closing as stale. this can be revisited in a new PR.

Issue #561, opened 2015-08-10
Rules that can fix the problems they find


myitcv:
Are there any plans to extend `tslint` in such a way that rules can also be accompanied by transforms that can safely fix the breaks they (the rules) find?

For example, consider `noConsecutiveBlankLinesRule`. It would be trivial to write a transform to delete lines in files that fell foul of this rule. 

Clearly not all breakages can safely be fixed, but most formatting related rules probably could.

Thoughts?

(apologies if there is a task discussing this somewhere, I couldn't find anything based on my searches thus far)

adidahiya:
Yes, this is something I'd definitely like to have, especially since ESLint supports it as well.

JKillian:
Although this is fairly obvious, there'd be at least two components to doing this:

1. A part where each rule specified what transforms should be made (as #1048 requests)
2. A feature that actually runs those transforms on the appropriate files/strings

JKillian:
https://github.com/Microsoft/vscode-tslint/pull/47 contains some work that's a start to this feature

ScottSWu:
Hi, following up on #1048, we would like for rules to suggest text-replacement fixes for rule failures.

Similar to codelyzer and error-prone, we propose two structs for fixes and replacements (a fix may contain multiple non-contiguous replacements), and that RuleFailure contains an array of suggested fixes. API changes include getters for fixes, and createFix / createReplacement functions for rules to suggest fixes.

This differs from the vscode plugin by allowing rules (including custom rules) to add fixes, rather than matching the diagnostics.

At the moment, this change could be limited to the API and not the command line - applying fixes and cli options might require more discussion.

Some work on this implementation has been done on [this fork](/ScottSWu/tslint/tree/fixes_main).

JKillian:
Excited to see your work on this issue @ScottSWu!

nojvek:
I've just started working on a globally installable npm module that can take output of tslint and automatically fix the errors. I haven't published it yet but will ready by end of month.

https://github.com/nojvek/tslint-fix

Ideally its just nice for tslint to have --fix option. I'm not sure what it takes to get it merged there. I created this because we had a large typescript base that had 1000+ tslint issues and I needed to quickly fix them.

Does this sort of usage make sense?

`tslint -c tslint.json src/**/*.ts | tslint-fix`

myitcv:
[...]

PR #1423, opened 2016-07-21
Add text replacement fixes for rules

ScottSWu:
* Created IFix and IReplacement structures for representing fixes
  * programWalker includes createFix and createReplacement
* Added optional parameter to createFailure for suggesting fixes
* Modified semicolon rule to suggest fixes (add / remove semicolon)
* Modified testing to check against a file with fixes applied
  * If there is a ".ts.fix" file, then the test will take the first suggested fix and apply to the test file (without markup) and compare

Addresses #561

### DIFF
--- a/src/language/rule/rule.ts
+++ b/src/language/rule/rule.ts
@@ -89,6 +89,18 @@ export interface IRule {
     applyWithWalker(walker: RuleWalker): RuleFailure[];
 }
 
+export interface IReplacement {
+    start: number;
+    length: number;
+    text: string;
+}
+
+export interface IFix {
+    ruleName: string;
+    description: string;
+    replacements: IReplacement[];
+}
+
 export class RuleFailurePosition {
     constructor(private position: number, private lineAndCharacter: ts.LineAndCharacter) {
     }
@@ -128,7 +140,8 @@ export class RuleFailure {
                 start: number,
                 end: number,
                 private failure: string,
-                private ruleName: string) {
+                private ruleName: string,
+                private fixes: IFix[] = []) {
 
         this.fileName = sourceFile.fileName;
         this.startPosition = this.createFailurePosition(start);
@@ -155,10 +168,15 @@ export class RuleFailure {
         return this.failure;
     }
 
+    public getFixes() {
+        return this.fixes;
+    }
+
     public toJson(): any {
         return {
             endPosition: this.endPosition.toJson(),
             failure: this.failure,
+            fixes: this.fixes,
             name: this.fileName,
             ruleName: this.ruleName,
             startPosition: this.startPosition.toJson(),
--- a/src/language/rule/rule.ts
+++ b/src/language/rule/rule.ts
@@ -89,16 +89,52 @@ export interface IRule {
     applyWithWalker(walker: RuleWalker): RuleFailure[];
 }
 
-export interface IReplacement {
-    start: number;
-    length: number;
-    text: string;
+export class Replacement {
+    constructor(private start: number, private length: number, private text: string) {
+    }
+
+    public getStart() {
+        return this.start;
+    }
+
+    public getLength() {
+        return this.length;
+    }
+
+    public getEnd() {
+        return this.start + this.length;
+    }
+
+    public getText() {
+        return this.text;
+    }
+
+    public apply(content: string) {
+        return content.substring(0, this.start) + this.text + content.substring(this.start + this.length);
+    }
 }
 
-export interface IFix {
-    ruleName: string;
-    description: string;
-    replacements: IReplacement[];
+export class Fix {
+    constructor(private ruleName: string, private description: string, private replacements: Replacement[]) {
+    }
+
+    public getRuleName() {
+        return this.ruleName;
+    }
+
+    public getDescription() {
+        return this.description;
+    }
+
+    public getReplacements() {
+        return this.replacements;
+    }
+
+    public apply(content: string) {
+        // sort replacements in reverse so that diffs are properly applied
+        this.replacements.sort((a, b) => b.getEnd() - a.getEnd());
+        return this.replacements.reduce((text, r) => r.apply(text), content);
+    }
 }
 
 export class RuleFailurePosition {
@@ -141,7 +177,7 @@ export class RuleFailure {
                 end: number,
                 private failure: string,
                 private ruleName: string,
-                private fixes: IFix[] = []) {
+                private fixes: Fix[] = []) {
 
         this.fileName = sourceFile.fileName;
         this.startPosition = this.createFailurePosition(start);
--- a/src/language/rule/rule.ts
+++ b/src/language/rule/rule.ts
@@ -173,7 +173,7 @@ export class RuleFailure {
                 end: number,
                 private failure: string,
                 private ruleName: string,
-                private fixes: Fix[] = []) {
+                private fix?: Fix) {
 
         this.fileName = sourceFile.fileName;
         this.startPosition = this.createFailurePosition(start);
@@ -200,15 +200,19 @@ export class RuleFailure {
         return this.failure;
     }
 
-    public getFixes() {
-        return this.fixes;
+    public hasFix() {
+        return this.fix !== undefined;
+    }
+
+    public getFix() {
+        return this.fix;
     }
 
     public toJson(): any {
         return {
             endPosition: this.endPosition.toJson(),
             failure: this.failure,
-            fixes: this.fixes,
+            fix: this.fix,
             name: this.fileName,
             ruleName: this.ruleName,
             startPosition: this.startPosition.toJson(),
--- a/src/language/rule/rule.ts
+++ b/src/language/rule/rule.ts
@@ -90,23 +90,23 @@ export interface IRule {
 }
 
 export class Replacement {
-    constructor(private start: number, private length: number, private text: string) {
+    constructor(private innerStart: number, private innerLength: number, private innerText: string) {
     }
 
-    public getStart() {
-        return this.start;
+    get start() {
+        return this.innerStart;
     }
 
-    public getLength() {
-        return this.length;
+    get length() {
+        return this.innerLength;
     }
 
-    public getEnd() {
-        return this.start + this.length;
+    get end() {
+        return this.innerStart + this.innerLength;
     }
 
-    public getText() {
-        return this.text;
+    get text() {
+        return this.innerText;
     }
 
     public apply(content: string) {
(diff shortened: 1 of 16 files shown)


JKillian:
@alexeagle, do you plan to review this? I doubt I'll be able to get to this until sometime next week at the earliest. Definitely a great feature I want to get implemented!

ScottSWu:
cc: @mprobst

alexeagle:
I've reviewed it on Scott's branch. I'll take another pass here, but the design already LGTM

src/test.ts
```diff
@@ -104,9 +109,39 @@ export function runTest(testDirectory: string, rulesDirectory?: string | string[
...
+        // test against fixed files
+        let fixedFileText: string;
+        let newFileText: string;
+        try {
+            const fixedFile = fileToLint.replace(/\.lint$/, FIXES_FILE_EXTENSION);
+            const stat = fs.statSync(fixedFile);
+            if (stat.isFile()) {
+                fixedFileText = fs.readFileSync(fixedFile, "utf8");
+                // accumulate replacements
+                const replacements = failures.reduce((acc, failure) => {
```
alexeagle: I'm not sure the reduce here is worth the syntactic and cognitive cost, it might be simpler to just iterate?

src/test.ts
```diff
@@ -104,9 +109,39 @@ export function runTest(testDirectory: string, rulesDirectory?: string | string[
...
+                    const fixes = failure.getFixes();
+                    if (fixes.length > 0) {
+                        return acc.concat(fixes[0].replacements);
+                    } else {
+                        return acc;
+                    }
+                }, [] as IReplacement[]);
+                // sort in reverse so that diffs are properly applied
+                replacements.sort((a, b) => (b.start + b.length) - (a.start + a.length));
+                newFileText = replacements.reduce((acc, r) => {
```
alexeagle: at some point, other places (eg. the ProseFormatter) will want to be able to apply fixes, so this probably belongs somewhere else. It could be refactored to another location in a later PR, when there is a second caller who wants to share it.

src/test.ts
```diff
@@ -122,20 +157,40 @@ export function consoleTestResultHandler(testResult: TestResult): boolean {
...
             console.log(colors.red(" Failed!"));
-            console.log(colors.green(`Expected (from ${FILE_EXTENSION} file)`));
+        }
+        if (!didMarkupTestPass) {
+            console.log(colors.green(`Expected markup (from ${MARKUP_FILE_EXTENSION} file)`));
+            console.log(colors.red("Actual (from TSLint)"));
+
+            didAllTestsPass = false;
+
+            for (const diffResult of markupDiffResults) {
```
alexeagle: consider extracting a helper function to print the failure, rather than duplicate this between markup failure and diff failure?

src/test.ts
ScottSWu: Turned Replacement into a class and added an apply function.

src/test.ts
ScottSWu: Done

src/test.ts
ScottSWu: Done

src/language/rule/rule.ts
```diff
@@ -89,6 +89,54 @@ export interface IRule {
     applyWithWalker(walker: RuleWalker): RuleFailure[];
 }
 
+export class Replacement {
+    constructor(private start: number, private length: number, private text: string) {
+    }
+
+    public getStart() {
```
JKillian: I'd be fine with making all of these real `get` methods. I find that this sort of method can get frustrating (but fine either way)

src/language/rule/rule.ts
```diff
@@ -89,6 +89,54 @@ export interface IRule {
...
+    }
+
+    public getReplacements() {
+        return this.replacements;
+    }
+
+    public apply(content: string) {
+        // sort replacements in reverse so that diffs are properly applied
+        this.replacements.sort((a, b) => b.getEnd() - a.getEnd());
+        return this.replacements.reduce((text, r) => r.apply(text), content);
```
JKillian: great use for `reduce`!

src/language/walker/ruleWalker.ts
```diff
@@ -82,6 +82,14 @@ export class RuleWalker extends SyntaxWalker {
         }
     }
 
+    public createFix(description: string, replacements: Replacement[]): Fix {
+        return new Fix(this.ruleName, description, replacements);
+    }
+
+    public createReplacement(start: number, length: number, text: string): Replacement {
+        return new Replacement(start, length, text);
+    }
```
JKillian: in the future, we could possible add some helper methods for typical fixes (`insertAt`, `delete`, stuff like that).

src/rules/semicolonRule.ts
```diff
@@ -58,6 +58,8 @@ export class Rule extends Lint.Rules.AbstractRule {
 
     public static FAILURE_STRING_MISSING = "Missing semicolon";
     public static FAILURE_STRING_UNNECESSARY = "Unnecessary semicolon";
+    public static FIX_STRING_MISSING = "Add semicolon";
```
JKillian: are fix strings necessary for your use case? It seems to me that the failure string would be an apt descriptor in most cases

src/test.ts
```diff
@@ -104,9 +109,36 @@ export function runTest(testDirectory: string, rulesDirectory?: string | string[
...
+                // accumulate replacements
+                let replacements: Replacement[] = [];
+                for (const failure of failures) {
+                    const fixes = failure.getFixes();
+                    if (fixes.length > 0) {
+                        replacements = replacements.concat(fixes[0].getReplacements());
+                    }
+                }
+                // sort in reverse so that diffs are properly applied
+                replacements.sort((a, b) => b.getEnd() - a.getEnd());
```
JKillian: This logic still seems somewhat duplicated to me, maybe there should be a static `Fix.applyAll(text: string, fixes: Fix[])` method?

src/test.ts
```diff
@@ -122,31 +154,43 @@ export function consoleTestResultHandler(testResult: TestResult): boolean {
...
         /* tslint:enable:no-console */
     }
 
     return didAllTestsPass;
 }
+
+function displayDiffResults(diffResults: diff.IDiffResult[], extension: string) {
+    /* tslint:disable:no-console */
+    console.log(colors.green(`Expected (from ${extension} file)`));
+    console.log(colors.red("Actual (from TSLint)"));
```
JKillian: Will we want any message here to differentiate that you're looking at fix or lint failure differences?

JKillian:
Looking good @ScottSWu.

Is the Replacement API robust enough? How tricky to use is it when you get into weird changes, like overlapping fixes? What about when multiple rules want to apply fixes to the same file? I know you looked into the way it's done in other similar tools.

Also, a test or two for your new test code would be great if possible! You could just add it to `test/check-bin.sh`

alexeagle:
Applying fixes from multiple rules at once sounds like trouble to me. In the worst case, two rules may conflict, and after applying one set of replacements, you are left with some that don't apply, and now the result isn't even syntactically valid.
I think we'd have to apply the fixes (either to the filesystem or an in-memory buffer) after running one rule, then take the result and pass it to the next rule. You may get different results depending on the order the rules are executed.

Within a single rule, I'm not sure there is a need for overlapping fixes. Fixes should prepend or append text where possible, rather than replacing a whole range (and it's hard to convert an AST subtree to text anyway)

src/rules/semicolonRule.ts
ScottSWu: The idea was that rules could suggest more than one fix (though I can't actually think of a good example for needing more than one). Should each rule return at most one fix?

src/test.ts
ScottSWu: Hmm, since these outputs are just for TSLint tests, is the extension enough or should it be more explicit?

src/rules/semicolonRule.ts
JKillian: I think one `Fix` per failure per rule per file makes sense?

src/test.ts
JKillian: Oh sorry, missed the `${extension}` part, that's fine! No need for changes

src/rules/semicolonRule.ts
ScottSWu: Done.

src/language/rule/rule.ts
ScottSWu: Done.

src/test.ts
ScottSWu: Done.

ScottSWu:
To add on a bit, no these replacement functions are not robust, but would be building blocks for an something that is. A single fix should not have overlapping replacements within itself, but also should be applied entirely or not at all (e.g. it doesn't make sense to replace one quote mark). Thus the Fix.apply and the Replacement.apply functions will be useful.

From what ESLint does, fixes are automatically applied, and only fixes that fail to be applied are displayed.

alexeagle:
Ping @JKillian - Scott's internship has only a couple more weeks so we'd like to get this moving

src/language/walker/ruleWalker.ts
```diff
@@ -82,6 +82,14 @@ export class RuleWalker extends SyntaxWalker {
         }
     }
 
+    public createFix(replacements: Replacement[]): Fix {
```
JKillian: I went back and forth on if we should have these wrapper methods at all (as the constructors have the exact same functionality). Do you think it makes sense to have them?

src/test.ts
```diff
@@ -104,9 +109,32 @@ export function runTest(testDirectory: string, rulesDirectory?: string | string[
...
 
+        // test against fixed files
+        let fixedFileText: string;
+        let newFileText: string;
+        try {
+            const fixedFile = fileToLint.replace(/\.lint$/, FIXES_FILE_EXTENSION);
+            const stat = fs.statSync(fixedFile);
+            if (stat.isFile()) {
+                fixedFileText = fs.readFileSync(fixedFile, "utf8");
+                const fixes: Fix[] = [];
```
JKillian: Believe you could simplify this a little:
`const fixes = failures.filter((f) => f.hasFix()).map((f) => f.getFix());`

src/language/walker/ruleWalker.ts
alexeagle: I'd prefer to expose only the typical fix methods. These are what we need for the current check and expose less API details to all the rules that might start using it.

so far we need
`appendText(n: node, t: text)` for putting a semicolon after the statement, and
`delete(start: number, length: number)` for deleting a semicolon?

ScottSWu: Fix is probably redundant. Eventually we should proxy replacements (including createReplacement) to make sure replacements at the same spots are applied in the right order. It turns out sorting is not guaranteed to be stable, but adding an additional ordering counter breaks ties correctly.

src/language/rule/rule.ts
```diff
@@ -89,6 +89,61 @@ export interface IRule {
...
+        return this.innerRuleName;
+    }
+
+    get replacements() {
+        return this.innerReplacements;
+    }
+
+    public apply(content: string) {
+        // sort replacements in reverse so that diffs are properly applied
+        this.innerReplacements.sort((a, b) => b.end - a.end);
```
JKillian: I'm thinking we could have a `Replacements.applyAll` that is parallel to `Fix.applyAll`?

JKillian:
@ScottSWu As far as `check-bin` tests, I was thinking perhaps something like [here](https://github.com/palantir/tslint/blob/dcdc22cab0287f814ef40c6b52cb3443fdbd5116/test/check-bin.sh#L97L102) just to make sure that the code paths dealing with `.fix` have a test that specifically hits them.

JKillian:
This is looking good though, I think we can merge this shortly. @ScottSWu, @alexeagle what do you plan to proceed with after this merges? Implementing fixes for more rules?

ScottSWu:
Got it, added a passing and failing check-bin tests.

I believe the next step is to add the --fix option to the CLI. Implementing more fixes for rules is definitely important. There's also adding more helper functions for creating replacements, and outputting fixes in the other formatters.

JKillian:
Sweet, future plans sound good to me. Merged this so things can more easily continue!

ScottSWu:
Awesome! Thanks for pushing this through!

PR #1426, opened 2016-07-22
Adding new rule to disallow non consecutive function overloads

rakatyal:


src/rules/consecutiveFunctionOverloadsRule.ts
```diff
@@ -0,0 +1,97 @@
...
+export class Rule extends Lint.Rules.AbstractRule {
+
+    /* tslint:disable:object-literal-sort-keys */
+    public static metadata: Lint.IRuleMetadata = {
+        ruleName: "consecutive-function-overloads",
+        description: "Enforces function overloads to be consecutive.",
+        optionsDescription: "Not configurable.",
+        options: null,
+        optionExamples: ["true"],
+        type: "style",
```
adidahiya: let's change this to `"typescript"` since it's typescript-specific

src/rules/consecutiveFunctionOverloadsRule.ts
```diff
@@ -0,0 +1,97 @@
...
+    /* tslint:enable:object-literal-sort-keys */
+
+    public static FAILURE_STRING_FACTORY = (name: string) => `All '${name}' signatures should be adjacent`;
+
+    public apply(sourceFile: ts.SourceFile): Lint.RuleFailure[] {
+        return this.applyWithWalker(new NoNonConsecutiveFunctionOverload(sourceFile, this.getOptions()));
+    }
+}
+
+function isStringOrNumericLiteral(kind: ts.SyntaxKind) {
```
adidahiya: prefer private helper functions to occur at the end of the file, after the walker class please

adidahiya: also, can you refactor this into a type guard?

```ts
function isLiteralExpression(node: ts.Node): node is ts.LiteralExpression {
    return node.kind === ts.SyntaxKind.StringLiteral || node.kind === ts.SyntaxKind.NumericLiteral;
}
```

generally, type guards are preferred over assertions whenever possible.

src/rules/consecutiveFunctionOverloadsRule.ts
```diff
@@ -0,0 +1,97 @@
...
+    return kind === ts.SyntaxKind.StringLiteral || kind === ts.SyntaxKind.NumericLiteral;
+}
+
+function getTextOfPropertyName(name: ts.PropertyName): string {
```
adidahiya: if you create the type guard I suggest above, this can be a little simpler:

```ts
switch (name.kind) {
    case ts.SyntaxKind.Identifier:
        return (name as ts.Identifier).text;
    case ts.SyntaxKind.ComputedPropertyName:
        const { expression } = (name as ts.ComputedPropertyName);
        if (isLiteralExpression(expression)) {
            return expression.text;
        }
     default:
        if (isLiteralExpression(name)) {
            return name.text;
        }
}
```

src/rules/consecutiveFunctionOverloadsRule.ts
```diff
@@ -0,0 +1,97 @@
...
+                return (<ts.LiteralExpression> (<ts.ComputedPropertyName> name).expression).text;
+            }
+        default:
+            return undefined;
+    }
+}
+
+class NoNonConsecutiveFunctionOverload extends Lint.BlockScopeAwareRuleWalker<{}, ScopeInfo> {
+
+    public createScope(): any {
```
adidahiya: please narrow this return type to `{}`

src/rules/consecutiveFunctionOverloadsRule.ts
```diff
@@ -0,0 +1,97 @@
...
+        return undefined;
+    }
+
+    public createBlockScope(): ScopeInfo {
+        return new ScopeInfo();
+    }
+
+    public visitInterfaceDeclaration(node: ts.InterfaceDeclaration): void {
+        const members = node.members;
+        for (const member of members) {
```
adidahiya: `members` is only used once; prefer inlining here as `node.members`

test/rules/consecutive-function-overloads/test.ts.lint
```diff
@@ -0,0 +1,63 @@
...
+    12();
+    ~~~~~  [All '12' signatures should be adjacent]
+}
+
+interface b4 {
+    a();
+    b(): void;
+    ["a"](v: number): void;
+    ~~~~~~~~~~~~~~~~~~~~~~~  [All 'a' signatures should be adjacent]
+}
```
adidahiya: minor: remove trailing whitespace and add an EOF newline

test/rules/consecutive-function-overloads/tslint.json
```diff
@@ -0,0 +1,5 @@
+{
+  "rules": {
+    "consecutive-function-overloads": true
+  }
+}
```
adidahiya: minor: please add an EOF newline

test/rules/consecutive-function-overloads/test.ts.lint
```diff
@@ -0,0 +1,63 @@
...
+    "a"();
+}
+
+interface i4 {
+    a();
+    ["a"]();
+}
+
+
+// bad
```
adidahiya: can you add some tests for nested interfaces?

```ts
interface foo {
    a: string;
    bar: {
        a: number;
        ...
    }
}
```

README.md
```diff
@@ -254,6 +254,7 @@ Core rules are included in the `tslint` package.
 * `arrow-parens` requires parentheses around the parameters of arrow function definitions.
 * `ban` bans the use of specific functions. Options are ["object", "function"] pairs that ban the use of object.function().
 * `class-name` enforces PascalCased class and interface names.
+* `consecutive-function-overloads` enforces function overloads to be consecutive.
```
adidahiya: how about naming this `adjacent-overload-signatures`?

make sure to name the walker class appropriately as well; in this case it would be `AdjacentOverloadSignaturesWalker`

adidahiya:
Thanks for the PR @rakatyal, in general I like the idea of this rule. Needs some small tweaks and additional tests before merging.

test/rules/consecutive-function-overloads/test.ts.lint
rakatyal: Good catch! I don't think this code will work for nested interfaces right now since it's only sweeps interfaceDeclarations. I probably would have to check if an interface member is an object and then recurse for them. Do you have any other suggestion?

adidahiya: yep, I think you'll want to recurse and create a new block scope for the nested member

adidahiya: also, take a look at the [object-literal-sort-keys rule](https://github.com/palantir/tslint/blob/master/src/rules/objectLiteralSortKeysRule.ts) because it has to handle a similar scenario for nested objects

rakatyal:
@adidahiya: Thanks for the detailed review. I have refactored the code according to your suggestions and to handle the case for nested interfaces. Please have a look again.

adidahiya:
lgtm 👍

PR #1433, opened 2016-07-26
Use correct type for tagged string function parameter

JKillian:
This allows the code to compile correctly with TS 1.8.x and TS 2.x

See https://github.com/Microsoft/TypeScript/pull/9855 also


adidahiya:
👍

PR #1435, opened 2016-07-27
New Rule: prefer-for-of

chrismbarr:
This is a cleaner commit of #1340.  The code is identical, but fewer commits that should be easier to merge with master.


src/rules/preferForOfRule.ts
```diff
@@ -0,0 +1,112 @@
...
+
+    public static FAILURE_STRING = "This standard for loop could be replaced with a for(... of ...) loop";
+
+    public apply(sourceFile: ts.SourceFile): Lint.RuleFailure[] {
+        const languageService = Lint.createLanguageService(sourceFile.fileName, sourceFile.getFullText());
+        return this.applyWithWalker(new ForOfWalker(sourceFile, this.getOptions(), languageService));
+    }
+}
+
+class ForOfWalker extends Lint.RuleWalker {
```
adidahiya: should be called `PreferForOfWalker` to match the rule name

src/rules/preferForOfRule.ts
```diff
@@ -0,0 +1,112 @@
...
+    constructor(sourceFile: ts.SourceFile, options: Lint.IOptions, private languageService: ts.LanguageService) {
+        super(sourceFile, options);
+        this.languageService = languageService;
```
adidahiya: don't need this line since you have `private languageService` in the method signature

src/rules/preferForOfRule.ts
```diff
@@ -0,0 +1,112 @@
...
+    }
+
+    public visitForStatement(node: ts.ForStatement) {
+        this.handleForStatement(node);
```
adidahiya: this indirection seems unnecessary; can you inline `this.handleForStatement` here?

src/rules/preferForOfRule.ts
```diff
@@ -0,0 +1,112 @@
...
+        if (arrayAccessNode.kind === ts.SyntaxKind.PropertyAccessExpression && arrayAccessNode.getLastToken().getText() === "length") {
+            const incrementorVariable = node.incrementor.getFirstToken();
+            const arrayToken = arrayAccessNode.getChildAt(0);
+            const loopSyntax = node.statement.getChildAt(1);
+
+            // Find all usages of the incrementor variable
+            const fileName = this.getSourceFile().fileName;
+            const highlights = this.languageService.getDocumentHighlights(fileName, incrementorVariable.getStart(), [fileName]);
+            // There are three usages when setting up the for loop,
+            // so remove those form the count to get the count inside the loop block
```
adidahiya: typo: "from"

src/rules/preferForOfRule.ts
```diff
@@ -0,0 +1,112 @@
...
+            // This means that the incrementor variable is not used on its own anywhere and is ONLY
+            // used to access the array item.
+            if (matchCount >= incrementorCount) {
+                const failure = this.createFailure(node.getStart(), node.getWidth(), Rule.FAILURE_STRING);
+                this.addFailure(failure);
+            }
+        }
+    }
+
+    private locateArrayNodeForLoop(forLoop: ts.ForStatement): ts.Node {
```
adidahiya: I would either rename this to `locateArrayNodeInForLoop` or omit the `ForLoop` part (since it's indicated in the type signature

src/rules/preferForOfRule.ts
```diff
@@ -0,0 +1,112 @@
...
+        let arrayAccessNode = forLoop.condition.getChildAt(2);
+        // If We haven't found it, maybe it's not a standard for loop, try looking in the initializer for the array
+        // Something like `for(var t=0, len=arr.length; t < len; t++)`
+        if (arrayAccessNode.kind !== ts.SyntaxKind.PropertyAccessExpression) {
+            forLoop.initializer.getChildren().forEach((initNode: ts.Node) => {
```
adidahiya: can we use a for-of loop here?

src/rules/preferForOfRule.ts
```diff
@@ -0,0 +1,112 @@
...
+                // look in `var t=0, len=arr.length;`
+                if (initNode.kind === ts.SyntaxKind.SyntaxList) {
+                    initNode.getChildren().forEach((initVar: ts.Node) => {
```
adidahiya: for-of loop here too?

src/rules/preferForOfRule.ts
```diff
@@ -0,0 +1,112 @@
...
+                        // look in `t=0, len=arr.length;`
+                        if (initVar.kind === ts.SyntaxKind.VariableDeclaration) {
+                            initVar.getChildren().forEach((initVarPart: ts.Node) => {
```
adidahiya: for-of loop here too?

src/rules/preferForOfRule.ts
```diff
@@ -0,0 +1,112 @@
...
+        description: "Recommends a for of loop over a standard for loop if the index is only used to access the array being looped over.",
+        rationale: "A for(... of ...) loop is easier to implement and read when the index is not needed.",
+        optionsDescription: "Not configurable.",
+        options: null,
+        optionExamples: ["true"],
+        type: "typescript",
+    };
+    /* tslint:enable:object-literal-sort-keys */
+
+    public static FAILURE_STRING = "This standard for loop could be replaced with a for(... of ...) loop";
```
adidahiya: since it's a lint _failure_, I think the messaging should be stronger (currently "could" makes it kind of weak).

How about "Expected a for-of loop instead of a for loop for simple iteration"

chrismbarr:
Good suggestions, thanks!  Check my changes and see if that fixes everything

test/rules/prefer-for-of/test.ts.lint
```diff
@@ -0,0 +1,63 @@
...
+        if (obj.hasOwnProperty(z)) {
+            console.log(z);
+        }
+    }
+    
+    //For of loops are allowed
+    for (var m of arr) {
+        console.log(m);
+    }
+}
```
JKillian: Can we make sure that your rule doesn't crash w/ more unusual loops? For example, maybe `for (;;) {}`, loops w/o blocks `for(let i = 0; i< arr.length; i++) method(arr[i]);`, and prefix incrementors `for(let i = 0; i < arr.length; ++i) { arr[i].whatever();`

JKillian:
Sorry for the delay in getting to this! Can we add a few more test cases to make sure everything is working well, and also merge `master` into this to get it up to date?

README.md
```diff
@@ -335,6 +335,7 @@ Core rules are included in the `tslint` package.
 * `one-variable-per-declaration` disallows multiple variable definitions in the same statement.
   * `"ignore-for-loop"` allows multiple variable definitions in for loop statement.
 * `only-arrow-functions` disallows traditional `function () { ... }` declarations, preferring `() => { ... }` arrow lambdas.
+* `prefer-for-of` recommends a [`for(... of ...)` loop](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/for...of) over a standard `for` loop if the index is only used to access the array being looped over.
```
JKillian: You can just get rid of this all together now, rules are no longer listed in the README

chrismbarr:
Good call!  I had to make some fixes for those added loop variations.

Right now the tests for this new rule pass.... but I'm getting unrelated failures for another rule.  I'm not sure what's up with that.
image

chrismbarr:
Also, one thing I'm unsure of with this rule that I realized today is that I locate the incrementor (`i` in `i++` for example) and then I look for highlight spans of it.  If I re-use the same variable name `i` in various loops, it seems to find ALL usages of it with the same name, not just the ones localized to the current loop.

Is there a way to make it only look for togen usages only with a certain scope?

JKillian:
> If I re-use the same variable name i in various loops, it seems to find ALL usages of it with the same name, not just the ones localized to the current loop.

This is correct behavior actually: when you use `var`, the variable actually exists in the entire scope of the containing function. If you use `let` or `const`, the variable is scoped to the for-loop specifically.

JKillian:
As far as the test errors, that was caused by https://github.com/palantir/tslint/pull/1475 I think. I'll have that fixed in a minute or two

chrismbarr:
I've merged in master again, but I still get the exact same error locally

JKillian:
> Is there a way to make it only look for token usages only with a certain scope?

Never really answered this question, sorry! I figure you could just find all occurrences, but just filter them to only the ones inside the for loop. At the worse you could use the position of the token and start and end of the for loop body.

chrismbarr:
@JKillian I didn't realize there was a new check for the license agreement thing. I took care of that this morning.

JKillian:
Sorry for such a delay, are you happy with this @chrismbarr, think it's complete?

chrismbarr:
I think it is, as long as you all think it looks good and the tests cover as many cases as possible

PR #1438, opened 2016-07-27
Prepare release v3.14.0-dev.1

JKillian:


adidahiya:
Should we add these new rules to `tslint:latest` before cutting the release?

adidahiya:
Also do we want to get #1433 in this release?

JKillian:
Good call, definitely wanted #1433 in this release. I don't actually know how this built...

I want to look though the new rules and see which make sense for `tslint:latest`, although wasn't sure if I'd actually add them in this release or in the next release (just because this is `dev.1`)

JKillian:
It's also going to take some work to merge `next` back into `master` for the main release I'm afraid. Some of the new typechecking code isn't compatible with TS 1.8. (Unless we want to hold off on adding that to master?)

adidahiya:
> Some of the new typechecking code isn't compatible with TS 1.8

which part specifically?

JKillian:
Haven't tested the whole way yet, but I think [this field](https://github.com/palantir/tslint/blob/785e7c45fd58e0e1fccef84ed99d85865f3ba0d3/src/tslint.ts#L63) isn't accepted in the 1.8 `ParseConfigHost` type

JKillian:
Actually, we may be good with the current `tslint:latest` config:

`arrow-parens`: is a rule I want to use, but I'm not sure it should go in the config.
`max-file-line-count`: is useful, but not something that should be on by default.
`no-unsafe-finally`: is already included in `tslint:latest`
`no-for-in-array`: is globally useful, but requires type-checking
`object-literal-key-quotes`: this is the one we possibly should add

adidahiya:
cool, let's leave the config as-is and just pull in master

PR #1441, opened 2016-07-27
Add new rule to enforce headers matched by regexp

ScottSWu:
- Checks for comments at the very beginning of files
- Matches comment contents with header regexp
- Ignores #! at the beginning of files

Relates to #155 

cc: @alexeagle


src/rules/enforceHeaderRule.ts
```diff
@@ -0,0 +1,61 @@
+import * as ts from "typescript";
+
+import * as Lint from "../lint";
+
+export class Rule extends Lint.Rules.AbstractRule {
+    /* tslint:disable:object-literal-sort-keys */
+    public static metadata: Lint.IRuleMetadata = {
+        ruleName: "enforce-header",
```
adidahiya: can we call this rule "file-header" instead? "enforce" is a little redundant because all our rules enforce _something_.

src/rules/enforceHeaderRule.ts
```diff
@@ -0,0 +1,61 @@
...
+        optionsDescription: "Regular expression to match the header.",
+        options: {
+            type: "string",
+        },
+        optionExamples: ['"true", "Copyright \\d{4}"'],
+        type: "style",
+    };
+    /* tslint:enable:object-literal-sort-keys */
+
+    public static FAILURE_STRING = "missing header";
```
adidahiya: "Missing file header"

adidahiya:
overall looks pretty good, just left minor comments. @JKillian can you also take a look?

src/rules/enforceHeaderRule.ts
ScottSWu: Done.

src/rules/enforceHeaderRule.ts
ScottSWu: Done.

JKillian:
Looks fine to me, thanks @ScottSWu!

PR #1443, opened 2016-07-28
Added additional type constraint for findFormatter function parameter

DovydasNavickas:
I've added added additional type constraint for `formatter` in ILinterOptionsRaw interface and `findFormatter` function parameter name as it accepts the type Function, but was constrained to `string` only.


JKillian:
Looks good to me, thanks @DovydasNavickas!

PR #1447, opened 2016-08-01
Require semicolon for function declaration with no body

andy-ms:
I suggest that all function declarations *without bodies* be required to end in a semicolon.
What do you think of this rule?
This would apply to both overloads and to functions in `.d.ts` declarations.


JKillian:
I agree with this, although I'm a little worried because it's technically a breaking change. Think consumers would mind this change? I'd rather not bump a major version for it

JKillian:
Slating this for an upcoming 4.0 release (in a couple weeks probably?)

andy-ms:
Great!

PR #1454, opened 2016-08-02
Fixing potential bug

rakatyal:


JKillian:
Hi @rakatyal, thanks for the PR! Is there a corresponding issue for this? Could you add a test case that would be broken before your PR and is now working correctly?

src/rules/adjacentOverloadSignaturesRule.ts
```diff
@@ -58,7 +58,7 @@ class AdjacentOverloadSignaturesWalker extends Lint.RuleWalker {
             if (member.name !== undefined) {
                 const methodName = getTextOfPropertyName(member.name);
                 if (methodName !== undefined) {
-                    if (seen[methodName] && last !== methodName) {
+                    if (hasProperty(seen, methodName) && last !== methodName) {
```
adidahiya: why not just make this `seen.hasOwnProperty(methodName)`?

rakatyal:
@JKillian: There isn't any corresponding issue. I have added the corresponding test case.

adidahiya:
Thanks for revising, but the rationale behind this fix is still not immediately clear from the diff -- can you explain it in words at least, if not in a code comment?

rakatyal:
I think when looping over the members of the node, the parents' object properties are also included. The rule checks if the given method name has been encountered and due to this when someone tries to override known methods like toString(), toLowerCase() which are defined in parents' objects, it fails even if they are only defined once.
Hence we have to use hasOwnProperty instead of just using seen[methodName].

PR #1459, opened 2016-08-07
Update "typedef" rule metadata

### LINKED ISSUE #1451: Add docs for typedef rule (option: arrow-call-signature)
I couldn't find them [here](http://palantir.github.io/tslint/rules/typedef/).

glen-84:
Fixes #1451.


JKillian:
Thanks @glen-84! This change will be live with the next release of TSLint when I update the site next.

Issue #1353, opened 2016-06-28
Empty xml report with checkstyle


Quantumplation:
### Bug Report

- __TSLint version__: `3.11.0`
- __TypeScript version__: `1.8.10`
- __Running TSLint via__: CLI

```
./node_modules/.bin/tslint src/**/*.ts --exclude **/*-spec.ts -o ./_lint/report.xml -t checkstyle
```

#### Actual behavior

File `report.xml` is created, but is empty

#### Expected behavior

File `report.xml` should have contents, something like:
```xml
<?xml version="1.0" encoding="utf-8"?>
<checkstyle version="4.3">
</checkstyle>
```

My build server fails because the produced file is invalid XML

adidahiya:
Do you get any failures reported with other formatters? Try quoting your globs in the CLI (`'src/**/*.ts`, etc.). If the linter is getting files and reporting zero failures but not outputting any XML, this sounds like a bug ([see these tests](https://github.com/palantir/tslint/blob/master/test/formatters/checkstyleFormatterTests.ts)).

Quantumplation:
If I introduce a linting violation, the command (unchanged) produces a valid checkstyle xml file, it's only when there are no linting errors in the project.

caugner:
Same problem and expectation (cf. initial comment) here.

JKillian:
This seems like a general TSLint w/ regards to formatters and not just an issue with the checkstyle one. [If there are no violations, nothing gets written to stdout.](https://github.com/palantir/tslint/blob/e23b1885fd8caca734c31078d70433640c422176/src/tslint-cli.ts#L254)

I'm not sure why this is the current behavior, I'm up for changing this unless anyone can think of a reason why it's better as it is now.

caugner:
> I'm up for changing this unless anyone can think of a reason why it's better as it is now.

I agree with this change, because the current behavior makes the wrong assumption that formatters don't need to emit anything on success.

It should be noted that custom formatters could _potentially_ make use of this assumption and report a failure even on success (e.g. if the reporter prints _"There were N failures: (...)"_ and CI checks for the string `There were [0-9]+ failures`). As this is only a _hypothetical_ problem and the included standard formatters are not affected by this, it should suffice to note this in the release notes, so that users with this issue have a higher chance of quickly resolving it.

caugner:
This issue [also affects the pmd reporter](https://github.com/palantir/tslint/blob/master/src/formatters/pmdFormatter.ts#L23).

[...]

PR #1462, opened 2016-08-07
Always write lint result. (#1353)

caugner:
Ensures that the result is always written to the output stream, even if there were no failures.


caugner:
Sorry, I created the PR against the wrong branch and will therefore close it.

Issue #656, opened 2015-09-13
Formatter processed on a per file basis


lbesson:
As far as I understand, rules are run on a per file basis. This means that formatter is called once for each file.

This is a problem if for example we want to output checkstyle file, since the xml and checkstyle tag must be added by other means.

Else, we would get output such as

```xml
<?xml version="1.0" encoding="utf-8"?>
<checkstyle version="4.3">
  <file name="file1.ts">
    <error line="1" "column="1" "severity="error" "blah" />
    <error line="2" "column="1" "severity="error" "bluh" />
  </file>
</checkstyle>
<checkstyle version="4.3">
  <file name="file2.ts">
    <error line="1" "column="1" "severity="error" "blah" />
    <error line="2" "column="1" "severity="error" "bluh" />
  </file>
</checkstyle>
```

which is incorrect

lbesson:
Looks similar to https://github.com/palantir/tslint/issues/379

lbesson:
Should probably be solved once https://github.com/palantir/tslint/issues/680 is ok?

Sweetchuck:
I wrote a little npm package which implements 2 new formatters for _tslint_, `yaml` and `checkstyle`.
As @lbesson pointed out that there is a problem with the checkstyle format because the formatter called with one file at a time and all the output saved into one file (or stdout). This method leads to an invalid XML file.

Luckily this method allows to generate a syntactically correct multi-document YAML output and it can be converted to any format including a correct checkstyle XML file.

This npm package also can be a temporary solution for #250 (Suggestion: Add checkstyle formatter)

https://github.com/Sweetchuck/npm-tslint-formatters

JKillian:
#379 contains some new discussion on this issue - discussion can continue here, this is something that should get fixed!

EarthCitizen:
When I originally looked at this and commented in #379, I was not paying attention to the fact that the Linter constructor wants the file name AND the contents of the file. This makes refactoring class Linter a bit difficult. If there needs to be backwards compatibility with a public API, then perhaps mark that class deprecated and start over in a new class.

If that were to happen, the new class should get the file contents itself when it is linting and not receive them in the constructor.

caugner:
I can see two solutions for this issue:

1. Let `#processFile` return the results per file without calling the formatter (or writing to the outputStream respectively), collect all results and hand it to the formatter (and write it to the outputstream) only once at the end of the tslint run.
[...]

PR #1463, opened 2016-08-07
Always write lint result. (#1353)

caugner:
Ensures that the result is always written to the output stream, even if there were no failures.

### DIFF
--- a/src/tslint-cli.ts
+++ b/src/tslint-cli.ts
@@ -251,11 +251,11 @@ const processFile = (file: string, program?: ts.Program) => {
 
     const lintResult = linter.lint();
 
-    if (lintResult.failureCount > 0) {
-        outputStream.write(lintResult.output, () => {
+    outputStream.write(lintResult.output, () => {
+        if (lintResult.failureCount > 0) {
             process.exit(argv.force ? 0 : 2);
-        });
-    }
+        }
+    });
 };
 
 // if both files and tsconfig are present, use files


JKillian:
[Fixes #1353]

JKillian:
Does this give reasonable output for all formatters when there are no failures? For example, I'm wondering if the msBuild and prose formatters should really be outputting a single `\n` when there are no failures.

caugner:
As I see it, this PR is only about correctly writing out the formatter's result in any case.

If there are formatters that don't give reasonable output for an empty failure list, this would be a bug in the formatter, which should be addressed separately. But as far as I can see, there aren't any obvious bugs in the included formatters, and the trailing `\n` actually does make sense, because text files [should always end with a newline](http://stackoverflow.com/questions/729692/why-should-text-files-end-with-a-newline).

JKillian:
Ah, I do see one issue actually, although it's related to another bug. TSLInt via the CLI doesn't handle output very well for multiple files: https://github.com/palantir/tslint/issues/656

With this change, formatters will output lots of extraneous output, because they'll output something for every file without errors. For the formatters mentioned above, we'll get a blank line for each file without failures, which will end up being a lot of blank lines.

What do you think @caugner? If we had a fix for #656 I would merge this right away, but I'm a little worried given #656.

JKillian:
Closing this as it's included in #1472

alexeagle:
True that text files should end with a newline, but if I run tslint on the command line, there is no text file, just a non-empty stdout.

PR #1464, opened 2016-08-07
Cyclomatic Complexity Rule

### LINKED ISSUE #1411: Cyclomatic Complexity
Clicked post to early, will flesh this out and re-open

bencoveney:
Resolves #1411 

Roughly matches the corresponding ESLint and JSHint rule with the exception that `&&` is now considered a code branch.


src/rules/cyclomaticComplexityRule.ts
```diff
@@ -0,0 +1,203 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import * as Lint from "../lint";
+import * as ts from "typescript";
+
+export class Rule extends Lint.Rules.AbstractRule {
+
+    public static ANONYMOUS_FAILURE_STRING = "The cyclomatic complexity of is higher than the threshold";
```
markwongsk: Wording?

src/rules/cyclomaticComplexityRule.ts
```diff
@@ -0,0 +1,203 @@
...
+    public static NAMED_FAILURE_STRING = "The cyclomatic complexity of is higher than the threshold for the function: ";
```
JKillian: wording here as well

src/rules/cyclomaticComplexityRule.ts
```diff
@@ -0,0 +1,203 @@
...
+
+    public static DEFAULT_THRESHOLD = 20;
+    public static MINIMUM_THRESHOLD = 1;
```
JKillian: Would this ever make sense? Should we go with 5 or something?

src/configs/latest.ts
```diff
@@ -16,6 +16,7 @@
  */
 
 export const rules = {
+    "cyclomatic-complexity": true,
```
JKillian: Don't want to enable this by default, because I'm not sure cyclomatic complexity is something everyone would want to check by default. Thoughts?

README.md
```diff
@@ -261,6 +261,7 @@ Core rules are included in the `tslint` package.
     * `"check-lowercase"` enforces the rule that the first non-whitespace character of a comment must be lowercase, if applicable.
     * `"check-uppercase"` enforces the rule that the first non-whitespace character of a comment must be uppercase, if applicable.
 * `curly` enforces braces for `if`/`for`/`do`/`while` statements.
+* `cyclomatic-complexity` enforces a threshold of cyclomatic complexity.
```
JKillian: (can remove this entirely now when you merge in master)

test/rules/cyclomatic-complexity/specifiedThreshold/test.ts.lint
```diff
@@ -0,0 +1,361 @@
...
+    const condition1 = true ? "true" : "false";
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+    const condition2 = true ? "true" : "false";
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+    const condition3 = true ? "true" : "false";
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+}
+~ [The cyclomatic complexity of is higher than the threshold]
+
+class MyClass {
```
JKillian: can we also add a test for a constructor?

JKillian:
Looking good @bencoveney! Just a few small changes needed before this can merge

src/configs/latest.ts
bencoveney: I am assuming in this case that I should leave it listed but marked as false? Using typedef in recommended.ts as an example.

bencoveney:
Anticipating a CI failure, I think there is some rule additions that are causing the linting to fail on unrelated code. I can re-merge and update once they are fixed.

JKillian:
Sorry about that @bencoveney, I believe #1511 should have fixed the issue you were seeing

bencoveney:
So I believe that's all the comments so far addressed.

The only one I have deviated from slightly is the recommendation to up the minimum threshold from 1 to 5 or similar. It makes sense to disallow 1 as this would not allow any branches in the code however 5 might be slightly too high in my opinion as there would be logical values which would be prevented for example my tests use a limit of 3.

With this considered I have increased the limit from 1 to 2 which would be the lowest reasonable value AFAICT. Whether anyone would actually want to enforce a limit of 2 on real code would probably be debatable...

JKillian:
Looking good @bencoveney, and apologies for the slowness in getting back to this. I'm thinking the failure messages should display the measured complexity and the threshold complexity - this would make it easier for users to adjust their code.

bencoveney:
Error messages now take the following formats:

The function has a cyclomatic complexity of 4 which is higher than the threshold of 3

The function myNamedFunction has a cyclomatic complexity of 4 which is higher than the threshold of 3

JKillian:
Looks good, thanks @bencoveney!

PR #1466, opened 2016-08-08
Allow source imports to be evaluated similar to named imports

### LINKED ISSUE #1431: Option to evaluate ordering of import sources case-insensitively
### Bug/Enchancement Report

- __TSLint version__: `3.13.0`
- __TypeScript version__: `2.0`
- __Running TSLint via__: CLI
- __Rule__: `ordered-imports`

This rule currently seems to consider import sources case-insensitively. For instance it complains if I have something like -

```
import {X} from "foolish";
import {X} from "fooList"; // failure
```

Either we could always consider them case-insensitively, or have an option to do so.

markwongsk:
This fixes #1431

### DIFF
--- a/src/rules/orderedImportsRule.ts
+++ b/src/rules/orderedImportsRule.ts
@@ -36,13 +36,22 @@ export class Rule extends Lint.Rules.AbstractRule {
                 however you like, e.g. by first- vs. third-party or thematically.`,
         optionsDescription: Lint.Utils.dedent`
             You may set the \`"named-imports-order"\` option to control the ordering of named
-            imports (the \`{A, B, C}\` in \'import {A, B, C} from "foo"\`.)
+            imports (the \`{A, B, C}\` in \`import {A, B, C} from "foo"\`).
 
             Possible values for \`"named-imports-order"\` are:
 
             * \`"case-insensitive'\`: Correct order is \`{A, b, C}\`. (This is the default.)
             * \`"lowercase-first"\`: Correct order is \`{b, A, C}\`.
             * \`"lowercase-last"\`: Correct order is \`{A, C, b}\`.
+
+            You may set the \`"source-imports-order"\` option to control the ordering of source
+            imports (the \`"foo"\` in \`import {A, B, C} from "foo"\`).
+
+            Possible values for \`"source-imports-order"\` are:
+            * \`"case-insensitive'\`: Correct order is \`"Bar"\`, \`"baz"\`, \`"Foo"\`. (This is the default.)
+            * \`"lowercase-first"\`: Correct order is \`"baz"\`, \`"Bar"\`, \`"Foo"\`.
+            * \`"lowercase-last"\`: Correct order is \`"Bar"\`, \`"Foo"\`, \`"baz"\`.
+
         `,
         options: {
             type: "object",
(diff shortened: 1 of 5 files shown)


src/rules/orderedImportsRule.ts
```diff
@@ -51,16 +60,23 @@ export class Rule extends Lint.Rules.AbstractRule {
...
+                    type: "string",
+                    enum: ["case-insensitive", "lowercase-first", "lowercase-last"],
+                },
             },
             additionalProperties: false,
         },
-        optionExamples: ["true", '[true, {"named-imports-order": "lowercase-first"}]'],
+        optionExamples: [
+            "true",
+            '[true, {"named-imports-order": "lowercase-first", "source-imports-order": "lowercase-last"}]',
```
adidahiya: is it actually useful to set a different option for the import names vs. the import sources? I guess if we wanted to consolidate those options into one, it might be a breaking change... but this feels a bit ugly

markwongsk: My rationale was:

1. Breaking change if options are consolidated (renamed).
2. The initial implementation is actually 'lowercase-last' for source-imports and 'case-insensitive' for named-imports, which seems to indicate a use-case envisioned, at least, by the author.

I don't mind changing if you guys think that we should only have one option though.

adidahiya: what do you think @danvk?

for now I think having these separately configurable is probably fine; we'll impose a sane default

danvk: I'd be fine with separate options. Module names tend to be more uniform (mostly lowercase, hyphenated) so I doubt this setting will matter much either way.

adidahiya: @danvk that's true for external dependencies, but much less true for relative path imports of local modules

JKillian: although it's good form to name all local files in that way as well to avoid filesystem case issues and the like,

src/rules/orderedImportsRule.ts
```diff
@@ -36,13 +36,22 @@ export class Rule extends Lint.Rules.AbstractRule {
...
-            imports (the \`{A, B, C}\` in \'import {A, B, C} from "foo"\`.)
+            imports (the \`{A, B, C}\` in \`import {A, B, C} from "foo"\`).
 
             Possible values for \`"named-imports-order"\` are:
 
             * \`"case-insensitive'\`: Correct order is \`{A, b, C}\`. (This is the default.)
             * \`"lowercase-first"\`: Correct order is \`{b, A, C}\`.
             * \`"lowercase-last"\`: Correct order is \`{A, C, b}\`.
+
+            You may set the \`"source-imports-order"\` option to control the ordering of source
```
JKillian: I think this should be `import-source-order` or `import-sources-order`

src/rules/orderedImportsRule.ts
```diff
@@ -51,16 +60,23 @@ export class Rule extends Lint.Rules.AbstractRule {
...
+            "true",
+            '[true, {"named-imports-order": "lowercase-first", "source-imports-order": "lowercase-last"}]',
+        ],
         type: "style",
     };
     /* tslint:enable:object-literal-sort-keys */
 
     public static NAMED_IMPORTS_UNORDERED = "Named imports must be alphabetized.";
-    public static IMPORT_SOURCES_UNORDERED = "Import sources within a group must be alphabetized.";
+    public static SOURCE_IMPORTS_UNORDERED= "Source imports within a group must be alphabetized.";
```
JKillian: Also think this should be reverted to how it was

JKillian:
Other than naming, looks good @markwongsk! Happy to merge it after that's tweaked

src/rules/orderedImportsRule.ts
markwongsk: Done

src/rules/orderedImportsRule.ts
markwongsk: Done

JKillian:
Sweet, thanks @markwongsk!

PR #1468, opened 2016-08-09
Fix typo in whitespace rule enum

apacala:
This change fixes a typo in `check-separator` option of whitespace rule.


JKillian:
Thanks @apacala!

Issue #1096, opened 2016-04-07
Expose an API that accepts ts.SourceFile or ts.Program directly


Arnavion:
Currently `Linter` accepts a filename and file contents, then creates a `ts.CompilerHost` and `ts.Program` internally to get the `ts.SourceFile` objects. I'm integrating tslint into a build script that already has its own custom CompilerHost and Program objects to compile the project, and it would be more performant if I could get Linter to use my Program object or the SourceFile objects from it directly.

adidahiya:
cool, makes sense, we're on board. this would also bring us closer to supporting running TSLint in a browser environment (as the compiler is capable of doing).

adidahiya:
partial progress has been made on this via #1363. the `Linter` class now accepts a `ts.Program` but still requires file contents.

JKillian:
Actually, I think it might be possible now to just pass a `ts.Program` and not pass the actual textual source. You'd just pass `undefined` to the `source` parameter of `Linter` and make sure to pass something as the `program` argument. This does feel a little strange given how the API is set up now, admittedly.

cc @ScottSWu on this

ScottSWu:
:thumbsup: that works. 

With the upcoming multilinter PR, would there multiple kinds of linters for different use cases? unless there is some way to package everything in a single interface + helper functions?

JKillian:
cc @caugner here. I think it might end up being best to keep things as just one main Linter class, but i'm not sure yet

caugner:
I would suggest separating things as follows:

* The **tslint core** should *lint a source* (a string or some sort of input stream) according to a configuration and *return the (raw) result* (with references to lines/positions in that source). It should not know anything about file names or file systems (and should not maintain results between lint runs, either).
* The **tslint client** should delegate the retrieval of a source to a **tslint source provider**, pass the source(s) to the **tslint core**, aggregate the returned results and pass these results to a **tslint result formatter**.
* A **tslint result formatter** should be opened first, then receive results and should then be closed (so it can write a header and footer or buffer the content, if required).

We could then have three **tslint source providers**:

1. A file path source provider which would provide the content of a file that it reads from the file system.
2. A `ts.Program` source provider which would provide the content of a file that it reads according to the `ts.Program` parameters.
3. (new) A string source provider which would provide the source with which it was instantiated (would fix this issue).

ajafff:
Given that the file's name is important for typescript to know how to parse the file, I don't think the concept described above would work.
[...]

PR #1472, opened 2016-08-10
Call formatter once for all file results

### LINKED ISSUE #656: Formatter processed on a per file basis

caugner:
Fixes #656.

I have extracted a `MultiLinter` which allows linting a sequence of files in such a way that the behavior of `Linter` is maintained without changing its interface by using the special case where the sequence of files includes exactly one file.

At the next major version bump, the `Linter` could be replaced with the `MultiLinter`.


src/tslint-cli.ts
```diff
@@ -295,4 +296,7 @@ if (argv.project != null) {
     }
 }
 
-glob.sync(files, { ignore: argv.e }).forEach((file) => processFiles(files, program));
+files = files
+  .map((file: string) => glob.sync(file, { ignore: argv.e }))
+  .reduce((a: string[], b: string[]) => a.concat(b));
+processFiles(files, program);
```
caugner: This change was meant to go in the first commit.

src/tslint.ts
```diff
@@ -49,95 +45,25 @@ class Linter {
...
-            format: this.options.formatter,
-            output: output,
-        };
-    }
-
-    private containsRule(rules: RuleFailure[], rule: RuleFailure) {
-        return rules.some((r) => r.equals(rule));
+        const multiLinter: MultiLinter = new MultiLinter(this.options, this.program);
+        multiLinter.lint(this.fileName, this.source, this.options.configuration);
+        return multiLinter.getResult();
```
caugner: The idea is to separate instance options from (optional) per file configurations.

src/tslintMulti.ts
```diff
@@ -0,0 +1,155 @@
...
+    /**
+     * Returns a list of source file names from a TypeScript program. This includes all referenced
+     * files and excludes declaration (".d.ts") files.
+     */
+    public static getFileNames(program: ts.Program): string[] {
+        return program.getSourceFiles().map(s => s.fileName).filter(l => l.substr(-5) !== ".d.ts");
+    }
+
+    constructor(private options: IMultiLinterOptions, private program?: ts.Program) {
+        // Empty
```
caugner: Is there a better way to write this, maybe omit the block? I'm not totally familiar with TypeScript yet.

test/tsconfig.json
```diff
@@ -148,6 +148,7 @@
         "../src/test/utils.ts",
         "../src/tslint-cli.ts",
         "../src/tslint.ts",
+        "../src/tslintMulti.ts",
```
caugner: This file was changed automatically when running `grunt`.

src/tslint-cli.ts
```diff
@@ -254,11 +254,11 @@ const processFiles = (files: string[], program?: ts.Program) => {
...
 
-    if (lintResult.failureCount > 0) {
-        outputStream.write(lintResult.output, () => {
+    outputStream.write(lintResult.output, () => {
+        if (lintResult.failureCount > 0) {
             process.exit(argv.force ? 0 : 2);
-        });
-    }
+        }
+    });
```
caugner: Strictly speaking, this is out of scope, but I adopted this change from PR #1463 anyway.

JKillian:
Will review this later, but I'm up for doing a major version bump soon and this could be a kay part of it.

src/tslintMulti.ts
JKillian: Nope, this is the way to do it!

src/tslintMulti.ts
```diff
@@ -0,0 +1,155 @@
...
+    }
+
+    public lint(fileName: string, source: string, configuration: any = DEFAULT_CONFIG): void {
```
JKillian: What do you think about still allowing a configuration on the `IMultiLinterOptions` interface? Then this function would fall back to that if no configuration was provided here, and lastly fall back to the default config.

JKillian: Another thing to consider here is that `source` is actually optional if a program was provided before

JKillian: I actually have an idea that this function should return the formatted output just for the file currently being linted. This way, if being used as part of a build, lint failures could start being displayed before all files are linted, and this is great when using a formatter designed for console output. The `getResult()` method can still exist as is, to provide a way to get all results at once

caugner: I'm not particularly keen on extending the scope of this PR, but let me say this:

> this function should return the formatted output just for the file currently being linted

This is *exactly* what the (Single)Linter did so far: Lint one file and return the (formatted) result.
Adopting this behavior would exactly contradict the intention of this PR: To separate linting from returning the formatted output (once for all files linted so far).

> if being used as part of a build, lint failures could start being displayed before all files are linted

I'm afraid this would require adopting my [second proposed solution](https://github.com/palantir/tslint/issues/656#issuecomment-238471972), while this PR covers the first proposed solution. Otherwise this would be incompatible with any formatter that requires header and footer.

If you're using console output, you could still achieve this by instantiating one `MultiLinter` per file (which is what the `(Single)Linter` would do). Returning *formatted* output for each file *and also* once at the end would be redundant (and a waste of computation time, if you ask me).

> The getResult() method can still exist as is, to provide a way to get all results at once

If the result wasn't formatted, I wouldn't mind, but as it is, this would be redundant work. In general, I agree that linting should be separated from the result which should be separated from the formatting, but this is out of the scope of this PR. If we change too many things at once, we will very likely introduce bugs.

> Another thing to consider here is that source is actually optional if a program was provided before (see #1096)

You're right, it needs to be optional (unfortunately the API is already quite confusing and this will add to it).

JKillian: One tweak that should simplify this PR a small bit is that we can merge the `MultiLinter` code into `Linter` and get rid of (Single)Linter all together.

I'm not overly concerned about having streaming results by default, as you mentioned it's still possible to achieve by linting one file at a time. I don't desire to make formatters stateful to achieve this. Given this, we can leave `lint` and `getResult` as they are. 

With those things off the table, I believe the remaining major issue is the API. Between CLI options, tsconfig.json options, the `ILinterOption` interface, and "configurations", things are somewhat confusing like you stated. For the sake of this PR and keeping it limited, I think [this comment](https://github.com/palantir/tslint/pull/1472#discussion_r75405096) is still relevant, as well as the optionality of `source`. I think we should be able to leave other changes for a separate PR.

JKillian: Just saw [your comment](https://github.com/palantir/tslint/issues/1096#issuecomment-243886842) on architecture/API structure. I don't have time to think about it deeply tonight, but in general it looked good to me. I think the last paragraph of my comment above still holds for the scope of this PR though

caugner: > What do you think about still allowing a configuration on the IMultiLinterOptions interface?
> Then this function would fall back to that if no configuration was provided here, and lastly fall back to the default config.

Actually, I would prefer if we could omit the `#lint` method's `configuration` parameter at all.

Currently, this (still untyped) configuration can contain a rules directory and a list of rules, but I wonder why these parameters should be specified for one particular file? It would therefore be cleaner to attach the whole configuration including options (by merging these) to the instance of the linter. If one needs to lint one file with specific options, a separate linter instance could still be used.

caugner: > One tweak that should simplify this PR a small bit is that we can merge the MultiLinter code into Linter and get rid of (Single)Linter all together.

Would you mind if we do this in another PR? This change would probably require a couple of other changes (including to the test setup), so I tend to think it would be more safe to separate this.

JKillian: > Would you mind if we do this in another PR? This change would probably require a couple of other changes (including to the test setup), so I tend to think it would be more safe to separate this.

Yep, makes sense to me, that seems fine.

> Actually, I would prefer if we could omit the `#lint` method's `configuration` parameter at all. It would therefore be cleaner to attach the whole configuration including options (by merging these) to the instance of the linter.

I'm fine with this as well - it seems like a reasonable breaking change to include as part of 4.0. We have to be careful and consider all the ways TSLint gets used - as parts of IDE addons, as part of build tools, from the CLI, etc. But I think your ideas are solid - an IDE plugin could create a new instance of the linter for linting each file one by one as it's open; a build tool can lint all files together; etc.

caugner:
@JKillian Let me know if this PR needs any additional work done before you can merge it.

JKillian:
@caugner Sorry for the delay, merged the PR

src/tslint-cli.ts
alexeagle: This has the surprising side effect that a lint with nothing to report now prints a `0x0a` to stdout, when it used to have no output.

PR #1475, opened 2016-08-11
Require semicolon for type alias

andy-ms:
This would require all `type t = foo;` to end with a semicolon.

### DIFF
--- a/src/language/walker/syntaxWalker.ts
+++ b/src/language/walker/syntaxWalker.ts
@@ -286,6 +286,10 @@ export class SyntaxWalker {
         this.walkChildren(node);
     }
 
+    protected visitTypeAliasDeclaration(node: ts.TypeAliasDeclaration) {
+        this.walkChildren(node);
+    }
+
     protected visitTypeAssertionExpression(node: ts.TypeAssertion) {
         this.walkChildren(node);
     }
@@ -582,6 +586,10 @@ export class SyntaxWalker {
 
             case ts.SyntaxKind.TryStatement:
                 this.visitTryStatement(<ts.TryStatement> node);
+                break;
+
+            case ts.SyntaxKind.TypeAliasDeclaration:
+                this.visitTypeAliasDeclaration(<ts.TypeAliasDeclaration> node);
                 break;
 
             case ts.SyntaxKind.TypeAssertionExpression:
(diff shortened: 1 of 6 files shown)


adidahiya:
👍  thanks @andy-ms!

JKillian:
Hmm, this also is a breaking change, we should either (temporarily) revert this, or just start merging other breaking changes for 4.0

ScottSWu:
Oops, looking at the log, it looks like the new tests are just missing in the `test.ts.fix` files added with suggesting fixes.

andy-ms:
I noticed that the `.fix` files have CRLF line endings while the `.lint` files have LF line endings, at least on my checkout. I was able to get tests working by adding `results.fixesFromMarkup = results.fixesFromMarkup.replace(/\r\n/g, "\n");` to `consoleTestResultHandler` in test.ts (and by adding the missing `type t = number;` lines), but there is likely a better solution.

ScottSWu:
Interesting, it looks like `diff.diffLines` [does consider both line endings](/kpdecker/jsdiff/blob/master/src/diff/line.js), but only ignores them in the diff if `{newlineIsToken: true}` is specified. That argument isn't in the typings file, but is implemented.

JKillian:
Hmm, I'm confused why the Circle build didn't break as well, considering the `.fix` files didn't change?

Looks like the diff options you mentioned are documented though, so I'm happy to use them in our code: https://github.com/kpdecker/jsdiff#api

JKillian:
Going to revert this PR temporarily for now (https://github.com/palantir/tslint/pull/1477), but we'll get it back in for 4.0

andy-ms:
Should I make a second PR with test fixes?

JKillian:
I'm not sure if I can "re-open" a merged PR on GH, so a new one would be great!

JKillian:
#1478 created which adjusts the appropriate `.fix` files. @ScottSWu - have any idea why the CircleCI build passed on this PR in the first place?

andy-ms:
The branch I had didn't have the `.fix` files on it at all, as it was out-of-date with the latest master. So a CI that just ran the contents of my branch (and not the potential merged contents) would succeed.

JKillian:
Ohh, I didn't realize it was off an out of date master, that makes perfect sense. I guess the underlying issue here then is to get CircleCI to run the tests on the result of the would-be-merge as well.

JKillian:
Seems as if [CircleCI doesn't support this unfortunately](https://discuss.circleci.com/t/show-test-results-for-prospective-merge-of-a-github-pr/1662?u=jkillian).

Issue #516, opened 2015-07-21
Show more meaningful error when sourceFile doesn't load


adidahiya:
This error has come up a handful of times in recent issues that users have been filing (like #515, #490). Situations where this error should be made more descriptive:
- [ ] supplying files with the wrong extension, like `.ts.lint`

adidahiya:
Noting that we should prioritize this ER. Updated description with a list of places to improve the error message.

PR #1480, opened 2016-08-13
Improve an error message on invalid source file

rhysd:
When input file is not a TypeScript code, `getSourceFile()` returns`undefined` but tslint doesn't handle it properly.  As the result, tslint raises an error which is not easy to understand when invalid soruce file is passed.

```
$ tslint mocha.opts
```

It raised below error:

```
/path/to/node_modules/tslint/lib/language/walker/ruleWalker.js:18
        this.limit = this.sourceFile.getFullWidth();
```

I improved this error message by checking returned value from `getSourceFile` is valid or not. Improved error message is below:

```
Error: Invalid source code: test/mocha.opts
```

I didn't write a test for this yet because I couldn't find where I should add tests for `src/tslint.ts`.

### DIFF
--- a/src/tslint.ts
+++ b/src/tslint.ts
@@ -91,6 +91,10 @@ class Linter {
             sourceFile = getSourceFile(this.fileName, this.source);
         }
 
+        if (sourceFile === undefined) {
+            throw new Error(`Invalid source file: ${this.fileName}. Ensure that the files supplied to lint have a .ts or .tsx extension.`);
+        }
+
         // walk the code first to find all the intervals where rules are disabled
         const rulesWalker = new EnableDisableRulesWalker(sourceFile, {
             disabledIntervals: [],


adidahiya:
nice, this partially addresses #516

src/tslint.ts
```diff
@@ -91,6 +91,10 @@ class Linter {
             sourceFile = getSourceFile(this.fileName, this.source);
         }
 
+        if (sourceFile === undefined) {
+            throw new Error("Invalid source code: " + this.fileName);
```
adidahiya: can we be more descriptive here and suggest why the input is considered invalid? how about:

```ts
throw new Error(`Invalid source file "${this.fileName}". Ensure that the files supplied to lint have a .ts or .tsx extension.`);
```

rhysd: I'm not sure that tsc only checks file extension.  `sourceFile` may be `undefined` because of other than file extension.

adidahiya: last time I touched this code IIRC `tsc` _does_ care about the file extension. having the wrong extension is a common error I see in issue reports so that's why I suggested it for a small improvement to the error message

rhysd: I understood.  I'll update the commit soon. Thanks :smile:

rhysd:
@adidahiya 

I fixed the point you commented :smile:

adidahiya:
thanks @rhysd!

Issue #1099, opened 2016-04-09
false positive: no-shadowed-variable


unional:
tslint: 3.7.1

```ts
export default function applyMixins(
    derivedCtor: new (...args: any[]) => any,
    baseCtors: Array<new (...args: any[]) => any>): void {
    // ...
    });
};
```
error: shadowed variable: 'args'

bencoveney:
Have another set of reproduction steps if it helps. Presumably the same issue:

```typescript
class MyClass { };

declare type MyConstructor = new (myVariable: any) => MyClass;

function MyFunction(constructor: MyConstructor)
{
	const myVariable = 1;
	console.log(myVariable);
}
```

adidahiya:
@bencoveney which identifier is reported as shadowed in that example?

bencoveney:
Sorry, the error message is:

`Shadowed variable: 'myVariable'`

IllusionMH:
@adidahiya, @JKillian looks like parameters from `ConstrustorType` and `FunctionType` nodes shouldn't be added to scope. What do you think on this kind of fix?

adidahiya:
@IllusionMH sounds reasonable. I assume that those syntax kinds are distinct from constructor/function declarations? i.e. this would still produce a lint failure:

```ts
function MyFunction(x: any) {
  let x: string;
      ^ Shadowed variable 'x'
}
```

JKillian:
Fixed by #1482

PR #1482, opened 2016-08-15
Fixes for no-shadowed-variable rule

### LINKED ISSUE #1258: False positive for no-shadowed-variable with multiple index definitions
### Bug Report

- __TSLint version__: `3.8.1`
- __Running TSLint via__: gulp-tslint

```ts
interface IFoo {
    [i: string]: any;
    [i: number]: any;
}
```

with `tslint.json`:

```json
{
    "rules": {
        "no-shadowed-variable": true
    }
}
```

#### Actual behavior

(no-shadowed-variable) file.ts[6, 6]: shadowed variable: 'i'

#### Expected behavior

This should be fine, yes?

IllusionMH:
Fixes #1258 

Also I'm working on fix for #1099. Is it appropriate to send several fixes for one rule in one PR?

### DIFF
--- a/src/rules/noShadowedVariableRule.ts
+++ b/src/rules/noShadowedVariableRule.ts
@@ -85,8 +85,9 @@ class NoShadowedVariableWalker extends Lint.BlockScopeAwareRuleWalker<ScopeInfo,
 
     public visitParameterDeclaration(node: ts.ParameterDeclaration) {
         const isSingleParameter = node.name.kind === ts.SyntaxKind.Identifier;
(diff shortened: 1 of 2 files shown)


adidahiya:
@IllusionMH re: multiple fixes in one PR -- use your discretion. if they're related fixes then 👍

IllusionMH:
I've reverted previous commit because there is better way to skip index signatures which is similar to other cases. Also added few test cases for index signatures.

Also added fix for #1099 to this PR since it is similar to function types (was implemented previously) and index signatures.

JKillian:
Looks good to me, thanks @IllusionMH!

PR #1488, opened 2016-08-16
Object literal shorthand rule

### LINKED ISSUE #304: Enforce usage of object literal property value shorthand syntax
See: http://ariya.ofilabs.com/2013/02/es6-and-object-literal-property-value-shorthand.html

danvk:
Closes #304


palantirtech:
Thanks for your interest in [palantir/tslint](https://github.com/palantir/tslint), @danvk! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

adidahiya:
@danvk can we make the name a little more verbose, `object-literal-shorthand`? I don't really like eslint's name for this rule.

also thanks for signing the CLA. as you can see that's a new thing :)

test/utilsTests.ts
```diff
@@ -35,4 +35,15 @@ describe("Utils", () => {
         assert.deepEqual(objectify(1), {});
         assert.deepEqual(objectify({foo: 1, mar: {baz: 2}}), {foo: 1, mar: {baz: 2}});
     });
+
+    it("dedent", () => {
```
adidahiya: nice!

src/rules/objectShorthandRule.ts
```diff
@@ -0,0 +1,42 @@
...
+        ruleName: "object-shorthand",
+        description: "Enforces use of ES6 object literal shorthand when possible.",
+        options: null,
+        optionExamples: ["true"],
+        type: "style",
+    };
+    /* tslint:enable:object-literal-sort-keys */
+
+    public static LONGHAND_PROPERTY = "Expected property shorthand.";
+    public static LONGHAND_METHOD = "Expected method shorthand.";
```
adidahiya: can we make the error messages more verbose too, like so?

- "Expected property shorthand in object literal."
- "Expected method shorthand in object literal."

test/rules/object-shorthand/test.ts.lint
```diff
@@ -0,0 +1,30 @@
...
+  [y]() {},
+  z
+};
+
+const arrows = {
+  x: (y) => y  // this is OK.
+};
+
+const quotes = {
+  "foo-bar": function() {},
```
adidahiya: what about when the `function` is named? what's the expected behavior?

test/rules/object-shorthand/test.ts.lint
```diff
@@ -0,0 +1,30 @@
+const bad = {
+  w: function() {},
+  ~~~~~~~~~~~~~~~~  [method]
+  x: function *() {},
+  ~~~~~~~~~~~~~~~~~~  [method]
+  [y]: function() {},
+  ~~~~~~~~~~~~~~~~~~  [method]
+  z: z
+  ~~~~  [property]
```
adidahiya: can we add more tests to verify there are no false positives? like `a: 4`

adidahiya:
some of my comments got collapsed in an outdated diff but they're still relevant -- please click through

danvk:
Thanks for the quick review @adidahiya. I've addressed all your comments. Good catch on named function expressions. eslint allows them and now I do, too.

danvk:
At the risk of side-tracking this PR, I am curious what you guys think of https://github.com/eslint/typescript-eslint-parser as an alternative way of running lint rules like this one that aren't TS-specific. It would be great to use the AirBnb style guide with TS.

danvk:
ping @adidahiya @JKillian anything else you need from me on this?

JKillian:
Looks good, thanks @danvk! I haven't given the ESLint parser a try yet, have you had any luck with it? I believe it strips all TS-specific elements from the AST?

danvk:
@JKillian I believe that's correct. You should really take `typescript-eslint-parser` for a spin. It's quite easy to set up and gives a mixture of great and less-great results.

* For style rules (e.g. `object-curly-spacing`, `no-extra-semi`, `space-infix-ops`, `indent`, `padded-blocks`, ...) it's perfect.
* Other rules like `no-undef`, `no-unused-vars` and `dot-notation`, it's hopeless. You can import symbols exclusively to use them as types, which it will never be able to understand.
* Rules that are purely about ES6 semantics also work well, e.g. `guard-for-in`, `no-restricted-syntax` and `prefer-arrow-callback`.

I think some sort of combination of eslint+typescript-eslint-parser and tslint could be very powerful. There's no sense in replicating all the eslint rules that are independent of TS (like `space-infix-ops`). But there are many rules that eslint can't get right in a TypeScript context.

JKillian:
Wrapping ESLint in some way and using its style rules is an interesting idea. Do the style rules actually work that well though? For example, does the AST that ESLint builds include interface definitions or generic constructs, both elements that wouldn't be normally found in an ES AST?

danvk:
The rules that are purely about style (whitespace / indentation / variable naming) mostly work great.

But unless their plan is different than what I think, there will always be some rules they can't implement because they require type information, e.g. `no-undef`. tslint will always be needed for those sorts of rules.

Another nice thing about the typescript-eslint-parser approach is that you get `eslint --fix` for free.

adidahiya:
@danvk there's discussion going on here about this: https://github.com/eslint/typescript-eslint-parser/issues/61

JamesHenry:
Thanks so much for linking to that issue, @adidahiya, and thanks for trying out the parser @danvk!

Currently, that thread is definitely the best place to learn about how we feel the combination of the typescript-eslint-parser and eslint-typescript-plugin will be able to facilitate full linting of TypeScript code with ESLint.

I appreciate that it is not necessarily very clear from the existing documentation what the scope of the project is. We will definitely rectify that once we get closer to a full "beta".

Please note that there is no limit to the ambition here, so for example @danvk: 

> But unless their plan is different than what I think, there will always be some rules they can't implement

...it most definitely is! You can see that we have already debunked the false preconceptions about what is possible, and not possible, in that issue.

@JKillian:

> For example, does the AST that ESLint builds include interface definitions or generic constructs

Yes, it already does! You can see how the AST from https://github.com/eslint/typescript-eslint-parser/blob/master/tests/fixtures/typescript/basics/class-with-implements-generic-multiple.src.ts includes `typeParameters` here: https://github.com/eslint/typescript-eslint-parser/blob/master/tests/fixtures/typescript/basics/class-with-implements-generic-multiple.result.js#L107

Things are progressing very well, it is simply early on in the process.

At this stage, we are able to _parse_ all stable TypeScript (i.e. currently 1.8.10) syntax (including TSX). In addition to that, we do already have some type information recorded in the ESTree-compatible AST (matching Flow's terminology, and as linked above) and I am most of the way through adding full decorator support.

We are also currently spec'ing out how we can elegantly support things such as the full TypeScript TypeChecker being used within rules (you can already see a POC of what those kind of rules could do in that thread).

Thanks again for taking the interest in the parser, and for your work on TSLint. Everybody in the TypeScript community is going to benefit from having both TSLint and ESLint available to them!

JKillian:
Thanks for the detailed response @JamesHenry! I responded in part over in https://github.com/eslint/typescript-eslint-parser/issues/61 - it seemed easiest to keep discussion in one thread over there

PR #1498, opened 2016-08-19
Add ArrayType rule to enforce 'T[]' or 'Array<T>'

ScottSWu:
* Add ArrayType / ArrayTypeNode to SyntaxWalker

This rule doesn't use type checking to detect array aliases because the alias was likely defined purposefully.

cc: @alexeagle

### DIFF
--- a/src/rules/arrayTypeRule.ts
+++ b/src/rules/arrayTypeRule.ts
@@ -1,0 +1,76 @@
+import * as ts from "typescript";
+
+import * as Lint from "../lint";
+
+const OPTION_ARRAY = "array";
+const OPTION_GENERIC = "generic";
+
+export class Rule extends Lint.Rules.AbstractRule {
+    /* tslint:disable:object-literal-sort-keys */
+    public static metadata: Lint.IRuleMetadata = {
+        ruleName: "array-type",
+        description: "Requires using either 'T[]' or 'Array<T>' for arrays.",
+        optionsDescription: Lint.Utils.dedent`
+            One of the following arguments must be provided:
+
+            * \`"${OPTION_ARRAY}"\` enforces use of 'T[]'.
+            * \`"${OPTION_GENERIC}"\` enforces use of 'Array<T>'.`,
+        options: {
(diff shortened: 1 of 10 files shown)


palantirtech:
Thanks for your interest in [palantir/tslint](https://github.com/palantir/tslint), @ScottSWu! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

danvk:
Not necessarily something that needs to be implemented in this PR, but our style guide is a little more nuanced than either of these options. For simple (non-generic) types we use `T[]`. If the type becomes more complex, we switch to `Array<>` for clarity, e.g. `Array<{[key: string]: T}>`.

src/rules/arrayTypeRule.ts
```diff
@@ -0,0 +1,76 @@
...
+        optionsDescription: Lint.Utils.dedent`
+            One of the following arguments must be provided:
+
+            * \`"${OPTION_ARRAY}"\` enforces use of 'T[]'.
+            * \`"${OPTION_GENERIC}"\` enforces use of 'Array<T>'.`,
+        options: {
+            type: "string",
+            enum: [OPTION_ARRAY, OPTION_GENERIC],
+        },
+        optionExamples: ["[true, array]", "[true, generic]"],
```
JKillian: You can use the `OPTION_` variables here as well (no need to make a commit just to change this though)

JKillian:
Looks good to me @ScottSWu, if it looks good to @alexeagle as well I'm happy to merge this

alexeagle:
Looks great to me.

Re. the CLA:
Scott's internship ended Friday :(
But all these commits were authored as a google employee, and we've just signed your corporate CLA, so maybe that resolves it.

ScottSWu:
Oops, I've already removed the relevant email from my account (if that matters).

adidahiya:
Confirmed receipt of the corporate CLA from our end 👍

PR #1504, opened 2016-08-22
[docs] 3.15.0 docs update

JKillian:
All changes are automatically generated from the rule metadata via the `grunt docs` command.


adidahiya:
👍

Issue #1506, opened 2016-08-22
Update tslint:latest configuration for 3.15 release


adidahiya:
let's add

- `adjacent-overload-signatures`
- `object-literal-shorthand`
- `only-arrow-functions` with `allow-declarations` option enabled
- `ordered-imports` with `"import-sources-order": "case-insensitive"` option
- `arrow-parens`

JKillian:
I'm not sure I think that `arrow-parens` provides enough intrinsic value to add. Some won't like the extra keystrokes tradeoff for a little bit of visual contrast.

The other four I'm 👍 with

adidahiya:
found some more:

- `no-for-in-array`
- `object-literal-key-quotes` with `"as-needed"` option

JKillian:
Not the first, `no-for-in-array` requires typechecking. (Need to update the docs to show which rules need typechecking). The other seems good to me

PR #1508, opened 2016-08-22
v3.15.1 with updated tslint:latest config

adidahiya:
addresses #1506

### DIFF
--- a/src/configs/latest.ts
+++ b/src/configs/latest.ts
@@ -16,8 +16,14 @@
  */
 
 export const rules = {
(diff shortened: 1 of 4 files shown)


src/configs/latest.ts
```diff
@@ -16,8 +16,14 @@
  */
 
 export const rules = {
+    "adjacent-overload-signatures": true,
+    "no-for-in-array": true,
```
adidahiya: gotta remove this

PR #1510, opened 2016-08-23
variable-names: allow-aliases.

mprobst:
This option allows aliasing symbols, as long as the left hand side name
is identical to the right hand side name.

The use case is interoperating with legacy code that uses namespaces,
but still being able to abbreviate long namespaces in an ES6 module:

    const SomeClass = some.long.namespace.SomeClass;

Another use case is exposing enums or other PascalCased objects to a
template in an Angular application (or similar environments):

    class MyCtrl {
      SomeEnum = SomeEnum;
    }

### DIFF
--- a/src/rules/variableNameRule.ts
+++ b/src/rules/variableNameRule.ts
@@ -85,8 +85,12 @@ class VariableNameWalker extends Lint.RuleWalker {
     public visitBindingElement(node: ts.BindingElement) {
         if (node.name.kind === ts.SyntaxKind.Identifier) {
             const identifier = <ts.Identifier> node.name;
-            this.handleVariableNameFormat(identifier);
             this.handleVariableNameKeyword(identifier);
+            // A destructuring pattern that does not rebind an expression is always an alias, e.g. `var {Foo} = ...;`.
+            // Only check if the name is rebound (`var {Foo: bar} = ...;`).
+            if (node.parent.kind !== ts.SyntaxKind.ObjectBindingPattern || node.propertyName) {
+                this.handleVariableNameFormat(identifier, node.initializer);
+            }
         }
         super.visitBindingElement(node);
     }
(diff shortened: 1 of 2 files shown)


palantirtech:
Thanks for your interest in [palantir/tslint](https://github.com/palantir/tslint), @mprobst! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

mprobst:
Company CLA has been mailed out, as far as I know.

mprobst:
Not sure what's going on with the CI failure, that seems unrelated to my change?

JKillian:
Sorry about that, #1511 should've fixed things now if you merge in master

mprobst:
Rebased.

Re CLA, Google has signed the Corporate CLA for all Google employees (including me), and mailed it to opensource@palantir.com.

adidahiya:
Confirmed receipt of the Corporate CLA 👍

mprobst:
Friendly ping! :-)

JKillian:
Sorry for the delay @mprobst. I'm going back on forth if this should just be baked in to the rule instead of another option. I'm hesitant to add too many options, it's sort of a weird set of options as it is. What do you think?

src/rules/variableNameRule.ts
```diff
@@ -39,6 +40,7 @@ export class Rule extends Lint.Rules.AbstractRule {
               * \`"${OPTION_LEADING_UNDERSCORE}"\` allows underscores at the beginning (only has an effect if "check-format" specified)
               * \`"${OPTION_TRAILING_UNDERSCORE}"\` allows underscores at the end. (only has an effect if "check-format" specified)
               * \`"${OPTION_ALLOW_PASCAL_CASE}"\` allows PascalCase in addtion to camelCase.
+              * \`"${OPTION_ALLOW_ALIASES}"\`: does not check the names of aliases, e.g. allows \`let FooBar = FooBar;\`
             * \`"${OPTION_BAN_KEYWORDS}"\`: disallows the use of certain TypeScript keywords (\`any\`, \`Number\`, \`number\`, \`String\`,
             \`string\`, \`Boolean\`, \`boolean\`, \`undefined\`) as variable or parameter names.`,
         options: {
```
JKillian: If we keep the option, can you add it to the schema below?

mprobst:
@JKillian I think you're right. Please take another look, I removed the option and just made this the default behaviour. There's one tricky thing: if you have `let {X} = {X: 1};`, tslint now no longer warns, but I guess that's WAI? It's a corner case in any case.

JKillian:
@mprobst agreed that `let {X} = {X: 1}` doesn't need to be handled specially. Thanks for the PR!

PR #1515, opened 2016-08-25
New feature: lint js files

### LINKED ISSUE #515: Add option to allow linting of .js files
This is a question/feature request.

I am working in a project where we have TypeScript for our main code and JavaScript for tests. I would like to lint both with tslint. As TypeScript is a superset of Javascript I do not see conceptual issues with that approach. I also think that using the same linter with the same configuration can help maintaining some consistency between the two.

Unfortunately, I get the following error message when trying to run tslint on  a JavaScript file:

```
/home/xi/project/node_modules/tslint/bin/tslint-cli.js:43043
            this.limit = this.sourceFile.getFullWidth();
                                         ^
TypeError: Cannot call method 'getFullWidth' of undefined
    at EnableDisableRulesWalker.RuleWalker (/home/xi/project/node_modules/tslint/bin/tslint-cli.js:43043:42)
    at EnableDisableRulesWalker.SkippableTokenAwareRuleWalker (/home/xi/project/node_modules/tslint/bin/tslint-cli.js:43586:20)
    at new EnableDisableRulesWalker (/home/xi/project/node_modules/tslint/bin/tslint-cli.js:43707:20)
    at Linter.lint (/home/xi/project/node_modules/tslint/bin/tslint-cli.js:43976:31)
    at processFile (/home/xi/project/node_modules/tslint/bin/tslint-cli.js:44134:29)
    at Object.<anonymous> (/home/xi/project/node_modules/tslint/bin/tslint-cli.js:44144:5)
    at Module._compile (module.js:456:26)
    at Object.Module._extensions..js (module.js:474:10)
    at Module.load (module.js:356:32)
    at Function.Module._load (module.js:312:12)
```

So there is really three things I would like to know:
-   Is it a good idea to use tslint for JavaScript?
-   Can you make tslint output a more useful error message in the case described above? How can I help with that?
-   Can you make tslint work with JavaScript? How can I help with that?

YuichiNukiyama:
## Related issue
# Fixes #515
## Usage

I add _jsRule_ option to tslint.json.  We can distinguish the rules that apply to js or ts depending on this option.

```
{
    "jsRule": {
        "class-name": true
    }
}
```
## Good point
- simple implimentation
- this is _not_ breaking change
- It easier to distinguish the rules that apply to js or ts
## Bad point
- There is no way to tell each rules which file(ts or js) call them
- There is a rule that can not be applied to js (ex. arrow-parens rule can't apply js files)
- The greater the overlap in the configuration file in some cases 
## Note

Now, I test only class-name and curly rule. If you think this PR is good, I add tests for another rules.


palantirtech:
Thanks for your interest in [palantir/tslint](https://github.com/palantir/tslint), @YuichiNukiyama! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

JKillian:
@YuichiNukiyama What do you think of a system where each rule has metadata that's something like `typescriptOnly: false`? We could still keep `jsRules`, but that would only be as an override. By default TSLint would use all the settings in `rules` when linting JS files if no configuration is specified in `jsRules`.

YuichiNukiyama:
@JKillian Metadata is good idea. But I won't use all the settings in *rules* when there is no *jsRules* option. A lot of developer use other linter for linting js. So we should use unobtrusive setting. If they don't specified any settings in *jsRule*, We lint nothing.

JKillian:
@YuichiNukiyama But why would they pass their `.js` files to TSLint if they were linting them with another tool? It seems to me that if they provide `.js` files, they should get linted with the regular rules unless they opt-out by overriding a setting in `jsRules`

YuichiNukiyama:
@JKillian You've given a misunderstanding ,because of my poor english. Since the TypeScript-only rule is large, the error occurs frequently when we apply the all rule options. And now many developer use eslint to lint js files. I afraid that these developer apply the jslint to js.

JKillian:
Let's look at a concrete example to see what makes the most sense.

Say we start [with this configuration](https://github.com/palantir/tslint/blob/49e75d71178abdc778bae6d816948cbe4007e506/src/configs/recommended.ts). We then automatically skip any rules that only make sense for TS. For example, `typedef-whitespace` and `no-namespace`. We then, by default, apply the remaining rules to JS files being linted.

Are you saying that this wouldn't work well because a lot of the remaining rules wouldn't work correctly on JS files?

YuichiNukiyama:
@JKillian I add tests for all rules.
We can use following rules (sorry for my large commit).

- alignRule
- banRule
- classNameRule
- curlyRule
- eoflineRule
- fileHeaderRule
- forinRule
- indentRule
- jsdocFormatRule
- labelPositionRule
- labelUndefinedRule
- linebreakStyleRule
- maxFileLineCountRule
- maxLineLengthRule
- newParensRule
- noArgRule
- noBitwiseRule
- noConditionalAssignmentRule
- noConsecutiveBlankLinesRule
- noConsoleRule
- noConstructRule
- noDebuggerRule
- noDefaultExportRule
- noDuplicateKeyRule
- noDuplicateVariableRule
- noEmptyRule
- noEvalRule
- noForInArrayRule
- noInvalidThisRule
- noNullKeywordRule
- noReferenceRule
- noShadowedVariableRule
- noStringLiteralRule
- noSwitchCaseFallThroughRule
- noTrailingWhitespaceRule
- noUnreachableRule
- noUnsafeFinallyRule
- noUnusedExpressionRule
- noUnusedNewRule
- noUseBeforeDeclareRule
- noVarKeywordRule
- objectLiteralKeyQuotesRule
- objectLiteralShorthandRule
- objectLiteralSortKeysRule
- oneLineRule
- oneVariablePerDeclarationRule
- onlyArrowFunctionsRule
- orderedImportsRule
- quotemarkRule
- radixRule
- restrictPlusOperandsRule
- semicolonRule
- switchDefaultRule
- trailingCommaRule
- tripleEqualsRule
- useIsnanRule
- useStrictRule
- variableNameRule
- whitespaceRule

YuichiNukiyama:
@JKillian I add JS recommendations.

JKillian:
Sorry @YuichiNukiyama, things have been really busy lately! @adidahiya, do you have time to weigh in on this / take a look?

adidahiya:
@YuichiNukiyama we should have time to review this soon, I think @nchen63 can take a look. Could you resolve the merge conflicts?

YuichiNukiyama:
@adidahiya Sure. I resolved conflict.
@nchen63  Please review my PR.

src/rules/arrowParensRule.ts
```diff
@@ -29,6 +29,7 @@ export class Rule extends Lint.Rules.AbstractRule {
         options: null,
         optionExamples: ["true"],
         type: "style",
+        typescriptOnly: true,
```
nchen63: can this rule work with js also

src/rules/noRequireImportsRule.ts
```diff
@@ -29,6 +29,7 @@ export class Rule extends Lint.Rules.AbstractRule {
         options: null,
         optionExamples: ["true"],
         type: "maintainability",
+        typescriptOnly: true,
```
nchen63: js also?

src/rules/commentFormatRule.ts
```diff
@@ -47,6 +47,7 @@ export class Rule extends Lint.Rules.AbstractRule {
         },
         optionExamples: ['[true, "check-space", "check-lowercase"]'],
         type: "style",
+        typescriptOnly: true,
```
YuichiNukiyama: This Rule don't work well in js.

src/rules/arrowParensRule.ts
YuichiNukiyama: This Rule don't work well in js.

src/rules/noRequireImportsRule.ts
YuichiNukiyama: This Rule don't work well in js.

YuichiNukiyama:
@nchen63 
Some of the rules don' work well in js file. If we want use these rules in js, We should fill rules itself.  
But I think that isn't scope of this PR.

src/rules/arrowParensRule.ts
nchen63: arrow functions supported in es6

src/rules/noRequireImportsRule.ts
nchen63: also useful in es6 with the new imports

src/rules/arrowParensRule.ts
YuichiNukiyama: done.

YuichiNukiyama: done.

src/configs/recommended.ts
```diff
@@ -127,4 +127,87 @@ export const rules = {
...
+    "class-name": true,
+    "curly": true,
+    "eofline": true,
+    "forin": true,
+    "indent": [true, "spaces"],
+    "jsdoc-format": true,
+    "label-position": true,
+    "label-undefined": true,
+    "max-line-length": [true, 120],
+    "member-access": true,
```
nchen63: does this rule make sense for js?

src/configs/recommended.ts
```diff
@@ -127,4 +127,87 @@ export const rules = {
...
+    "no-shadowed-variable": true,
+    "no-string-literal": true,
+    "no-switch-case-fall-through": false,
+    "no-trailing-whitespace": true,
+    "no-unreachable": true,
+    "no-unused-expression": true,
+    "no-unused-new": true,
+    // disable this rule as it is very heavy performance-wise and not that useful
+    "no-use-before-declare": false,
+    "no-var-keyword": true,
```
nchen63: this only applies to ES6. I don't think most Javascript users will be targeting that version

src/configs/recommended.ts
adidahiya: no, it's not supported in ES.

src/configs/recommended.ts
```diff
@@ -127,4 +127,87 @@ export const rules = {
...
+    "no-console": [true,
+        "debug",
+        "info",
+        "log",
+        "time",
+        "timeEnd",
+        "trace",
+    ],
+    "no-construct": true,
+    "no-constructor-vars": false,
```
YuichiNukiyama: **parameter properties** is typescript only feature. So, I removed it.

src/configs/recommended.ts
YuichiNukiyama: I removed this.

nchen63:
Loading tslint.json that extends another config such as "recommended", won't load the base js rules. Please update `extendConfigurationFile()` to include them

nchen63:
Also update DEFAULT_CONFIG

YuichiNukiyama:
@nchen63 I fixed configration.ts. This commit pass ci. But somewhat there is no self-confidence.(I can't test on local.)

YuichiNukiyama:
Thank you for your help 😊

nchen63:
You're welcome. Thanks for the contribution!

src/rules/noDuplicateKeyRule.ts
```diff
@@ -31,6 +31,7 @@ export class Rule extends Lint.Rules.AbstractRule {
         options: null,
         optionExamples: ["true"],
         type: "functionality",
+        typescriptOnly: false,
```
adidahiya: This rule was deleted on the master branch because it is covered by the compiler. I'm going to delete it again; I don't want to maintain it _just_ for JS files.

adidahiya: #1671

nchen63: this was removed in `next`, but not in `master`

PR #1524, opened 2016-08-27
JetBrains //noinspection whitelisted in comment-format space rule

nomaed:
JetBrains IDE’s (IntelliJ IDEA, WebStorm, PhpStorm, etc.) disabling
inspections by adding `//noinspection [rulenames]` before the
expression. This violates the rule of single-line comments must start
with a space.

### DIFF
--- a/src/rules/commentFormatRule.ts
+++ b/src/rules/commentFormatRule.ts
@@ -129,12 +129,18 @@ function startsWithSpace(commentText: string) {
         return true; // comment is "//"? Technically not a violation.
     }
 
+    const commentBody = commentText.substring(2);
+
     // whitelist //#region and //#endregion
-    if ((/^#(end)?region/).test(commentText.substring(2))) {
+    if ((/^#(end)?region/).test(commentBody)) {
+        return true;
+    }
+    // whitelist JetBrains IDEs' "//noinspection ..."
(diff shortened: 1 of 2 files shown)


JKillian:
Code looks good @nomaed! Mind adding  a line to [the test here](https://github.com/palantir/tslint/blob/98d1b2962fbc019c33fbdd72f1e3a5e2ba74ac10/test/rules/comment-format/lower/test.ts.lint#L18) so we can verify that everything works correctly now?

adidahiya:
I don't think we should encode JetBrains semantics into the rule by default... can we make it a configurable option?

nomaed:
@adidahiya I wanted to add an option for this, but then I saw that there is already another exception (`//#region` and `//#endregion`) so I figured it's pretty much the same thing.
After all, `//noinspection ...` is not actually a comment, but a command, so it makes sense that these things are an exception, by default.

JKillian:
I'm sort of happy just having it baked-in. Nobody's going to accidentally have `//noinspection` comments scattered around their codebase unless they're using JetBrains stuff. Having _too_ many small configuration options can be overwhelming, seems like it's best to just "do the right thing" by default when possible.

adidahiya:
Alright, cool sounds good 👍

JKillian:
Thanks @nomaed!

Issue #1512, opened 2016-08-23
object-literal-sort-keys rule ignores quoted keys


JKillian:
I believe [this line](https://github.com/palantir/tslint/blob/e0bd8c9867bc4a420d867a9535e51c1d47b25d41/src/rules/objectLiteralSortKeysRule.ts#L65) is the culprit:

```
if (keyNode.kind === ts.SyntaxKind.Identifier) {
```

Fairly sure this causes the key to get ignored if it's a string and not an Identifier.

PR #1529, opened 2016-08-30
Fixed object-literal-sort-keys rule ignores quoted keys #1512

bolatovumar:
New pull request as I deleted the old repository and branch that I was issuing a pull request from originally.


adidahiya:
👍

adidahiya:
are there tests for this fix?

adidahiya:
nice, thanks @bolatovumar!

PR #1539, opened 2016-09-05
Add array-simple option to array-type rule

### LINKED ISSUE #1526: array-type rule requires array in places that are less readable
### Bug Report

- __TSLint version__: HEAD

```ts
Array<{length: number, deferred: Deferred<any[]>}>
Array<Document|string>
Array<(event: Event) => void>
TableRow<Array<string|number|null>>
```

```json
(include if relevant)
{
  "rules": {
    "array-type": [true, "array"]
  }
}
```

#### Actual behavior

Replaces with
```ts
{length: number, deferred: Deferred<any[]>}[]
(Document|string)[]
((event: Event) => void)[]
TableRow<(string|number|null)[]>
```

#### Expected behavior

These are more readable in the original form with `Array<T>`. 

@mprobst explains:
precisely, if your type in the grammar (https://github.com/Microsoft/TypeScript/blob/master/doc/spec.md#a1-types) is not just a PredefinedType (number, string, etc) or a TypeReference (some.namepsace.MyFunkyType), then you should be allowed to use Array<MyType>.

ScottSWu:
Only use '[]' notation for 'simple' types.

Fixes #1526

### DIFF
--- a/src/rules/arrayTypeRule.ts
+++ b/src/rules/arrayTypeRule.ts
@@ -14,9 +14,9 @@ export class Rule extends Lint.Rules.AbstractRule {
         optionsDescription: Lint.Utils.dedent`
             One of the following arguments must be provided:
 
-            * \`"${OPTION_ARRAY}"\` enforces use of 'T[]'.
-            * \`"${OPTION_GENERIC}"\` enforces use of 'Array<T>'.
-            * \`"${OPTION_ARRAY_SIMPLE}"\` enforces use of 'T[]' if 'T' is a simple type (primitive or type reference).`,
+            * \`"${OPTION_ARRAY}"\` enforces use of \`T[]\` for all types T.
+            * \`"${OPTION_GENERIC}"\` enforces use of \`Array<T>\` for all types T.
+            * \`"${OPTION_ARRAY_SIMPLE}"\` enforces use of \`T[]\` if \`T\` is a simple type (primitive or type reference).`,
         options: {
             type: "string",
             enum: [OPTION_ARRAY, OPTION_GENERIC, OPTION_ARRAY_SIMPLE],
@@ -29,7 +29,7 @@ export class Rule extends Lint.Rules.AbstractRule {
     public static FAILURE_STRING_ARRAY = "Array type using 'Array<T>' is forbidden. Use 'T[]' instead.";
     public static FAILURE_STRING_GENERIC = "Array type using 'T[]' is forbidden. Use 'Array<T>' instead.";
     public static FAILURE_STRING_ARRAY_SIMPLE = "Array type using 'Array<T>' is forbidden for simple types. Use 'T[]' instead.";
-    public static FAILURE_STRING_GENERIC_SIMPLE = "Array type using 'T[]' is forbidden for simple types. Use 'Array<T>' instead.";
+    public static FAILURE_STRING_GENERIC_SIMPLE = "Array type using 'T[]' is forbidden for non-simple types. Use 'Array<T>' instead.";
 
     public apply(sourceFile: ts.SourceFile): Lint.RuleFailure[] {
         const alignWalker = new ArrayTypeWalker(sourceFile, this.getOptions());
@@ -85,8 +85,6 @@ class ArrayTypeWalker extends Lint.RuleWalker {
     }
 
     private isSimpleType(nodeType: ts.TypeNode) {
-        // tslint:disable-next-line:no-console
-        console.log(ts.SyntaxKind[nodeType.kind]);
         switch (nodeType.kind) {
             case ts.SyntaxKind.AnyKeyword:
             case ts.SyntaxKind.ArrayType:
@@ -95,9 +93,17 @@ class ArrayTypeWalker extends Lint.RuleWalker {
             case ts.SyntaxKind.NumberKeyword:
             case ts.SyntaxKind.StringKeyword:
             case ts.SyntaxKind.SymbolKeyword:
-            case ts.SyntaxKind.TypeReference:
             case ts.SyntaxKind.VoidKeyword:
                 return true;
+            case ts.SyntaxKind.TypeReference:
+                // TypeReferences must be non-generic or be another Array with a simple type
+                const node = nodeType as ts.TypeReferenceNode;
+                const typeArgs = node.typeArguments;
+                if (!typeArgs || typeArgs.length === 0 || node.typeName.getText() === "Array" && this.isSimpleType(typeArgs[0])) {
+                    return true;
+                } else {
+                    return false;
+                }
             default:
                 return false;
         }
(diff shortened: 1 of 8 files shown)


src/rules/arrayTypeRule.ts
```diff
@@ -14,18 +15,21 @@ export class Rule extends Lint.Rules.AbstractRule {
...
-        optionExamples: ["[true, array]", "[true, generic]"],
+        optionExamples: [`[true, ${OPTION_ARRAY}]`, `[true, ${OPTION_GENERIC}]`, `[true, ${OPTION_ARRAY_SIMPLE}]`],
         type: "style",
     };
     /* tslint:enable:object-literal-sort-keys */
 
     public static FAILURE_STRING_ARRAY = "Array type using 'Array<T>' is forbidden. Use 'T[]' instead.";
     public static FAILURE_STRING_GENERIC = "Array type using 'T[]' is forbidden. Use 'Array<T>' instead.";
+    public static FAILURE_STRING_ARRAY_SIMPLE = "Array type using 'Array<T>' is forbidden for simple types. Use 'T[]' instead.";
+    public static FAILURE_STRING_GENERIC_SIMPLE = "Array type using 'T[]' is forbidden for simple types. Use 'Array<T>' instead.";
```
alexeagle: this one should say "for non-simple types"

src/rules/arrayTypeRule.ts
```diff
@@ -14,18 +15,21 @@ export class Rule extends Lint.Rules.AbstractRule {
             One of the following arguments must be provided:
 
             * \`"${OPTION_ARRAY}"\` enforces use of 'T[]'.
-            * \`"${OPTION_GENERIC}"\` enforces use of 'Array<T>'.`,
+            * \`"${OPTION_GENERIC}"\` enforces use of 'Array<T>'.
```
alexeagle: maybe clarify these two with "for all types T"

src/rules/arrayTypeRule.ts
```diff
@@ -14,18 +15,21 @@ export class Rule extends Lint.Rules.AbstractRule {
...
+            * \`"${OPTION_ARRAY_SIMPLE}"\` enforces use of 'T[]' if 'T' is a simple type (primitive or type reference).`,
```
alexeagle: nit: should these single quotes be backticks? for when this doc is presented in markdown?

src/rules/arrayTypeRule.ts
```diff
@@ -35,46 +39,67 @@ export class Rule extends Lint.Rules.AbstractRule {
...
             }
-            this.addFailure(this.createFailure(node.getStart(), node.getWidth(), Rule.FAILURE_STRING_ARRAY, fix));
         }
 
         super.visitTypeReference(node);
     }
+
+    private isSimpleType(nodeType: ts.TypeNode) {
+        // tslint:disable-next-line:no-console
+        console.log(ts.SyntaxKind[nodeType.kind]);
```
alexeagle: should this still be here? does it spam the console for tslint CLI?

test/rules/array-type/array-simple/test.ts.lint
```diff
@@ -0,0 +1,50 @@
...
+type Arr<T> = Array<T>;
+              ~~~~~~~~  [Array type using 'Array<T>' is forbidden for simple types. Use 'T[]' instead.]
+
+// Ignore user defined aliases
+let yyyy: Arr<Array<Arr<string>>[]> = [[[["2"]]]];
+              ~~~~~~~~~~~~~~~~~~                   [Array type using 'Array<T>' is forbidden for simple types. Use 'T[]' instead.]
+
+interface ArrayClass<T> {
+    foo: Array<T>;
+         ~~~~~~~~  [Array type using 'Array<T>' is forbidden for simple types. Use 'T[]' instead.]
```
alexeagle: might be nice to also have a dotted name, like
`namespace a {
  class B {}
}`
`Array<a.B>` should be disallowed

alexeagle:
thanks, Scott! can't wait to tsunami this one again.

JKillian:
Thanks for reviewing this @alexeagle! Will merge after corrections

alexeagle:
Just ran this fix across google's codebase.

I noticed that
`Array<FilterItem<T>>` => `FilterItem<T>[]` which I think we don't mean to be a simple type?
especially in a case like
`DataValue<Array<DataValue<any>>>` => `DataValue<DataValue<any>[]>`
where the `Array` is just one of several collection generics, and it gets such different treatment.

ScottSWu:
Hmm, I marked all TypeReferences, so I'm guessing it should only be TypeReferences that are not generics.

In that case, should `Array<Array<number>` still be `number[][]`?

alexeagle:
`number[][]` seems like the right representation, but then what about if I have a `List` type in my app?

`List<number[]>` vs `List<Array<number>>`
 and  
`List<number>[]` vs.  `Array<List<number>>`

If we simply apply to non-generic TypeReferences we get
`Array<number[]>`   (note you'd have to be aware that `number[]` is actually a generic)
`List<number[]>`
`Array<List<number>>`

I think the real readability issue is that the syntax sugar is postfix, so if it follows a very long expression, like
`let longThing: MultiMap<Set<A>, NonIntersectingPair<B>, Comparator<A, B>>[];`
You would have liked to read the `Array` bit at the beginning because by the time you encounter it, you think you've already understood the type.

So on that grounds, I think treating generics as a "non-simple" type is the right thing to do, even if we lose a bit of sugar for `number[][]`

alexeagle:
@evmar comments?

JKillian:
If I had to vote in the cases you gave above, I prefer the following (bolded) options:

`Array<Array<number>` vs **`number[][]`**
**`List<number[]>`** vs `List<Array<number>>`
`List<number>[]` vs.  **`Array<List<number>>`**

alexeagle:
@ScottSWu would you like me to take this one over?

ScottSWu:
Sorry - PTAL

The new commit follows @JKillian's preferences with `number[][]`. A type is simple if it is non-generic or is an Array with another simple type argument.

alexeagle:
Thanks @ScottSWu I ran it across all our code again. Looks really good!

Found bugs:
1)
`pointsLineData as any as[number, number][]` became `pointsLineData as any asArray<[number, number]>` (syntax error)

2)
```
interface EAP {
  '.b.SpawnInfo.spawnInfo': {argument: string[];};
}
```
became
```
interface EAP {
  '.b.SpawnInfo.spawnInfo': {argument: [];};
[]
```

ScottSWu:
The first is fixable, but the second looks weird - I'll add that as a test case, but could that be a syncing issue?

alexeagle:
I doubt it's sync in this case (I own the code that was mangled and haven't changed it) but as long as there's a test case, tslint is not responsible for what happens :)

ScottSWu:
Added both test cases. For the first bug, I added a space for the `as` keyword only - couldn't think of other situations where a type follows an alphanumeric character.

alexeagle:
LGTM thanks a lot Scott! Looking forward to turning this one on.

alexeagle:
ping @JKillian for LGTM

adidahiya:
thanks @ScottSWu!

Issue #1121, opened 2016-04-12
false positive: jsdoc-format


unional:
TSLint: 3.7.3

```js
  /*********
   * Chain *
   ********/
```
Not really JSDoc.... :rose:

nomaed:
I agree.
I often use 
```
/*****************************************************************/
```
for visual clarity to separate various sections of classes (for example, a separator between properties and methods, or publics and privates).

I keep getting error although these aren't JSDocs.

I suggest that detection of a JSDoc should check for any whitespace after the `/**` sequence.
I made this change in my local build, and I will add a PR for it.

JKillian:
Fixed by #1543

PR #1543, opened 2016-09-08
JSDoc rule should not block arbitrary comments like /********/

nomaed:
# Problem
There are 2 prime examples of comments that confuse tslint JSDoc detection regex:
A separator:
```typescript
class MyClass {
    public prop1;
    public prop2;
    /********************************************/
    constructor() { ... }
    public method1() { ... }
    public method2() { ... }
    /********************************************/
    private internal1() { ... }
}
```
and
```typescript
/**************
 * my-file.ts *
 **************/
```

These can be used as visual separator or guides, and are definitely not JSDocs.
Was also reported in issue #1121 several months ago.

# Solution
Instead of matching any `/^\s*\/\*\*/`, I suggest that when there are more than 2 asterisks in a row, it should not be tested as a JSDoc comment.
Thus, the regex is updated to `/^\s*\/\*\*([^*]|$)/` which would accept `/**` followed by anything but another `*`.

## matched as jsdoc
```typescript
/** (space) */ // valid, empty
/**text */ // invalid
/**_ */ // invalid
/** (new-line) // valid
 */
```

## not jsdoc
```typescript
/***/
/*/
/*******/
/******
*******/
```

### DIFF
--- a/src/rules/jsdocFormatRule.ts
+++ b/src/rules/jsdocFormatRule.ts
@@ -76,8 +76,8 @@ class JsdocWalker extends Lint.SkippableTokenAwareRuleWalker {
         const firstLine = lines[0];
         let jsdocPosition = currentPosition;
 
-        // regex is: start of string, followed by any amount of whitespace, followed by /**
-        const isJsdocMatch = firstLine.match(/^\s*\/\*\*[^*]/);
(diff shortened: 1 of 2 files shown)


JKillian:
This seems reasonable, thanks @nomaed!

PR #1548, opened 2016-09-12
[options] adding exceptions to max-line-length rule (closes #1495)

### LINKED ISSUE #1495: rule(max-line-length) cannot be disabled within a block comment
### Bug Report
- **TSLint version**: `3.14.0`
- **TypeScript version**: `2.0.0`
- **Running TSLint via**: `gulp (via gulp-tslint)`

``` ts
/*
  // tslint:disable-next-line:max-line-length
  const lengthy = 'Lorem ipsum dolor sit amet, consectetur adipiscing elit. Phasellus et mauris turpis. Phasellus tincidunt...';
*/
```

``` json
{
}
```
#### Actual behavior

```
(max-line-length) path/to/file.ts[123, 1]: Exceeds maximum line length of 120
```
#### Expected behavior

Rule is disabled and there is no error.

jmlopez-rod:
Adding `ignoreUrls`, `ignoreImports`, and `ignoreComments` options to
the rule.

### DIFF
--- a/src/rules/maxLineLengthRule.ts
+++ b/src/rules/maxLineLengthRule.ts
@@ -19,6 +19,8 @@ import * as ts from "typescript";
 
 import * as Lint from "../lint";
 
+import { IDisabledInterval } from "../language/rule/rule";
+
 export class Rule extends Lint.Rules.AbstractRule {
     /* tslint:disable:object-literal-sort-keys */
     public static metadata: Lint.IRuleMetadata = {
@@ -38,6 +40,8 @@ export class Rule extends Lint.Rules.AbstractRule {
     };
     /* tslint:enable:object-literal-sort-keys */
 
+    public static URL_REGEXP = /[^:/?#]:\/\/[^?#]/;
+
     public static FAILURE_STRING_FACTORY = (lineLimit: number) => {
         return `Exceeds maximum line length of ${lineLimit}`;
     };
@@ -53,8 +57,9 @@ export class Rule extends Lint.Rules.AbstractRule {
     }
 
     public apply(sourceFile: ts.SourceFile): Lint.RuleFailure[] {
+        const ruleArgs = this.getOptions().ruleArguments;
         const ruleFailures: Lint.RuleFailure[] = [];
-        const lineLimit = this.getOptions().ruleArguments[0];
+        const lineLimit = ruleArgs[0];
         const lineStarts = sourceFile.getLineStarts();
         const errorString = Rule.FAILURE_STRING_FACTORY(lineLimit);
         const disabledIntervals = this.getOptions().disabledIntervals;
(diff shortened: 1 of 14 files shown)


test/rules/max-line-length/ignore-imports/test.ts.lint
```diff
@@ -0,0 +1,29 @@
...
+import {
+    obj1,
+    obj2,
+    obj3,
+    obj4,
+} from 'my-favorite-module/with/lots/of/deep/nested/modules';
+
+import {
+    obj1, obj2, obj3, obj4, just, trying, to, be, a, rebel, here
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~ [Exceeds maximum line length of 50]
```
JKillian: shouldn't this be okay?

src/rules/maxLineLengthRule.ts
```diff
@@ -19,6 +19,8 @@ import * as ts from "typescript";
 
 import * as Lint from "../lint";
 
+import { IDisabledInterval } from "../language/rule/rule";
+
 export class Rule extends Lint.Rules.AbstractRule {
```
JKillian: Can we declare the rule options as constants, and then use those constants in the rule metadata and in the code? 

(For example, [similar to this](https://github.com/palantir/tslint/blob/master/src/rules/tripleEqualsRule.ts#L22L51))

src/rules/maxLineLengthRule.ts
```diff
@@ -19,6 +19,8 @@ import * as ts from "typescript";
...
     /* tslint:disable:object-literal-sort-keys */
     public static metadata: Lint.IRuleMetadata = {
```
JKillian: Let's update this as well - this info is used to generate the docs website when a release is published

src/rules/maxLineLengthRule.ts
```diff
@@ -71,11 +76,73 @@ export class Rule extends Lint.Rules.AbstractRule {
                 // (and thus we are not actually over the limit).
                 const ruleFailure = new Lint.RuleFailure(sourceFile, from, to - 1, errorString, this.getOptions().ruleName);
                 if (!Lint.doesIntersect(ruleFailure, disabledIntervals)) {
-                    ruleFailures.push(ruleFailure);
+                    let report = true;
+                    if (ruleArgs.indexOf("ignoreUrls") > -1) {
```
JKillian: You can just use `this.hasOption(OPTION_IGNORE_URLS)` here

test/rules/max-line-length/ignore-imports/test.ts.lint
jmlopez-rod: This shouldn't be ignored. If they choose to do

```
import { obj1, obj2, obj3, obj4, just, trying, to, be, a, rebel, here } from '...';
```

then that is ok, but the moment they decide to break the line then we have options to prevent exceeding the line limit. I usually prefer to place each import on its own line when using this form:


```
import {
  obj1,
  obj2,
  obj3,
  obj4,
  just,
  trying,
  to,
  be,
  a,
  rebel,
  here,
} from '...';
```

I decided to go with this approach because in the past I have seen a really long list increasing on the same line and it gets hard to see the changes. So really the rule should just apply to the string part of the import which we don't have much control over.

src/rules/maxLineLengthRule.ts
jmlopez-rod: Ah, that can happen. Got it, will apply the change.

src/rules/maxLineLengthRule.ts
jmlopez-rod: I think this can only be used on a walker. I was going to move the logic in the `apply` function to the walker so I'll be able to do this there.

jmlopez-rod:
@JKillian , I have refactored the rule so that it may be easier to add other rules in the future. That is, in case we want to add other eslint options that I'm not adding such as disabling the rule in strings and templates. The only other thing I added was support to ignore a user defined pattern. This is particularly useful when using `require` (another form of imports). I have added tests for this as well.

I hope the update to the documentation was done properly.

jmlopez-rod:
@JKillian I just noticed that the current code as it stands including my PR does not cover the case when users use tabs instead of spaces. That is we are assuming that `\t` counts for 1 character. In eslint they have a `tabWidth` option which defaults to 4 to properly account for tab use. I bring this up mainly because I would like to push for getting rid of the arguments and instead just use an object to declare the exceptions. Obviously since the original rule only has one argument we must leave it as it. But once we add `ignore-urls`, `ignore-imports` and `ignore-comments` as arguments we will have to keep them and the rule schema will just get larger. Thoughts?

adidahiya:
@JKillian can you take another look this week?

jmlopez-rod:
Hi guys, this PR has diverted a bit. I have released this rule in the tslint-eslint-rules repo (https://github.com/buzinas/tslint-eslint-rules/blob/master/src/docs/rules/terMaxLenRule.md). The rule does the same as the one in eslint. I'm closing this PR to help you clean up. If you guys want  to make improvements to the rule you peek into the source file. Thank you for your comments @JKillian , I learned a few things things from them.

JKillian:
Apologies @jmlopez-rod for not not fully seeing this through the whole way, thanks for the understanding! I haven't had much extra time to work on TSLint recently (although @nchen63 has been doing a great job!). Glad you got things merged over in `tslint-eslint-rules`!

PR #1549, opened 2016-09-12
New rule - banLibraryImport

use-strict:
This rule checks for require/import statements with a user-specified list of banned strings. This is useful to prevent accidental inclusions of entire libraries, when tree shaking is not possible. E.g. importing "material-ui" instead of "material-ui/lib/something".

### DIFF
--- a/src/rules/banLibraryImportRule.ts
+++ b/src/rules/banLibraryImportRule.ts
@@ -1,0 +1,112 @@
+import * as ts from "typescript";
+import * as Lint from 'tslint/lib/lint';
+
+export class Rule extends Lint.Rules.AbstractRule {
+    public static FAILURE_STRING = "require(\"{0}\") will result in the entire \"{0}\" library being included. If this is intentional, you can suppress the rule with a tslint:ignore directive.";
+
+    public isEnabled(): boolean {
+        if (super.isEnabled()) {
+            return !!this.getOptions().ruleArguments.length;
+        }
+
+        return false;
+    }
+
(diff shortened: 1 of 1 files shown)


palantirtech:
Thanks for your interest in [palantir/tslint](https://github.com/palantir/tslint), @use-strict! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

JKillian:
@use-strict What if you want to ban all imports of `"material-ui"` but also `"material-ui/anything"`?

use-strict:
@JKillian , you just specify the rule as a blacklist: e.g. `[true, "material-ui", "material-ui/anything"]` . That has to be an exact match though, no wildcards. I assume you are asking about the latter possibility. Haven't thought about that yet.

nchen63:
what do you think about renaming this to `import-library-blacklist`? `ban-library-import` implies banning all libraries. The PR is missing a few things like:

* license header
* Rule.metadata
* tests

SimonSchick:
@use-strict would you mind me forking your fork and implement the changes requested by the palantir guys? We want to use this feature productively and want to avoid having an additional dependency for just this rule in the future.

use-strict:
@SimonSchick , go ahead, thanks :)

SimonSchick:
@use-strict I will do so now, since I will fork your fork you still have to accept the palantir CLA!

adidahiya:
closing in favor of #1841, which just merged

PR #1553, opened 2016-09-13
print nothing when no failures

alexeagle:
https://github.com/palantir/tslint/pull/1472/commits/8fd8723586480f2396574072e5b467f8c096240e introduced a change
where the proseFormatter writes a single newline to stdout when there are no failures

### DIFF
--- a/src/formatters/proseFormatter.ts
+++ b/src/formatters/proseFormatter.ts
@@ -20,7 +20,7 @@ import {RuleFailure} from "../language/rule/rule";
 
 export class Formatter extends AbstractFormatter {
     public format(failures: RuleFailure[]): string {
-        if (!failures.length) {
+        if (failures.length === 0) {
             return "";
         }
 
--- a/src/formatters/proseFormatter.ts
+++ b/src/formatters/proseFormatter.ts
@@ -20,6 +20,10 @@ import {RuleFailure} from "../language/rule/rule";
 
 export class Formatter extends AbstractFormatter {
     public format(failures: RuleFailure[]): string {
+        if (!failures.length) {
+            return "";
+        }
+
         const outputLines = failures.map((failure: RuleFailure) => {
             const fileName = failure.getFileName();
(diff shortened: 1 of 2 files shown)


palantirtech:

JKillian:
Wanted to note that there's been some discussion on this before: https://github.com/palantir/tslint/pull/1463#issuecomment-238346484

@caugner was of the opinion that a newline should be written when there are no failures

src/formatters/proseFormatter.ts
```diff
@@ -20,6 +20,10 @@ import {RuleFailure} from "../language/rule/rule";
 
 export class Formatter extends AbstractFormatter {
     public format(failures: RuleFailure[]): string {
+        if (!failures.length) {
```
JKillian: TSLint norms is to do an explicit comparison, `(failures.length === 0)`.

Just curious, do you have that feature enabled that lets TSLint maintainers make modifications to your PR branch? I haven't gotten a chance to try it out yet, and something like this is the ideal scenario

JKillian:
However, I'm leaning towards agreeing with you @alexeagle. Obviously, some formatters will always have output - for example, a JSON formatter would always output `{}`. However, it doesn't make sense for a formatter, like `prose` to emit a blank line. Similarly, a program like grep doesn't print a blank line when there are no results found.

alexeagle:
The reason I found this is that we have some internal tests for our tslint integration. We assert on the exit code by sensing whether there is any output, this way we don't need some other golden file saying what the exit code should be.
I could trim() that (in bash) before that logic, but I figured that non-empty output in the success case was probably an accident.

alexeagle:
Allow edits from maintainers. is checked - so give that a try?

JKillian:
Hmm, can't get it working for some reason. I'm going to try again tomorrow, but if I can't figure it out I'll merge this, no need for any changes on your end

PR #1557, opened 2016-09-18
Stylish Formatter: output all file names

nomaed:
Currently the stylish formatter prints *only* the first fileName, and then a list of lint errors from all files without specifying to which file they belong.
After this fix, each time a new file is reached, a new-line will be added, and then the name of the file, so each group of lint failures will be group with their fileName.

### DIFF
--- a/src/formatters/stylishFormatter.ts
+++ b/src/formatters/stylishFormatter.ts
@@ -26,20 +26,18 @@ export class Formatter extends AbstractFormatter {
             return "\n";
         }
 
-        const outputLines: Array<string> = [];
-
-        const positionMaxSize = this.getPositionMaxSize(failures);
-        const ruleMaxSize     = this.getRuleMaxSize(failures);
+        const outputLines: string[] = [];
+        const positionMaxSize       = this.getPositionMaxSize(failures);
+        const ruleMaxSize           = this.getRuleMaxSize(failures);
 
         let currentFile: string;
 
         for (const failure of failures) {
-            // File name for each new file
             const fileName = failure.getFileName();
+
+            // Output the name of each file once
             if (currentFile !== fileName) {
-                if (currentFile) {
-                    outputLines.push("");
-                }
+                outputLines.push("");
                 outputLines.push(fileName);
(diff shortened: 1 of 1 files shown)


src/formatters/stylishFormatter.ts
```diff
@@ -26,15 +26,24 @@ export class Formatter extends AbstractFormatter {
             return "\n";
         }
 
-        const fileName        = failures[0].getFileName();
+        const outputLines: Array<string> = [];
```
adidahiya: please format the type def as `string[]`

src/formatters/stylishFormatter.ts
```diff
@@ -26,15 +26,24 @@ export class Formatter extends AbstractFormatter {
...
-            fileName,
-        ];
+        let currentFile: string;
 
         for (const failure of failures) {
+            // File name for each new file
+            const fileName = failure.getFileName();
+            if (currentFile !== fileName) {
+                if (currentFile) {
+                    outputLines.push("");
```
adidahiya: does this just happen once, at the beginning of the lint output? if so, I think this can be refactored out of the loop.

nomaed: It happens every time a new file is reached.
The idea is that whenever we start a new file, we add an empty line before it to separate the output from the previous list of lint errors.
I'll refactor it out of the loop by adding the first file-name before it.

nomaed: @adidahiya Actually, if I move the first file out of the loop, I still need to add a test for whether we're at the first iteration or not, so it won't really make it cleaner.

adidahiya: I probably should've commented on L40 -- why would `currentFile` be undefined multiple times?

nomaed: It wouldn't be. It would be only for the first file, which is the case I'm trying to catch here, to avoid adding an empty line.
It's either that, or adding the line even at the first time, and then unshifting it out of the outputLines before returning it. Although then there is a need to checked whether the array is not empty and etc.

PR #1558, opened 2016-09-18
Add new formatter: fileslist

nomaed:
This formatter does not output any lint error details. Instead, it lists all of the files that had lint errors.
This is useful for creating lists of files that need to be taken care
of, or for bulk-opening all of the files in an editor from CLI.

### Prose
```shell
$ tslint --project tsconfig.json
/Users/dude/crc-calc/src/lib/crcData.ts[139, 4]: file should end with a newline
/Users/dude/crc-calc/src/lib/crcData.ts[1, 14]: variable name must be in camelcase or uppercase
/Users/dude/crc-calc/src/lib/crc.ts[31, 5]: comment must start with a space
/Users/dude/crc-calc/src/lib/crc.ts[35, 2]: file should end with a newline
/Users/dude/crc-calc/src/lib/crc.ts[17, 7]: Forbidden bitwise operation
/Users/dude/crc-calc/src/lib/crc.ts[20, 14]: Forbidden bitwise operation
/Users/dude/crc-calc/src/lib/crc.ts[20, 31]: Forbidden bitwise operation
/Users/dude/crc-calc/src/lib/crc.ts[12, 3]: Multiple variable declarations in the same statement are forbidden
src/crc-calc.ts[14, 1]: Consecutive blank lines are forbidden
src/crc-calc.ts[2, 1]: <reference> is not allowed, use imports
```

### Files-list
```shell
$ tslint --project tsconfig.json -t fileslist
/Users/dude/crc-calc/src/lib/crcData.ts
/Users/dude/crc-calc/src/lib/crc.ts
src/crc-calc.ts
```

## Usage examples:
Opening all files that require fixing in an editor:
```shell
$ atom $(tslint --project tsconfig.json -t fileslist)
  (or)
$ vi $(tslint --project tsconfig.json -t fileslist)
```
When using Perforce RCS, checking out all files for edit:
```shell
$ tslint --project tsconfig.json -t fileslist | xargs p4 edit
```

PR #1563, opened 2016-09-21
Doesn't print \n if no failures

nomaed:
Addresses review [#875140](https://github.com/palantir/tslint/pull/1558#discussion_r79750793) in PR #1558 for compatibility with discussion item #1553


adidahiya:
thanks @nomaed!

PR #1564, opened 2016-09-21
Documentation for formatters

### LINKED ISSUE #1378: Better documentation for built-in formatters
We don't really document them anywhere.

related: #1376

bencoveney:
Resolves #1378.

- Added IFormatterMetadata and populated for each formatter.
- Altered BuildDocs.ts script to generate docs for formatters as well as rules.
- Added formatter pages to `/docs` site (layout, navigation etc).

I have also regenerated the Jekyll pages to show the formatter output. This has also created a lot of changes in the docs pages for Rules which I have left in but can remove them if it would be better to keep the PR tidy.

Pic of resulting docs pages attached.

2016-09-21 20_28_58-formatter_ json - opera


JKillian:
@bencoveney it'd be great if you could remove the rules related changes as you suggested. Will help for the review of this PR, but other reason is that those changes aren't actually available in the latest release of TSLint yet, so better not to put them on the live docs site!

bencoveney:
I have removed the changes to the rules documentation and added meta-data for new rules.

Issue #822, opened 2015-11-25
Improve rule curly with one-line if without braces


tkrotoff:
Having one-line if and else without braces is common:

``` TypeScript
// One-line without braces
if (foo) doSomething();
else doSomethingElse();

// Multi-line with braces
if (foo) {
  doSomething1();
  doSomething2();
} else {
  doSomethingElse1();
  doSomethingElse2();
}
```

ESLint allows this with [`"curly": [2, "multi-line"]`](http://eslint.org/docs/rules/curly.html). Would be nice for TSLint to support it.

NeverwinterMoon:
This would be nice to have. We've been using one-liners without curly braces for years across multiple projects. Now, after moving to TS and enabling TSLint, this aforementioned problem arose.

devrelm:
I'm starting to implement this right now. I'd like to clear up the terminology a little bit and suggest these options: `same-line-multi-line`, `same-line-single-line`, `next-line-multi-line`, and `next-line-single-line`. An additional `all` option will be an alias for all four, which should mirror existing behavior. The idea behind this is that the rule is only enforced on a single statement coming after a control-flow statement, so the rule's options are coined to describe the statement itself, where that statement might reside and how many lines the statement covers.

Note that the rule doesn't specify where it can reside or how many lines a statement is allowed use — that would be handled by another rule. Instead, it simply dictates whether or not it should have curly braces in some particular case. I got as granular as possible so that later we can add something like a `same-line` option that is an alias for both `same-line-single-line` and `same-line-multi-line`.

#### In Scope:

* `{ "curly": true }` - existing behavior
* `{ "curly": [true, "all"] }` - existing behavior
* `{ "curly": [true, "same-line-single-line"] }` - only single-line if-statement expressions get brackets:
    ```
    // Good
    if (foo) { doSomething(); }
    if (foo) doSomething(
        someParam,
        someOtherParam);
    if (foo)
        doSomething();
    if (foo)
        doSomething(
            someParam,
            someOtherParam);

    // Bad
    if (foo) doSomething();
    if (foo) { doSomething(
        someParam,
        someOtherParam);
    }
    if (foo) {
        doSomething();
    }
    if (foo) {
        doSomething(
            someParam,
            someOtherParam);
    }
    ```
* `{ "curly": [true, "next-line-single-line"] }`
    ```
    // Good
    if (foo) {
        doSomething();
    }
    if (foo)
        doSomething(
            someParam,
            someOtherParam);
[...]

PR #1565, opened 2016-09-22
Improve Curly rule with more granular options

devrelm:
In #822, there was a request to improve the Curly rule with the ability to have single-line unbraced statements. This PR adds the options per [my comment on that issue](https://github.com/palantir/tslint/issues/822#issuecomment-247307510).

One thing to note: the presence of an option in the config will require brackets for that case, but the absence of an option does not require the absence of brackets for that case.


devrelm:
Are there any change requests for this?

tyv:
would love to see this functionality in release!

src/rules/curlyRule.ts
```diff
@@ -54,76 +95,150 @@ export class Rule extends Lint.Rules.AbstractRule {
...
-    public visitDoStatement(node: ts.DoStatement) {
-        if (!isStatementBraced(node.statement)) {
-            this.addFailureForNode(node, Rule.DO_FAILURE_STRING);
+    private validateStatementPositioning(node: ts.Node,
+                                         kind: ts.SyntaxKind,
+                                         isSameLine: boolean,
+                                         isSingleLine: boolean,
+                                         isBraced: boolean) {
+
+        if (((this.getOptions().length === 0) || this.hasOption(OPTION_ALL)) && !isBraced) {
```
nchen63: parse options one time when rule is loaded and store flags on the class.  Should also treat [true] as [true, "all"]

src/rules/curlyRule.ts
```diff
@@ -19,6 +19,12 @@ import * as ts from "typescript";
 
 import * as Lint from "../lint";
 
+const OPTION_ALL = "all";
```
nchen63: rename to "always" to be consistent with other rules (I realize this is different from eslint)

docs/rules/curly/index.html
```diff
@@ -12,12 +12,47 @@
...
   to be executed only if `foo === bar`. However, he forgot braces and `bar++` will be executed
   no matter what. This rule could prevent such a mistake.
-optionsDescription: Not configurable.
-options: null
+optionsDescription: |
+
+  The rule may be set to `true`, or to any combination of these arguments:
+
+  * `"all"` requires braces for all control-flow statements
+  * `"same-line-single-line"` requires braces for control-flow statements
```
nchen63: my issue with these settings is that it doesn't consider that there may be a rule in the future that requires that the brace *not* be there. All the names here assume that they all combine together to a set of places where braces are required.

nchen63: Also, sorry for not giving feedback earlier in the process. I know a lot of work went into this PR, but I don't think these are the set of options that we need.

How about this set of rules, which the user selects one of:
* always
* allow-same-line: allows for cases where the body and control flow statement are in one line. This addresses #822 

Then later, we can consider discussing:
* allow-single-line: allows for cases where the body is in a single line and may be on a different line from the control flow statement
* never-single-line
* never-same-line

alexeagle:
want!
@devrelm are you still pushing this PR forward?

devrelm:
@alexeagle I am! However, the holidays are the busy season for my employer (and my family,) so I haven't had much free time to make the modifications. I'll likely get back to this pr in a few weeks.

danvk:
@devrelm think you'll get back to this any time soon? I'd be happy to take over the PR from here if not.

devrelm:
@dankv: As it happens, my busy season is winding down now, so I have a bit more free time.

I'll resume dedicating some time to this tonight and see how long it takes to update per the suggestions.

docs/rules/curly/index.html
devrelm: I can agree with this, though yeah, it pretty much blows a lot of my implementation away. I'll try not to get to caught up on any sunk-cost fallacy, though, since having a good implementation trumps however much time was spent creating a poor implementation.

Besides, at this point there appears to have been some changes made to `curlyRule.ts` that would cause a merge conflict. I haven't yet checked how big the changes are, so I might have to do a decent-sized refactor anyway.

My choices for my original implementation were basically because I was getting caught up on creating granular rules that we could build off of.

For instance, your "allow-same-line" and "allow-single-line" might be implemented with (or be an aliases for):

* allow-same-line: `[true, "next-line-single-line", "next-line-multi-line"]`
* allow-single-line: `[true, "same-line-single-line", "next-line-single-line"]`

The granular options I created definitely don't allow for the "never" cases, though. However, it wouldn't require a complete rewrite, and would basically just require more `if` statements inside of `validateStatementPositioning()` (along with the additional config constants, docs, strings, etc. -- of course.)

Anyway, food for thought. I'll clean some stuff up over the next few nights and update the pr.

adidahiya:
Following up one of the sub-discussions above https://github.com/palantir/tslint/pull/1565#discussion_r95894307...

I'm mostly in agreement with @nchen63 here. Very granular options make this harder to maintain (think about the auto fixers too, those will be _essential_ to turning on this rule) and fork the coding style of the community (we're starting to see tslint enabled in DefinitelyTyped, which is pretty great). I think there's generally some demand to move towards fewer formatting options in JS/TS; look at projects like the recently-announced [prettier](http://jlongster.com/A-Prettier-Formatter) formatter. I certainly don't want to reject your work outright -- there are some useful options here in this PR -- but this large degree of customization doesn't belong in the core rule.

devrelm:
That's all perfectly reasonable, especially regarding the auto-fixers which I don't think were a thing (at least not a released thing) when I did the original work.

adidahiya:
@devrelm friendly ping, any progress here?

devrelm:
@adidahiya Yep. Sorry for the repeated delays. I tried to dive back into it last month, but between the conflicts and the atrophy of knowledge around my own changes, I realized that I had to dedicate more than a couple hours at a time to it to get it all in my head at the same time. I put in a ticket at work, and it's in the sprint that just started today, so expect an update within the next 2 weeks.

It might come in the form of a new pull request, since this one has a ton of conflicts due to the updates around auto-fixers and some other refactoring that was done last fall.

devrelm:
@adidahiya @nchen63 I've opened up #2334 with updates per the discussion here. It's pretty much done from scratch and should be much easier to follow. There's some issues with the tests running, but we can discuss that over on the other PR.

PR #1568, opened 2016-09-23
Add fix for no-unused-variable, only for imports

alexeagle:
cc @chuckjaz


palantirtech:

alexeagle:
cc @ScottSWu too - we talked about this one a couple times

alexeagle:
cc @paulvanbrenk I think this handles more cases than https://github.com/Microsoft/TypeScript/blob/pvb/discoverCodeActions/src/services/codefixes/unusedIdentifierFixes.ts
is that up-to-date? Where do we want to land this feature? I don't think tslint and typescript need two independent implementations...

alexeagle:
Ping!

alexeagle:
Note, my fix for #1569 is based on this one

PR #1571, opened 2016-09-23
fix: do not flag nested or .d.ts namespaces.

mprobst:
Note: I've been unable to test this properly as on my machine grunt for some reason picks up tsc 2.0.3, which fails to compile tslint.


mprobst:
Friendly ping?

mprobst:
Thanks @adidahiya!

PR #1575, opened 2016-09-23
Compile with TypeScript 1.8.10

adidahiya:
Breaking changes in compiler APIs broke the master branch. This is a simple workaround to get compilation working, but it forces us to depend on TS 2.0 (it doesn't compile with TS 1.8).

What do you think about bumping the peer dependency @JKillian?

### DIFF
--- a/README.md
+++ b/README.md
@@ -391,7 +391,7 @@ Creating a new release
 ----------------------
 <sup>[back to ToC &uarr;](#table-of-contents)</sup>
 
-1. Bump the version number in `package.json` and `src/tslintMutli.ts`
+1. Bump the version number in `package.json` and `src/tslintMulti.ts`
 2. Add release notes in `CHANGELOG.md`
 3. Run `grunt` to build the latest sources
 4. Commit with message `Prepare release <version>`
--- a/README.md
+++ b/README.md
@@ -391,7 +391,7 @@ Creating a new release
 ----------------------
 <sup>[back to ToC &uarr;](#table-of-contents)</sup>
 
-1. Bump the version number in `package.json` and `src/tslint.ts`
+1. Bump the version number in `package.json`, `src/tslint.ts`, and `src/tslintMutli.ts`
 2. Add release notes in `CHANGELOG.md`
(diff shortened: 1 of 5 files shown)


alexeagle:
We had to work around this too - we compile Angular using TS 1.9 internally.
https://github.com/angular/angular/blob/979657989bab4fb93faf6a207d30760faa9bd696/tools/%40angular/tsc-wrapped/src/tsc.ts#L70

adidahiya:
Update: I changed this PR to compile tslint with TS 1.8. This should fix the master branch. I tested the compiled linter with a peer of TS 2.0.3 and it still works.

README.md
```diff
@@ -391,7 +391,7 @@ Creating a new release
 ----------------------
 <sup>[back to ToC &uarr;](#table-of-contents)</sup>
 
-1. Bump the version number in `package.json`, `src/tslint.ts`, and `src/tslintMutli.ts`
+1. Bump the version number in `package.json` and `src/tslintMutli.ts`
```
alexeagle: spelling

JKillian:
Just so I understand, `getDirectories` is required by the TS 2.0 typings, but not really required at runtime, so TS 2.0 is fine as a peer? (And a similar story with `fileExists`?)

alexeagle:
What is the timeline for compiling tslint with TS 2.0?
Internally we use a single TS version for all compilation, which means I can't sync unmodified tslint into our third party libraries anymore.

adidahiya:
@alexeagle I just released v4.0.0-dev.0, https://github.com/palantir/tslint/pull/1596 (unfortunately it includes some of the v4 breaking changes) -- would that work for your build?

alexeagle:
I still need the local modification to uncomment lines prefixed with "include this field when compiling with TS 2.0" but it's not a big deal.

adidahiya:
that release comes from the `next` branch, which does compile with 2.0, so it includes the new compiler host options: https://github.com/palantir/tslint/blob/next/src/language/utils.ts#L32

PR #1591, opened 2016-10-02
Make only-arrow-functions allow generator functions

andy-hanson:
Currently the only way to write a generator function is with the `function` keyword, so it makes sense to allow this without requiring a flag.
This also appears to be the behavior of [eslint](http://eslint.org/docs/rules/prefer-arrow-callback).


palantirtech:
Thanks for your interest in [palantir/tslint](https://github.com/palantir/tslint), @andy-hanson! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

PR #1602, opened 2016-10-05
Add "any" order transform for "ordered-imports" rule

### LINKED ISSUE #1600: Optionally disable `named-imports-order` check of `ordered-imports` rule
It would be useful if the `named-imports-order` check could be disabled, so that only the import sources order is checked.

Either by allowing `named-imports-order` to be set to `true` or to `"any"` (or similar.

I'm happy to create a PR request for this if it seems seems sensible.

janslow:
`TRANSFORMS#any` transforms every input name to the empty string, so that they always equal to each other, thus all names are always sorted.

Fixes #1600


palantirtech:
Thanks for your interest in [palantir/tslint](https://github.com/palantir/tslint), @janslow! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

PR #1608, opened 2016-10-07
Check for newly introduced diagnostics before reporting failures (for unused import)

### LINKED ISSUE #1569: no-unused-variable rules incorrectly marks required import
### Bug Report
- **TSLint version**: `3.15.1 and HEAD`
- **TypeScript version**: `2.0`
- **Running TSLint via**: (please choose one) CLI

`tsconfig.json`

``` json
{ "compilerOptions": { "declaration": true } }
```

`modA.ts`

``` ts
export class OpaqueToken {
  private _desc: string;
}
```

`modB.ts`

``` ts
import {OpaqueToken} from './modA';
export class B {
  static BB = new OpaqueToken();
}
```

`use.ts`

```
import {OpaqueToken} from './modA';
import {B} from './modB';

export class Use {
  static PROP = [B.BB];
}
```

``` json
{
    "rules": {
      "no-unused-variable": true
    }
}
```
#### Actual behavior

The import of `OpaqueToken` is flagged as unused in `use.ts`. This causes my PR #1568 to break type-checking Angular when I auto-remove unused imports, with the error

```
[ts] Public static property 'PROP' of exported class has or is using name 'OpaqueToken' from external module "/Users/alexeagle/tmp123/modA" but cannot be named.
(property) Use.PROP: OpaqueToken[]
```
#### Expected behavior

Since we are producing `.d.ts` files, the OpaqueToken type appears in `use.d.ts` even though it does not appear in `use.ts`. Therefore TS requires that the symbol be imported so it knows what import to write in the `.d.ts`.
This means the import is not unused, even though a naive syntactic walk of `use.ts` doesn't show the usage.

alexeagle:
When a program is available, verify that fixes for no-unused-variable don't introduce new diagnostics. If a fix breaks the program, don't report that failure at all.

This re-uses TypeScript to verify that our failures are legitimate, without having to re-implement the TS logic for whether an import is actually unused. This capability can be used in other rules as well, ensuring that accepting tslint fixes won't break type-checking.

The downside is that it makes the check slower, especially in a codebase with many unused imports, since each file must be checked (and those with spurious failures must have each unused import checked).

Fixes #1569


palantirtech:

alexeagle:
See also discussion in the linked issue. TypeScript can fix this particular issue in the future, but tslint might still benefit from this capability in other cases.

alexeagle:
Ping!

src/language/languageServiceHost.ts
```diff
@@ -19,6 +19,57 @@ import * as ts from "typescript";
...
+                }
+            },
+        };
+    const langSvc = ts.createLanguageService(host, ts.createDocumentRegistry());
+    (langSvc as any).editFile = host.editFile;
+    return langSvc;
+}
+
+export function checkEdit(ls: ts.LanguageService, sf: ts.SourceFile, newText: string) {
+    if (ls.hasOwnProperty("editFile")) {
```
JKillian: typeguard function would be nice here

src/test.ts
```diff
@@ -74,6 +81,9 @@ export function runTest(testDirectory: string, rulesDirectory?: string | string[
                         return ts.createSourceFile(filenameToGet, fileText, compilerOptions.target);
                     } else if (filenameToGet === fileCompileName) {
                         return ts.createSourceFile(fileBasename, fileTextWithoutMarkup, compilerOptions.target, true);
+                    } else if (fs.existsSync(path.resolve(path.dirname(fileToLint), filenameToGet))) {
```
JKillian: just curious, what's this case for?

PR #1616, opened 2016-10-09
Rename no-constructor-vars to no-parameter-properties

### LINKED ISSUE #1296: Rename no-constructor-vars to no-parameter-properties
This is more consistent with the official name for this feature:

```ts
constructor(private foo: string) {
}
```

http://www.typescriptlang.org/docs/handbook/classes.html

adidahiya:
Resolves #1296.


msssk:
It looks like `ruleName` and `title` still [need to be updated](https://github.com/palantir/tslint/blob/231a6a9335cde1ed754571bb6f124cc1197a5aa3/docs/rules/no-parameter-properties/index.html).

adidahiya:
@msssk thanks for the heads up -- our docs generally stay a little bit behind because we don't want to publish github pages docs for unreleased versions of TSLint

PR #1618, opened 2016-10-09
Deprecate no-unused-variable rule

### LINKED ISSUE #1481: Start to deprecate no-unused-variable rule
I just noticed these new compiler checks in v2.0: [`--noUnusedParameters` and `--noUnusedLocals`](https://github.com/Microsoft/TypeScript/wiki/What%27s-new-in-TypeScript#flag-unused-declarations-with---nounusedparameters-and---nounusedlocals). These essentially make the `no-unused-variable` rule obsolete. Since it's a very core rule, I don't think we should remove it right away.

Proposal:

- remove `no-unused-variable` from `tslint:latest` & `tslint:recommended` in v4.0
- mark the rule as deprecated in v4.0
  - maybe in the failure message?
- remove the rule implementation completely in TSLint v5.0

adidahiya:
Resolves #1481. Adds a `deprecationMessage` field to rule metadata.

### DIFF
--- a/src/ruleLoader.ts
+++ b/src/ruleLoader.ts
@@ -48,6 +48,10 @@ export function loadRules(ruleConfiguration: {[name: string]: any},
                 const ruleSpecificList = (ruleName in enableDisableRuleMap ? enableDisableRuleMap[ruleName] : []);
                 const disabledIntervals = buildDisabledIntervalsFromSwitches(ruleSpecificList, allList);
                 rules.push(new Rule(ruleName, ruleValue, disabledIntervals));
+
+                if (Rule.metadata && Rule.metadata.deprecationMessage) {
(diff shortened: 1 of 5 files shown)


src/ruleLoader.ts
```diff
@@ -49,6 +49,10 @@ export function loadRules(ruleConfiguration: {[name: string]: any},
                 const disabledIntervals = buildDisabledIntervalsFromSwitches(ruleSpecificList, allList);
                 rules.push(new Rule(ruleName, ruleValue, disabledIntervals));
             }
+
+            if (Rule.metadata && Rule.metadata.deprecationMessage) {
```
adidahiya: this should actually go into the `else` block above.

PR #1619, opened 2016-10-09
Build system enhancements

adidahiya:
- Remove `typings/` and TSD, install ambient types from NPM
- Remove grunt, use NPM scripts instead
- Enable new compiler options `--noUnusedParameters` and `--noUnusedLocals`


.gitignore
```diff
@@ -13,3 +13,6 @@ typings/.basedir.ts
 # vim swap files
 *.swo
 *.swp
+
+# grunt-ts, see See https://github.com/grunt-ts/grunt-ts/issues/77
+.baseDir.ts
```
adidahiya: this can go away now

Issue #1400, opened 2016-07-13
"Variable used before declaration" with renamed import with the same name as another import


tomduncalf:
### Bug Report
- **TSLint version**: `3.13.0`
- **TypeScript version**: `2.0.0`
- **Running TSLint via**: CLI and Visual Studio Code
#### TypeScript code being linted

``` ts
import { Test as ITest } from './TestInterface'

import Test from './Test'
```

with `tslint.json` configuration:

``` json
{
  "rules": {
    "no-use-before-declare": true
  }
}

```
#### Actual behavior

`index.ts[1, 10]: variable 'Test' used before declaration`
#### Expected behavior

No errors. This works correctly with Typescript 1.8.10. 

Example repo at https://github.com/tomduncalf/tslint-bug

tomduncalf:
Off topic – excellent issue template by the way :)

IllusionMH:
The root case for this issue - change in behavior of `this.languageService.getDocumentHighlights` that start to return all `propertyNames` from imports even if they have alias.

I've tracker this to function `populateSearchSymbolSet` which has different results in TS 1.8.10 (symbol for Test) and 2.0.3 (symbols for Test and unknown)
```
getDocumentHighlights
getSemanticDocumentHighlights
getReferencedSymbolsForNode
getReferencesInNode
populateSearchSymbolSet
```
I didn't found any notices about such breaking changes in `languageService`

@JKillian, @adidahiya possible fix is to track positions of imported properties that has alias and don't show error if highlighted position matches position of tracked items. This won't introduce any problems for TS1.8.

However this method is used across several other rules and may cause other regressions.
Do you know where it is possible to find API documentation for all TS components?

Is it possible that this is regression in TS?

joscha:
This error was gone in TS 2.1.x but reappared in TS 2.2.

A type declaration for CSS modules as follows:
```
declare module "*.css" {
  const styles: any;
  export default styles;
}
```

and an import in `bla.tsx` via:
```
import styles from './some.css';
```

results in:
```
bla.tsx[6, 21]: variable 'styles' used before declaration
```

joscha:
A workaround for this is to add as many number of lines before the `declare module "*.css"` as there are potentially import statements in any of your tsx files.

joscha:
seems to be fixed in tslint 5.1

adidahiya:
@joscha thanks for the confirmation, that makes sense since we switched to `tsc`'s implementation of no-use-before-declare.

[...]

PR #1620, opened 2016-10-09
Workaround for #1400. Excludes imported properties from error

IllusionMH:
As I stated in [comment](https://github.com/palantir/tslint/issues/1400#issuecomment-249458412) for issue #1400 - difference with TS 2.x because `languageService.getDocumentHighlights` highlights not only nodes for `ITest` but also for `unknown`. I don't know if it is regression in TS or how it is supposed to filtered in new way.

This fix tracks positions for imported properties and prevents errors if it is "highlighted" as use before declarations.

I'd be glad to improve fix if you have any suggestions.

### DIFF
--- a/src/rules/noUseBeforeDeclareRule.ts
+++ b/src/rules/noUseBeforeDeclareRule.ts
@@ -46,6 +46,8 @@ export class Rule extends Lint.Rules.AbstractRule {
 type VisitedVariables = {[varName: string]: boolean};
 
 class NoUseBeforeDeclareWalker extends Lint.ScopeAwareRuleWalker<VisitedVariables> {
+    private importedPropertiesPositions: number[] = [];
+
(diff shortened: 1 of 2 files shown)


palantirtech:
Thanks for your interest in [palantir/tslint](https://github.com/palantir/tslint), @IllusionMH! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

IllusionMH:
PR #1608 fixes issue #1400 if `linterOptions` have `"typeCheck": true` and both `TestInterface.ts` and `Test` are provided.
Therefore this PR is only relevant for typeless (or untyped. What is correct version?) rules.

nchen63:
@IllusionMH very nice

PR #1622, opened 2016-10-10
More intuitive configuration composition

### LINKED ISSUE #1503: Change applicability order of "extends" array
### Enhancement Request
- **TSLint version**: `3.14.0`
- **TypeScript version**: `1.8.10`
- **Running TSLint via**: CLI

`"extends"` config parameter supports an _array_ of "base" configurations to be passed. The current behavior of applying base configurations could be more intuitive. Right now the layout is like this:

``` json
{
  "extends": [
    "./base-rules-overrides.json",
    "./base-rules.json"
  ],
  "rules": {
    "override the overrides": ""
  }
}
```

Whereas the proposition is to reverse the order in which "extensions" are applied:

``` json
{
  "extends": [
    "./base-rules.json",
    "./base-rules-overrides.json"
  ],
  "rules": {
    "override the overrides": ""
  }
}
```

See #1487 for more context.

adidahiya:
Resolves #1503.

### DIFF
--- a/src/configuration.ts
+++ b/src/configuration.ts
@@ -219,22 +219,21 @@ export function extendConfigurationFile(targetConfig: IConfigurationFile,
     const nextConfigRulesDirectory = arrayify(nextConfigSource.rulesDirectory);
     combinedConfig.rulesDirectory = configRulesDirectory.concat(nextConfigRulesDirectory);
 
-    combinedConfig.rules = {};
-    for (const name of Object.keys(objectify(targetConfig.rules))) {
-        combinedConfig.rules[name] = targetConfig.rules[name];
-    }
-    // next config source overwrites the target config object
-    for (const name of Object.keys(objectify(nextConfigSource.rules))) {
-        combinedConfig.rules[name] = nextConfigSource.rules[name];
-    }
+    const combineProperties = (targetProperty: any, nextProperty: any) => {
+        let combinedProperty: any = {};
+        for (const name of Object.keys(objectify(targetProperty))) {
+            combinedProperty[name] = targetProperty[name];
+        }
+        // next config source overwrites the target config object
+        for (const name of Object.keys(objectify(nextProperty))) {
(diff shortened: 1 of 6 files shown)


test/config/tslint-custom-rules-with-two-dirs.json
```diff
@@ -2,6 +2,7 @@
   "rulesDirectory": ["../files/custom-rules-2", "../files/custom-rules/"],
   "rules": {
     "always-fail": true,
-    "no-fail": true
+    "no-fail": true,
+    "rule-two": true
```
JKillian: what was this for? It seems to break the tests because there is no rule-two in existence for [this test](https://github.com/palantir/tslint/blob/eadbca9389439ab8685ebe0a50fd63a3fe340930/test/check-bin.sh#L63)

src/configuration.ts
```diff
@@ -189,19 +189,21 @@ function resolveConfigurationPath(filePath: string, relativeTo?: string) {
...
-    for (const name of Object.keys(objectify(baseConfig.rules))) {
-        combinedConfig.rules[name] = baseConfig.rules[name];
+    for (const name of Object.keys(objectify(targetConfig.rules))) {
+        combinedConfig.rules[name] = targetConfig.rules[name];
     }
-    for (const name of Object.keys(objectify(config.rules))) {
-        combinedConfig.rules[name] = config.rules[name];
+    // next config source overwrites the target config object
+    for (const name of Object.keys(objectify(nextConfigSource.rules))) {
+        combinedConfig.rules[name] = nextConfigSource.rules[name];
```
nchen63: please pick up changes from master and update combinedConfig.jsRules

nchen63:
should probably get this in for 4.0 since it's breaking

test/config/tslint-custom-rules-with-two-dirs.json
nchen63: removed this. Did was not hitting a unique test case

Issue #892, opened 2015-12-22
Rule suggestion: require `await` of async functions


cevek:
For example: 

```
async function foo(){
}
async function bar(){
    foo(); // I need absent await warning 
    return 234;
}
```

I need just simple example how to start, find method declaration and check that is async function

JKillian:
I'm not sure off the top of my head what exactly you'll need to do, but [atom-typescript](https://github.com/TypeStrong/atom-typescript) has a cool AST viewer which helps a lot when trying to write lint rules (see image below). 

However, I'm not sure this is possible at the moment. TSLint lints files on a file by file basis currently, so it your function comes from another file, it has no way of knowing that it's an async function. 

If you're only dealing with function declarations and calls from the same file, I imagine you could do something like this:

Override `visitCallExpression` and find all references to that function in the file, [similarly to here](https://github.com/palantir/tslint/blob/e9d0eae07d6048ec4ba872e874f958d6dac85444/src/rules/noUnusedVariableRule.ts#L281). If you can find a reference that's a function declaration / function expression, see if it contains the `async` flag. If it does, and if the call expression isn't part of an `async` statement, generate a lint failure. This is a very rough sketch though, I'm sure there'd be more complications / there might be better language services features you can use.

![image](https://cloud.githubusercontent.com/assets/1400151/11988246/c7b3571e-a9ba-11e5-8f68-8a3347d3ee5f.png)

adidahiya:
blocked on #680

aseemk:
We just hit a bug where we weren't properly awaiting an `async void` function. This lint rule would be great!

One addition to the rule: it'd be nice if the rule automatically saw if you were legitimately wanted to run something in parallel, not `await` it — but you should still be `then`ing or `catch`ing errors.

```ts
async function foo() { ... }

async function bar() {
    // This should fail:
    foo();

    // But these should pass:
    await foo();

    foo().then(
        () => { ... },
        (err) => { ... }
    )

    foo()
        .catch((err) => { ... });

    // Maybe ideally this should fail too, since it could lead to silent errors:
    foo().then(() => { ... });
```

adidahiya:
yep, this feature request is now unblocked, accepting PRs

andy-hanson:
@JoshuaKGoldberg @cevek I still suspect this is resolved by #1632. See my comment: https://github.com/palantir/tslint/pull/1632#issuecomment-273357419

JoshuaKGoldberg:
[...]

PR #1632, opened 2016-10-16
New rule: no-floating-promises

JoshuaKGoldberg:
Promises must be stored or passed to a function.

The idea here is that floating (unhandled) Promises can cause unexpected behavior. For example, in tests, they might resolve at unexpected times (like after another test following their parent test has started).

Fixes https://github.com/Microsoft/tslint-microsoft-contrib/issues/174 (PRing here because I'm not sure they support the type checker).

### DIFF
--- a/test/rules/no-floating-promises/promises/test.ts.lint
+++ b/test/rules/no-floating-promises/promises/test.ts.lint
@@ -1,0 +1,141 @@
+class Promise {
+    then(): Promise;
+}
+
+function returnsPromiseFunction() {
+    return new Promise();
+}
+
+const returnsPromiseVariable = () => new Promise();
+
+class ReturnsPromiseClass {
+    returnsPromiseMemberFunction() {
+        return new Promise();
+    }
+
+    returnsPromiseMemberVariable = () => new Promise();
+
+    static returnsPromiseStaticFunction = () => new Promise();
+}
+
+let a = returnsPromiseFunction();
+let b = returnsPromiseVariable();
+let c = new ReturnsPromiseClass().returnsPromiseMemberFunction();
+let d = new ReturnsPromiseClass().returnsPromiseMemberVariable();
+let e = ReturnsPromiseClass.returnsPromiseStaticFunction();
+
+returnsPromiseFunction();
+~~~~~~~~~~~~~~~~~~~~~~~~       [0]
+
+returnsPromiseFunction().then();
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~       [0]
+
+returnsPromiseVariable();
+~~~~~~~~~~~~~~~~~~~~~~~~       [0]
+
+returnsPromiseVariable().then();
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~       [0]
+
+new ReturnsPromiseClass().returnsPromiseMemberFunction();
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~       [0]
+
+new ReturnsPromiseClass().returnsPromiseMemberFunction().then();
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~       [0]
+
+new ReturnsPromiseClass().returnsPromiseMemberVariable();
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~       [0]
+
+new ReturnsPromiseClass().returnsPromiseMemberVariable().then();
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~       [0]
+
+ReturnsPromiseClass.returnsPromiseStaticFunction();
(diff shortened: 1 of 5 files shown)


JoshuaKGoldberg:
Note: added configurable options in https://github.com/palantir/tslint/pull/1632/commits/6f932ddd093eba38d25c8b99c714462c713ca10f so additional class names of Promises can be added.

test/rules/no-floating-promises/promises/test.ts.lint
```diff
@@ -0,0 +1,58 @@
+class Promise { }
```
nchen63: add test cases for `returnsPromiseFunction().then()` and calling static Promise functions

JoshuaKGoldberg:
Following up on this PR: I haven't had time to get to it, but will in the next month.

andy-hanson:
This should fix #892?

JoshuaKGoldberg:
@andy-hanson unfortunately no. Specifically ensuring async functions are awaited is a different matter than generally ensuring functions that return Promises have their results dealt with.

The following code should be flagged by 892's solution, but not by this one:

```typescript
async function getSomething() {
    return Promise.resolve(7);
}

async function runTest() {
    const result = getSomething();
    return result; // you may think this is a number, but it's actually `Promise<number>`
}
```

andy-hanson:
Are you referring to syntax? Semantically "async functions" and "functions that return Promises" are the same, right?

I don't think there's anything wrong with the result you show here. Returning a Promise from an async function is OK. `return result;` and `return await result;` should hvae the same effect, and it won't change the return type since async functions always return promises anyway.

I interpreted the issue there as just not wanting any `foo();` where the return type of `foo()` is `Promise<void>` rather than `void`.

alitaheri:
@andy-hanson Actually it is recommended that you return the promise directly as appending await will create a promise that is resolved with the result of await expression. that means one unnecessary idempotent promise created.

Arguably it should better be **disallowed**. `return await asyncStuff()` is not only semantically equivalent to `return asyncStuff()` but also is less memory/cpu efficient.

ajafff:
@alitaheri 
> Arguably it should better be disallowed. return await asyncStuff() is not only semantically equivalent to return asyncStuff() but also is less memory/cpu efficient.

I'm currently working on a rule for this.

JoshuaKGoldberg:
@nchen63 I believe your feedback has been resolved. Could you please take another look?

I switched from a whitelist of allowed parent types to a blacklist, since there are much more places using a Promise generator would be allowed.

JoshuaKGoldberg:
Rebased, but now the build is failing with `node_modules/@types/underscore/index.d.ts(3824,43): error TS2304: Cannot find name 'Element'.`. Guessing that's external to this PR?

nchen63:
try rebuild without cache

andy-hanson:
Do you know if there's a way to permanently clear the cache in CircleCI? I have to rebuild without cache every time I push.

JoshuaKGoldberg:
Passed CircleCI, thanks for the tip.

nchen63:
@JoshuaKGoldberg thanks!

RobFreundlich:
When turning on no-floating-promises, code like the following is getting flagged as a problem:

```typescript
something_that_produces_a_promise().then(() => fulfillPromise());
function fulfillPromise()
{
  // do some stuff ...
  resolve(whatever);
}
```

That seems wrong to me - the ```Promise``` is getting handled by a ```then()```.  Granted, the ```then()``` technically returns a ```Promise```, but if TSLint is going to complain about that, then isn't this rule sort of impossible to *not* trigger?

Am I misunderstanding something, or is this a bug?

Thanks!

JoshuaKGoldberg:
So your source looks something like this?

```typescript
return new Promise((resolve, reject) => {
    something_that_produces_a_promise().then(() => fulfillPromise());

    function fulfillPromise() {
        // do some stuff ...
        resolve(whatever);
    }
});
```

Is there a reason you can't use `async`/`await` or rewrite it to not use `resolve`?

```typescript
return something_that_produces_a_promise()
    .then(() => fulfillPromise())
    .then(() => {
        // do some stuff ...
        return whatever;
    });
```

I'm not convinced there's a frequently valid use case for using `resolve`/`reject`. `async`/`await` simplifies everything, and even if you can't use it yet, it's (as I understand it) more Promise-like to return things through instead of manually calling `resolve`s.

Also, this feels like a bug/feature request that might be better as a separate issue?

RobFreundlich:
I'd love to use ```async```/```await```, but we have to support (at least for a while) older versions of IE that don't do well when debugging them.  /sigh

The code does look like you described, though, and now I'm wondering whether I can just rewrite it to not use ```resolve(whatever)```.  As long as I declare ```whatever``` before the return, it's available (I know, it's available even before its declaration unless I turn on *that* TSLint rule, which I'd love to, but I got overruled on that debate, but I have using variables before they're declared, but I digress), that will work.  

I imagine I could also just return ```whatever``` directly from ```fulfillPromise()`` as well, right?  (Still figuring out all of the intricacies of Promises).

I'll give that a shot and let you know how it turns out.  Thanks!

sbking:
@JoshuaKGoldberg The following code is flagged:

    async function run () {
      return Promise.resolve()
    }

    run()

How are you supposed to call the top level async function in an application? The following also fails:

    run.then(() => console.log('foo'))

What am I missing? Do you have to have a regular JavaScript file to call the top level function which returns a promise?

JoshuaKGoldberg:
Yup, either works. This should ideally only happen once per application.

andy-hanson:
I think we should allow a promise as long as its errors are caught. See #2774.

PR #1634, opened 2016-10-16
Track and restore initial or previous rule state for one line lint switches (fixes #1624) 

### LINKED ISSUE #1624: disable-line affects "global" disable behavior for following lines
### Bug Report
- **TSLint version**: `^3.15.1`
- **TypeScript version**: `2.0.3`
- **Running TSLint via**: (please choose one) `grunt-tslint` and `TSLint for Visual Studio Code`

``` ts
//tslint:disable:no-string-literal 

class MyClass {}
const myClass = new MyClass();

// no linting problems, because of "global" disable
myClass["xxx"]();

// no linting problems because of "global" or "line" disable
myClass["xxx"]();  //tslint:disable-line:no-string-literal 

// now the linter tells me that object access via string literals is disallowed
myClass["xxx"](); 
```
#### Actual behavior

The `disable-line` tslint statement also changes the behavior of the `disable` tslint statement. So, for all following lines it's as if `tslint:enable` has been called.

True, in my code example above there's actually no need for `tslint:disable-line`, because it's already disabled by the "global" `disable`. So this behavior won't happen a lot in most use cases, but it just bit me (I copied/pasted a line with `disable:line`).
#### Expected behavior

I would expect that  `disable-line` really only affects the behavior of the current line. I don't want my "global" `tslint:disable` being re-enabled by `tslint:disable-line`.

IllusionMH:
This PR makes `EnableDisableRulesWalker` aware of initial rules configuration for file.
This allows track rule state during file parsing and restore proper rule state after it was disabled for single line (with `disable-line` or `disable-next-line` switchers).

Todo:
- [x] Handle `tslint:(disable|enable)` cases
- ~~Add tests for `enable-line` and `enable-next-line` switchers~~ (it was early morning and I've missed that this rules aren't documented)
- [x] Do not add new entries to `enableDisableRuleMap` if rule state doesn't change

ETA: October, 23rd


IllusionMH:
I've updated this pull request with rework for how enabled/disabled states are tracked for file.
Notable changes:

- Rule switchers are tracked only for rules that are enabled in configuration (rules that wouldn't be executed in any case)
- Switch positions are only added in case when state for rule changes (this allow to be sure that states in `enableDisableRuleMap` always alternate which simplifies generation of disabled intervals)
- If switch is performed for all rules - all of them are manually switched if needed (further simplifies generation of disabled intervals)

I've removed test cases for `enable-line` and `enable-next-line` from TODO list since this feature isn't documented for TSLint (ESLint also).
@adidahiya @JKillian should this feature be supported (I'll add test cases if needed) or may be it should show warnings for developers (since it might be error prone to count on rule that are enabled only for one line in disabled interval) ?

At this point I think that work is finished and PR is ready for review.

IllusionMH:
I haven't found any docs on my question so I want to clarify it here: What is proper way for this repo to make branch up to date with `master` - `git rebase master` or `git merge master`?

nchen63:
It doesn't matter if you merge or rebase since we just squash the commit before merging into master

IllusionMH:
Rebased and updated PR with recent changes in `master`.

src/tslintMulti.ts
```diff
@@ -164,19 +164,19 @@ class MultiLinter {
 
     private getEnabledRules(fileName: string, source?: string, configuration: IConfigurationFile = DEFAULT_CONFIG): IRule[] {
         const sourceFile = this.getSourceFile(fileName, source);
+        const isJs = fileName.substr(-3) === ".js";
```
nchen63: yes, good catch

IllusionMH:
@nchen63 thank you for review.

> The enable-line and enable-next-line cases still work right? Like nothing blows up?

Yes. Actually there already were [tests for `enable-line`/`enable-next-line`](https://github.com/palantir/tslint/blob/master/test/rules/_integration/enable-disable/test.ts.lint#L58-L62) switchers.

> factor out and put in Utils

`src/utils.ts` contains only basic utility functions without any rules related functionality.
Therefore I decided to move logic from instance `isEnabled` method to static method which still belongs to `AbstractRule`.
But if you insist - I'll move it to utils.

nchen63:
@IllusionMH Looks good. Thanks!

PR #1638, opened 2016-10-17
Allow fast null checks in no-unused-expression rule

IllusionMH:
This PR adds `allow-fast-null-checks` option for `no-unused-expression` rule which allows to use logical operators `&&` and `||` to perform fast check for null(falsy) values and than call methods on non-null values or functions for side effects (e.g. `e && e.preventDefault()` in event handlers that for some reasons might be called without event object).

New functionality implemented as option to allow teams forbid fast null checks in their code in favor of more verbose implementations.

Implementation allows expression statements with `&&` and `||` operators (may be nested or in parentheses) which has call expression in right most position , but doesn't checks if this call has side effects (it is developers duty).


nchen63:
I'm a little uncomfortable with letting this syntax to creep into TypeScript, but I totally get how it's so much easier and that it's optional. @adidahiya?

adidahiya:
@nchen63 seems ok if the behavior is is hidden behind a rule option. We probably won't enable it in the recommended config.

IAfanasov:
PR waiting for review for 8 months. pity

IllusionMH:
It was merged half year ago. Only label left.

gnapse:
This is documented to allow things like `e && e.preventDefault()`. But what about `fn && fn()`? It does not seem to allow that to pass. Is it intentional? I'd propose that that should be added too if it's not supported so far.

suchanlee:
@gnapse could you create an issue to track that? would welcome a PR for that!

gnapse:
@suchanlee I created the issue (#4021) as requested. I'd love to work on a PR myself, but I'm totally new not just to tslint but to typescript in general. That's no reason to back off, but a hint on where to look for how to work on this would be greatly needed and appreciated. Would love to help!

PR #1643, opened 2016-10-19
Adds semicolon option `ignore-bound-class-methods`

### LINKED ISSUE #1476: Semicolon rule should flag unnecessary semicolons on class member functions
We changed the behavior of the semicolon rule to allow omission of the trailing semicolon after bound class member functions:

``` ts
class Foo {
  public renderFoo = () => {
    return null;
  }
   ^ right here
}
```

I propose that we actually _ban_ semicolons at that location in TSLint 4.0. We can support the old functionality with a new option `allow-bound-class-methods`.

nchen63:
also changes default behavior by requiring semicolons to not come after bound class methods unless this setting is used

closes #1476


palantirtech:
Thanks for your interest in [palantir/tslint](https://github.com/palantir/tslint), @nchen63! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

nchen63:
@JKillian @adidahiya

docs/rules/semicolon/index.html
```diff
@@ -10,6 +10,7 @@
```
JKillian: this file update shouldn't be in this PR, it should only happen when you're updating the docs as a whole (on release)

src/rules/semicolonRule.ts
```diff
@@ -21,6 +21,7 @@ import * as Lint from "../lint";
 
 const OPTION_ALWAYS = "always";
 const OPTION_NEVER = "never";
+const OPTION_ALLOW_BOUND_CLASS_METHODS = "allow-bound-class-methods";
```
JKillian: maybe it should be "skip-bound-class-methods"?

JKillian: oh wait, we already have an "ignore-interfaces" option, this should be "ignore-bound-class-methods" then

src/rules/semicolonRule.ts
```diff
@@ -21,6 +21,7 @@ import * as Lint from "../lint";
 
 const OPTION_ALWAYS = "always";
 const OPTION_NEVER = "never";
+const OPTION_ALLOW_BOUND_CLASS_METHODS = "ignore-bound-class-methods";
```
JKillian: naming of variable

PR #1644, opened 2016-10-19
Added a completed-docs rule

JoshuaKGoldberg:
Fixes https://github.com/Microsoft/tslint-microsoft-contrib/issues/162.

Enforces non-blank comments for any or all of classes, functions,
methods, and/or properties.

### DIFF
--- a/src/rules/completedDocsRule.ts
+++ b/src/rules/completedDocsRule.ts
@@ -1,0 +1,120 @@
+/**
+ * @license
+ * Copyright 2013 Palantir Technologies, Inc.
+ *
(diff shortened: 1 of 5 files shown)


JoshuaKGoldberg:
*(this also requires the type checker)*

nchen63:
Thanks for the contribution!

JoshuaKGoldberg:
Awesome, thanks for taking a look!

The original tslint-microsoft-contrib request had a number of places it could request them exist on. I didn't put them in this initial PR for fear of doing extra work, but now have filed #1713 to fill it out.

PR #1645, opened 2016-10-19
Update docs

nchen63:


adidahiya:
as I mentioned in #1296, let's not do this until we make a stable release of 4.0 on the `latest` npm dist-tag.

adidahiya:
@nchen63 can you update this PR (or make a new one if you prefer)?

PR #1649, opened 2016-10-22
npm-debug.log in .gitignore

nomaed:
npm-debug.log file is generated by any npm tasks that returns with an error-code.
This file should not remain in git.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,9 +7,9 @@
 /lib/
 node_modules/
 tscommand*.txt
+npm-debug.log
 # created by grunt-ts for faster compiling
 typings/.basedir.ts
-npm-debug.log
 
 # vim swap files
 *.swo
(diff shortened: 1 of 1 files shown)


nomaed:
Follow up questions:

1. The `created by grunt-ts for faster compiling` section in `.gitignore`, is it still needed?
2. I noticed that after running the full build, there are plenty files modified under `docs/*`, even on files that I haven't modified. Should these be ignored in git as well and auto-generated? Should they be submitted? What's the story with them?
3. Related to no. 2: for running a full build, instead of `grunt` I am using `npm run verify`. Is that correct?

adidahiya:
sorry for the confusion; there are still some references to grunt after I moved the build system to NPM. `npm run verify` is the right script for a full build. related: 1631

PR #1650, opened 2016-10-22
Option added to no-consecutive-blank-lines

nomaed:
Added an optional argument to `no-consecutive-blank-lines` rule to allow specifying the maximum number of allowed blank lines to appear in sequence, instead of only allowing a hard limit of 1.

If no option is specified, the old value of 1 is used, thus the rule is fully backwards compatible.

Similar to ESLint’s `no-multiple-empty-lines` rule “Max” option:
http://eslint.org/docs/rules/no-multiple-empty-lines


src/rules/noConsecutiveBlankLinesRule.ts
```diff
@@ -43,6 +67,9 @@ class NoConsecutiveBlankLinesWalker extends Lint.SkippableTokenAwareRuleWalker {
     public visitSourceFile(node: ts.SourceFile) {
         super.visitSourceFile(node);
 
+        const options = this.getOptions();
+        const allowedBlanks = options && options[0] || Rule.DEFAULT_ALLOWED_BLANKS;
```
nchen63: should this match line 57? `options` is not an array

nomaed: The `options` property of the  `NoConsecutiveBlankLinesWalker` class _is_ an array, and it is the rule's arguments. In it's parent's parent:
```typescript
export class RuleWalker extends SyntaxWalker {
    ...
    private options: any[];
    ...

    constructor(private sourceFile: ts.SourceFile, options: IOptions) {
        ...
        this.options = options.ruleArguments;
        ...
    }

    ...
}
```

nchen63:
Thanks @nomaed!

Issue #1611, opened 2016-10-09
indent rule applied inside of multi-line `template` strings


nonplus:
### Bug Report
- **TSLint version**: `3.15.1`
- **TypeScript version**: `2.0.3`
- **Running TSLint via**: CLI
#### TypeScript code being linted

``` ts
// (second line in the template string has 3 leading spaces)
let sql = `SELECT *
   FROM user
   WHERE name = 'bob'`;
```

with `tslint.json` configuration:

``` json
{
    "rules": {
        "indent": [
            true,
            "tabs"
        ]
}
```
#### Actual behavior

TSLint logs an `(indent) db/test.ts[1, 1]: tab indentation expected` error.
#### Expected behavior

No error should be logged.
#### Workaround:

``` ts
/* tslint:disable:indent */
let sql = `SELECT *
   FROM user
   WHERE name = 'bob'`;
/* tslint:enable:indent */
```
#### Notes:

Indentation should be ignored inside the literal portion of a template string (`InTemplateHeadOrNoSubstitutionTemplate` and `InTemplateMiddleOrTail`).

However, I think it makes sense to check for indentation **within template substitution**.  So the following should be considered an indentation error:

``` ts
let name = "Bob";
// (second line in the template string has 3 leading spaces)
let s = `Hello ${
   name
}, how are you?`;
```

JoshuaKGoldberg:
Looks like this was fixed! 🎉

PR #1651, opened 2016-10-22
Skip over template string when checking indent (#1611)

nonplus:


palantirtech:
Thanks for your interest in [palantir/tslint](https://github.com/palantir/tslint), @nonplus! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

nchen63:
I think this would be more straightforward if you overrode visitTemplateExpression(node) to get the template start/end offsets.

nchen63:
Thanks for the contribution! Regarding my earlier comment, I don't think it's necessary

PR #1654, opened 2016-10-26
New Rule: no-useless-files

chrismbarr:
#### PR checklist
- [ ] Addresses an existing issue
- [x] New feature
- [x] Includes tests
- [ ] Documentation update

A new rule named `no-useless-files` has been added.  It reports violations for file that have no content (or only whitespace), and files that only contain comments.
#### Reasoning/History

My organization used this rule internally for a bit to help locate and remove these useless files.  Sometimes an entire file was commented out instead of being removed from source control with the thinking "I'll get back to this later", but of course that never happened and now we just had these useless files cluttering the project.

I feel like my tests are formatted correctly, but they just aren't passing for some reason.  Sometimes it seems to be treating the lines that only contain `~~~~` characters as code, and it underlines those underlines.

Here's what it's showing me:
2016-10-26 09_11_30-c__windows_system32_cmd exe

I'm happy to do the work to make these tests pass, I just need a little help understanding why these are failing right now.

### DIFF
--- a/docs/rules/no-useless-files/index.html
+++ b/docs/rules/no-useless-files/index.html
@@ -1,0 +1,12 @@
+---
+ruleName: no-useless-files
+description: Disallows empty files, files that only contain whitespace, and files that only contain comments.
+optionsDescription: Not configurable.
+options: null
+optionExamples:
+  - 'true'
+type: typescript
+optionsJSON: 'null'
+layout: rule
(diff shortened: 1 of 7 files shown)


chrismbarr:
Just realized this PR is still sitting here - any idea what the deal is with these tests?

PR #1656, opened 2016-10-26
New Rule: max-classes-per-file

chrismbarr:
#### PR checklist
- [ ] Addresses an existing issue
- [x] New feature
- [x] Includes tests
- [ ] Documentation update

A new rule named `max-classes-per-file` has been added.  This rule will disallow a file to contain more classes than is specified in the configuration.  The rule can also be configured with a regex pattern to ignore file names that match a certain pattern.
#### Reasoning

It mostly comes back to the singe-responsibility principle.  In general if a file contains multiple classes, it might be doing too many things.  A better practice might be to limit each file to only contain a single class
#### Example Configurations

This configuration would limit all TS files to contain no more than 1 class.  Any file with 2+ classes will report a violation for each additional class in a file.

``` json
{
  "max-classes-per-file": [true, 1]
}
```

This configuration would allow a maximum of two classes per file, except for files that match the specified pattern. This specific example would allow more than 2 classes in files named `*.model.ts`, `*.tests.ts` or `*.spec.ts`

``` json
{
  "max-classes-per-file": [
    true,
    2,
    {"ignore-filename-pattern": "\\.+(model|tests|spec)\\.ts$"}
  ]
}
```


docs/rules/max-classes-per-file/index.html
```diff
@@ -0,0 +1,56 @@
+---
```
adidahiya: please remove the generated docs from this PR; we'll add it when we make a release that contains this new rule.

chrismbarr: Sorry about that, I'll remove it.  last time I made a PR to this repo these weren't a thing, so it must be new (but I like it!)

chrismbarr:
Ah yeah, good point about the `ignore-filename-pattern` option since no other rule uses that.  This was a rule that my team used internally on an angular app where we also enforced file naming conventions.  This way all our angular controllers, services, etc. would just have a single class, but files named `*.model.ts` or `*.tests.ts` were allowed to contain multiple classes when the rule was configured that way.

I'll remove this option for the PR, but if it's excluded I'll likely want to keep a fork of it with this option as a custom rule that my team will use.

docs/rules/max-classes-per-file/index.html
chrismbarr: FYI, it's a little confusing that we need to run the `validate` NPM script, but then know not to check in certain files that it generates.

chrismbarr:
OK, your suggestions have been implemented now.  I was unaware of class expressions, so I learned about that and added a visitor & tests for those.

I thought about adding a boolean option called `ignore-class-expressions`, but really a class is a class, and if you set a max per file then the syntax used to define the class shouldn't matter all that much.  If that option existed and was turned on, then lazy developers on a team might try to switch syntax to get around this violation as a quick "fix". That's just my $0.02, but I could add that option if anyone else thought it was useful

Issue #1640, opened 2016-10-18
auto-fixer for import-ordering rule


adidahiya:
@nchen63

nchen63:
@adidahiya @JKillian I'm worried about the future of fixers. In the near future, there will be fixes that overlap each other, but the way it's implemented now as a formatter, it runs only once and doesn't easily allow for multiple passes. One way around this might be to detect if 2 replacements overlap each other and then rerun the entire linter, but then you run into a halting problem. Maybe it shouldn't be a type of formatter

JKillian:
cc @JoshuaKGoldberg

PR #1657, opened 2016-10-28
Implement fixer for ordering import declarations

nchen63:
#### PR checklist
- [x] Addresses an existing issue: #1640
- [x] Includes tests
- [ ] Documentation update 

_need a place to list fixers in docs_

Auto-fixer for import declarations, and the named declarations inside of them.

(optional)

### DIFF
--- a/src/rules/orderedImportsRule.ts
+++ b/src/rules/orderedImportsRule.ts
@@ -109,6 +109,19 @@ function findUnsortedPair(xs: ts.Node[], transform: (x: string) => string): [ts.
     return null;
 }
 
+function sortByKey<T>(xs: T[], getSortKeyCallback: (x: T) => string): T[] {
+    return xs.sort((a, b) => {
+        const transformedA = getSortKeyCallback(a);
+        const transformedB = getSortKeyCallback(b);
+        if (transformedA > transformedB) {
+            return 1;
+        } else if (transformedA < transformedB) {
+            return -1;
+        }
+        return 0;
+    });
+}
+
 // Transformations to apply to produce the desired ordering of imports.
 // The imports must be lexicographically sorted after applying the transform.
 const TRANSFORMS: {[ordering: string]: (x: string) => string} = {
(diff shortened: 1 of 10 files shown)


src/rules/orderedImportsRule.ts
```diff
@@ -109,6 +109,19 @@ function findUnsortedPair(xs: ts.Node[], transform: (x: string) => string): [ts.
     return null;
 }
 
+function sortByKey<T>(xs: T[], getSortKeyCallback: (x: T) => string): T[] {
```
JKillian: i'd recommend just calling this `getSortKey`, since it returns a string and not a function

src/rules/orderedImportsRule.ts
```diff
@@ -109,6 +109,19 @@ function findUnsortedPair(xs: ts.Node[], transform: (x: string) => string): [ts.
...
+    return xs.sort((a, b) => {
+        const transformedA = getSortKeyCallback(a);
+        const transformedB = getSortKeyCallback(b);
+        if (transformedA > transformedB) {
+            return 1;
+        } else if (transformedA < transformedB) {
+            return -1;
+        }
+        return 0;
```
JKillian: Trivia: all these comparisons can be replaced with ` return transformedA.localeCompare(transformedB)`. it is slower though, and mainly only important if you're comparing non-english strings

src/rules/orderedImportsRule.ts
```diff
@@ -155,25 +171,118 @@ class OrderedImportsWalker extends Lint.RuleWalker {
...
         }
         super.visitNode(node);
     }
 }
+
+interface ImportDeclaration {
+    node: ts.ImportDeclaration;
+    nodeStartOffset: number;    // start position of node within source file
+    text: string;
+    sourceFile: string;
```
JKillian: I'd recommend calling this something like `sourcePath` or just `source`, ad it doesn't always refer to a file and `getSourceFile`/`SourceFile` is used by TS with a different meaning in TSLint

src/rules/orderedImportsRule.ts
nchen63: Yeah, I used localeCompare, but it ignores casing, so it changes the behavior from what it was before

JKillian: In hindsight, beyond performance concerns, you might be better off doing things this way anyways. I believe `localeCompare` could sort strings differently depending on the users locale and then we wouldn't have a deterministic way of sorting things, which would be bad

PR #1659, opened 2016-10-28
Make sure vscode uses locally provided Typescript

nomaed:
By default, VSCode uses its bundled TypeScript version which may change depending on VSCode version.
Also, it shows warnings when the globally installed TS version differs from the bundled one.

The solution is to instruct VSCode to always use the local TS version that is installed from packages.json in node_modules.


JKillian:
Makes sense, thanks @nomaed!

Issue #1635, opened 2016-10-16
npm run test:bin fails on Windows platform


IllusionMH:
### Bug Report
- **TSLint version**: `4.0.0-dev`
- **TypeScript version**: `2.0.3`
- **Running TSLint via**: CLI
- **OS**: Windows 10

`npm run test:bin` fails on Windows (even in Git Bash).
#### Actual behavior

```
$ npm run test:bin

> tslint@4.0.0-dev test:bin D:\web\projects\tslint
> ./test/check-bin.sh &> /dev/null

The syntax of the command is incorrect.

npm ERR! Windows_NT 10.0.14946
npm ERR! argv "C:\\Program Files (x86)\\Nodist\\v\\nodev6.5.0\\node.exe" "C:\\Program Files (x86)\\Nodist\\bin\\node_modules\\npm\\cli.js" "run" "test:bin"
npm ERR! node v6.5.0
npm ERR! npm  v3.5.2
npm ERR! code ELIFECYCLE
npm ERR! tslint@4.0.0-dev test:bin: `./test/check-bin.sh &> /dev/null`
npm ERR! Exit status 1
npm ERR!
npm ERR! Failed at the tslint@4.0.0-dev test:bin script './test/check-bin.sh &> /dev/null'.
npm ERR! Make sure you have the latest version of node.js and npm installed.
npm ERR! If you do, this is most likely a problem with the tslint package,
npm ERR! not with npm itself.
npm ERR! Tell the author that this fails on your system:
npm ERR!     ./test/check-bin.sh &> /dev/null
npm ERR! You can get information on how to open an issue for this project with:
npm ERR!     npm bugs tslint
npm ERR! Or if that isn't available, you can get their info via:
npm ERR!     npm owner ls tslint
npm ERR! There is likely additional logging output above.

npm ERR! Please include the following file with any support request:
npm ERR!     D:\web\projects\tslint\npm-debug.log
```

If I remove `&> /dev/null` from this script it still fails, but with another error

```
$ npm run test:bin

> tslint@4.0.0-dev test:bin D:\web\projects\tslint
> ./test/check-bin.sh

'.' is not recognized as an internal or external command,
operable program or batch file.

npm ERR! Windows_NT 10.0.14946
npm ERR! argv "C:\\Program Files (x86)\\Nodist\\v\\nodev6.5.0\\node.exe" "C:\\Program Files (x86)\\Nodist\\bin\\node_modules\\npm\\cli.js" "run" "test:bin"
[...]

PR #1667, opened 2016-11-01
Reimplement tests for TSLint bin file

IllusionMH:
#### PR checklist

- [X] Addresses an existing issue: #1635
  - [ ] Includes tests
- [ ] Documentation update

This PR contains new approach to run tests for TSLint CLI executable.
New test are based on Mocha framework and replace old approach with shell script.
Benefits of new implementation:
- Uses same tools as other test in project
- Supports all platforms (previously were skipped on Windows)
- Allow checks of `stdout` and `stderr` content with ease

Possible drawbacks: May be slower than shell script

There no direct documentation for this update, but I can update build steps in `readme.md` since they are related. Moreover I propose to remove `docs` from `verify` script (use separate script to generate docs) and use `npm run verify` instead of `grunt`.

~~PS. Tests with first commit may fail on node v4 and v5 because of wrong exit codes.~~ Looks like everything is OK on both platforms.


nchen63:
lgtm

JKillian:
This is cool! Thanks @IllusionMH!

Issue #1639, opened 2016-10-18
using tslint cli as npm script has inconsistent path behavior for windows and linux


acdcjunior:
### Bug Report
- **TSLint version**: `~3.15.1`
- **TypeScript version**: `~2.0.3`
- **Running TSLint via**: CLI via NPM script

TypeScript code being linted: (irrelevant)
with `tslint.json` configuration: (irrelevant)
#### `package.json` configuration:

``` json
  "scripts": {
    "unquoted": "tslint src/**/*.ts",
     "quoted": "tslint 'src/**/*.ts'"
  },
```
#### Actual behavior

Unconsistend behavior when running @ windows or linux (and not working at all on windows).
#### Expected behavior

Consistent behavior across platforms.
### Workaround

Use both, I guess:

``` json
  "scripts": {
    "unquoted_and_quoted": "tslint src/**/*.ts && tslint 'src/**/*.ts'"
  },
```
# How to reproduce:

Check readme and files: https://github.com/acdcjunior/tslint-windows-path-inconsistency
(pasting README here for reference:)
## Unquoted path (`npm run unquoted`)
- @windows 10
  - All good: three linting errors!
  
  ``` bash
  $ npm run unquoted
  
  > tslint-windows-path-inconsistency@1.0.0 unquoted D:\github\tslint-windows-path-inconsistency
  > tslint src/**/*.ts
  
  src/FirstLevel.ts[1, 12]: " should be '
  src/second/SecondLevel.ts[1, 12]: " should be '
  src/second/third/ThirdLevel.ts[1, 12]: " should be '
  
  npm ERR! Windows_NT 10.0.14393
  ...
  ```
- @centos 7
  - **Only one error when three are expected!**
  
  ``` bash
  [box@box tslint-windows-path-inconsistency]$ npm run unquoted
  
  > tslint-windows-path-inconsistency@1.0.0 unquoted /box/github/tslint-windows-path-inconsistency
  > tslint src/**/*.ts
  
  src/second/SecondLevel.ts[1, 12]: " should be '
  
  npm ERR! Linux 3.10.0-229.el7.x86_64
  ...
  ```
## Quoted path  (`npm run quoted`)
- @windows 10
  - **Bad: no error whatsoever!**
  
  ``` bash
  $ npm run quoted
  
  > tslint-windows-path-inconsistency@1.0.0 quoted D:\github\tslint-windows-path-inconsistency
  > tslint 'src/**/*.ts'
  
  $
  ```
[...]

PR #1678, opened 2016-11-03
Normalize glob matching between Linux and Windows during TSLint build. Fix lint error.

IllusionMH:
#### PR checklist

- [X] ~~New feature,~~ bugfix, ~~or enhancement~~
  - [ ] Includes tests
- [ ] Documentation update

Currently builds on AppVeyor fails due to lint errors however Circle CI builds are fine which is strange
Keeping in mind discussion on bug #1639 I've checked it on Debian "Jessie" (Docker image for `node:6`) and looks like glob `src/**/*.ts` passed without quotes is expanded and matches files only in nested folders (e.g. `src/configs/latest.ts`), but not directly in `src` folder (e.g. `src/tslintMulti.ts `).

Since Windows doesn't expand glob patterns by itself only one glob string with patters is passed to `glob` module which correctly matches all files and leads to fail on lint error.

This PR normalizes behavior between platforms by preventing glob expand and delegates in to `glob` module.

`next` branch has same root problem, however there are other lint errors. Should I create second PR to `next` branch or it should be done by team members only?


JKillian:
Thanks @IllusionMH! This PR is fine, we'll fix those other lint errors when we merge things next. We should recommend that people put double quotes around their globs, it seems to be a source of confusion for many (including us haha).

PR #1679, opened 2016-11-04
Add support for glob patterns in single quotes for files and exclude arguments

IllusionMH:
#### PR checklist

- [X] Addresses an existing issue: #1639 _(related)_
- [X] ~~New feature, bugfix, or~~ enhancement
  - [X] Includes tests
- [ ] Documentation update

There were three ways to pass globs in npm scripts were mentioned in #1639:

1. Without quotes `src/**/*.ts`
  - works on Windows (`src/**/*.ts` is passed to process and matching is done with `glob` module)
  - doesn't match all files on Linux (see #1678)

2. With single quotes `'src/**/*.ts'`
  - works on Linux (`src/**/*.ts` is passed to process and matched with `glob` module)
  - doesn't match anything on Windows since process receives quotes string as argument `'src/**/*.ts'` and `glob` returns no matches with quotes

3. With double quotes `"src/**/*.ts"`
  - works both on Windows and Linux (`src/**/*.ts` is passed to process and matching is done with `glob` module)
  - requires escaping in `package.json file`  (e.g. `"lint": "tslint \"src/**/*.ts\""`) which hurts readability and might lead to errors if code editor doesn't notify about errors immediately

First case is OS specific and can't be addressed in TSLint package, but second might be properly handled.

This PR simply clears single quotes that may appear on Windows platform if glob is wrapped in single quotes in npm script.
Main intent is to increase interoperability between platforms and make npm scripts easy to write/read.

### DIFF
--- a/src/tslint-cli.ts
+++ b/src/tslint-cli.ts
@@ -26,7 +26,7 @@ import {
     DEFAULT_CONFIG,
     findConfiguration,
 } from "./configuration";
-import {consoleTestResultHandler, runTest} from "./test";
+import { consoleTestResultHandler, runTest } from "./test";
 import * as Linter from "./tslintMulti";
 
 let processed = optimist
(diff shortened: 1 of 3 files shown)


nchen63:
@IllusionMH thanks!

Issue #1674, opened 2016-11-03
semicolon rule should not complain about unnecessary semicolons on single-line lambdas


adidahiya:
see https://github.com/Microsoft/TypeScript/pull/12021#issuecomment-258173897

PR #1680, opened 2016-11-04
Treat one-line arrow funcs like a normal statement regarding semicolons

nchen63:
#### PR checklist

- [x] Addresses an existing issue: #1674 
  - [x] Includes tests
- [ ] Documentation update

Only apply rule of arrows and semicolons on multiline functions

(optional)

### DIFF
--- a/src/rules/semicolonRule.ts
+++ b/src/rules/semicolonRule.ts
@@ -121,7 +121,9 @@ class SemicolonWalker extends Lint.RuleWalker {
 
     public visitPropertyDeclaration(node: ts.PropertyDeclaration) {
         const initializer = node.initializer;
-        if (initializer && initializer.kind === ts.SyntaxKind.ArrowFunction) {
+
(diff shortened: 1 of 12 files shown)


src/rules/semicolonRule.ts
```diff
@@ -121,7 +121,9 @@ class SemicolonWalker extends Lint.RuleWalker {
 
     public visitPropertyDeclaration(node: ts.PropertyDeclaration) {
         const initializer = node.initializer;
-        if (initializer && initializer.kind === ts.SyntaxKind.ArrowFunction) {
+
+        // check if this is a multi-line arrow function (`[^]` in the regex matches all characters including CR & LF)
+        if (initializer && initializer.kind === ts.SyntaxKind.ArrowFunction && /\{[^]*\n/.test(node.getText())) {
```
adidahiya: this condition is getting pretty long; I think the last test with the regex belongs in its own simple `function`; that would allow you to move the comment elsewhere as well

Issue #1681, opened 2016-11-04
The error for invalid tslint.json files could be prettier


JoshuaKGoldberg:
As mentioned by @JKillian in #1676, it would be nice to have some sort of `validate` function and/or interface(s) generated for each rule. If someone uses an invalid `tslint.json` perhaps TSLint should complain? We can define invalid as having unknown or incorrect (e.g. a string instead of a number) options for a rule.

JKillian:
Perhaps a duplicate of https://github.com/palantir/tslint/issues/889? (Should've mentioned it in the other thread, sorry!) But maybe this is different - validating the whole file vs. just validating the rule options?

JoshuaKGoldberg:
Validating the whole file could be considered separate from the individual rules, sure. 

When you have technically invalid JSON the error isn't completely pretty:

```cmd
tslint --config tslint.json index.ts

undefined:4
    "rules": {
    ^

SyntaxError: Unexpected string in JSON at position 33
    at Object.parse (native)
    at loadConfigurationFromPath (C:\Code\tslint\lib\configuration.js:130:31)
    at Object.findConfiguration (C:\Code\tslint\lib\configuration.js:70:12)
    at LintCli.processFiles (C:\Code\tslint\lib\lintCli.js:43:49)
    at processFiles (C:\Code\tslint\lib\tslint-cli.js:132:30)
    at Object.<anonymous> (C:\Code\tslint\lib\tslint-cli.js:176:1)
    at Module._compile (module.js:541:32)
    at Object.Module._extensions..js (module.js:550:10)
    at Module.load (module.js:458:32)
    at tryModuleLoad (module.js:417:12)
```

...but yes, this is otherwise mostly a dup of #889.

PR #1686, opened 2016-11-05
Slightly better error for failed tslint.json parses

JoshuaKGoldberg:
Fixes https://github.com/palantir/tslint/issues/1681

`configuration`'s `findConfiguration` now returns a result wrapper with the attempted `path` as well as `error?` and `results?`. `tslint-cli` will log the failure nicely.


JoshuaKGoldberg:
@nchen63 added

test/executable/executableTests.ts
```diff
@@ -79,6 +79,17 @@ describe("Executable", function() {
                 done();
             });
         });
+
+        it("exits with code 1 if config file is invalid", (done) => {
+            execCli(["-c", "test/config/tslint-invalid.json", "src/tslint.ts"], (err, stdout, stderr) => {
+                assert.isNotNull(err, "process should exit with error");
+                assert.strictEqual(err.code, 1, "error code should be 1");
+
+                assert.include(stderr, "Invalid option for configuration", "stderr should contain notification about invalid json");
```
nchen63: this test isn't exercising the code in this PR. `Invalid option for configuration` is written when the config file is missing. I was expecting `Failed to load <path>: <error message>`

JoshuaKGoldberg: You're right - that was sloppy. Should be fixed.

nchen63:
@JoshuaKGoldberg thanks!

Issue #1631, opened 2016-10-15
README.md still references grunt for building


JoshuaKGoldberg:
It's all been converted to npm scripts, yes? It'd be nice to have a list of recommended steps.

PR #1687, opened 2016-11-05
Mentioned git --config settings in the README

JoshuaKGoldberg:
Also replaced `grunt` with some new `npm run` commands.

Cloning without endline config settings on Windows results in failed lint checks during tests because of different endline behavior. These are the same settings as https://github.com/microsoft/tslint-microsoft-contrib/#development.

### DIFF
--- a/README.md
+++ b/README.md
@@ -373,9 +373,10 @@ Development
 #### Quick Start
 
 ```bash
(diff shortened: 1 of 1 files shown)


adidahiya:
fixes #1631

PR #1688, opened 2016-11-05
Split CLI into tslint-cli and a Runner class

### LINKED ISSUE #1675: Expose CLI functionality as a separate class
_**Context**: I'm working on [autotslint](https://github.com/autolint/autotslint), a CLI tool to take in TSLint failures and, when possible, fix them in the files._

_**Ask**_: Can the existing logic in `tslint-cli.ts` be exposed publicly as a usable class and documented in the readme? Ideally we should be able to do something like the following:

Autotslint will need to be able to generate "waves" of TSLint failures: given the standard configurations that can be given via tslint-cli, it should generate failures for rules. It's possible to do this using the existing require("tslint") linter and some work, but would be much simpler if the existing CLI logic could be re-used.

```typescript
import {LintCli} from "tslint/lib/lintCli");

const options = {
    exclude: "**/*.d.ts",
    rules: {
        "variable-name": true,
        "quotemark": [true, "double"]
    }
};

const linter = new LintCli(options);

const result = linter.processFiles(["**/*.ts"]/*, ts.createProgram(...) */);
```

The CLI class would just be a refactor of tslint-cli.ts, not a change in any existing behavior. That being said, it would be nice to switch the options to have shorthand as aliases so the interface could use strings like `"config"` instead of `"c"`.

JoshuaKGoldberg:
Fixes #1675.

tslint-cli.ts takes in arguments using optimist and converts them to a friendlier object to pass to a CliRunner. The rest of the logic that used to be in tslint-cli.ts is now in the CliRunner.

Notes:
* I really wanted to make `CliRunner::run` return a `Promise<number>`, but didn't want to add new dependencies or change the output target.
* I tried to stick to the existing code style for the cut-and-pasted code. If you'd like it changed I'm up for that too.
* I'm not confident in `CliRunner` as a name. Alternate suggestions, anyone?

### DIFF
--- a/src/runner.ts
+++ b/src/runner.ts
@@ -1,0 +1,255 @@
+/**
+ * @license
+ * Copyright 2013 Palantir Technologies, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import * as fs from "fs";
+import * as glob from "glob";
+import * as path from "path";
+import * as ts from "typescript";
+
+import {
+    CONFIG_FILENAME,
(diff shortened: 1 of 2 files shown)


JoshuaKGoldberg:
👐 merge conflicts! I'll wait to resolve for if/until a design review passes.

JoshuaKGoldberg:
@nchen63 thanks for the ping, updated and renamed `cliRunner.ts`/`CliRunner` to `runner.ts`/`Runner`.

nchen63:
lgtm

nchen63:
@JoshuaKGoldberg thanks!

PR #1692, opened 2016-11-06
Add vscode launch.json

nchen63:
make tslint debuggable out of the box for vscode

### DIFF
--- a/src/tsconfig.json
+++ b/src/tsconfig.json
@@ -6,7 +6,7 @@
         "noUnusedParameters": true,
         "noUnusedLocals": true,
         "declaration": true,
-        "sourceMap": true,
+        "sourceMap": false,
         "target": "es5",
         "outDir": "../lib"
     }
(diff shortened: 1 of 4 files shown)


src/tsconfig.json
```diff
@@ -6,7 +6,7 @@
         "noUnusedParameters": true,
         "noUnusedLocals": true,
         "declaration": true,
-        "sourceMap": true,
+        "sourceMap": false,
```
adidahiya: why do we have to set this to false?

nchen63: oh, I was just undoing a change in https://github.com/palantir/tslint/commit/1b76b85f17fbda8ffb1f63d2cd7c48d4f3aa1be1. Since we only need the source maps in ./build for debugging, and don't need it in ./lib. Doesn't hurt to set it to true, but need to set npmignore for .js.map

adidahiya: ah, ok, nvm, carry on then

Issue #990, opened 2016-02-22
Use update-notifier to notify users of new versions in a non-intrusive way


adidahiya:
https://www.npmjs.com/package/update-notifier

davieschoots:
Hi @adidahiya, 

I would like to help by adding this feature to the tslint cli, can you let me know if this is ok?

Do have some questions, but I'll ask them if you're ok with it :)

Kind regards,

Davie

adidahiya:
Sure, sounds good to me, we can discuss here

davieschoots:
Cool, thank you!

I'm wondering what your thoughts are about the default message and the interval for the check on updates. 

I think for the update check once every week or 2 weeks should be fine (the default it's once a day). 

Are you ok with this and do you prefer having a custom message or leave the default?

davieschoots:
Hi @nchen63, this issue should be on closed, right?

PR #1696, opened 2016-11-07
Use update-notifier to notify of new versions #990

davieschoots:
#### PR checklist

- [x] Use update-notifier to notify users of new versions in a non-intrusive way #990
- [ ] Includes tests
- [ ] Documentation update

I've added the functionality of the [update-notifier](https://github.com/yeoman/update-notifier) to have it display the update message when there is a new tslint version available. 

The functionality is inside a new file called 'updateNotifier.ts', to keep it separate and because I need to require the plugin as well as the package.json. The new function is implemented via the cli.

Do you think implementing the function call at this level is correct? 

My reasons for placing it here is because it then runs every time. The interval for checking a new version is set to three days (default is once a day).

I've tested it by setting the 'updateCheckInterval' to 0 and by lowering the version inside the package.json to '3.15.0', so it would trigger the check.

This triggers the image below:
update-message

### DIFF
--- a/package.json
+++ b/package.json
@@ -41,7 +41,8 @@
     "glob": "^7.1.1",
     "optimist": "~0.6.0",
     "resolve": "^1.1.7",
-    "underscore.string": "^3.3.4"
+    "underscore.string": "^3.3.4",
+    "update-notifier": "^1.0.2"
   },
   "devDependencies": {
     "@types/chai": "^3.4.34",
--- a/src/tslint-cli.ts
+++ b/src/tslint-cli.ts
@@ -28,6 +28,7 @@ import {
 } from "./configuration";
 import { consoleTestResultHandler, runTest } from "./test";
 import * as Linter from "./tslintMulti";
+import { updateNotifierCheck } from "./updateNotifier";
 
 let processed = optimist
     .usage("Usage: $0 [options] file ...")
@@ -267,6 +268,11 @@ const processFiles = (files: string[], program?: ts.Program) => {
             process.exit(argv.force ? 0 : 2);
         }
     });
+
+    if (lintResult.format === "prose") {
+        // Check to see if there are any updates available
+        updateNotifierCheck();
+    }
 };
 
 // if both files and tsconfig are present, use files
--- a/src/updateNotifier.ts
+++ b/src/updateNotifier.ts
@@ -1,0 +1,36 @@
+/**
+ * @license
+ * Copyright 2016 Palantir Technologies, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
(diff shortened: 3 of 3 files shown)


palantirtech:
Thanks for your interest in [palantir/tslint](https://github.com/palantir/tslint), @davieschoots! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

nchen63:
Nice. This may not work well with apps that consume tslint via a special formatter. For example, if the JSON formatter was used, the output would become invalid. Can you make the check only occur with prose output format?

davieschoots:
Hi @nchen63, sure no problem. I've moved and pushed the change and think this is what you wanted.

nchen63:
it looks like update-notifier asks users to install using -g, but the running tslint may not be a global install

davieschoots:
Yes, there is a default option message which we can override. I'm happy to add this if you want. What would be your suggestion as text or do you think it's not desirable just to get rid of the -g?

nchen63:
I'd prefer to point people to the change log since there may be breaking changes that require more than doing an npm install.

Also, I have a concern about whether this will work in Windows since I saw unix type commands in the referenced libraries

nchen63:
also one of the files is missing the license header

davieschoots:
Sorry, I'll copy the license header and update the new file. 

About the message, we could do something like below;
'There is a new update of tslint, please check our change log before updating'

I don't think it won't work on windows, but will definitely test it to see if it works. 

Also I can remove the require for package.json, then use tslint as a fixed name value and get the version from Linter.VERSION instead of using it from the package.

nchen63:
not working in windows is a deal breaker

davieschoots:
Yes, it sure is. I'm setting up a test right now :)

davieschoots:
Hi @nchen63 , I've did a test on windows 10 on cmd and powershell and it looks good. 

cmd
powershell

IllusionMH:
Also looks good in Git Bash 👍 
git-bash-update-notifier

Is it possible to detect is it local (from `dependencies` or `devDependencies`) or global installation to provide proper update instruction (`message` option)? Since updating global won't fix outdated version on CI.

davieschoots:
I've added all the changes:
- removed extra spaces
- added license
- added the desired text for when an update is available

I'll still want to do some final testing on windows

davieschoots:
@nchen63 I tested it on windows in powershell, cmd and git bash. 

screen shot 2016-11-10 at 20 50 00

This is my first PR in a large codebase and I want to thank you for your patience!

@IllusionMH Thanks for testing and the review!

nchen63:
very nice. Thanks for the contribution. Happy to help.

PR #1697, opened 2016-11-07
Use --fix instead of -t applyFixes

### LINKED ISSUE #1684: Use `--fix` instead of `-t applyFixes`
Before we 4.0, I'd like to change the parameters so that fixes are not applied using a formatter

- ApplyFixes is not a typical formatter. Formatters are usually used to format the linter errors
- Using a formatter limits the number of cycles that the fixes can be applied in since all fixes are batched and applied at one time. Multiple cycles are needed since fixes can overlap.
- If we move to an AST transform based fixer in the future, fixes will most likely be done at the time the linting error is generated.
- Consistent with eslint

nchen63:
#### PR checklist

- [x] Addresses an existing issue: closes #1684 
  - [x] Includes tests
- [x] Documentation update

When fixes are applied, they are applied first. If a rule triggers a fix, the AST is reloaded to prevent one fix from overwriting another fix. When all fixes are complete, the sourceFile is linted again if there were fixes. That way, we don't show errors that are already fixed and the positions will be correct.

here's what it looks like when using the prose formatter (the only formatter to output that there are fixes) when there is 1 issue fix and 1 remaining issue that can't be fixed:
```
Fixed 1 error(s) in ../a.ts

../a.ts[6, 1]: Exceeds maximum line length of 140
```


src/language/formatter/formatter.ts
```diff
@@ -47,5 +47,5 @@ export interface IFormatterMetadata {
 export type ConsumerType = "human" | "machine";
 
 export interface IFormatter {
-    format(failures: RuleFailure[]): string;
+    format(failures: RuleFailure[], fixes?: RuleFailure[]): string;
```
adidahiya: is the idea that `fixes` are only supplied to this method if `--fix` is enabled in the CLI? can you leave a comment explaining that?

src/tslint-cli.ts
```diff
@@ -50,6 +50,9 @@ let processed = optimist
             alias: "exclude",
             describe: "exclude globs from path expansion",
         },
+        fix: {
+            describe: "Fixes some linting errors (may overwrite linted files)",
```
adidahiya: "some" isn't great language -- can you just use the same description you have on L159

src/tslintMulti.ts
```diff
@@ -89,54 +90,34 @@ class MultiLinter {
...
-                ruleFailures = rule.applyWithProgram(sourceFile, this.program);
-            } else {
-                ruleFailures = rule.apply(sourceFile);
-            }
-            for (let ruleFailure of ruleFailures) {
-                if (!this.containsRule(this.failures, ruleFailure)) {
-                    this.failures.push(ruleFailure);
-                }
+        // make a 2nd pass if there were any fixes because the positions may be off        
+        if (!this.options.fix || this.fixes.length > 0) {
```
adidahiya: the first part of this conditional expression looks unintuitive to me -- don't you want to only continue if `this.options.fix` is true?

nchen63: You want to run the linter at least once if `fix` is false. Will try to clarify

src/language/formatter/formatter.ts
nchen63: yep

nchen63:
@adidahiya made changes

adidahiya:
heads up @alexeagle

PR #1701, opened 2016-11-10
Use denormalized path in test (fixes Windows build)

IllusionMH:
#### PR checklist

- [X] ~~New feature,~~ bugfix, ~~or enhancement~~
  - [ ] Includes tests
- [ ] Documentation update

Currently build fails since normalized path returned from Node.js API (with `\` as path separator) is compared to deormalized one (with `/` as path separator, which is also valid on Windows) which was picked from TypeScript.

This PR fixes build on Windows by converting normalized path to denormalized and using it in comparison.

I've reviewed 3 possible solutions:

1. Normalize `fileName`'s when they are stored in `RuleFailure` class (used only in formatters) 

1. Normalize only absolute `fileName`'s when they are stored in `RuleFailure` class (probably rare case, but matches problem in test)

1. Use denormalized path in test

While first two solution fix problem, I think they bring almost no value and will lead to inconsistency in paths output between platforms.

### DIFF
--- a/test/executable/executableTests.ts
+++ b/test/executable/executableTests.ts
@@ -123,12 +123,12 @@ describe("Executable", function() {
             execCli(["-c", "test/files/multiple-fixes-test/tslint.json", tempFile, "--fix"],
                 (err, stdout) => {
                     const content = fs.readFileSync(tempFile, "utf8");
-                    // compare against file name which will be retyrned by formatter (used in TypeScript)
-                    const denornalizedFileName = denormalizeWinPath(tempFile);
+                    // compare against file name which will be returned by formatter (used in TypeScript)
+                    const denormalizedFileName = denormalizeWinPath(tempFile);
                     fs.unlinkSync(tempFile);
                     assert.strictEqual(content, "import * as y from \"a_long_module\";\nimport * as x from \"b\";\n");
                     assert.isNull(err, "process should exit without an error");
-                    assert.strictEqual(stdout, `Fixed 2 error(s) in ${denornalizedFileName}`);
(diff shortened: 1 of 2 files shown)


test/executable/executableTests.ts
```diff
@@ -123,10 +123,12 @@ describe("Executable", function() {
             execCli(["-c", "test/files/multiple-fixes-test/tslint.json", tempFile, "--fix"],
                 (err, stdout) => {
                     const content = fs.readFileSync(tempFile, "utf8");
+                    // compare against file name which will be retyrned by formatter (used in TypeScript)
```
nchen63: typo `retyrned`

test/executable/executableTests.ts
```diff
@@ -123,10 +123,12 @@ describe("Executable", function() {
...
+                    const denornalizedFileName = denormalizeWinPath(tempFile);
                     fs.unlinkSync(tempFile);
                     assert.strictEqual(content, "import * as y from \"a_long_module\";\nimport * as x from \"b\";\n");
                     assert.isNull(err, "process should exit without an error");
-                    assert.strictEqual(stdout, `Fixed 2 error(s) in ${tempFile}`);
+                    assert.strictEqual(stdout, `Fixed 2 error(s) in ${denornalizedFileName}`);
```
nchen63: typo `denornalizedFileName`

IllusionMH:
@nchen63 thank you for review. Next time I will look more carefully to avoid such typos.

nchen63:
@IllusionMH no problem. Thanks for the contribution!

PR #1707, opened 2016-11-11
Extend adjacent-overload-signatures to work in classes, source files, modules, and namespaces

andy-ms:
#### PR checklist

  - [X] Includes tests
- [ ] Documentation update

The adjacent-overload-signatures rule now works in situations other than interfaces and type literals.
I don't think this needs a documentation update, since the existing documentation doesn't say anything about being specific to just those.


nchen63:
nice, thanks!

PR #1717, opened 2016-11-14
Add recommended rules

### LINKED ISSUE #1507: Update tslint:recommended configuration for 4.0 release
#Let's track the breaking changes we want to make to the `tslint:recommended` config here.
- [ ] TODO: list of changes
  - [x] **deprecate** `no-unused-variable` #1618 
  - [x] **remove** `no-duplicate-key` #885 
  - [x] **remove** `label-undefined` #877 
  - [x] **remove** `use-strict` #678 
  - [x]  **remove** `no-unreachable` #661 
  - [ ] **add** `no-unsafe-finally`
  - [ ] **add** `ordered-imports`
    - compared to current `tslint:latest`, we want to **change** the `named-imports-order` option to `"case-insensitive"`
  - [ ] **add** new ones from #1506 
  - [ ] **add** `max-classes-per-file`
  - [ ] **add** `prefer-for-of`
  - [ ] **add** `array-type`
  - [ ] **add** `adjacent-overload-signatures`
  - [ ] **add** `object-literal-key-quotes`
  - [ ] **add** `object-literal-shorthand`?
  - [ ] **add** `only-arrow-functions`/`allow-declarations`

---

Also, I was thinking it might be useful for users to have access to an older stable configuration so that they don't have to address all of 4.0's breaking changes at once. Perhaps a new config called `tslint:3.x` that is identical to the current `tslint:latest`?

nchen63:
fixes #1507

### DIFF
--- a/src/configs/latest.ts
+++ b/src/configs/latest.ts
@@ -16,16 +16,6 @@
  */
 
 export const rules = {
-    "adjacent-overload-signatures": true,
-    "cyclomatic-complexity": false,
-    "no-unsafe-finally": true,
-    "object-literal-key-quotes": [true, "as-needed"],
-    "object-literal-shorthand": true,
-    "only-arrow-functions": [true, "allow-declarations"],
-    "ordered-imports": [true, {
-        "import-sources-order": "case-insensitive",
-        "named-imports-order": "lowercase-last",
-    }],
 };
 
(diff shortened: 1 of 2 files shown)


adidahiya:
how about `array-type` and `arrow-parens`?

PR #1718, opened 2016-11-14
Removes `no-duplicate-variable` from the default & recommended configs

elwayman02:
Per the [docs](https://palantir.github.io/tslint/rules/no-duplicate-variable/) for `no-duplicate-variable` itself:

> This rule is only useful when using the var keyword - the compiler will detect redeclarations of let and const variables.

Given that `no-var-keyword` is already included in both configs, `no-deuplicate-variable` becomes redundant since `var` will never be allowed in those configs.

### DIFF
--- a/src/configs/recommended.ts
+++ b/src/configs/recommended.ts
@@ -53,7 +53,6 @@ export const rules = {
     ],
     "no-construct": true,
     "no-debugger": true,
-    "no-duplicate-variable": true,
     "no-empty": true,
     "no-eval": true,
     "no-internal-module": true,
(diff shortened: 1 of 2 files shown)


palantirtech:
Thanks for your interest in [palantir/tslint](https://github.com/palantir/tslint), @elwayman02! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

elwayman02:
Done.

PR #1720, opened 2016-11-14
Simplify API

### LINKED ISSUE #1595: Simplify typings for API consumers
right now we have both `lib/tslint.d.ts` and `lib/lint.d.ts`. let's try to consolidate them into one file.

via #1592

nchen63:
- Remove `tslint.ts` and interfaces associated with single file `Linter`
- Rename `tslintMulti.ts` -> `linter.ts`
- Rename `lint.ts` -> `index.ts`
- Remove old `ILinterOptions` and rename `IMultiLinterOptions` `ILinterOptions`
- Reference `index.js/index.d.ts` in `package.json`

fixes #1595


adidahiya:
this looks good, but I forgot about @JKillian's comment here: https://github.com/palantir/tslint/issues/1595#issuecomment-251567070

how much effort will be required for third-party tools to transition to using the new `Linter` class?

nchen63:
It's pretty straightforward to migrate to the new `Linter` class. You can see the changes that I made to the `test.ts` class here: https://github.com/palantir/tslint/pull/1720/files#diff-c1e1ffc130aead13aa199b62448b178c

- move the `configuration` property off of `ILinterOptions` and pass it into `lint()`
- move filename and source from the constructor to `lint()`
- you have to call `Linter::getResult()` to get the result instead of being returned from `lint()`

nchen63:
I also made the new linter throw an error if `configuration` was included on the ILinterOptions object since *not* moving it would still work but use the default config

panuhorsmalahti:
I noticed that 'findConfiguration' API seems to have been changed a bit (or maybe the function which uses it's result). Anyway, I had to change my code to use the `.results` property of the return value of `findConfiguration`. Maybe this will help someone.

Issue #629, opened 2015-09-01
Cleaner Configuration of Rules


JKillian:
There are currently a few different possible levels of complexity of options for rules:
1. enabled/disabled only (ex: `no-debugger`)
2. flag options (ex: `check-whitespace`, each option indicates whether something should be checked or not)
3. fixed length argument lists (ex: `max-line-length`)
4. variable length argument lists (ex: `ban`)
5. key/value pairs, nested lists and key/value pairs, etc. (no current rules)

The current Boolean/Array paradigm of declaring options only supports the first two of this optimally. It also is functional for the third option, but doesn't really model it well. It's not ideal at all for the fourth case. 

There are more and more feature requests which require a more descriptive way to list options (https://github.com/palantir/tslint/issues/562 and https://github.com/palantir/tslint/issues/522), and it makes sense to have a system that better facilitates rules like this.

Any thoughts on how we should do this?

alexkrauss:
Most importantly, any refactoring here should make rule arguments statically typed in some way. That is, a rule should declare which arguments it knows about, and what type they have.

With the current model, it is pretty much impossible to implement, e.g., a GUI that allows configuration of tslint rules.

JKillian:
@vvakame wondering if you have any feedback regarding this? I know `tsfmt` provides an option to read from `tslint.json`

vvakame:
You do not have to care! I'll support new tslint.json.

JKillian:
I think something similar to how ESLint does thing might work well. Here's [an example](https://github.com/eslint/eslint/blob/87caa64bfcc76639f71c45bcb832e53237b07f2d/lib/rules/brace-style.js#L216L229) of how a rule declares its configuration schema.

JKillian:
Proposal for new config schema (taken from #345):

I'm thinking about implementing this soon, I like the idea of each rule taking an object as configuration. (This is also very nice code-wise, since rules only accept a boolean or array right now, we can implement this feature with no backwards-compat breakage.)

I like a format similar to what's specified above:

```
"rules": {
  "align": {
    "mode": "off",
    "options": ["parameters", "arguments"]
  }
}
```

`"mode"`: could be either `"off"`, `"warn"`, `"error"`, or `"on"`. If `"on"`, the default level of warning would be used.

Open questions:
  * Should we stick with the ESLint way of things and also allow `0`, `1`, `2`, and `3` instead of "off", "warn", "error", "on", etc?
  * Do we want an "on" option at all?
[...]

Issue #345, opened 2015-03-12
Add severity of errors


lachvoj:
Please add severity of errors into rules to distinguish if it's functionality issue or just code readability issue.

Lexicality:
I support this - I would very much like to be able to fail the build due to `class-name` violations, but not `comment-format`.
It'd be great if you could do it in the same way as ESLint where you can specify `off`, `warning` or `error` when activating a rule in the config.

readme42:
:+1: this would be great!

JKillian:
:+1: I like this idea as well

pgatilov:
:+1: please!

JKillian:
Related to #629 perhaps. Rules could have a default severity that could be overriden in `tslint.json`

JoshuaKGoldberg:
+1, this would be great

Would it be feasible to add a `"warnings"`option to `tslint.json`s?

JKillian:
I'm thinking about implementing this soon, I like the idea of each rule taking an object as configuration. (This is also very nice code-wise, since rules only accept a boolean or array right now, we can implement this feature with no backwards-compat breakage.)

I like a format similar to what's specified above:

```
"rules": {
  "align": {
    "mode": "off",
    "options": ["parameters", "arguments"]
  }
}
```

`"mode"`: could be either `"off"`, `"warn"`, or `"error"`.  Open questions:
  * Should we stick with the ESLint way of things and also allow `0`, `1`, and `2`?
  * I like the name "mode" because it's easy to type, but should we pick a different name?

I also like how this proposal makes it easy to add in additional options for things like auto-fixing, as mentioned above.

Lexicality:
I'm _fairly_ sure the number severities in ESLint are backwards compatibility only. Unless you use the same rule format (eg an array rather than an object) there's not much point I would say.
Perhaps "level" instead of mode?

JKillian:
Realized #629 is a better issue to discuss rule configuration schema issues, so moved some comments over there!

adidahiya:
shipped in 5.0

ErikCupal:
@adidahiya Sorry if I missed something, but how to turn all errors back to warnings in version 5.0? I tried this so far..

```json
{
    "defaults": {
        "severity": "warning"
    },
    ...
}
```

adidahiya:
@ErikCupal `"defaultSeverity": "warning"`. see https://palantir.github.io/tslint/usage/configuration/

nchen63:
[...]

PR #1738, opened 2016-11-17
Support for ERROR & WARN per rule

olore:
#### PR checklist

- [x] Addresses an existing issue: #629 #345 
  - [x] Includes tests
- [x] Documentation update

Allow a "mode" for each rule in tslint.json. Anything other than "off" and the rule will run, a value of "warn" will exit with a return code of 0
```
  "rules": {
    "interface-name": {
      "severity": "off"
    },
    "max-line-length": {
      "severity": "error",
      "options": 140
    },
    "member-ordering": {
      "severity": "warn",
      "options": [
        "public-before-private",
        "static-before-instance",
        "variables-before-functions"
      ]
    }
}
```

* I have a few questions, please see them as comments within the diff


palantirtech:
Thanks for your interest in [palantir/tslint](https://github.com/palantir/tslint), @olore! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

test/external/tslint-test-custom-rules/rules/ruleOneRule.js
```diff
@@ -12,6 +12,7 @@ var Rule = (function (_super) {
     Rule.prototype.apply = function (sourceFile) {
         return this.applyWithWalker(new NoFailWalker(sourceFile, this.getOptions()));
     };
+    Rule.prototype.isWarning = function() { return false; };
```
olore: I couldn't figure out where these files were generated from, so I just modified by hand.

test/formatters/stylishFormatterTests.ts
```diff
@@ -45,19 +43,11 @@ describe("Stylish Formatter", () => {
 
         const maxPositionTuple = `${maxPositionObj.line + 1}:${maxPositionObj.character + 1}`;
 
-        const expectedResult = colors.enabled ?
```
olore: this test was never run with colors disabled, so I removed the ternary

src/language/rule/abstractRule.ts
```diff
@@ -29,6 +30,8 @@ export abstract class AbstractRule implements IRule {
 
         if (Array.isArray(value) && value.length > 1) {
             ruleArguments = value.slice(1);
+        } else if (value.options) {
+            ruleArguments = arrayify(value.options);
```
olore: Allow the tslint.json maintainer to no be forced to put a single value in an array

olore:
@IllusionMH That makes a lot of sense to me. In fact it's where I started, but because of the huge number of changes, I backed it out. What do you think about moving RuleFailure to RuleViolation? A RuleViolation can either be a WARNING or an ERROR.

To me, it's a cleaner solution, I was just apprehensive about making so many changes (100+ files).

IllusionMH:
I think there is no need for 100+ files change (although this level of changes probably will be required for rule test configuration files).

Probably I incorrectly used
> extending `RuleFailure` interface with `mode` property

which lead to incomprehension.
I meant add `mode/level` property directly to this interface.

My rough estimation of changes to add `level` into `RuleFailure `.
 (I'll use it instead of `mode` since I prefer this option and I think there were no final word on term name)

File `src/language/rule/rule.ts`
    * Add `level` to `IOptions`
    * Add `level` to `RuleFailure` `constructor`/`toJSON`
    * Add `getLevel` method

File `src/language/rule/abstractRule.ts`
    * Add level to `this.options` initialization

File `src/language/walker/ruleWalker.ts`
    * Add private member `level` from `options.level`
    * Add `level` to `createFailure` method
    * Add `getLevel` method

File `src/linter.ts`
    * Use filter failures array to get counts

Update formatters and utils tests (14 files) and `eofline` rule since they use `RuleFailure` constructor directly.

At this moment I'm not 100% sure if this is the best solution, since `RuleWalker` will store `level` however don't directly rely on it 😞

olore:
Thanks @IllusionMH ! We were on the same page. I ended up with so many changes because I couldn't get past the name RuleFailure, so I renamed it RuleViolation. Each Violation can either be an ERROR or a WARNING (aka a RuleLevel).

Hopefully this makes sense. Appreciate the feedback and suggestions.

olore:
After reading through some of the PRs and Issues, I think I underestimated how many custom rules are out there, and changing RuleFailure to RuleViolation will affect all of them :(

I will change back to RuleFailure

tslint.json
```diff
@@ -1,20 +1,32 @@
...
-      "variables-before-functions"
-    ],
-    "no-unused-variable": false,
-    "variable-name": [true,
-      "ban-keywords",
-      "check-format",
-      "allow-leading-underscore",
-      "allow-pascal-case"
-    ]
+    "interface-name": {
```
IllusionMH: Probably it is too early to change `tslint.json` for project since it uses latest stable package.

IllusionMH: It uses it from `node_modules`:
1. `package.json` refers it as `tslint` so this executable should be in path (which was done by npm) (just replace `lint` script to `which tslint`)

2. All rules in new format are just disabled (check `isEnabled` logic in `AbstractRule`)
Simply use 10 for `max-line-length` rule and there will be no violations. Restore previous format and you'll see all of them.

3. You can alter constructor in `node_modules/tslint/lib/language/rule/abstractRule.js` to print all not enabled rules - you'll see all of rules that has new format.

olore: You are 100% correct. I was running the linter directly to use the new format, I should have been using `npm run lint`. It's a little weird to use the project to lint itself 😄 , but I get it.

olore:
@adidahiya @nchen63 @IllusionMH Sorry I have been distracted by other things, but I want to get back to this. What can I do to help get this rolling to the next step.

Appreciate your time and any feedback you may have.

test/formatters/pmdFormatterTests.ts
```diff
@@ -33,10 +33,10 @@ describe("PMD Formatter", () => {
...
 
         const failures = [
-            new RuleFailure(sourceFile, 0, 1, "first failure", "first-name"),
-            new RuleFailure(sourceFile, 2, 3, "&<>'\" should be escaped", "escape"),
-            new RuleFailure(sourceFile, maxPosition - 1, maxPosition, "last failure", "last-name"),
-            new RuleFailure(sourceFile, 0, maxPosition, "full failure", "full-name"),
+            new RuleFailure(sourceFile, 0, 1, "first failure", RuleLevel.ERROR, "first-name"),
+            new RuleFailure(sourceFile, 2, 3, "&<>'\" should be escaped", RuleLevel.ERROR, "escape"),
+            new RuleFailure(sourceFile, maxPosition - 1, maxPosition, "last failure", RuleLevel.ERROR, "last-name"),
+            new RuleFailure(sourceFile, 0, maxPosition, "full failure", RuleLevel.ERROR, "full-name"),
```
nchen63: should change priority depending on level (3 = error, 4 = warning according to https://pmd.github.io/pmd-5.3.3/customizing/rule-guidelines.html). Not sure why it was written with priority=1 which means critically broken. Also a breaking change

olore:
@nchen63 Thank you for the feedback! I will try to address these items this week.

test/formatters/checkstyleFormatterTests.ts
```diff
@@ -38,12 +38,12 @@ describe("Checkstyle Formatter", () => {
         const maxPosition2 = sourceFile2.getFullWidth();
 
         const failures = [
-            new RuleFailure(sourceFile1, 0, 1, "first failure", "first-name"),
-            new RuleFailure(sourceFile1, 2, 3, "&<>'\" should be escaped", "escape"),
-            new RuleFailure(sourceFile1, maxPosition1 - 1, maxPosition1, "last failure", "last-name"),
-            new RuleFailure(sourceFile2, 0, 1, "first failure", "first-name"),
-            new RuleFailure(sourceFile2, 2, 3, "&<>'\" should be escaped", "escape"),
-            new RuleFailure(sourceFile2, maxPosition2 - 1, maxPosition2, "last failure", "last-name"),
+            new RuleFailure(sourceFile1, 0, 1, "first failure", RuleLevel.ERROR, "first-name"),
```
olore: done

test/formatters/msbuildFormatterTests.ts
```diff
@@ -16,7 +16,7 @@
 
 import * as ts from "typescript";
 
-import {IFormatter, RuleFailure, TestUtils} from "../lint";
+import {IFormatter, RuleFailure, RuleLevel, TestUtils} from "../lint";
```
olore: done. Is there a way to indicate it's a breaking change for documentation purposes? Or should I just note it in the CHANGELOG.md ?

src/language/rule/abstractRule.ts
```diff
@@ -34,18 +34,30 @@ export abstract class AbstractRule implements IRule {
...
+        if (ruleConfigValue.level !== "off" && ruleConfigValue.level !== "none") {
+            return true;
+        }
+
         return false;
     }
 
     constructor(ruleName: string, private value: any, private disabledIntervals: IDisabledInterval[]) {
         let ruleArguments: any[] = [];
+        let ruleLevel = RuleLevel.ERROR;
```
olore: What are you thinking here? Allow the config file to specify a default level for all? Sounds doable.

test/formatters/msbuildFormatterTests.ts
nchen63: Yeah, breaking means that this will have to wait for the v5.0 release. Can you put something in CHANGELOG.md and comment it out? Something like:

```
<!--
v5.0.0
---

* **BREAKING CHANGES**
    *  The severity level is now configurable and the default severity may have changed:
        - "msbuild" was "warning", but now defaults to "error"
        - "vso" ......
-->

src/language/rule/abstractRule.ts
olore: Still working on this

test/formatters/msbuildFormatterTests.ts
olore: Done

olore:
1) Regarding the change from RuleLevel -> RuleSeverity - I'm OK with it, but since naming is tricky should we try to get consensus?

2) Regarding the defaultLevel - it's a little more nuanced than I realized. Could this be done in a separate PR, but still in time for 5.0 ?

Thanks again for the assistance @nchen63 !

adidahiya:
@olore 

> Regarding the change from RuleLevel -> RuleSeverity - I'm OK with it, but since naming is tricky should we try to get consensus?

ESLint calls it "severity", so 👍  for aligning with that.

> Regarding the defaultLevel - it's a little more nuanced than I realized. Could this be done in a separate PR, but still in time for 5.0 ?

Yes, a separate PR sounds good.

olore:
@nchen63 All requested changes have been made. 

Three outstanding items -
1. CircleCI is failing, but `npm run verify` works locally. I'm still digging at it, but if anyone has a suggestion, I'd appreciate it.
1. I could use a hand with the CHANGELOG, I added info, but it isn't exactly pretty.
2. Should the update to the schemastore (mentioned as non blocking) be done such that both config file formats are valid or just the new one?

nchen63:
the tests fail on my machine too. It's failing because `tslint-test-custom-rules` is installed  in`test/config/node_modules` which can be found by configs in `test/config` while the config file for the failing test lives in `test/config-legacy`, which can't see the node module.

honestly, you should get rid of the `config-legacy` directory since it's just a copy of `config` but with the severity specified. It's not necessary to test every permutation of every feature with every other feature since it just adds more noise and would grow tests exponentially. Just add a few more tests to ensure that inheritance for severity is tested well. 

some examples:

- true -> severity: none
- severity: error -> severity: warn
- severity: error -> severity: error
- false -> severity: none
- true -> severity: warn

this might get complicated later on with inheriting default severity. For example, does a new default severity just override the old default or does it wipe out all of the custom levels? Do we need a "default" severity?

olore:
Thanks for the insight. That shouldn't be a problem. I was apprehensive about killing the legacy stuff, but it definitely felt like extra baggage.

Now you feel my pain with trying to do a default severity 😄

nchen63:
I don't think you should necessarily kill the legacy stuff. You should actually leave it mostly unchanged since they test order of inheritance more than anything else.

olore:
I believe I have addressed all items. Thanks for your input.

adidahiya:
ping @nchen63 -- what are the remaining actions here and what is the feature rollout plan?

nchen63:
@olore nice work

we are now on our way to 5.0!

tugberkugurlu:
Has this been released?

JKillian:
@tugberkugurlu not in a regular release, but you can try it in this prerelease: https://github.com/palantir/tslint/releases/tag/5.0.0-dev.0

prigara:
We're working on the TSLint integration in WebStorm. Can you please point me to a comment or code describing the difference between `warn`, `warning` and `none`. Thanks!

nchen63:
`warn` and `warning` are the same thing. `none` means that the rule will be ignored

PR #1739, opened 2016-11-18
Need to add @types/node to reg dependencies since we use `Error` interface

nchen63:
Otherwise you see compilation errors on some of our d.ts files when installing tslint as a npm library


adidahiya:
This doesn't look right. `@types` should _never_ be in the regular dependencies block. Consumers need to install node typings on their own.

PR #1740, opened 2016-11-18
Revert "Need to add @types/node to reg dependencies since we use `Error` interface"

adidahiya:
Reverts palantir/tslint#1739

### DIFF
--- a/package.json
+++ b/package.json
@@ -35,7 +35,6 @@
     "verify": "npm-run-all clean compile lint test docs"
   },
   "dependencies": {
-    "@types/node": "^6.0.45",
     "colors": "^1.1.2",
     "diff": "^3.0.1",
     "findup-sync": "~0.3.0",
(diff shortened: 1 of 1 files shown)


nchen63:
Yeah, it's not the right thing to do, but we should still address this error when compiling when tslint is installed as a library:

```
node_modules/tslint/lib/configuration.d.ts(1,1): error TS2688: Cannot find type definition file for 'node'.
```

adidahiya:
@nchen63 how exactly did you get that error -- did you set up a new test project? We could make a note in the docs

nchen63:
Yeah, I set up a test project and installed from an `npm pack` package

nchen63:
I feel like there should be a way to make it work that doesn't require users to install node typings. Pre-2.0 typescript did not generate the `/// <reference types="node" />` header which causes this issue. If we can avoid exporting functions that use the `Error` interface, that would work too

PR #1741, opened 2016-11-18
Create local Error interface so that node doesn't need to be externally referenced

nchen63:
The problem: when using TSLint as an NPM library, you see error:

```
node_modules/tslint/lib/configuration.d.ts(1,1): error TS2688: Cannot find type definition file for 'node'.
```

Why: @types/node augments the `Error` interface that was initially defined by TypeScript's `lib.d.ts`. Since the use of this interface is exported by TSLint, the app using TSLint must also include @types/node. Sadly, the property added by Node is currently already defined by `lib.d.ts`

The fix: Don't export the @types/node `Error` interface and define/use our own similar interface.

### DIFF
--- a/src/test/lintError.ts
+++ b/src/test/lintError.ts
@@ -14,7 +14,7 @@
  * limitations under the License.
  */
 
-export { Error } from "../configuration";
+import { Error } from "../configuration";
 
 export interface PositionInFile {
    line: number;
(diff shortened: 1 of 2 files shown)


src/test/lintError.ts
```diff
@@ -14,6 +14,8 @@
  * limitations under the License.
  */
 
+export { Error } from "../configuration";
```
adidahiya: Why does this compile? This import is just an interface, it doesn't define a call or constructor signature. I think L45 is still referencing node's `Error` constructor.

nchen63: arg, you're right

PR #1745, opened 2016-11-19
Allow null keyword in a type position (originally #1277)

nchen63:
This checkin got lost in `next`


src/linter.ts
```diff
@@ -41,7 +41,7 @@ import { arrayify, dedent } from "./utils";
  * Linter that can lint multiple files in consecutive runs.
  */
 class Linter {
-    public static VERSION = "4.0.0-dev.2";
+    public static VERSION = "4.0.0";
```
adidahiya: this line change looks wrong

PR #1749, opened 2016-11-19
Fix parameter type of hasModifier

ajafff:
#### PR checklist

  - [ ] Includes tests
- [ ] Documentation update

Avoid compiler error with strictNullChecks.
hasModifier already checks for null and undefined.
This commit adds types null and undefined to the modifiers parameter.

### DIFF
--- a/src/language/utils.ts
+++ b/src/language/utils.ts
@@ -77,8 +77,8 @@ export function scanAllTokens(scanner: ts.Scanner, callback: (s: ts.Scanner) => 
 /**
  * @returns true if any modifier kinds passed along exist in the given modifiers array
  */
-export function hasModifier(modifiers: ts.ModifiersArray|null|undefined, ...modifierKinds: ts.SyntaxKind[]) {
-    if (modifiers == null || modifierKinds == null) {
+export function hasModifier(modifiers: ts.ModifiersArray|undefined, ...modifierKinds: ts.SyntaxKind[]) {
+    if (modifiers === undefined || modifierKinds.length === 0) {
         return false;
     }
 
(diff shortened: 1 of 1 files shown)


palantirtech:
Thanks for your interest in [palantir/tslint](https://github.com/palantir/tslint), @ajafff! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

src/language/utils.ts
```diff
@@ -77,7 +77,7 @@ export function scanAllTokens(scanner: ts.Scanner, callback: (s: ts.Scanner) =>
 /**
  * @returns true if any modifier kinds passed along exist in the given modifiers array
  */
-export function hasModifier(modifiers: ts.ModifiersArray, ...modifierKinds: ts.SyntaxKind[]) {
+export function hasModifier(modifiers: ts.ModifiersArray|null|undefined, ...modifierKinds: ts.SyntaxKind[]) {
```
adidahiya: why would it be `null`? it could only be `undefined`, right? with this change, the condition on the next line can also be changed to be more strict, i.e. `modifiers === undefined || modifierKinds === undefined`

ajafff: you're right, fixed
I also noticed, that modifierKinds can never be null, so I changed the condition to check for `.length === 0`

nchen63:
I'm not sure I understand why this change is needed. We don't do strict null checks within the tslint project

ajafff:
But I use strictNullChecks in my custom rules project. Without this change
tsc forces me to check node.modifiers for undefined before passing it to
hasModifier, although the function already does that.

Am 24.11.2016 01:45 schrieb "Noah Chen" <notifications@github.com>:

nchen63:
got it. Thanks!

PR #1757, opened 2016-11-20
Add exceptions support for comment-format rule (fixes #562)

### LINKED ISSUE #562: comment-format rule "check-lowercase" option should allow whitelist of exceptions
I'd like to allow exceptions to the `check-lowercase` rule option via a whitelist that looks something like:

``` json
"comment-format": [true,
    ["check-lowercase", "TODO", "HACKHACK"],
    "check-space"
]
```

IllusionMH:
#### PR checklist

- [X] Addresses an existing issue: #562 
- [X] ~New feature, bugfix, or~ enhancement
  - [X] Includes tests
- [ ] Documentation update

This PR adds support for exceptions list to comment-format rule.
Implementation uses array of exceptions as last argument.

1. Is there a need to escape special characters in exception list to avoid broken RegEx'es?
2. Does exceptions array in last position good choice until #1738 implemented?

P.S. Is it possible to set tests timeout for executable to 3000ms? This will help me a lot, since some tests take ~2500ms under when I run tests in Docker container.

### DIFF
--- a/src/rules/commentFormatRule.ts
+++ b/src/rules/commentFormatRule.ts
@@ -39,20 +39,25 @@ export class Rule extends Lint.Rules.AbstractRule {
         options: {
             type: "array",
             items: {
-                type: "string",
-                enum: ["check-space", "check-lowercase", "check-uppercase"],
+                oneOf: [{
+                    type: "string",
+                    enum: ["check-space", "check-lowercase", "check-uppercase"],
+                }, {
+                    type: "array",
+                    items: { type: "string" },
+                }],
             },
             minLength: 1,
-            maxLength: 3,
+            maxLength: 4,
         },
-        optionExamples: ['[true, "check-space", "check-lowercase"]'],
+        optionExamples: ['[true, "check-space", "check-uppercase"], [true, "check-lowercase", ["TODO", "HACK"]]'],
         type: "style",
         typescriptOnly: false,
     };
     /* tslint:enable:object-literal-sort-keys */
 
-    public static LOWERCASE_FAILURE = "comment must start with lowercase letter";
-    public static UPPERCASE_FAILURE = "comment must start with uppercase letter";
+    public static LOWERCASE_FAILURE = "comment must start with lowercase letter or word from exceptions list";
+    public static UPPERCASE_FAILURE = "comment must start with uppercase letter or word from exceptions list";
     public static LEADING_SPACE_FAILURE = "comment must start with a space";
 
     public apply(sourceFile: ts.SourceFile): Lint.RuleFailure[] {
@@ -61,6 +66,22 @@ export class Rule extends Lint.Rules.AbstractRule {
 }
 
 class CommentWalker extends Lint.SkippableTokenAwareRuleWalker {
+    private exceptionsPattern: RegExp | null;
+
+    constructor(sourceFile: ts.SourceFile, options: Lint.IOptions) {
+        super(sourceFile, options);
+
+        const optionsList = this.getOptions() as Array<string | string[]>;
+        const possibleExceptions = optionsList[optionsList.length - 1];
+        if (Array.isArray(possibleExceptions)) {
+            const wordsPattern = possibleExceptions.map(String).join("|");
+            // regex is "start of string"//"any amount of whitespace"("any word from exception list") followed by non alphanumeric character
+            this.exceptionsPattern = new RegExp(`^\\/\\/\\s*(${wordsPattern})\\W`);
+        } else {
+            this.exceptionsPattern = null;
+        }
+    }
+
     public visitSourceFile(node: ts.SourceFile) {
         super.visitSourceFile(node);
         Lint.scanAllTokens(ts.createScanner(ts.ScriptTarget.ES5, false, ts.LanguageVariant.Standard, node.text), (scanner: ts.Scanner) => {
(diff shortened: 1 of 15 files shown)


nchen63:
@IllusionMH I extended timeout for executable to 4s in #1766

src/rules/commentFormatRule.ts
```diff
@@ -39,20 +39,25 @@ export class Rule extends Lint.Rules.AbstractRule {
         options: {
             type: "array",
             items: {
-                type: "string",
-                enum: ["check-space", "check-lowercase", "check-uppercase"],
+                oneOf: [{
+                    type: "string",
+                    enum: ["check-space", "check-lowercase", "check-uppercase"],
+                }, {
+                    type: "array",
```
nchen63: would it be better to just make this a single regex string since the code makes it into a reg expression anyways? Also, people might see weird behavior if they don't know it's a regex.

nchen63: on second thought, it's probably better the way it is...

IllusionMH: I've checked list of ESLint rules and they use RE to provide patterns, however they use object with "ignorePattern" key to provide it.

As alternative approach I can implement exception argument as object with two possible keys `ignoreWords` - array of strings to ignore and `ignorePattern` - string with RE to ignore, instead of array of words.

IllusionMH:
> @IllusionMH I extended timeout for executable to 4s in #1766

@nchen63 thank you!

I've noticed that I forgot to update  `optionsDescription`, and I will submit fix later this week (and address any other requests).

src/rules/commentFormatRule.ts
nchen63: yes, would suggest an object with key for future extensibility

IllusionMH: Ok. I'll rework this PR to use object instead of array.
And one more question: Is current state of `options` in rule metadata correct for case with array?

I not sure if I correctly get this format especially when there should be 2 types of items in array - strings and array (unfortunately I can't find docs with description or sources that process it)

nchen63: I'm not that familiar with JSON schema, but don't you want `anyOf` instead of `oneOf`?

IllusionMH: Thanks for the tip, I'll check Schema docs then. (haven't used JSON Schema before and didn't realize that TSLint uses it)

adidahiya:
@IllusionMH could you please enable Circle CI on your fork?

IllusionMH:
@adidahiya added. But for some reason it doesn't run build on all 3 Node.js versions.
Also I haven't synced this branch with `master`.

adidahiya:
Your CI build looks fine, it is running different node versions in 3 containers (maybe you went and enabled multiple containers in the hour since you posted your last comment :))

IllusionMH:
@adidahiya yes, now it runs in 3 containers, however failed/stopped few times when I tried to setup them.

Moreover build on `master` runs fine in my fork (both have same last commit https://github.com/palantir/tslint/commit/d0533553fed521fd252a951f8dbdebe337e19c9a) 😃

IllusionMH:
I've rebased this branch and resolved conflicts.

@nchen63, @adidahiya is there anything that should be addressed in this PR?

src/rules/commentFormatRule.ts
```diff
@@ -35,24 +43,60 @@ export class Rule extends Lint.Rules.AbstractRule {
...
+            '[true, "check-lowercase", {"ignorePattern": "STD\\w{2,3}\\b"}]',
+        ],
         type: "style",
         typescriptOnly: false,
     };
     /* tslint:enable:object-literal-sort-keys */
 
-    public static LOWERCASE_FAILURE = "comment must start with lowercase letter";
-    public static UPPERCASE_FAILURE = "comment must start with uppercase letter";
+    public static LOWERCASE_FAILURE = "comment must start with lowercase letter, word from exceptions list or exceptions pattern";
```
nchen63: Could you print the list of exceptions or regex exception in the error? I think most people wouldn't know what an "exceptions list" is when seeing this error.

IllusionMH: Thanks! Good point.
Is it ok if there will be 3 different messages instead of one generic?
`comment must start with lowercase letter`
`comment must start with lowercase letter or word from exceptions list: TODO, HACK`
`comment must start with lowercase letter or match exceptions pattern: std(in|out|err)\b`

And same for uppercase option.

nchen63: 3 different messages are ok.

I'd avoid using the word "exceptions" since it could confuse people with the kind of exception you throw. Something like `Comment must start with a lowercase letter, "TODO", or "HACK"`

IllusionMH:
@nchen63 I've updated messages.
Message for ignore words was simplified to avoid differences for cases where single or several words are specified.
Message for ignore pattern was extended to specify that start of the comment will be matched against regexp and not a whole comment.

nchen63:
@IllusionMH thanks!

nchen63:
I realized as I was compiling the release notes that the option `ignoreWords` should probably be `ignore-words`

IllusionMH:
@nchen63 you are right. `typedef-whitespace` has config object with hyphens in keys.
I have stupid question: when rule docs should be updated? In every pull request or they are generated right before release?
Docs don't describe requirement here and I remember cases when docs weren't required between 3.15 and 4.0 releases.

nchen63:
Docs are generated and published with the release. I've recently added the generated docs to .gitignore, so it's not possible to update in any case.

IllusionMH:
@nchen63 thanks, PR is on the way

Issue #1761, opened 2016-11-21
Rule object-literal-key-quotes mistakenly flags the empty string as not needing quotes


gustavderdrache:
### Bug Report

- __TSLint version__: `3.15.1`
- __TypeScript version__: `2.1.1`
- __Running TSLint via__: both grunt-tslint and VS Code

#### TypeScript code being linted

```ts
const bug = {
  '': 'bug',
};
```

with `tslint.json` configuration:

```json
{
  "rules": {
    "object-literal-key-quotes": [true, "as-needed"]
  }
}
```

#### Actual behavior

TSLint complains:
```
const bug = {
  '': 'bug',
  ~~         [Unnecessarily quoted property '' found.]
};
```

#### Expected behavior

No error; the empty string as a property should always be quoted. I found this bug when filling out a Backbone router, the key '' corresponds to the home page.

nchen63:
closed by #1762

PR #1762, opened 2016-11-21
Special case the empty string in object-literal-key-quotes rule

gustavderdrache:
#### PR checklist

- [X] Addresses an existing issue: #1761
  - [X] Includes tests
- [ ] Documentation update

* The bug in #1761 appears to be rooted in `IDENTIFIER_NAME_REGEX` using a `*` instead of `+` repeat modifier. This has been addressed.
* I added a test case for `''` in the `"as-needed"` files.

I don't know enough about the spirit of "consistent" or "consistent-as-needed" to know if this case should be flagged as a failure or not:

```typescript
const uncertain = {
  foo: 'bar',
  '': 'baz',
};
```

Right now, this test fails in both cases. It seems like this is correct, but I welcome confirmation. Once I have it I'll update the tests.

### DIFF
--- a/src/rules/objectLiteralKeyQuotesRule.ts
+++ b/src/rules/objectLiteralKeyQuotesRule.ts
@@ -77,7 +77,7 @@ export class Rule extends Lint.Rules.AbstractRule {
 }
 
 // This is simplistic. See https://mothereff.in/js-properties for the gorey details.
-const IDENTIFIER_NAME_REGEX = /^(?:[\$A-Z_a-z])*$/;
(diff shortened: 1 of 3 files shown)


palantirtech:
Thanks for your interest in [palantir/tslint](https://github.com/palantir/tslint), @gustavderdrache! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

nchen63:
@gustavderdrache thanks!

PR #1764, opened 2016-11-21
Fatal errors from API should be unhandled and caught in tslint-cli

### LINKED ISSUE #1759: When tslint.json uses extends, some rules can't be disabled
### Bug Report

- __TSLint version__: `4.0.1`
- __Running TSLint via__: `gulp-tslint` v7.0.0

When I use this configuration
```json
{
  "rulesDirectory": [
    "node_modules/tslint-eslint-rules/dist/rules"
  ],
  "extends": [
    "./tslint-eslint-rules.json"
  ],
  "rules": {
    "comment-format": [ false ],
    "quotemark": [ false ]
  }
}
```

I get violations for those 2 rules, even though they are set to be disabled. (I probably get more violations than this, but I have it limited to only show 50 at a time).  However, if I comment out the contents of the `rulesDirectory` and `extends` arrays like this:

```json
{
  "rulesDirectory": [
    //"node_modules/tslint-eslint-rules/dist/rules"
  ],
  "extends": [
    //"./tslint-eslint-rules.json"
  ],
  "rules": {
    "comment-format": [ false ],
    "quotemark": [ false ]
  }
}
```

 those rules stay disabled and do not report violations as expected.

nchen63:
fixes #1759

3rd party apps should be able to see the fatal errors thrown by the API. #1686 changed it so that the error was caught and added as a property to IConfigurationLoadResult. It's not reasonable for 3rd party apps to be aware of this property, so we should follow convention and throw an unhandled exception that the app itself can handle.

This change creates a new FatalError class which is caught in `tslint.cli`.

### DIFF
--- a/docs/_data/rules.json
+++ b/docs/_data/rules.json
@@ -850,7 +850,7 @@
   },
   {
     "ruleName": "no-unused-variable",
-    "deprecationMessage": "Use the compiler options --noUnusedParameters and --noUnusedLocals instead.",
+    "deprecationMessage": "Use the tsc compiler options --noUnusedParameters and --noUnusedLocals instead.",
     "description": "Disallows unused imports, variables, functions and private class members.",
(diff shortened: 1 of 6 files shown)


nchen63:
@JoshuaKGoldberg fyi

PR #1771, opened 2016-11-22
Don't exit on missing rules

### LINKED ISSUE #1763: Tune down missing implementation error to warning
### Bug Report

- __TSLint version__: `4.0.1`
- __Running TSLint via__: CLI

Since the plan is depreciating `no-unused-variables` and others at 4.0 and remove at 5.0, it should not cause error at runtime right now:

```sh
        throw new Error(ERROR_MESSAGE);
        ^

Error:
            Could not find implementations for the following rules specified in the configuration:
            no-duplicate-key
no-unreachable
            Try upgrading TSLint and/or ensuring that you have all necessary custom rules installed.
            If TSLint was recently upgraded, you may have old rules configured which need to be cleaned up.
```

Most people uses existing config packages (`tslint-config-standard`, `tslint-config-typings`, `tslint-config-unional`, etc), harsh failing causes many builds to fail unnecessarily.

🌷

nchen63:
fixes #1763

### DIFF
--- a/src/ruleLoader.ts
+++ b/src/ruleLoader.ts
@@ -20,7 +20,8 @@ import * as path from "path";
 import {camelize} from "underscore.string";
 
 import {getRulesDirectories} from "./configuration";
-import {IDisabledInterval, IRule} from "./language/rule/rule";
+import { IDisabledInterval, IRule } from "./language/rule/rule";
+import { dedent } from "./utils";
 
 const moduleDirectory = path.dirname(module.filename);
 const CORE_RULES_DIRECTORY = path.resolve(moduleDirectory, ".", "rules");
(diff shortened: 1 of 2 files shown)


src/ruleLoader.ts
```diff
@@ -20,7 +20,8 @@ import * as path from "path";
 import {camelize} from "underscore.string";
 
 import {getRulesDirectories} from "./configuration";
-import {IDisabledInterval, IRule} from "./language/rule/rule";
+import { IDisabledInterval, IRule } from "./language/rule/rule";
```
bondz: Would be best if you follow the indentation of the file, braces aren't spaced.

src/ruleLoader.ts
```diff
@@ -20,7 +20,8 @@ import * as path from "path";
...
+import { dedent } from "./utils";
```
bondz: Same here.

Issue #1699, opened 2016-11-08
Rule proposal: require `async` for `Promise` returning methods.


AndyMoreland:
I'd like to have a lint rule that requires the `async` modifier be attached to any method that returns a `Promise`.

Reasoning:

`foo(): Promise<T>` can (technically) either return a rejected promise or bubble a standard synchronous Error object. No sane API is designed to use both error paths, because it requires any consumer use two separate error handling constructs at any exception handling point.

In my codebase I'd like to require that any method that returns a `Promise` also be marked `async`, because `async` methods automatically catch synchronous exceptions and convert them into rejected promises. This would prevent accidental synchronous exceptions in async codepaths.

andy-hanson:
Should be fixed by #1779.

Issue #1445, opened 2016-07-29
Error: no-for-in-array requires type checking


Delagen:
### Bug Report
- **TSLint version**: `3.14.0`
- **TypeScript version**: `2.0.0`
- **Running TSLint via**: Node.js API
#### TypeScript code being linted

with `tslint.json` configuration:

``` json
{
    "linterOptions": {
        "typeCheck": true
    },
    "rules": {
        "no-for-in-array": true
    }
}
```
#### Actual behavior

Error: no-for-in-array requires type checking
    at Rule.TypedRule.apply (...\node_modules\tslint\lib\language\rule\typedRule.js:14:15)
    at Linter.lint (...\node_modules\tslint\lib\tslint.js:65:37)
    ...
#### Expected behavior

No errors

adidahiya:
`linterOptions` in tslint.json is actually a private API. If using the CLI, you must use the `--type-check` flag. If using the Node.js APIs, you'll need to follow a pattern similar to [src/test.ts](https://github.com/palantir/tslint/blob/master/src/test.ts).

Delagen:
This is ugly and must be public

adidahiya:
Well, it's not a documented API... how did you discover it? We reserve the right to experiment with APIs in our project before committing to them...

Delagen:
Reading code of sample configurations, and reading code is the best documentation in most javascript projects. Developers update documentation very slow, released features may be documented after long time, some projects does not provide documentation about Node API use totally prefer to offer only CLI use, that is ugly.

Delagen:
Many build tools plugins use Node.JS api, and it's functionality must not differ from CLI, because CLI does not provide ability to control process of output etc. 
For example: https://github.com/panuhorsmalahti/gulp-tslint/blob/master/index.js

adidahiya:
> reading code is the best documentation in most javascript projects

Sure, that's partially true -- that's why I pointed out [src/test.ts](https://github.com/palantir/tslint/blob/master/src/test.ts) in my comment above.

---

I agree that the Node.js API is useful and should be documented. We _do_ have documentation for parts that are well-supported, __including type checking__, [right here in the README](https://github.com/palantir/tslint#type-checking). Also, being a typescript project, we naturally have documentation in the form of typings declaration files published in the NPM package.

[...]

PR #1779, opened 2016-11-23
Rule: `Promise`-returning methods must be `async`

buu700:
#### PR checklist

- [X] Addresses an existing issue: #1699
  - [x] Includes tests
- [ ] Documentation update

Added the rule proposed in the linked issue.

Not sure if my methods of detecting async and Promise are optimal (wasn't sure where to find detailed API documentation on the stuff under `ts`).


palantirtech:
Thanks for your interest in [palantir/tslint](https://github.com/palantir/tslint), @buu700! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

src/rules/promiseAsyncRule.ts
```diff
@@ -0,0 +1,63 @@
...
+        super.visitFunctionExpression(node);
+    }
+
+    public visitMethodDeclaration(node: ts.MethodDeclaration): void {
+        this.test(node);
+        super.visitMethodDeclaration(node);
+    }
+
+    private test(node: ts.FunctionExpression|ts.MethodDeclaration): void {
+        const isAsync   = node.getText().split(/[\(=]/)[0].match(/(^|\s)async($|\s)/) !== null;
```
nchen63: check for async using 
```
Lint.hasModifier(node.modifiers, ts.SyntaxKind.AsyncKeyword)
```

nchen63:
unit tests?

src/rules/promiseAsyncRule.ts
```diff
@@ -0,0 +1,63 @@
...
+    }
+
+    public visitMethodDeclaration(node: ts.MethodDeclaration): void {
+        this.test(node);
+        super.visitMethodDeclaration(node);
+    }
+
+    private test(node: ts.FunctionExpression|ts.MethodDeclaration): void {
+        const isAsync   = Lint.hasModifier(node.modifiers, ts.SyntaxKind.AsyncKeyword);
+        const isPromise = node.type.getText().indexOf("Promise<") === 0;
```
nchen63: searching the body for `Promise<` can be pretty flaky. Can you use `ApplyWithProgram` and get the actual return type?

src/rules/promiseAsyncRule.ts
```diff
@@ -0,0 +1,63 @@
...
+        super.visitFunctionExpression(node);
+    }
+
+    public visitMethodDeclaration(node: ts.MethodDeclaration): void {
+        this.test(node);
+        super.visitMethodDeclaration(node);
+    }
+
+    private test(node: ts.FunctionExpression|ts.MethodDeclaration): void {
+        const isAsync   = Lint.hasModifier(node.modifiers, ts.SyntaxKind.AsyncKeyword);
```
adidahiya: nit: please avoid aligning `=` across statements like this; we don't generally do this elsewhere since we find it unmaintainable.

src/rules/promiseAsyncRule.ts
adidahiya: this looks like it's testing just the return type of a function. it would work in some cases, but yeah I generally agree that this rule should use the type checker (unless maybe if the `typedef` rule is enabled? in that case you'd be forced to declare `: Promise<T>` everywhere and a simple syntactic check is sufficient... food for thought)

buu700:
Any idea what's wrong with my latest commit? I'm not sure what to fix based on the most recent CircleCI failure message.

buu700:
Not sure what's going on here, but it looks like my unit tests are failing because the return type of every `Promise`-returning function/method is being considered `any` (the `number`-returning functions are fine). Maybe a missing type definition?

src/rules/promiseAsyncRule.ts
```diff
@@ -0,0 +1,72 @@
...
+
+    public visitMethodDeclaration(node: ts.MethodDeclaration): void {
+        this.test(node);
+        super.visitMethodDeclaration(node);
+    }
+
+    private test(node: ts.FunctionExpression|ts.MethodDeclaration): void {
+        const tc = this.getTypeChecker();
+
+        const returnType = tc.typeToString(
```
nchen63: i'm fine with this.

test/rules/promise-async/test.ts.lint
```diff
@@ -0,0 +1,38 @@
+const nonAsyncPromiseFunction = function(p: Promise<void>) { return p; };
```
nchen63: you need `declare class Promise<T>{}` for this to work. Otherwise, the type checker will not know what a Promise<> is and consider it to be `any`

src/rules/promiseAsyncRule.ts
```diff
@@ -0,0 +1,72 @@
...
+        this.test(node);
+        super.visitFunctionExpression(node);
+    }
+
+    public visitMethodDeclaration(node: ts.MethodDeclaration): void {
+        this.test(node);
+        super.visitMethodDeclaration(node);
+    }
+
+    private test(node: ts.FunctionExpression|ts.MethodDeclaration): void {
```
nchen63: change the type to simply `ts.SignatureDeclaration`

src/rules/promiseAsyncRule.ts
```diff
@@ -0,0 +1,72 @@
...
+
+    public static FAILURE_STRING = "functions that return promises must be async";
+
+    public applyWithProgram(sourceFile: ts.SourceFile, program: ts.Program): Lint.RuleFailure[] {
+        return this.applyWithWalker(new PromiseAsyncWalker(sourceFile, this.getOptions(), program));
+    }
+}
+
+class PromiseAsyncWalker extends Lint.ProgramAwareRuleWalker {
+    public visitFunctionExpression(node: ts.FunctionExpression): void {
```
nchen63: also need `visitArrowFunction` and `visitFunctionDeclaration`

test/rules/promise-async/test.ts.lint
```diff
@@ -0,0 +1,38 @@
...
+                                ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~    [0]
+
```
nchen63: also need cases 

```
x.callback(() => return new Promise<void>());
```

and 

```

function x() {
    return new Promise<int>();
}
```

which hit the new `visit*` cases I suggested above

test/rules/promise-async/test.ts.lint
adidahiya: that type is provided by `lib.es2015.d.ts`; I would prefer using that if possible instead of constructing our own type

src/rules/promiseAsyncRule.ts
```diff
@@ -0,0 +1,72 @@
...
+ * limitations under the License.
+ */
+
+import * as ts from "typescript";
+import * as Lint from "../index";
+
+export class Rule extends Lint.Rules.TypedRule {
+    /* tslint:disable:object-literal-sort-keys */
+    public static metadata: Lint.IRuleMetadata = {
+        ruleName: "promise-functions-async",
```
buu700: Oops, sorry about that. Any preference on whether I change the ruleName or file name, or is there an entirely different name that would better fit your established conventions?

src/rules/promiseAsyncRule.ts
```diff
@@ -0,0 +1,72 @@
...
+        description: "Requires any function or method that returns a promise to be marked async.",
+        rationale: "Ensures that a function can only either return a rejected promise or throw an Error object.",
```
buu700: How's this? (Is Markdown syntax supported in these descriptions?)

> Ensures that each function is only capable of 1) returning a rejected promise, or 2) throwing an Error object. In contrast, non-`async` `Promise`-returning functions are technically capable of either. This rule removes a requirement for consuming code to handle both cases.

src/rules/promiseAsyncRule.ts
nchen63: I prefer `promise-functions-async`. @adidahiya?

test/rules/promise-async/test.ts.lint
buu700: What would be the best way to get those types included in this test?

src/rules/promiseAsyncRule.ts
```diff
@@ -0,0 +1,72 @@
...
+        optionsDescription: "Not configurable.",
+        options: null,
+        optionExamples: ["true"],
+        type: "functionality",
```
adidahiya: let's classify this as a typescript-only rule. I know the categories aren't perfectly mutually exclusive, but I'd like to highlight the fact that you get this lint rule only via usage of the TypeScript compiler services.

src/rules/promiseAsyncRule.ts
```diff
@@ -0,0 +1,72 @@
...
+        typescriptOnly: false,
```
adidahiya: despite my comment on the above line, the rule _can_ be used with either TS or JS code, so this line is correct 👍

src/rules/promiseAsyncRule.ts
adidahiya: nit: prefer omitting the `: void` return typedef on functions/methods

test/rules/promise-async/test.ts.lint
adidahiya: good question... I don't think this has come up before. perhaps we can add a `"compilerOptions"` block to our test `tslint.json` files?

N.B. as you can see from this thread (https://github.com/palantir/tslint/issues/1445), it might be a good idea to put `"compilerOptions"` under a namespace in the test `tslint.json` so as to clearly indicate it is a private API for use in this repo only and not a public API for tslint users yet.

buu700: Got it, thanks. For this PR, mind if I just put `declare interface Promise<T>;` at the top of the test file? That at least seems to resolve the immediate issue.

src/rules/promiseAsyncRule.ts
buu700: Hm, interesting; I'd actually meant to (and forgotten to) change that as part of my last commit. How would a rule that requires type checking work with JS?

src/rules/promiseAsyncRule.ts
buu700: Cool, I'll go with that then.

test/rules/promise-function-async/test.ts.lint
```diff
@@ -0,0 +1,65 @@
...
+                                      ~~~~~~~~~~~~~~~~~~~~~~~~~    [0]
+
+// 'async' 'Promise'-returning arrow functions are allowed
+const asyncPromiseArrowFunction = async (p: Promise<void>) => p;
+
+// non-'async' non-'Promise'-returning arrow functions are allowed
+const nonAsyncNonPromiseArrowFunction = (n: number) => n;
+
+class Test {
+    public nonAsyncPromiseMethodA(p: Promise<void>) {
```
nchen63: for big functions, seems excessive to highlight the whole thing including the body. How about just highlighting the function signature?

test/rules/promise-function-async/test.ts.lint
```diff
@@ -0,0 +1,65 @@
+declare class Promise<T>{}
+
+const nonAsyncPromiseFunctionExpressionA = function(p: Promise<void>) { return p; };
```
nchen63: for every test that causes a lint failure, you should have a matching test that passes

test/rules/promise-async/test.ts.lint
adidahiya: @buu700 yeah that's fine; @nchen63 can look into the longer term solution after this PR merges

src/rules/promiseAsyncRule.ts
buu700: Ah okay, cool.

nchen63:
@buu700 looks good. Thanks!

PR #1781, opened 2016-11-24
Narrowed no-internal-module failure to the keyword

### LINKED ISSUE #939: no-internal-module rule should narrow its reported error to just one line
currently we report an error on the whole module, but it would be better to report the failure span just on the `module` keyword and the module name.

JoshuaKGoldberg:
Fixes #939.

A couple of notes:
* @adidahiya suggested also including the module name in the failure. This doesn't since IMO the failure is just on the `module` keyword, not the name... but I'm not opposed to also including the name if asked.
* This blindly uses `node.getText().indexOf("module")` to determine where to start the error, since the logic for parsing `declare ` and `declare    ` could get a little complex.


JoshuaKGoldberg:
That was fast, thanks!

PR #1787, opened 2016-11-24
Added final onComplete, returns after onCompletes

### LINKED ISSUE #1785: Return after all onComplete calls in runner.ts
When `Runner` is imported/required and run, `onComplete` may no longer pipe to `process.exit`. That means it can't guarantee execution is stopped after calling `onComplete`.

Note that this doesn't affect the CLI - just external usage.

JoshuaKGoldberg:
Fixes #1785.

### DIFF
--- a/src/runner.ts
+++ b/src/runner.ts
@@ -151,6 +151,7 @@ export class Runner {
             if (!fs.existsSync(this.options.project)) {
                 console.error("Invalid option for project: " + this.options.project);
                 onComplete(1);
+                return;
             }
             program = Linter.createProgram(this.options.project, path.dirname(this.options.project));
             if (files.length === 0) {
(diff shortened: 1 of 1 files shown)


nchen63:
@JoshuaKGoldberg requiring people to return after calling onComplete seems fragile, and using a callback isn't necessary if it occurs when the function is returning anyways. Why not just set a return type that can include the exit code?

JoshuaKGoldberg:
For the async code in the `this.outputStream.write(lintResult.output, () => {` callback, I didn't see a way to do that without fancier logic. Promises seem like a better way of dealing with this, would that be something you're willing to take a dependency on?

nchen63:
True, unfortunately there isn't a writeSync method

PR #1792, opened 2016-11-27
Fix merge issue for `promise-function-async`

nchen63:


src/rules/promiseFunctionAsyncRule.ts
```diff
@@ -40,8 +40,8 @@ export class Rule extends Lint.Rules.TypedRule {
 
     public static FAILURE_STRING = "functions that return promises must be async";
 
-    public applyWithProgram(sourceFile: ts.SourceFile, program: ts.Program): Lint.RuleFailure[] {
-        return this.applyWithWalker(new PromiseAsyncWalker(sourceFile, this.getOptions(), program));
+    public applyWithProgram(sourceFile: ts.SourceFile, langSvc: ts.LanguageService): Lint.RuleFailure[] {
```
adidahiya: minor: please avoid using abbreviations; it's tough to have concrete rules around them, so we generally just avoid them. in this case this variable could be `ls` or `languageService`

PR #1799, opened 2016-11-29
Added no-magic-numbers rule

SimonSchick:
#### PR checklist

  - [x] Includes tests
- [x] Documentation update

Added a rule to prevent _magic numbers_.

See the added tests for examples.

Basic usage: `"no-magic-numbers": true` disallows all except 0 and 1/-1.
Custom allow list: `"no-magic-numbers": [true, 1337, 1338]"` Only 1337 and 1338 are allowed.

I am not very familiar with tslint and the TS API and it would love to get feedback on the filter method.

### DIFF
--- a/src/rules/noMagicNumbersRule.ts
+++ b/src/rules/noMagicNumbersRule.ts
@@ -60,6 +60,7 @@ export class Rule extends Lint.Rules.AbstractRule {
     };
 
     public static DEFAULT_ALLOWED = [
+        -1,
         0,
         1,
     ];
@@ -85,11 +86,26 @@ class NoMagicNumbersWalker extends Lint.RuleWalker {
     }
 
     public visitNode(node: ts.Node) {
-        if (node.kind === ts.SyntaxKind.NumericLiteral && !Rule.ALLOWED_NODES[node.parent.kind]) {
-            if (!this.allowed[node.getText()]) {
+        const isUnary = this.isUnaryNumericExpression(node)
+        if (node.kind === ts.SyntaxKind.NumericLiteral && !Rule.ALLOWED_NODES[node.parent.kind] || isUnary) {
+            let text = node.getText();
+            if (!this.allowed[text]) {
                 this.addFailure(this.createFailure(node.getStart(), node.getWidth(), Rule.FAILURE_STRING));
             }
         }
-        super.visitNode(node);
+        if (!isUnary) {
+            super.visitNode(node);
+        }
+    }
+
+    /**
+     * Checks if a node is an unary expression with on a numeric operand.
+     */
+    private isUnaryNumericExpression(node: ts.Node): boolean {
+        if (node.kind !== ts.SyntaxKind.PrefixUnaryExpression) {
+            return false;
+        }
+        const unaryNode = (<ts.PrefixUnaryExpression>node);
+        return unaryNode.operator === ts.SyntaxKind.MinusToken && unaryNode.operand.kind === ts.SyntaxKind.NumericLiteral;
     }
 }
--- a/src/rules/noMagicNumbersRule.ts
+++ b/src/rules/noMagicNumbersRule.ts
@@ -1,0 +1,79 @@
+/**
+ * @license
+ * Copyright 2016 Palantir Technologies, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import * as ts from "typescript";
+
+import * as Lint from "../index";
+
+export class Rule extends Lint.Rules.AbstractRule {
+    /* tslint:disable:object-literal-sort-keys */
+    public static metadata: Lint.IRuleMetadata = {
+        ruleName: "no-magic-numbers",
+        description: "Disallows the use constant number values outside of variable assignment. -1, 0 and 1 are allowed by default.",
+        rationale: Lint.Utils.dedent`
+            Magic numbers should be avoided as they often lack documentation, forcing
+            them to be stored in variables gives them implicit documentation.`,
+        optionsDescription: "A list of allowed number.",
+        options: {
+            type: "array",
+            items: {
+                type: "number",
+            },
+            minLength: 1,
+        },
+        optionExamples: ["true", "[true, 1, 2, 3]"],
+        type: "typescript",
+        typescriptOnly: false,
+    };
+    /* tslint:enable:object-literal-sort-keys */
+
+    public static FAILURE_STRING = "'magic numbers' are no allowed";
+
+    public static ALLOWED_NODES = [
+        ts.SyntaxKind.ExportAssignment,
+        ts.SyntaxKind.FirstAssignment,
+        ts.SyntaxKind.LastAssignment,
+        ts.SyntaxKind.PropertyAssignment,
+        ts.SyntaxKind.ShorthandPropertyAssignment,
+        ts.SyntaxKind.VariableDeclaration,
+        ts.SyntaxKind.VariableDeclarationList,
+    ];
+
+    public static DEFAULT_ALLOWED = [
+        0,
+        1,
+    ];
+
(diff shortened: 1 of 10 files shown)


palantirtech:
Thanks for your interest in [palantir/tslint](https://github.com/palantir/tslint), @SimonSchick! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

src/rules/noMagicNumbersRule.ts
```diff
@@ -0,0 +1,81 @@
...
+
+import * as ts from "typescript";
+
+import * as Lint from "../index";
+
+export class Rule extends Lint.Rules.AbstractRule {
+    /* tslint:disable:object-literal-sort-keys */
+    public static metadata: Lint.IRuleMetadata = {
+        ruleName: "no-magic-numbers",
+        description: "Disallows the use constant number values outside of variable assignment. -1, 0 and 1 are allowed by default.",
```
nchen63: `DEFAULT_ALLOWED` does not include -1

nchen63: Description should say that the default values are set if there are no list of numbers is supplied. When I read this, it sounded like the default values are unioned with the user-supplied list

src/rules/noMagicNumbersRule.ts
```diff
@@ -0,0 +1,81 @@
...
+        rationale: Lint.Utils.dedent`
+            Magic numbers should be avoided as they often lack documentation, forcing
+            them to be stored in variables gives them implicit documentation.`,
+        optionsDescription: "A list of allowed number.",
```
nchen63: `number` -> `numbers`

src/rules/noMagicNumbersRule.ts
```diff
@@ -0,0 +1,81 @@
...
+    }
+}
+
+class NoMagicNumbersWalker extends Lint.RuleWalker {
+    public visitNode(node: ts.Node) {
+        if (node.kind === ts.SyntaxKind.NumericLiteral &&
+            Rule.ALLOWED_NODES.indexOf(node.parent.kind) === -1) {
+            const options = this.getOptions();
+
+            const allowed: number[] = options.length > 0 ? options : Rule.DEFAULT_ALLOWED;
```
nchen63: read `options` and do this logic in a constructor so you don't have to do this for every node

src/rules/noMagicNumbersRule.ts
```diff
@@ -0,0 +1,81 @@
...
+        },
+        optionExamples: ["true", "[true, 1, 2, 3]"],
+        type: "typescript",
+        typescriptOnly: false,
+    };
+    /* tslint:enable:object-literal-sort-keys */
+
+    public static FAILURE_STRING = "'magic numbers' are no allowed";
+
+    public static ALLOWED_NODES = [
```
nchen63: Use object literal with computed property names for faster lookups:
```
public static ALLOWED_NODES = {
    [ts.SyntaxKind.ExportAssignment]: true
    ...
}
```
```
Rule.ALLOWED_NODES[node.parent.kind]
```

nchen63:
Looks pretty good in general. Have some minor changes

src/rules/noMagicNumbersRule.ts
SimonSchick: It is as the minus sign is not part of number literal.

src/rules/noMagicNumbersRule.ts
SimonSchick: Ah, I wasn't entirely sure if `this.getOptions` might change during operation and would thus break things. Thanks for clearing that up.

src/rules/noMagicNumbersRule.ts
```diff
@@ -0,0 +1,95 @@
...
+        const allowed: { [prop: string]: boolean } = {};
+        allowedArray.forEach((value) => {
+            allowed[value] = true;
+        });
+        this.allowed = allowed;
+    }
+
+    public visitNode(node: ts.Node) {
+        if (node.kind === ts.SyntaxKind.NumericLiteral && !Rule.ALLOWED_NODES[node.parent.kind]) {
+            if (!this.allowed[node.getText()]) {
```
nchen63: this should be changed to handle negative numbers. If someone configures the rule to allow `-1`, it would never be matched.

test/rules/no-magic-numbers/default/test.ts.lint
```diff
@@ -0,0 +1,30 @@
+console.log(-1, 0, 1);
```
nchen63: add test for floats

src/rules/noMagicNumbersRule.ts
SimonSchick: I'll have to look into that, I'm a little short on time right now tho...

src/rules/noMagicNumbersRule.ts
```diff
@@ -0,0 +1,95 @@
...
+        [ts.SyntaxKind.LastAssignment]: true,
+        [ts.SyntaxKind.PropertyAssignment]: true,
+        [ts.SyntaxKind.ShorthandPropertyAssignment]: true,
+        [ts.SyntaxKind.VariableDeclaration]: true,
+        [ts.SyntaxKind.VariableDeclarationList]: true,
+        [ts.SyntaxKind.EnumMember]: true,
+        [ts.SyntaxKind.PropertyDeclaration]: true,
+    };
+
+    public static DEFAULT_ALLOWED = [
```
adidahiya: minor: can you just make this a one liner? i.e. `... = [0, 1];`

src/rules/noMagicNumbersRule.ts
```diff
@@ -0,0 +1,95 @@
...
+    }
+}
+
+class NoMagicNumbersWalker extends Lint.RuleWalker {
+    private allowed: { [prop: string]: boolean };
+    constructor (sourceFile: ts.SourceFile, options: IOptions) {
+        super(sourceFile, options);
+
+        const configOptions = this.getOptions();
+        const allowedArray: number[] = configOptions.length > 0 ? configOptions : Rule.DEFAULT_ALLOWED;
```
adidahiya: please be more specific in the variable name and omit "Array" since it is redundant with the typdef. call this `const allowedNumbers` instead.

src/rules/noMagicNumbersRule.ts
```diff
@@ -0,0 +1,95 @@
...
+        1,
+    ];
+
+    public apply(sourceFile: ts.SourceFile): Lint.RuleFailure[] {
+        return this.applyWithWalker(new NoMagicNumbersWalker(sourceFile, this.getOptions()));
+    }
+}
+
+class NoMagicNumbersWalker extends Lint.RuleWalker {
+    private allowed: { [prop: string]: boolean };
```
adidahiya: what is the point of this object? its values are always `true`.

SimonSchick: https://github.com/palantir/tslint/pull/1799#discussion_r90358780 it's used a lookup map (which is faster).

SimonSchick:
I think I addressed all comments for now.

src/rules/noMagicNumbersRule.ts
```diff
@@ -0,0 +1,107 @@
...
+
+import { IOptions } from "../language/rule/rule";
+
+export class Rule extends Lint.Rules.AbstractRule {
+    /* tslint:disable:object-literal-sort-keys */
+    public static metadata: Lint.IRuleMetadata = {
+        ruleName: "no-magic-numbers",
+        description: Lint.Utils.dedent`
+            Disallows the use constant number values outside of variable assignments.
+            When no list of allowed values is specified, 0 and 1 are allowed by default.`,
```
nchen63: this doesn't match `DEFAULT_ALLOWED`

src/rules/noMagicNumbersRule.ts
```diff
@@ -0,0 +1,107 @@
...
+            },
+            minLength: 1,
+        },
+        optionExamples: ["true", "[true, 1, 2, 3]"],
+        type: "typescript",
+        typescriptOnly: false,
+    };
+    /* tslint:enable:object-literal-sort-keys */
+
+    public static FAILURE_STRING = "'magic numbers' are no allowed";
```
nchen63: `no` -> `not`

src/rules/noMagicNumbersRule.ts
nchen63: maybe add a comment

src/rules/noMagicNumbersRule.ts
SimonSchick: Not anymore, right 😛

nchen63:
@SimonSchick looks good. Thanks!

PR #1800, opened 2016-11-29
Fix adjacentOverloadSignatures for constructors

zerovox:
#### PR checklist

  - [x] Includes tests
- [ ] Documentation update

adjacent-overload-signatures did not work for nodes without names, such as constructors or empty lines. For empty lines, that's not really an issue we care about for this rule, but we probably want to catch constructors that are not adjacent.

(optional)


adidahiya:
nice, thanks @zerovox

PR #1801, opened 2016-11-29
add `prefer-const` rule

### LINKED ISSUE #491: "prefer-const" rule to discourage unnecessary use of "let"
I'd like a new rule that prevents declaring variables as `let` _unless that reference is later mutated via a left-hand-side assignment_. `const` should be used instead for better readability if you intend to never re-assign the variable.

Some code examples:

``` ts
// bad
for (let i of [1, 2, 3]) {
  //
}

// good
for (const i of [1, 2, 3]) {
  //
}

// bad
let {createReadStream} = require("fs");

// good, unless the recently-introduced "no-require-imports" rule is also enabled
const {createReadStream} = require("fs");
```

nchen63:
fixes #491


src/language/walker/blockScopeAwareRuleWalker.ts
```diff
@@ -58,13 +63,13 @@ export abstract class BlockScopeAwareRuleWalker<T, U> extends ScopeAwareRuleWalk
 
         if (isNewBlockScope) {
             this.blockScopeStack.push(this.createBlockScope());
+            this.onBlockScopeStart();
```
nchen63: was there a reason why this wasn't in `isNewBlockScope`?

adidahiya:
@gscshoyru any chance you'd be able to review this?

nchen63:
does anyone have opinions about destructured variables which has some reassigned and some not?

```
let [a,b] = [1,2];
a = 4
```

currently it errors and underlines `b`, but I could see an argument where no error is shown

nchen63:
Tested on Slate and got 7793 fixes without breaking compilation

test/rules/prefer-const/test.ts.lint
```diff
@@ -0,0 +1,120 @@
+const a = 1;
```
ajafff: can you add tests for the following cases, just to be safe for future changes?
```typescript
let arr = [];
    ~~~ [lorem ipsum ...]
arr.push(0);
arr[1] = 1;
arr.length = 1;
```

src/rules/preferConstRule.ts
```diff
@@ -0,0 +1,219 @@
...
+                }
+            }
+        }
+        super.visitIdentifier(node);
+    }
+
+    private handleLHSExpression(node: ts.Expression) {
+        node = unwrapParentheses(node);
+        if (node.kind === ts.SyntaxKind.Identifier) {
+            this.markAssignment(<ts.Identifier> node);
```
adidahiya: nit: prefer the `as ts.Identifier` syntax for type assertions

src/rules/preferConstRule.ts
```diff
@@ -0,0 +1,219 @@
...
+}
+
+function isAssignmentOperator(token: ts.Node) {
+    return token.kind >= ts.SyntaxKind.FirstAssignment
+        && token.kind <= ts.SyntaxKind.LastAssignment;
+}
+
+function unwrapParentheses(node: ts.Expression) {
+    while (node.kind === ts.SyntaxKind.ParenthesizedExpression) {
+        node = (<ts.ParenthesizedExpression> node).expression;
```
adidahiya: same nit here, `as ts.ParenthesizedExpression`

src/rules/preferConstRule.ts
```diff
@@ -0,0 +1,219 @@
...
+        return false;
+    }
+}
+
+function isAssignmentOperator(token: ts.Node) {
+    return token.kind >= ts.SyntaxKind.FirstAssignment
+        && token.kind <= ts.SyntaxKind.LastAssignment;
+}
+
+function unwrapParentheses(node: ts.Expression) {
```
adidahiya: this should probably go in language/utils.ts

src/rules/preferConstRule.ts
```diff
@@ -0,0 +1,219 @@
...
+
+    protected visitBinaryExpression(node: ts.BinaryExpression) {
+        if (isAssignmentOperator(node.operatorToken)) {
+            this.handleLHSExpression(node.left);
+        }
+        super.visitBinaryExpression(node);
+    }
+
+    protected visitEndOfFileToken(node: ts.Node) {
+        this.onBlockScopeEnd();
```
nchen63: I originally made that change, but it caused a compilation error for a project which generated d.ts files:

```
import {a, A} from './a';

export class B {
  static thing = a;
}
```

In this code, `A` was flagged by TSLint as unnecessary to export, but since `a` is a var and `A` is the type of `a`, it needed to import `A` also.

src/rules/preferConstRule.ts
nchen63: maybe that should be a rule

adidahiya: it is, let's enable it :D https://palantir.github.io/tslint/rules/no-angle-bracket-type-assertion/

nchen63:
I added `SourceFile` as a scope boundary even though it is only a boundary if the file is module. This shouldn't have any side effects, since TSLint lints files one at a time and can't consider global cross-file variables anyways.

adidahiya:
> I added SourceFile as a scope boundary even though it is only a boundary if the file is module.

TS language services ought to have an API that tells you whether a file is a module or a script; if not, can you request one (and make this a follow-up task)? Would prefer to be careful in this implementation.

nchen63:
It looks like the `ts.SourceFile` type populates `externalModuleIndicator` if it is a module

adidahiya:
@nchen63 cool, so will you address that in a follow up PR?

nchen63:
opened #1804

PR #1813, opened 2016-12-02
False positive in `prefer-for-of` rule when element assigned

### LINKED ISSUE #1807: prefer-for-of false positive when index is used to modify array
### Bug Report

- __TSLint version__: `4.0.2`

```ts
const arr = [1, 2, 3];
for (let i = 0 ; i < arr.length ; ++i) {
    arr[i] = 8;
}
```

```json
{
    "rules": {
        "prefer-for-of": true
    }
}
```

#### Actual behavior

`Expected a 'for-of' loop instead of a 'for' loop with this simple iteration`

#### Expected behavior

No errors, since `i` is being used in a way that wouldn't be possible in a for-of loop.

nchen63:
fixes #1807

### DIFF
--- a/src/language/utils.ts
+++ b/src/language/utils.ts
@@ -125,6 +125,14 @@ export function someAncestor(node: ts.Node, predicate: (n: ts.Node) => boolean):
     return predicate(node) || (node.parent && someAncestor(node.parent, predicate));
 }
 
+export function isAssignment(node: ts.Node) {
+    if (node.kind === ts.SyntaxKind.BinaryExpression) {
+        const binaryExpression = node as ts.BinaryExpression;
+        return binaryExpression.operatorToken.kind >= ts.SyntaxKind.FirstAssignment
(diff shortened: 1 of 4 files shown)


src/language/utils.ts
```diff
@@ -125,6 +125,14 @@ export function someAncestor(node: ts.Node, predicate: (n: ts.Node) => boolean):
     return predicate(node) || (node.parent && someAncestor(node.parent, predicate));
 }
 
+export function isAssignment(node: ts.Node) {
+    if (node.kind === ts.SyntaxKind.BinaryExpression) {
+        const binaryExpression = node as ts.BinaryExpression;
+        return binaryExpression.operatorToken.kind >= ts.SyntaxKind.FirstAssignment
+            && binaryExpression.operatorToken.kind <= ts.SyntaxKind.LastAssignment;
+    }
```
adidahiya: I feel like this function should always produce a return value, i.e. have an `else { return false; }` at the end.

Issue #1088, opened 2016-04-05
Request: no-empty should ignore empty catch blocks


RomkeVdMeulen:
I'm getting a `no-empty` error on this:

``` js
try {
  doSomethingThatMayFail();
} catch (e) {
}
```

but not this:

``` js
try {
  doSomethingThatMayFail();
} catch (e) {
  // ignore this
}
```

Of course it's open to interpretation, but I would expect the first example to also be okay. What do you think?

m-abs:
I'd like this too. +1

JKillian:
This could be added as an option to the rule: `allow-empty-catch`

adidahiya:
I don't really feel comfortable making this the only exception to the `no-empty` rule. Why should this be treated specially?

ikokostya:
> I don't really feel comfortable making this the only exception to the no-empty rule. Why should this be treated specially?

Because add comment to every empty block is annoying. Eslint has similar option http://eslint.org/docs/rules/no-empty#options. Why not make this behavior configurable for users?

adidahiya:
> Why not make this behavior configurable for users?

alright, sure, go for it, we'll accept a PR (but it won't make it into `tslint:recommended`).

RomkeVdMeulen:
Hasn't there been a PR for this already which got reverted? Why not change it back?

Izhaki:
Here's another case:

```
            try {
                await reject(this.dummyTask, 'Reject reason')
            } catch (err) {}
```

We want to reject a promise in an async block (in tests).

PR #1817, opened 2016-12-04
allow empty catch

YuichiNukiyama:
#### PR checklist

- [x] Addresses an existing issue: #1088

Allow empty catch block in no-empty rule.


adidahiya:
hey, wait a sec, why are we making this the _only_ exception to the no-empty rule? that doesn't make sense to me. why shouldn't you be expected to throw in an explanatory comment to circumvent the rule in the same way you have to do so for other empty blocks?

nchen63:
The linked ticket was marked `Accepting PRs`, implying tacit endorsement. That said, this does seem like an unnecessary option.

adidahiya:
Sorry about that, I should've removed the label (I didn't add it in the first place). Would prefer to revert this change.

Issue #1683, opened 2016-11-04
Add Babel code frame to stylishFormatter output


mohsen1:
[Babel code frame](https://www.npmjs.com/package/babel-code-frame) makes it easy to print line/col numbers with a nice pointer for lint errors.

adidahiya:
looks pretty cool, I'm always down for better error messages!

mohsen1:
This is done

PR #1819, opened 2016-12-05
Add CodeFrame Formatter

mohsen1:
#### PR checklist

- [x] Addresses an existing issue: #1683 
  - [x] Includes tests
- [x] Documentation update

Add the new `codeFrame` formatter.

This is how it looks on real code: 

screen shot 2016-12-04 at 10 45 18 pm

A questions: Do you have a place to define modules for npm dependencies that don't have `@types` yet? I like to use `import` for babel-code-frame

### DIFF
--- a/docs/_data/formatters.json
+++ b/docs/_data/formatters.json
@@ -7,6 +7,13 @@
     "consumer": "machine"
   },
   {
+    "formatterName": "codeFrame",
+    "description": "Framed formatter which creates a frame of error code.",
+    "descriptionDetails": "\nPrints syntax highlighted code in a frame with a pointer to where\nexactly lint error is happening.",
+    "sample": "\nsrc/components/Payment.tsx\nParentheses are required around the parameters of an arrow function definition (arrow-parens)\n  21 |     public componentDidMount() {\n  22 |         this.input.focus();\n> 23 |         loadStripe().then(Stripe => Stripe.pay());\n     |                          ^\n  24 |     }\n  25 |\n  26 |     public render() {",
+    "consumer": "human"
+  },
+  {
     "formatterName": "filesList",
     "description": "Lists files containing lint errors.",
     "sample": "directory/myFile.ts",
@@ -43,13 +50,6 @@
     "description": "Human-readable formatter which creates stylish messages.",
     "descriptionDetails": "\nThe output matches that produced by eslint's stylish formatter. Its readability\nenhanced through spacing and colouring",
     "sample": "\nmyFile.ts\n1:14  semicolon  Missing semicolon",
-    "consumer": "human"
-  },
-  {
-    "formatterName": "codeFrame",
-    "description": "Framed formatter which creates a frame of error code.",
-    "descriptionDetails": "\nPrints syntax highlighted code in a frame with a pointer to where\nexactly lint error is happening.",
-    "sample": "",
     "consumer": "human"
   },
   {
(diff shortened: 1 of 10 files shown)


palantirtech:
Thanks for your interest in [palantir/tslint](https://github.com/palantir/tslint), @mohsen1! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

mohsen1:
Also working on https://github.com/DefinitelyTyped/DefinitelyTyped/pull/13102 to add `@types/babel-code-frame`

src/formatters/stylishFormatter.ts
```diff
@@ -20,6 +20,7 @@ import {IFormatterMetadata} from "../language/formatter/formatter";
 import {RuleFailure} from "../language/rule/rule";
 
 import * as colors from "colors";
+const codeFrame = require("babel-code-frame");
```
adidahiya: I think you'll need to disable the `no-var-requires` rule to get your build passing. This is fine for now, just leave a comment explaining that the typings for this are currently in progress.

test/formatters/stylishFormatterTests.ts
```diff
@@ -41,25 +42,71 @@ describe("Stylish Formatter", () => {
...
-            "1:1  first-name  first failure" + "\n" +
-            "1:3  escape      &<>\'\" should be escaped" + "\n" +
-            `${maxPositionTuple}  last-name   last failure` + "\n" +
-            "1:1  full-name   full failure" + "\n" +
-            "\n";
-
-        assert.equal(formatter.format(failures), expectedResult);
+        // const maxPositionObj = sourceFile.getLineAndCharacterOfPosition(maxPosition - 1);
+
+        // const maxPositionTuple = `${maxPositionObj.line + 1}:${maxPositionObj.character + 1}`;
```
adidahiya: please remove commented code

mohsen1:
@adidahiya fixed.

mohsen1:
This should be good to merge.

adidahiya:
@nchen63 can you take a look please

nchen63:
is there a reason this is a change to the stylish formatter and not a new formatter? The stylish formatter is based on the eslint stylish formatter, so this change might not be what people expect.

nchen63:
can we just name this one `codeframe` like eslint?

mohsen1:
Actually it made sense so I just made it a new formatter.

mohsen1:
I'm going to need help adding this new formatter.

nchen63:
I think the test is failing because you're missing the original logic for storing and getting `rawLines`. 

Also, could you store the name of the file instead of the entire file? It could take a lot of memory to store all the source code in RAM.

mohsen1:
@nchen63 I don't think have an extra reference will consume any memory. It's not a clone.

mohsen1:
@adidahiya Which files I need to update and which ones are automatically generated?

nchen63:
@mohsen1 You're keeping a reference to the contents of every source file, right? Normally, these would be eligible for GC, but if you hold on to the reference, it won't be cleaned up.

mohsen1:
I see. So each rule should not keep a strong reference of text file. With the file name, what's the right API to get the raw text?

nchen63:
Actually, you can leave it like this. I need to refactor code in this area anyways

nchen63:
code and docs look good. Thanks

Issue #253, opened 2014-12-18
New rule to only allow boolean statements in logical expressions


oefirouz:
Disallow if (expr) unless expr is a boolean.

So no if (obj)... rather if (obj != null)

adidahiya:
I think the new TS APIs have support for type information about nodes. This should be doable.

adidahiya:
blocked on #680

JKillian:
See https://github.com/Microsoft/TypeScript/issues/9702 as well, although as far as I can tell they haven't decided to implement this yet. This shouldn't be too hard to do now on our end!

Sammons:
I would really like this to help with not coercing promise-like types to booleans, but I think it would be really nice if we could explicitly enable certain types (strings, numbers) for coercion.

radykov:
> I think the new TS APIs have support for type information about nodes. This should be doable.
 
Would this be able to pick up things such as `x = y || z`;? 
I'm unfamiliar with TS APIs and what was meant by nodes

OliverJAsh:
This would be really handy and catch so many bugs.

adidahiya:
fixed via #1820

PR #1820, opened 2016-12-05
strict boolean expressions

subash-a:
#### PR checklist

- [x] Addresses an existing issue: #253 
- [x] Includes tests
- [x] Documentation update

Introduced a rule to type check boolean expressions using type aware rules. This rule checks if the given binary expression, if statement, while statement or for statement comprises of correctly formed boolean expressions or literals. This is a first cut of the implementation and would like to get feedback on how to improve this and finally make this rule a part of the tslint rule set.
@myitcv tagging you here since we both discussed this rule offline, @adidahiya any feedback is welcome on how we can improve this.

### DIFF
--- a/src/rules/strictBooleanExpressionsRule.ts
+++ b/src/rules/strictBooleanExpressionsRule.ts
@@ -1,0 +1,151 @@
+/**
+ * @license
+ * Copyright 2016 Palantir Technologies, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import * as ts from "typescript";
+import * as Lint from "../index";
+import * as utils from "../language/utils";
+
+export class Rule extends Lint.Rules.TypedRule {
+    /* tslint:disable:object-literal-sort-keys */
+    public static metadata: Lint.IRuleMetadata = {
+        ruleName: "strict-boolean-expressions",
+        description: `Usage of && or || operators should be with boolean operands and
+expressions in If, Do, While and For statements should be of type boolean`,
+        optionsDescription: "Not configurable.",
+        options: null,
+        optionExamples: ["true"],
+        type: "functionality",
+        typescriptOnly: true,
+        requiresTypeInfo: true,
+    };
+    /* tslint:enable:object-literal-sort-keys */
+
+    public static BINARY_EXPRESSION_ERROR = "Operands for the && or || operator should be of type boolean";
+    public static UNARY_EXPRESSION_ERROR = "Operand for the ! operator should be of type boolean";
+    public static STATEMENT_ERROR = "statement condition needs to be a boolean expression or literal";
+    public static CONDITIONAL_EXPRESSION_ERROR = "Condition needs to be a boolean expression or literal";
+
+    public applyWithProgram(srcFile: ts.SourceFile, langSvc: ts.LanguageService): Lint.RuleFailure[] {
+        return this.applyWithWalker(new StrictBooleanExpressionsRule(srcFile, this.getOptions(), langSvc.getProgram()));
+    }
+}
+
+type StatementType = ts.IfStatement|ts.DoStatement|ts.WhileStatement;
+
+class StrictBooleanExpressionsRule extends Lint.ProgramAwareRuleWalker {
+    private checker: ts.TypeChecker;
+
+    constructor(srcFile: ts.SourceFile, lintOptions: Lint.IOptions, program: ts.Program) {
+        super(srcFile, lintOptions, program);
+        this.checker = this.getTypeChecker();
+    }
+
+    public visitBinaryExpression(node: ts.BinaryExpression) {
+        let isAndAndBinaryOperator = node.operatorToken.kind === ts.SyntaxKind.AmpersandAmpersandToken;
+        let isOrOrBinaryOperator = node.operatorToken.kind === ts.SyntaxKind.BarBarToken;
+        if (isAndAndBinaryOperator || isOrOrBinaryOperator) {
+            let lhsExpression = node.left;
+            let lhsType = this.checker.getTypeAtLocation(lhsExpression);
+            let rhsExpression = node.right;
+            let rhsType = this.checker.getTypeAtLocation(rhsExpression);
+            if (!this.isBooleanType(lhsType)) {
+                if (lhsExpression.kind !== ts.SyntaxKind.BinaryExpression) {
+                    this.addFailure(this.createFailure(lhsExpression.getStart(), lhsExpression.getWidth(), Rule.BINARY_EXPRESSION_ERROR));
+                } else {
+                    this.visitBinaryExpression(<ts.BinaryExpression> lhsExpression);
+                }
+            }
+            if (!this.isBooleanType(rhsType)) {
+                if (rhsExpression.kind !== ts.SyntaxKind.BinaryExpression) {
+                    this.addFailure(this.createFailure(rhsExpression.getStart(), rhsExpression.getWidth(), Rule.BINARY_EXPRESSION_ERROR));
+                } else {
+                    this.visitBinaryExpression(<ts.BinaryExpression> rhsExpression);
+                }
+            }
+        }
+        super.visitBinaryExpression(node);
+    }
+
+    public visitPrefixUnaryExpression(node: ts.PrefixUnaryExpression) {
+        let isExclamationOperator = node.operator === ts.SyntaxKind.ExclamationToken;
+        if (isExclamationOperator) {
+            let expr = node.operand;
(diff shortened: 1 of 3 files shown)


palantirtech:
Thanks for your interest in [palantir/tslint](https://github.com/palantir/tslint), @subash-a! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

myitcv:
Looks great to me 👍

src/rules/strictBooleanExpressionsRule.ts
```diff
@@ -0,0 +1,116 @@
+import * as ts from "typescript";
+import * as Lint from "../index";
+
+export class Rule extends Lint.Rules.TypedRule {
+    public static BINARY_EXPRESSION_ERROR = "Invalid Expression: Operands for the && or || operator should be of type boolean";
+    public static UNARY_EXPRESSION_ERROR = "Invalid Expression: Operand for the ! operator should be of type boolean";
+    public static IF_STATEMENT_ERROR = "Invalid Expression: If statement condition needs to be a boolean expression or literal";
```
nchen63: Combine lines 7 to 11 into `Statement is not a boolean expression`. After that, combine some of the checker logic below to reduce code duplication.

src/rules/strictBooleanExpressionsRule.ts
```diff
@@ -0,0 +1,116 @@
+import * as ts from "typescript";
+import * as Lint from "../index";
+
+export class Rule extends Lint.Rules.TypedRule {
+    public static BINARY_EXPRESSION_ERROR = "Invalid Expression: Operands for the && or || operator should be of type boolean";
```
nchen63: Remove `Invalid Expression: `

test/rules/strict-boolean-expressions/test.ts.lint
```diff
@@ -0,0 +1,184 @@
+/* tslint:disable:no-unused-parameters */
```
adidahiya: also this is not a tslint rule

src/rules/strictBooleanExpressionsRule.ts
```diff
@@ -0,0 +1,116 @@
...
+                if (!this.isBooleanType(lhsType) || !this.isBooleanType(rhsType)) {
+                    // with the && or || operator the operands do not resolve to a boolean type
+                    this.addFailure(this.createFailure(node.getStart(), node.getWidth(), Rule.BINARY_EXPRESSION_ERROR));
+                }
+            } else {
+                // usage of binary operator && or || results in a non-boolean expr
+                this.addFailure(this.createFailure(node.getStart(), node.getWidth(), Rule.BINARY_EXPRESSION_ERROR));
+            }
+        }
+        super.visitBinaryExpression(node);
```
nchen63: same with other `super` calls below

src/rules/strictBooleanExpressionsRule.ts
```diff
@@ -0,0 +1,116 @@
...
+        // check if the expression is of boolean type
+        let expType = checker.getTypeAtLocation(node.condition);
+        if (!this.isBooleanType(expType)) {
+            this.addFailure(this.createFailure(node.getStart(), node.getWidth(), Rule.FOR_STATEMENT_ERROR));
+        }
+        super.visitForStatement(node);
+    }
+
+    private isBooleanType(btype: ts.Type): boolean {
+        let boolType = /* tslint:disable:no-bitwise */(btype.flags & ts.TypeFlags.BooleanLike)/* tslint:enable:no-bitwise */;
```
subash-a: @nchen63 this would not work because I am checking the type flags and not the node flags, there are two options available
1. I can leave this line as is
2. I can introduce a new function `isTypeFlagsSet` into `language/utils.ts` which can be used to check the type flags here
which option do you suggest I go with?

subash-a:
@nchen63 I have made all the changes suggested, but one of the comments had a problem with the implementation for which I have left you a note, will make changes based on your answer there. Thanks for the quick and detailed review

src/rules/strictBooleanExpressionsRule.ts
nchen63: Please add this to utils

src/rules/strictBooleanExpressionsRule.ts
```diff
@@ -0,0 +1,160 @@
...
+        }
+        if (!errorFlagged) {
+            super.visitPrefixUnaryExpression(node);
+        }
+    }
+
+    public visitIfStatement(node: ts.IfStatement) {
+        let hasError = this.checkStatement(node);
+        if (!hasError) {
+            super.visitIfStatement(node);
```
nchen63: Won't this skip the block/body of the `if` statement?

src/rules/strictBooleanExpressionsRule.ts
```diff
@@ -0,0 +1,160 @@
...
+        super.visitForStatement(node);
+    }
+
+    private checkStatement(node: StatementType): boolean {
+        let errorFlagged = false;
+        let expType = this.checker.getTypeAtLocation(node.expression);
+        if (!this.isBooleanType(expType)) {
+            switch (node.kind) {
+            case ts.SyntaxKind.IfStatement:
+                this.addFailure(this.createFailure(node.getStart(), node.getWidth(), `If ${Rule.STATEMENT_ERROR}`));
```
nchen63: Don't necessarily need different error message for each type of conditional, but this works

src/rules/strictBooleanExpressionsRule.ts
```diff
@@ -0,0 +1,160 @@
...
+        let isOrOrBinaryOperator = node.operatorToken.kind === ts.SyntaxKind.BarBarToken;
+        if (isAndAndBinaryOperator || isOrOrBinaryOperator) {
+            let nodeType = this.checker.getTypeAtLocation(node);
+            if (this.isBooleanType(nodeType)) {
+                let lhsExpression = node.left;
+                let lhsType = this.checker.getTypeAtLocation(lhsExpression);
+                let rhsExpression = node.right;
+                let rhsType = this.checker.getTypeAtLocation(rhsExpression);
+                if (!this.isBooleanType(lhsType) || !this.isBooleanType(rhsType)) {
+                    this.addFailure(this.createFailure(node.getStart(), node.getWidth(), Rule.BINARY_EXPRESSION_ERROR));
```
nchen63: The failure should underline the offending operand, not the entire statement

subash-a:
@nchen63 changed the code to point out only the offending expression and also changed the logic of checking statements even after error has been flagged to find further errors in the statement block.

subash-a:
@nchen63 However I had a question about the `super.visit` call for the various statements and expressions that I have in my code, is calling `super.visitIfStatement` (for example) mandatory after the error checking logic is executed? How does this affect the other rules, if it does at all, and would it stop errors from other rules showing up? Because if that is the case then what we are doing inside the `visitBooleanExpression` would stop other errors for the boolean expression to show up when there is an error in my rule and the other error would show up only after correcting the first error. Any suggestion on how we can structure this rule to avoid affecting other rules? perhaps @adidahiya and @JKillian can also pitch in with their thoughts on how best to do this.

nchen63:
we have rules like this that don't call super. See `noShadowedVariableRule.ts`

adidahiya:
Rules are evaluated individually in isolation, so omitting a `super` call will not affect other rules. If such a call is omitted at a non-leaf node in the AST, the only effect is that the children in that subtree are not linted for that rule.

test/rules/strict-boolean-expressions/test.ts.lint
```diff
@@ -0,0 +1,178 @@
...
+
+/*** Valid ***/
+!!boolFn();
+!boolExpr;
+!!boolType;
+
+/*** If Statement ***/
+/*** Invalid ***/
+if (numType) { /* statements */ }
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~ [If statement condition needs to be a boolean expression or literal]
```
nchen63: underline the condition, not the entire `if` statement. Same with others

test/rules/strict-boolean-expressions/test.ts.lint
```diff
@@ -0,0 +1,178 @@
...
+boolType || anyType;
+            ~~~~~~~  [Operands for the && or || operator should be of type boolean]
+boolType || numType;
+            ~~~~~~~  [Operands for the && or || operator should be of type boolean]
+strType || boolType;
+~~~~~~~              [Operands for the && or || operator should be of type boolean]
+bwrapType || boolType;
+~~~~~~~~~              [Operands for the && or || operator should be of type boolean]
+objType || boolType || enumType;
+~~~~~~~~~~~~~~~~~~~              [Operands for the && or || operator should be of type boolean]
```
nchen63: I don't think it makes sense to underline `objType || boolType`, `objType` should only be underlined. If we can avoid flagging this, we can go back to the way it was before using `errorFlagged` and call `super` since that address the issue of overlapping errors

subash-a:
Thank you @nchen63 and @adidahiya for that explanation, things are much clearer to me now, so based on the feedback left in the earlier comments I have made changes to eliminate the duplication of error messages and also narrow down the error message to point to the specific expression or identifier, please take a look and let me know if the changes are fine, this should work as good first cut for this rule and probably improved upon as we go. Please let me know what you guys think.

subash-a:
@nchen63 and @adidahiya Thanks for the review, how do we get the documentation updated, i.e add the description about the new rule at https://palantir.github.io/tslint/rules/

nchen63:
We don't update that site until a release is made. You can run `npm run verify` to generate the docs, but you don't need to

PR #1823, opened 2016-12-05
Compile tests before runing tests

mohsen1:
#### PR checklist

  - [ ] Includes tests
- [ ] Documentation update

Running `npm test` does not reflect changes in test, this will make sure we're always compiling tests before running them. 

(optional)

### DIFF
--- a/package.json
+++ b/package.json
@@ -28,6 +28,7 @@
     "lint": "npm-run-all -p lint:core lint:test",
     "lint:core": "tslint \"src/**/*.ts\"",
     "lint:test": "tslint \"test/**/*.ts\" -e \"test/**/*.test.ts\"",
+    "pretest": "npm run compile:test",
     "test": "npm-run-all test:pre -p test:mocha test:rules",
     "test:pre": "cd ./test/config && npm install",
     "test:mocha": "mocha --reporter spec --colors \"build/test/**/*Tests.js\" build/test/assert.js",


palantirtech:
Thanks for your interest in [palantir/tslint](https://github.com/palantir/tslint), @mohsen1! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

nchen63:
why can't you just run `compile:test`?

mohsen1:
`pretest` is triggered automatically by npm when running `npm t`. [docs](https://docs.npmjs.com/misc/scripts)

adidahiya:
Oh, right, I forgot about that. Then it probably makes sense to refactor the existing `npm run test:pre` script?

mohsen1:
@adidahiya that makes sense. native npm scripts pre stuff is better.

I noticed that this change will not do it for when you make changes in source code itself. It should compile source and test.

In other note, why did you guys decide to compile tests and source instead of using `ts-node/register`?

nchen63:
what benefit would there be to use `ts-node/register`? It still needs to compile

mohsen1:
@nchen63 you don't need to compile before test and in watch mode it's pretty fast due to TS compiler caching

nchen63:
I looked at the `ts-node` source, and it doesn't seem to involve watching. For it to be in watch mode, it would also need to keep the process alive, right?

mohsen1:
I used it with webpack and watch was really fast. Maybe it's the webpack...

nchen63:
Yeah probably WebPack.

PR #1826, opened 2016-12-06
Install babel-code-frame

mohsen1:
Add missing dependency that was left from #1819


mohsen1:
How tests were passing in #1819  I don't know!

nchen63:
@mohsen1 thanks again!

PR #1835, opened 2016-12-08
Use @types/babel-code-frame to import babel-code-frame

mohsen1:


SimonSchick:
This PR appears to have broken master, I noticed your CI didn't run, you might want to look into that 😛

IllusionMH:
Same here. Looks like it is still compiled to function module and should be imported in old way
```
import  = require('babel-code-frame');
```

`babel-code-frame` should export function as `default` (as in ES6 source) and has it compiled version
```
exports.default = function (rawLines, lineNumber, colNumber) {
 /// ...
};
```

but looks like last line of the file overrides it and forces return of function
```
module.exports = exports["default"];
```

And when required in REPL
```
> cF = require('babel-code-frame')
[Function]
> cF.default
undefined
```

SimonSchick:
Yea hence I also made this PR: https://github.com/DefinitelyTyped/DefinitelyTyped/pull/13246

Doesn't really explain how it managed to to bypass CI tho 😛

PR #1843, opened 2016-12-11
Add more simple types

andy-hanson:
#### PR checklist

  - [X] Includes tests
- [ ] Documentation update **(Not needed)**

Consider the `undefined` and `never` types to be simple types. This allows writing `undefined[]` or `never[]`.
The second commit changes tests to prove that I changed something. Could be removed if you think it's ugly.

[rule-update] `array-type` now consider `undefined` and `never` as simple types, allowing `undefined[]` and `never[]`

### DIFF
--- a/src/rules/arrayTypeRule.ts
+++ b/src/rules/arrayTypeRule.ts
@@ -94,10 +94,12 @@ class ArrayTypeWalker extends Lint.RuleWalker {
             case ts.SyntaxKind.ArrayType:
             case ts.SyntaxKind.BooleanKeyword:
             case ts.SyntaxKind.NullKeyword:
+            case ts.SyntaxKind.UndefinedKeyword:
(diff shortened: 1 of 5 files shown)


andy-hanson:
I would mind waiting for version 5 to do this.

ajafff:
What about `ts.SyntaxKind.LiteralType`?
That would allow for `true[]`, `1337[]`, `"foo"[]`.

I just want to add this to the discussion. I don't have an opinion wether they should be considered simple types or not.

adidahiya:
yeah it's a subjective call about what is considered "simple" here. IMO the examples you just proposed are a little confusing to read (partially because literals in type positions are a more recent addition to the language), so I would prefer writing `Array<true>`, `Array<1337>`, `Array<"foo">`. open to criticism here though.

PR #1848, opened 2016-12-11
Properly handle computed/static property names

### LINKED ISSUE #1831: `adjacent-overload-signatures` false positive with symbol names
### Bug Report

- __TSLint version__: `4.0.2`
- __Running TSLint via__: CLI

```ts
class Foo {
  readonly [Symbol.toStringTag] = 'Foo';

  bar() {
    throw new Error();
  }

  [Symbol.iterator]() {
    throw new Error();
  }
}

```

```json
{
  "rules": {
    "adjacent-overload-signatures": true
  }
}
```

#### Actual behavior

The adjacent-overload-signatures rule appears to coerce method names which are symbols (e.g., `Symbol.iterator`) into the `"undefined"`, so it can't thinks they are overloads instead of separate methods.

```
src/test.ts[8, 3]: All 'undefined' signatures should be adjacent
```

#### Expected behavior

The rule should not raise an error, as they are separate methods, not overloads.

andy-hanson:
Fixes  #1831

#### PR checklist

- [X] Addresses an existing issue:  #1831
  - [X] Includes tests
- [ ] Documentation update **(not needed)**

Previously we used the name of an overload as the key. This caused problems for static and non-static properties conflicting, so #1772 was necessary. But that solution would give ugly error messages for conflicting static overloads. The new solution passes the pretty name separately from the key, so it avoids that problem.

(#1831 occured because the solution in #1772 had a problem when `nameText` was undefined: it would compute `undefined + ""`, which isnt' itself undefined. This could be fixed by simply checking for `undefined` before adding `suffix`, but I think this more complex PR is a better solution in the long run.)

### DIFF
--- a/src/rules/adjacentOverloadSignaturesRule.ts
+++ b/src/rules/adjacentOverloadSignaturesRule.ts
@@ -58,9 +58,7 @@ class AdjacentOverloadSignaturesWalker extends Lint.RuleWalker {
     }
 
     public visitInterfaceDeclaration(node: ts.InterfaceDeclaration): void {
-        this.checkOverloadsAdjacent(node.members, (member) => {
-            return getTextOfPropertyName(member);
-        });
+        this.checkOverloadsAdjacent(node.members, getTextOfPropertyName);
         super.visitInterfaceDeclaration(node);
     }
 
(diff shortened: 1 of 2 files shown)


nchen63:
code looks good. need to resolve merge conflict

nchen63:
@andy-hanson thanks!

PR #1849, opened 2016-12-12
Treat a property `foo: () => void` like `foo = () => {}`: as a method declaration.

### LINKED ISSUE #1156: member-ordering: method declaration mistreated as a member
### Bug Report

- __TSLint version__: `3.8.0`
- __Running TSLint via__: Visual Studio Code

```ts
class Class {
  someMethod() { }

  // to be added in prototype at runtime
  mixinMethod: () => number;  // Error: Declaration of public instance member variable not allowed to appear after declaration of public instance member function
}
```

with `tslint.json`:

```json
{
  "rules": {
    "member-ordering": [true, "public-before-private", "static-before-instance", "variables-before-functions"]
  }
}
```

#### Actual behavior

tslint reports an error “Declaration of public instance member variable not allowed to appear after declaration of public instance member function” for `mixinMethod()`

#### Expected behavior

no error is reported

andy-hanson:
Fixes #1156

#### PR checklist

- [X] Addresses an existing issue: #1156
  - [X] Includes tests
- [ ] Documentation update (may be needed, but current documentation doesn't seem to mention anything about #969)

Consider a property with a function type the same way as a property initialized to an arrow function.

This is a breaking change, since people with this rule enabled will have to move all function-typed properties.
Maybe this should be put under an option. We can't really detect whether a property is meant to be used as an assignable field or as something to be mixed in to the prototype. However, it does seem more consistent with #969.

### DIFF
--- a/src/rules/memberOrderingRule.ts
+++ b/src/rules/memberOrderingRule.ts
@@ -262,8 +262,9 @@ export class MemberOrderingWalker extends Lint.RuleWalker {
     }
 
     public visitPropertyDeclaration(node: ts.PropertyDeclaration) {
-        const { initializer } = node;
-        const isFunction = initializer != null
+        const { initializer, type } = node;
+        const isFunction = type && type.kind === ts.SyntaxKind.FunctionType
+            || initializer != null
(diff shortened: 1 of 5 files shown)


adidahiya:
this is stale and there's not a lot of movement on the reported issue. closing for now; I think we can focus on other things

PR #1850, opened 2016-12-12
Add package.json scripts for easier development

andy-hanson:
#### PR checklist

  - [ ] Includes tests
- [ ] Documentation update
- [X] Infrastructure

Added scripts allowing one to run the compiler in `--watch` mode and run specific tests.

* Another idea: could `ruleTestRunner` generate mocha tests? Then we could use mocha's builtin functionality for filtering tests.
* Why does the `test` directory feature its own compilation of `src`, anyway? I was confused when recompiling `src` didn't affect tests.

### DIFF
--- a/docs/develop/contributing/index.md
+++ b/docs/develop/contributing/index.md
@@ -13,6 +13,8 @@ npm run compile
 npm run test
 ```
 
+From there you can use `npm run test:watch` and `npm run test:rules -- my-rule-name` to try out changes to a rule.
+
 #### `next` branch
 
 The [`next` branch of the TSLint repo](https://github.com/palantir/tslint/tree/next) tracks the latest TypeScript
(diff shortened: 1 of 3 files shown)


SimonSchick:
Is there a way to compile/run the unit tests for only a single rule? If not that'd be a very useful addition imo.

IllusionMH:
@SimonSchick I was achieving this by rewrite of `test/ruleTestRunner.ts` file for running single test folder (calculated based on `*Rule.ts` file name) from VSCode debugger.

If maintainers/collaborators think that it might be good approach (it will be implemented as separate file for debug purposes only), I'll submit PR for single rule folder debug this or (most probably) next weekend.

adidahiya:
@SimonSchick yeah, that's currently possible using the CLI:

```sh
// global tslint
// point to a dir that has tslint.json and .lint files
tslint --test test/rules/semicolon/always

// locally built tslint
./bin/tslint --test test/rules/semicolon/always
```

andy-hanson:
@adidahiya Maybe I should remove the changes to `ruleTestRunner.ts` and just have `contributing/index.md` tell people to use the commands you provided.

nchen63:
closing in favor of #2025

PR #1855, opened 2016-12-12
Add IntelliJ files to .gitignore

greglo:


SimonSchick:
Usually better practise to keep platform/IDE specific files in your global `.gitignore`.

adidahiya:
hehe @SimonSchick is right; I shouldn't have checked in the vim-specific lines into that file :)

politely declining this PR

PR #1866, opened 2016-12-14
Don't error for misplaced braces for 'else' if there are no braces

### LINKED ISSUE #400: Single-line if with no braces and "check-else"
"check-else" gives an error with this case:

``` typescript
if (condition1)
    doWork();
else
    doOtherWork();
```

There should a way to specify to TSLint that this is a valid coding convention.

andy-hanson:
Fixes #400.

#### PR checklist

- [X] Addresses an existing issue: #400
  - [X] Includes tests
- [ ] Documentation update

Stop `one-line` from reporting an error for `else` when there are no braces.


nchen63:
@andy-hanson thanks!

Issue #1867, opened 2016-12-14
Task: simplify uses of `createFailure`


andy-hanson:
As of #1845, `noUnusedVariableRule.ts` and `objectLiteralKeyQuotesRule.ts` are the only rules using `createFailure` directly. We should look into simplifying these.

edit: other rules using `createFailure`:

- [x] `banTypesRules.ts`
- [x] `importSpacingRule.ts`
- [x] `noFloatingPromisesRule.ts`
- [x] `noStringThrowRule.ts`

nchen63:
we can remove the deprecated functions now.

nchen63:
having second thoughts on removing the deprecated functions, which would break almost all 3rd party rules. I think we should leave these in.

adidahiya:
@nchen63 that's a good point; let's add a deprecation warning in the console in 5.0 though?

nchen63:
taking out of 5.0 release since we can deprecate later in 5.x, and we're still using the `addFailure` in some files

adidahiya:
fair enough

PR #1874, opened 2016-12-15
Refactor object-literal-key-quotes rule

andy-hanson:
#### PR checklist

- [X] Addresses an existing issue: #1867 (part of it)
  - [ ] Includes tests
- [ ] Documentation update

Rewrote `object-literal-key-quotes` to process all properties at once instead of using state on the Walker object. It no longer stores failures in an array, instead adding them immediately.

The line `node.properties.filter(({ kind }) => kind !== ts.SyntaxKind.MethodDeclaration);` is just for backwards-compatibility. Not sure when it would be OK to introduce a breaking change of this kind, but I don't see why MethodDeclarations should be an exception to the rule. It seems like a result of the fact that the previous implementation used `visitPropertyAssignment`, skipping over MethodDeclarations implicitly.

### DIFF
--- a/src/rules/objectLiteralKeyQuotesRule.ts
+++ b/src/rules/objectLiteralKeyQuotesRule.ts
@@ -87,9 +87,7 @@ class ObjectLiteralKeyQuotesWalker extends Lint.RuleWalker {
     }
 
     public visitObjectLiteralExpression(node: ts.ObjectLiteralExpression) {
-        // For backwards-compatibility only. Old versions of this rule did nothing for method declarations.
-        const properties = node.properties.filter(({ kind }) => kind !== ts.SyntaxKind.MethodDeclaration);
-
+        const { properties } = node;
         switch (this.mode) {
             case "always":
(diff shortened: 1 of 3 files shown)


src/rules/objectLiteralKeyQuotesRule.ts
```diff
@@ -76,85 +76,83 @@ export class Rule extends Lint.Rules.AbstractRule {
...
-        let state: IObjectLiteralState = {
-            hasQuotesNeededProperty: false,
-            quotesNotNeededProperties: [],
-            unquotedProperties: [],
-        };
-        // a nested object literal should store its parent state to restore when finished
-        let previousState = this.currentState;
-        this.currentState = state;
+        // For backwards-compatibility only. Old versions of this rule did nothing for method declarations.
+        const properties = node.properties.filter(({ kind }) => kind !== ts.SyntaxKind.MethodDeclaration);
```
nchen63: I don't think we need this filter anymore

nchen63:
@andy-hanson thanks!

PR #1882, opened 2016-12-16
Fix string quotes on docs rule array-type

jabooth:
#### PR checklist

- [ ] Includes tests
- [x] Documentation update

The generated docs page for this rule is missing string quotes, so a copy/paste of the rule into `tslint.json` fails.

(optional)

### DIFF
--- a/docs/rules/array-type/index.html
+++ b/docs/rules/array-type/index.html
@@ -15,9 +15,9 @@ options:
     - generic
     - array-simple
 optionExamples:
-  - '[true, array]'
-  - '[true, generic]'
-  - '[true, array-simple]'
+  - '[true, "array"]'
+  - '[true, "generic"]'
+  - '[true, "array-simple"]'
 type: style
 typescriptOnly: true
 layout: rule
--- a/src/rules/arrayTypeRule.ts
+++ b/src/rules/arrayTypeRule.ts
@@ -21,7 +21,7 @@ export class Rule extends Lint.Rules.AbstractRule {
             type: "string",
             enum: [OPTION_ARRAY, OPTION_GENERIC, OPTION_ARRAY_SIMPLE],
         },
-        optionExamples: [`[true, ${OPTION_ARRAY}]`, `[true, ${OPTION_GENERIC}]`, `[true, ${OPTION_ARRAY_SIMPLE}]`],
+        optionExamples: [`[true, "${OPTION_ARRAY}"]`, `[true, "${OPTION_GENERIC}"]`, `[true, "${OPTION_ARRAY_SIMPLE}"]`],
         type: "style",
         typescriptOnly: true,
     };


palantirtech:
Thanks for your interest in [palantir/tslint](https://github.com/palantir/tslint), @jabooth! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

jabooth:
@palantirtech wow that CLA is a little heavy-weight for a minor docs fix...  👎

adidahiya:
The CLA is automated; it is required for any change to any project in this organization. It's much simpler for us to apply it across the board rather than make a judgement for every single PR. Also this is common practice for the majority of open source projects maintained by a company or organization.

docs/rules/array-type/index.html
```diff
@@ -15,9 +15,9 @@
     - generic
     - array-simple
 optionExamples:
-  - '[true, array]'
-  - '[true, generic]'
-  - '[true, array-simple]'
+  - '[true, "array"]'
```
adidahiya: these docs are autogenerated from the [source code](https://github.com/palantir/tslint/blob/2218cfc63b8282b9082a89ce03b395c3d9daa4cd/src/rules/arrayTypeRule.ts#L24), you'll need to make the changes there.

jabooth: Shall I commit again to fix or amend existing commit?

jabooth:
@adidahiya yeah I know the deal, just wanted to voice the fact that it does increase friction for contributors (we had a CLA [and binned it](https://github.com/menpo/menpo/issues/667) because of that friction). 
 
Sure many big companies do have them, but the fact that Apple and Mozilla get away just fine without them makes me question whether it's really necessary to make contributors go through the explicit signing rather than just having a statement in `CONTRIBUTING.md`.

Some interesting discussion on this [hacker news thread](https://news.ycombinator.com/item?id=10896206) too.

Of course not expecting meaningful change from a Goliath like Palantir based on this, but felt like it should be said.

docs/rules/array-type/index.html
adidahiya: you can make an additional commit; we squash all PRs anyway

adidahiya:
Interesting; I didn't realize that Apple doesn't use a CLA (Mozilla doesn't really sell software, so it isn't as important for them). I do realize it creates a bit of friction, but I think the CLA form UX is pretty simple. We'd welcome any feedback on the form itself.

jabooth:
@adidahiya so from memory the steps were:

1. Auth with github
2. Receive email linking to CLA
3. Read doc, sign
4. Receive email confirmation
5. Clean up my inbox of two emails I don't need

I know it's silly as it really wasn't that arduous, but that was more effort than it took to issue the PR:

1. Edit file in github
2. Type commit message (bottom of same page no less)
2. Fill in PR template and hit submit

Now if your PR template had clearly stated something along the lines of

> Please note that by submitting this Pull Request you are agreeing to the Contributor Licence Agreement as outlined in [CONTRIBUTING.md](https://github.com/palantir/tslint/blob/master/LICENSE). Please do not submit this Pull Request unless you agree with the terms outlined there.

I would have followed the link, taken a glace to see if it was reasonable, and proceeded happily. I'm sure there's a legal question over whether that is sufficient, but the end result seems the same to me (IANAL!) but it's so much less effort on my part to glance over one text file rather than the whole auth/emails/webUI to sign doc procedure. I haven't committed to apple/swift but I think there workflow is likely something similar to that.

Issue #1844, opened 2016-12-11
New rules for linting type declarations


andy-ms:
We've been working on lint [rules](https://github.com/Microsoft/types-publisher/tree/master/src/tslint) for [DefinitelyTyped](https://github.com/DefinitelyTyped/DefinitelyTyped) and most may be of general use. Would you be interested in my making PRs for some of these rules?

| Name | Description | PR |
| --- | --- | --- |
| `forbidden-types` | This forbids the `Function`, `Object`, `Boolean`, `Number`, and `String` types. These have better alternatives (specific function types, `{}`, `boolean`, `number`, `string`). |  #2271 |
| `functional-interfaces` | This prefers `type Foo = () => void` over `interface Foo { (): void; }`. | #1891 |
| `interface-over-type-literal` | This prefers `interface Foo { x: number; }` over `type Foo = { x: number; }`. |  #1890 |
| `no-empty-interface` | This forbids `interface Foo { }`. (TypeScript is structurally typed, so this doesn't accomplish anything.) | #1889 |
| `no-public` | This forbids the `public` keyword. This could be an option for `member-access` to make it forbid instead of mandate. | #2247 |
| `unified-signatures` | This prefers `interface Foo { foo(x: number | string): void; }` over `interface Foo { foo(x: number): void; foo(x: string): void; }`. |  #1944 |
| `void-return` | This forbids `foo(x: void): number` but allows `foo(x: number): void`. This could be joined with `forbidden-types`. | N/A |

(`dt-header` and `no-single-declare-module` will probably not be useful outside of DefinitelyTyped.)

nchen63:
when would someone ever write `foo(x: void)`?
`forbidden-types` 


| Name | Description |
| --- | --- |
| `forbidden-types` | This is covered by the `ban rule`. |
| `functional-interfaces` | 👍 I like it  |
| `interface-over-type-literal` | 👍 I like it  |
| `no-empty-interface` | 👍 I like it |
| `no-public` | 🤔 I don't see the benefit of this one |
| `unified-signatures` | How would this work for more complex signatures like `interface Foo { foo(x: number, y: boolean): void; foo(x: string, y: SomeInterface): void; }`? Would it allow a bad combo like `foo(true, "bar")`?. |
| `void-return` | I don't understand why someone would write `foo(x: void): number` in the first place |

adidahiya:
I don't think the `ban` rule covers anything in type positions

nchen63:
right, I'll edit the table

[...]

PR #1890, opened 2016-12-16
Add `interface-over-type-literal` rule

andy-ms:
#### PR checklist

- [X] Addresses an existing issue: #1844
  - [X] Includes tests
- [X] Documentation update

Added the `interface-over-type-literal` rule. This forbids `type T = { ... }`, preferring `interface T { ... }`.


src/rules/noEmptyInterfaceRule.ts
```diff
@@ -27,7 +27,6 @@ export class Rule extends Lint.Rules.AbstractRule {
         rationale: "An empty interface is equivalent to its supertype (or `{}`).",
         optionsDescription: "Not configurable.",
         options: null,
-        optionExamples: ["true"],
```
andy-ms: This was left in by accident.

nchen63:
@andy-ms thanks!

itajaja:
Hi, just out of curiosity, can you point me to some rationale behind this?

andy-ms:
It's stylistic. Generally `interface T { foo(): void; }` and `type T = { foo(): void; }` mean the same thing (although an interface can be extended). I think the latter style was mostly coming from people familiar with flow, where there is a difference.

itajaja:
thanks for the explanation :)

PR #1891, opened 2016-12-16
Add `callable-types` rule

andy-ms:
#### PR checklist

- [X] Addresses an existing issue: #1844
  - [X] Includes tests
- [X] Documentation update

Added the `callable-types` rule, which recommends to use a function type over an interface or type literal with just a call signature.


nchen63:
@andy-ms thanks!

CodeOtter:
I've found having a module with common interfaces for callbacks improves code readability.  For example:

```
export interface IError {
  (
    message: string,
    data: any,
  ): void;
}

export interface IDone {
  (
    error: any,
    data: any,
  ): void;
}

export interface INext {
  (
    error: any,
    data: any,
  ): void;
}

export interface IPassthru {
  (
    data: any,
  ): void;
}
```

This rule appears to prevent this.  I'll be turning this rule off in my local tslint.json, but I'm curious what problem this rule is solving.  Thanks!

andy-ms:
You can just as easily use a type alias.
```ts
export type Error = (message: string, data: any) => void;
```
Also, in your example, if you used a specific function type instead of using a common one, you could avoid using `any`. You could also use a generic: `export type Error<T> = (message: string, data: T) => void`.

CodeOtter:
Oh my.  Most helpful, @andy-ms   Thanks for the breakdown!

Issue #1371, opened 2016-07-01
Stricter rule & option naming conventions


adidahiya:
I think it would be useful to do a pass over all the rule and option names for 4.0 so that we can enforce stricter conventions around them. As we add more and more rules to core TSLint (and we have more community rule sets), we should think about how we want these to appear in a user's `tslint.json` file so that they can make sense of it.

For example, an "ordered-imports" rule recently merged. We should be intentional about the name:
- Do we want it to appear next to other rules related to _ordering_ or _sorting_? Which of those words should we use, "order" or "sort"? I lean slightly towards the latter, so I would probably call it "sort-imports" like [ESLint does](http://eslint.org/docs/rules/sort-imports). Then it could be near other rules like "sort-object-literal-keys" (which would be a rename of the existing "object-literal-sort-keys"), etc.
- Or should it appear next to other rules related to _imports_? In this case we should call it "import-ordering" or "import-sorting" like in [my original proposal](https://github.com/palantir/tslint/issues/771).
- This has implications for the existing "member-ordering" rule as well.

One place I would take inspiration from is stylelint -- check out their [rule naming conventions here](https://github.com/stylelint/stylelint/blob/master/docs/user-guide/about-rules.md).

We'll need to flesh out a proposal for rule naming guidelines before implementing this breaking change across the rules list (sure, we could have a few releases where we deprecate existing names).

JKillian:
Good call - the [Stylelint guidelines](https://github.com/stylelint/stylelint/blob/master/docs/user-guide/about-rules.md) that you mentioned are a great example of a clear and well-documented specification for rule names.

zimme:
Something I've run into with community rules is that I've added multiple community rules packages and they have the same names for rules that work just a bit different and I would really like to be able to choose a specific rule. Maybe I've missed some documentation but I haven't found a way to do this.

e.g. Add `tslint-eslint-rules` and `tslint-misc-rules`, both have a rule `sort-imports` and they work different and I would like to use one over the other.

JoshuaKGoldberg:
There have been a lot of rules with names that start with `no-` but end up having the opposite convention requested. It also somewhat defeats the purpose of nicely alphabetized folders or config listings when half the rules start with the same three characters. Maybe this should be discouraged?

JoshuaKGoldberg:
### 💀 _**It's time!**_ 💀

**TSLint is being deprecated** and no longer accepting pull requests for major new changes or features. See #4534. 😱

[...]

PR #1893, opened 2016-12-17
Adds option to `arrow-parens` to disallow  parentheses for single arg

### LINKED ISSUE #1860: `arrow-parens` should have an option to require single-arg arrow functions to *never* have parentheses
Some of us like the slightly less verbose version that doesn't include parentheses for things like `map`/`filter`/etc. (i.e. highly functional code). Parentheses can get hard to read in TypeScript's relatively complex grammar, so it would be nice to require them to not exist. (Plus, I encountered this while migrating a JavaScript code base linted with ESLint to TypeScript.)

nchen63:
closes #1860


src/rules/arrowParensRule.ts
```diff
@@ -20,21 +20,28 @@ import * as ts from "typescript";
...
 export class Rule extends Lint.Rules.AbstractRule {
     /* tslint:disable:object-literal-sort-keys */
     public static metadata: Lint.IRuleMetadata = {
         ruleName: "arrow-parens",
         description: "Requires parentheses around the parameters of arrow function definitions.",
         rationale: "Maintains stylistic consistency with other arrow function definitions.",
-        optionsDescription: "Not configurable.",
-        options: null,
-        optionExamples: ["true"],
+        optionsDescription: Lint.Utils.dedent`if \`${AVOID_ON_SINGLE_PARAMETER}\` is specified, then arrow functions with one parameter
```
IllusionMH: Shouldn't it have line break right after ` ?

src/rules/arrowParensRule.ts
```diff
@@ -20,21 +20,29 @@ import * as ts from "typescript";
 import * as Lint from "../index";
 import { hasModifier } from "../language/utils";
 
+const AVOID_ON_SINGLE_PARAMETER = "avoid-on-single-parameter";
```
adidahiya: this should've used the rule option name I suggested here, "ban-single-arg-parens" https://github.com/palantir/tslint/issues/1860#issuecomment-267388635. We need to move towards a more predictable and consistent naming strategy for rules an options; this is diverging in the wrong direction. We don't use the token "avoid" anywhere else. See https://github.com/palantir/tslint/issues/1371

Issue #1747, opened 2016-11-19
Provide getStart method for RuleWalker


jmlopez-rod:
This is not a bug, this is a suggestion that may give us some performance).

After encountering a bug in typescript and doing some digging around I realized that in most of the rules (if not all of them) we use a form of `node.getStart()`. This triggers the following calls:

```ts
        NodeObject.prototype.getStart = function (sourceFile, includeJsDocComment) {
            return ts.getTokenPosOfNode(this, sourceFile, includeJsDocComment);
        };
```

Note that we never provide the source file so we are then calling `ts.getTokenPosOfNode(this)` which has this implementation:

```ts
function getTokenPosOfNode(node, sourceFile, includeJsDocComment) {
        // With nodes that have no width (i.e. 'Missing' nodes), we actually *don't*
        // want to skip trivia because this will launch us forward to the next token.
        if (nodeIsMissing(node)) {
            return node.pos;
        }
        if (isJSDocNode(node)) {
            return ts.skipTrivia((sourceFile || getSourceFileOfNode(node)).text, node.pos, /*stopAfterLineBreak*/ false, /*stopAtComments*/ true);
        }
        if (includeJsDocComment && node.jsDocComments && node.jsDocComments.length > 0) {
            return getTokenPosOfNode(node.jsDocComments[0]);
        }
        // For a syntax list, it is possible that one of its children has JSDocComment nodes, while
        // the syntax list itself considers them as normal trivia. Therefore if we simply skip
        // trivia for the list, we may have skipped the JSDocComment as well. So we should process its
        // first child to determine the actual position of its first token.
        if (node.kind === 286 /* SyntaxList */ && node._children.length > 0) {
            return getTokenPosOfNode(node._children[0], sourceFile, includeJsDocComment);
        }
        return ts.skipTrivia((sourceFile || getSourceFileOfNode(node)).text, node.pos);
    }
```

The key here is the last return statement: `(sourceFile || getSourceFileOfNode(node)).text`, since we are not providing the sourceFile this means that we call this bad boy all the time:

```ts
    function getSourceFileOfNode(node) {
        while (node && node.kind !== 256 /* SourceFile */) {
            node = node.parent;
        }
        return node;
    }
```

Which means that if we have a node and we call `node.getStart()`, we do not get the start of the node for free, in fact, this will be very costly for nodes that are very deeply nested.

Please see the following comment in the issue I reported in typescript:

[...]

PR #1896, opened 2016-12-18
Replace `node.getSourceFile()`, which climbs ancestors, with the `sourceFile` field on a RuleWalker.

andy-hanson:
#### PR checklist

  - [ ] Includes tests
- [ ] Documentation update

Made a minor change for performance. `getSourceFile` isn't good to call repeatedly from deep within a tree, especially when the source file is already a field of the class. https://github.com/Microsoft/TypeScript/blob/master/src/compiler/utilities.ts#L153

### DIFF
--- a/src/language/walker/ruleWalker.ts
+++ b/src/language/walker/ruleWalker.ts
@@ -44,6 +44,10 @@ export class RuleWalker extends SyntaxWalker {
         return this.sourceFile;
     }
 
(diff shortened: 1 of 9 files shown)


nchen63:
@andy-hanson thanks! This is the same underlying issue as https://github.com/palantir/tslint/issues/1747

Issue #1899, opened 2016-12-18
tslint-switch disable-next-line throws when next line ends with EOF


ajafff:
### Bug Report

- __TSLint version__: `4.0.2`
- __TypeScript version__: `2.1.4`
- __Running TSLint via__: cli, vscode

#### TypeScript code being linted

```ts
// tslint:disable-next-line
var foo = bar; // no newline at end of file
```

#### Actual behavior

```
Error: Debug Failure. False expression:
    at Object.assert ([...]/node_modules/typescript/lib/typescript.js:3177:23)
    at computePositionOfLineAndCharacter ([...]/node_modules/typescript/lib/typescript.js:4947:18)
    at Object.getPositionOfLineAndCharacter ([...]/node_modules/typescript/lib/typescript.js:4942:16)
    at SourceFileObject.getPositionOfLineAndCharacter ([...]/node_modules/typescript/lib/typescript.js:78360:23)
    at EnableDisableRulesWalker.getStartOfLinePosition ([...]/node_modules/tslint/lib/enableDisableRules.js:54:21)
    at EnableDisableRulesWalker.handlePossibleTslintSwitch ([...]/node_modules/tslint/lib/enableDisableRules.js:108:48)
...
```

PR #1902, opened 2016-12-18
Handle eof in line switches

ajafff:
#### PR checklist

- [x] Addresses an existing issue: #1899
  - [ ] Includes tests
- [ ] Documentation update

To get the end of the `next-line` switch, tslint searches for the start of the line after the next line. If there is no such line, tslint crashed.
This change checks, if the line in question exists. You can now use the next-line switch without even having a next line (which doesn't make sense at first glance, but is useful when linting live during development)

(optional)


nchen63:
can you turn on CircleCI?

nchen63:
@ajafff thanks!

Issue #1383, opened 2016-07-08
void functions should not have a return


Fank:
### Feature request
- **TSLint version**: `3.12.1`
- **TypeScript version**: `1.8.10`
- **Running TSLint via**: (please choose one) CLI / grunt-tslint / Atom
#### TypeScript code being linted

``` ts
export class MyTestClass {
    public foobar(): void {
        return this._voidTest();
    }

    private _voidTest(): void {
        // Something
    }
}
```
#### Actual behavior

Everything is fine
#### Expected behavior

The `return` in `MyTestClass.foobar` should show an error, because a `void` function/method does not return something.

adidahiya:
closing as I think #1903 mostly covers this use case; let me know if we should re-open

PR #1903, opened 2016-12-18
Add `no-void-expression` rule

andy-hanson:
#### PR checklist

- [X] Addresses an existing issue: #1383
  - [X] Includes tests
- [X] Documentation update

Added the `no-void-expression` rule, which requires that any expression of type `void` go on its own line as a statement.

### DIFF
--- a/docs/_data/rules.json
+++ b/docs/_data/rules.json
@@ -59,10 +59,16 @@
     "ruleName": "arrow-parens",
     "description": "Requires parentheses around the parameters of arrow function definitions.",
     "rationale": "Maintains stylistic consistency with other arrow function definitions.",
-    "optionsDescription": "Not configurable.",
-    "options": null,
+    "optionsDescription": "\nif `avoid-on-single-parameter` is specified, then arrow functions with one parameter \nmust not have parentheses if removing them is allowed by TypeScript.",
+    "options": {
+      "type": "string",
+      "enum": [
+        "avoid-on-single-parameter"
+      ]
+    },
     "optionExamples": [
-      "true"
+      "true",
+      "[true, avoid-on-single-parameter]"
     ],
     "type": "style",
     "typescriptOnly": false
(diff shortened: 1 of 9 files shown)


nchen63:
@andy-hanson thanks!

Martin-Wegner:
Can you please explain why this rule complains about `this.subscriptions.forEach((subscription: Subscription) => subscription.unsubscribe());`?

andy-hanson:
This rule flags any expression of type `void`. So if `subscription.unsubscribe()` is of type `void`, there will be a linter error.
Try using this:

```ts
for (const subscription of this.subscriptions) {
    subscription.unsubscribe();
}
```

Martin-Wegner:
@andy-hanson thanks but I don't understand in which case a forEach loop with a void expression will do harm?

andy-hanson:
The goal of the lint rule isn't to prevent crashes, it's to prevent things that are really statements from looking like expressions.

Martin-Wegner:
I thought the goal is

> void functions should not have a return

And `this.subscriptions.forEach((subscription: Subscription) => subscription.unsubscribe());` has no return statement and it does not look like as an expression...

andy-hanson:
Don't see why this should be limited to just the expression in a `return` statement. `const x = alert()` is also a code smell.
The code you have there *is* an expression, and its type is `void`.

Martin-Wegner:
I understand that a void function/method should not return anything and that a void function/method should not on the right side of an assignment.

But I really don't understand why a for-each loop with a void call is a bad thing (and a normal for loop with a void call is a good thing)... Thanks for your patience and help!

adidahiya:
I think most of the issue here is with the implicit return of

```ts
this.subscriptions.forEach((subscription: Subscription) => subscription.unsubscribe());
```

the code would work fine if you did

```ts
this.subscriptions.forEach((subscription: Subscription) => {
   subscription.unsubscribe()
});
```

I do think the first construct is reasonable... perhaps you could add a rule option to `allow-implicit-return`

Martin-Wegner:
@adidahiya thanks for making things clear :)

PR #1914, opened 2016-12-23
Use for..of to iterate in allBlockScopes in preferConstRule

mohsen1:


src/rules/preferConstRule.ts
```diff
@@ -155,8 +155,8 @@ class PreferConstWalker extends Lint.BlockScopeAwareRuleWalker<{}, ScopeInfo> {
     private markAssignment(identifier: ts.Identifier) {
         const allBlockScopes = this.getAllBlockScopes();
         // look through block scopes from local -> global
-        for (let i = allBlockScopes.length - 1; i >= 0; i--) {
-            if (allBlockScopes[i].incrementVariableUsage(identifier.text)) {
+        for (const blockScope of allBlockScopes) {
```
ajafff: We already had a for... of loop in the initial PR, which called .reverse() on the array. That was changed in favor of performance. 
With your change, the loop is from start to end. That won't work for shadowed variables, as the most recent scope is the last in that array. See comment above the loop

Issue #1886, opened 2016-12-16
Consecutive blank lines are forbidden should not apply inside template strings


corbinu:
### Bug Report

- __TSLint version__: `4.0.0`
- __TypeScript version__: `2.1.4`
- __Running TSLint via__: npm script

#### TypeScript code being linted

```ts
const str = `
    here is some text


`;
```

with `tslint.json` configuration:

```json
"rules": {
    "no-consecutive-blank-lines": [
        true
    ]
}
```

#### Actual behavior
Treats the consecutive blank lines inside the string as a violation

#### Expected behavior
Don't treat the consecutive blank lines inside the string as a violation

adidahiya:
sounds like a bug, accepting PRs

IllusionMH:
Is there any alternatives to `getTokenAtPosition` function?
There is easy fix with
```ts
const token = (ts as any).getTokenAtPosition(node, start) as ts.Node;
```
but it is not a part of public API (however `ts as any` already used in [noInferredEmptyObjectTypeRule.ts#L55](https://github.com/palantir/tslint/blob/master/src/rules/noInferredEmptyObjectTypeRule.ts#L55) )

I tried to use `LanguageService` but none of methods that has `AtPosition` returns needed info.

nchen63:
the `ts as any` was only used there to handle backcompat with TypeScript v2.0, since the field doesn't exist in Typescript v2.1.

Seems strange that `getTokenAtPosition` is not public. I don't think we should be using private functions. I guess the alternative would be to walk every node to find the template string nodes, then filter out the errors

ajafff:
I have a fix in my local fork. Will prepare a PR tomorrow(ish)

PR #1919, opened 2016-12-23
Allow consecutive blank lines in template strings

ajafff:
#### PR checklist

- [x] Addresses an existing issue: #1886 
  - [x] Includes tests
- [ ] Documentation update

Find all template strings in AST and use their start and end position as disabled interval to filter failures.

Expressions inside the template string are still checked for consecutive blank lines. See added test cases.


ajafff:
Seems like that build error `src/test.ts[86, 31]: Shadowed variable: 'fileText'` is caused by the new version of tslint

nchen63:
@ajafff thanks!

PR #1920, opened 2016-12-23
Fix build

ajafff:
Rename shadowed variable fileText to pass build.

Ref: #1919

#### PR checklist

  - [ ] Includes tests
- [ ] Documentation update

Build fails because tslint 4.2 now complains about shadowed variable `fileText`, so I renamed it.


nchen63:
@ajafff thanks!

Issue #1900, opened 2016-12-18
Refactor RuleWalker and friends


ajafff:
This is more like a tracking issue. If you are ok with the todos below, I would like to do the refactoring by myself.

__Move funtionality to make it reusable:__
*Why?* I think RuleWalker is too bloated for the common use case. There are all those unnecessary visitor methods on the call stack and everytime `walkChildren` is called, there is a new closure created.
Don't get me wrong: this is pretty convenient for beginners and I don't want to rebuild this from scratch. I just want to use my own and do the AST walking myself. But there is some functionality, which I need and which would belong to other modules or classes in my opinion.
- [x] move `isScopeBoundary` and `isBlockScopeBoundary` to `utils`
- [x] move `addFailure`, failure deduplication (`existsFailure`) and `disabledIntervals`-check to `Rule`

Of course `RuleWalker` should still work the same to be backwards compatible.

__Cleanup:__
- [ ] remove `skip()` and `position` property, as they seem to be not useful at all

__Performance:__
- [x] don't create a new array in `ScopeAwareRuleWalker#getAllScopes()` (`BlockScopeAwareRuleWalker#getAllBlockScopes()` doesn't do this either)
- [x] provide `sourceFile` argument to `node.getStart()`, `getWidth()`, `getText()`, etc. (see #1747)

__Convenience:__
- [x] add helper method to create `Fix` and use `ruleName` from `options`

jmlopez-rod:
These changes seem to be going in the right direction. I would also like to suggest the following:

In the `RuleWalker` make some private properties `public readonly`. Right now I'm writing my custom rule walker as follows:

```ts
class CustomRuleWalker extends SyntaxWalker {
  public readonly sourceFile: ts.SourceFile;
  public readonly sourceText: string;
  public readonly limit: number;
  public readonly options: any[];
  private failures: RuleFailure[];
  private disabledIntervals: IDisabledInterval[];
  private ruleName: string;

  constructor(sourceFile: ts.SourceFile, options: IOptions) {
    super();
    this.sourceFile = sourceFile;
    this.limit = sourceFile.text.length;
    this.sourceText = sourceFile.text.substring(0, this.limit);
    this.options = options.ruleArguments;
    this.failures = [];
    this.disabledIntervals = options.disabledIntervals;
    this.ruleName = options.ruleName;
  }

  public substring(start: number, end: number) {
    return this.sourceText.substring(start, end - start);
  }

[...]

PR #1923, opened 2016-12-23
Refactor walker

ajafff:
#### PR checklist

- [x] Addresses an existing issue: #1900
  - [ ] Includes tests
- [ ] Documentation update

See tracking issue #1900 

I made a separate commit for every change. If you want I can submit them as individual PRs.

There will be a followup PR with the breaking changes for cleanup. That includes removing the is*Boundary _methods_ from their classes and only use utility function instead.


src/language/walker/walker.ts
```diff
@@ -0,0 +1,26 @@
+/**
+ * @license
+ * Copyright 2013 Palantir Technologies, Inc.
```
nchen63: 2016

src/language/walker/ruleWalker.ts
```diff
@@ -68,21 +64,14 @@ export class RuleWalker extends SyntaxWalker {
         }
     }
 
-    public skip(node: ts.Node) {
```
nchen63: I'm really hope nobody is using this. It looks like it implemented with a bunch of other stuff and never used

src/language/walker/walker.ts
```diff
@@ -0,0 +1,26 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import * as ts from "typescript";
+
+import {RuleFailure} from "../rule/rule";
+
+export interface IWalker {
```
adidahiya: can we put this in `walker/index.ts` instead? the import from `/.../walker/walker` doesn't look as nice as `/.../walker`

src/language/walker/ruleWalker.ts
ajafff: I readded the method.
I will remove this in the followup PR with breaking changes

nchen63:
@ajafff thanks!

PR #1925, opened 2016-12-23
Fix restrict-plus-operands for ternary ops

### LINKED ISSUE #1922: restrict-plus-operands false positive with ternary operator
### Bug Report

- __TSLint version__: `4.2.0`
- __Running TSLint via__: CLI

```ts
const balls = 'hello' + (someBoolean ? 'a' : 'b') + (() => someBoolean ? 'c' : 'd')() + 'e';
```

```json
{
	"rules": {
		"restrict-plus-operands": true
	}
}
```

#### Actual behavior

`Operands of '+' operation must either be both strings or both numbers`

#### Expected behavior

No error, as with TS 2.0.10 and tslint 4.1.1.

nchen63:
fixes #1922

### DIFF
--- a/src/rules/restrictPlusOperandsRule.ts
+++ b/src/rules/restrictPlusOperandsRule.ts
@@ -18,6 +18,7 @@
 import * as ts from "typescript";
 
 import * as Lint from "../index";
+import { isTypeFlagSet } from "../language/utils";
 
 export class Rule extends Lint.Rules.TypedRule {
     /* tslint:disable:object-literal-sort-keys */
@@ -40,6 +41,8 @@ export class Rule extends Lint.Rules.TypedRule {
     }
 }
 
+type AddTypes = "string" | "number" | "invalid";
+
 class RestrictPlusOperandsWalker extends Lint.ProgramAwareRuleWalker {
     public visitBinaryExpression(node: ts.BinaryExpression) {
         if (node.operatorToken.kind === ts.SyntaxKind.PlusToken) {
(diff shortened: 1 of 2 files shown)


src/rules/restrictPlusOperandsRule.ts
```diff
@@ -40,6 +41,8 @@ export class Rule extends Lint.Rules.TypedRule {
     }
 }
 
+type AddTypes = "string" | "number" | "invalid";
```
adidahiya: I don't think we need a new type for this; I would prefer the union inlined in two places. This change makes the code more confusing

src/rules/restrictPlusOperandsRule.ts
```diff
@@ -60,12 +63,34 @@ class RestrictPlusOperandsWalker extends Lint.ProgramAwareRuleWalker {
     }
 }
 
-function getAddableType(node: ts.Expression, tc: ts.TypeChecker): "string" | "number" | "invalid" {
+function getAddableType(node: ts.Expression, tc: ts.TypeChecker): AddTypes {
```
adidahiya: I don't think this extra function in the call stack is necessary -- would prefer to inline usage of `getBaseTypeOfLiteralType` above, like so:

```ts
const tc = this.getTypeChecker();
const leftType = getBaseTypeOfLiteralType(ts.getTypeAtLocation(node.left));
...
```

src/rules/restrictPlusOperandsRule.ts
```diff
@@ -60,12 +63,34 @@ class RestrictPlusOperandsWalker extends Lint.ProgramAwareRuleWalker {
...
+    } else if (isTypeFlagSet(type, ts.TypeFlags.Union) && !isTypeFlagSet(type, ts.TypeFlags.Enum)) {
+        const types = (type as ts.UnionType).types.map(getBaseTypeOfLiteralType);
+        return allSame(types) ? types[0] : "invalid";
+    } else if (isTypeFlagSet(type, ts.TypeFlags.EnumLiteral)) {
+        return getBaseTypeOfLiteralType((type as ts.EnumLiteralType).baseType);
     }
     return "invalid";
 }
+
+function allSame(array: string[]) {
```
adidahiya: I think this would be nicer with `Array#reduce`

PR #1927, opened 2016-12-24
Add typeof-compare rule

mohsen1:
#### PR checklist

  - [x] Includes tests
- [x] Documentation update

Added `typeof-compare` rule. It makes sure result of `typeof` expressions are always compared to legal strings. I originally [asked TypeScript](https://github.com/Microsoft/TypeScript/issues/12697) to make comparing result of `typeof` to illegal strings a compiler error but per spec `typeof` can return anything! 

This is my first rule. I'm almost certain I'm missing stuff in this PR.


mohsen1:
How should I update documentation for this?

nchen63:
you can run the doc generator by running `npm run verify`

mohsen1:
@natgabb done

test/rules/typeof-compare/test.js.lint
```diff
@@ -0,0 +1,38 @@
...
+    }
+    const x = typeof foo === 'undefied'
+              ~~~~~~~~~~~~~~~~~~~~~~~~~      [typeof must be compared to correct value]
+}
+
+console.log(typeof window == 'object ')
+            ~~~~~~~~~~~~~~~~~~~~~~~~~~      [typeof must be compared to correct value]
+
+const string = "string";
+typeof bar === string
```
nchen63: needs more positive tests. Currently only tests against a string variable. Needs to test against directly against a string, against a template string, and a template string with a replacement variable. Also exercise different types of valid types

mohsen1:
Hmmm failures don't make sense to me

src/rules/typeofCompareRule.ts
```diff
@@ -0,0 +1,96 @@
...
+
+class ComparisonWalker extends Lint.RuleWalker {
+    private static LEGAL_TYPEOF_RESULTS = ["undefined", "string", "boolean", "number", "function", "object", "symbol"];
+
+    private static isCompareOperator(node: ts.Node): boolean {
+        return node.kind === ts.SyntaxKind.EqualsEqualsToken || node.kind === ts.SyntaxKind.EqualsEqualsEqualsToken;
+    }
+
+    private static isLegalStringLiteral(node: ts.Node) {
+        return ComparisonWalker.LEGAL_TYPEOF_RESULTS.indexOf(node.getText()) > -1;
```
nchen63: it's failing here because `node.getText()` includes the quotes. For example, `"string"` instead of what you want: `string`. If you use `node.text`, it will exclude the quotes

nchen63:
@mohsen1 thanks for the contribution!

Issue #1827, opened 2016-12-06
Enable stricter compiler options in TSLint source


adidahiya:
- [x] noImplicitReturns
- [x] noImplicitThis
- [x] strictNullChecks

PR #1930, opened 2016-12-24
`noImplicitReturns`, `noImplicitThis` and `strictNullChecks`

nchen63:


src/rules/completedDocsRule.ts
```diff
@@ -59,9 +59,10 @@ export class Rule extends Lint.Rules.TypedRule {
         const options = this.getOptions();
         const completedDocsWalker = new CompletedDocsWalker(sourceFile, options, langSvc.getProgram());
 
-        const nodesToCheck = this.getNodesToCheck(options.ruleArguments);
-        completedDocsWalker.setNodesToCheck(nodesToCheck);
-
+        if (options.ruleArguments !== undefined) {
```
ajafff: You can omit all those checks for `ruleArguments !== undefined` if you make ruleArguments a required property in IOptions. AbstractRule initializes it to an empty array anyways. IIRC the only place where it is omitted is the constructor of EnableDisableRulesWalker in linter.ts

nchen63:
fixes #1827

PR #1934, opened 2016-12-25
Semicolon comma fix (fixes #1856) 

### LINKED ISSUE #1856: Semicolon rule incorrectly fixes trailing commas
### Bug Report

- __TSLint version__: `4.0.2`
- __TypeScript version__: `2.0.3`
- __Running TSLint via__: CLI

#### TypeScript code being linted with `--fix`

```ts
interface ITest {
    trailingComma: string,
}
```

#### Actual behavior

A semicolon is appended instead of replacing the comma

```ts
interface ITest {
    trailingComma: string,;
}
```

#### Expected behavior

The comma is replaced with a semicolon

```ts
interface ITest {
    trailingComma: string;
}
```

See failing test case: https://github.com/greglo/tslint/commit/66e7022370a238134439548c3cff6262556b8e22

IllusionMH:
#### PR checklist

- [x] Addresses an existing issue: #1856
- [x] ~New feature,~ bugfix, ~or enhancement~
  - [x] Includes tests
- [ ] Documentation update

This PR adds new error if interface member has comma instead of semicolon after it and proper fix that replaces comma.
Developers still can use `ignore-interfaces` option to skip commas in interfaces.

Thanks to @greglo for failing test case.

### DIFF
--- a/src/rules/semicolonRule.ts
+++ b/src/rules/semicolonRule.ts
@@ -61,6 +61,7 @@ export class Rule extends Lint.Rules.AbstractRule {
     /* tslint:enable:object-literal-sort-keys */
 
     public static FAILURE_STRING_MISSING = "Missing semicolon";
+    public static FAILURE_STRING_COMMA = "Interface properties should be separated by semicolons";
     public static FAILURE_STRING_UNNECESSARY = "Unnecessary semicolon";
 
     public apply(sourceFile: ts.SourceFile): Lint.RuleFailure[] {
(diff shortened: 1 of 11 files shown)


src/rules/semicolonRule.ts
```diff
@@ -171,11 +172,20 @@ class SemicolonWalker extends Lint.RuleWalker {
         const always = !never && (this.hasOption(OPTION_ALWAYS) || (this.getOptions() && this.getOptions().length === 0));
 
         if (always && !hasSemicolon) {
-            const failureStart = Math.min(position, this.getLimit());
-            const fix = new Lint.Fix(Rule.metadata.ruleName, [
-                this.appendText(failureStart, ";"),
-            ]);
-            this.addFailureAt(failureStart, 0, Rule.FAILURE_STRING_MISSING, fix);
+            const lastChild = children[children.length - 1];
```
nchen63: is it possible for children.length === 0?

IllusionMH: I don't think that there are valid TS code where `PropertySignature` doesn't have any child nodes (it should have name at least). Even if it might be `undefined` in other cases, it will be accessed only inside `InterfaceDeclaration` where it can't be valid (I haven't find any such cases).

Issue #1584, opened 2016-09-29
`import* as React from 'react'` should throw a whitespace lint error


mohsen1:
### Bug Report
- **TSLint version**: `"tslint": "^3.15.1",`
- **TypeScript version**: `"typescript": "^2.0.3",`
- **Running TSLint via**: VSCode
#### TypeScript code being linted

``` ts
import* as React from 'react'
```

with `tslint.json` configuration:

``` json
{
    "extends": [
        "tslint:latest"
    ],
    "rules": {
      "quotemark": [
        true, "single", "avoid-escape"
      ]
    }
}

```
#### Actual behavior

No lint error
#### Expected behavior

Whitespace lint error

Issue #1872, opened 2016-12-14
Exclude length from imports via max-line-length


Ne-Ne:
Hi,

Question / Possible feature..?

---

gulp-tslint: 6.1.1 via - Visual Studio

Rule:

https://palantir.github.io/tslint/rules/max-line-length/

`tslint.json` configuration:

```json
"max-line-length": [
  true,
  150
],
```
---
Is there any existing way we can exclude imports from this rule?

```ts
import { item1, item2, item3, item4, item5, item6, item7, item8, item9, item10, item11, item12, item13, item14, item15, item16, item17, item18, item19, item20 } from "@angular/router";
```

Many thanks.

IllusionMH:
@Ne-Ne this is exact case for this rule.
In this case it is more preferable to write imports one in line (much easy merge/review changes in list) or you may load it as 
```ts
import * as routes from "@angular/router";
```

and then use required items as `routes.item2` when needed.

Or you can just use [rule flags](https://palantir.github.io/tslint/usage/rule-flags/) to disable this rule for single line.

adidahiya:
This has been requested before in #1123. I wouldn't mind having an option to disable the rule for import statements, but we won't enable in the built-in configurations.

ajafff:
Closing in favor of #1123

PR #1935, opened 2016-12-25
Add import-spacing rule

mohsen1:
This PR adds `import-spacing` rule which enforces proper spacing in import declarations. 

This is implementation of https://github.com/palantir/tslint/issues/1584

#### PR checklist

  - [x] Includes tests
- [x] Documentation update


mohsen1:
I don't see any failures in the output but my tests are failing. Can you help?

IllusionMH:
@mohsen1 I'm unable to create PR to your fork.

You can find changes that allowed me to fix test in commit https://github.com/palantir/tslint/commit/f37c74dc84e0cf136e266936e3b7cefeae413fc5

mohsen1:
@IllusionMH thanks for looking into it. 
What `nil` means in there?

JKillian:
`~nil` means that the error starts there but has zero width. In this case, it probably means that the error indicates that it starts after the last character of the line (where the line break character would be I suppose). In general it's nicer to avoid outputting errors like this if possible.

IllusionMH:
@mohsen1 `~nil` is described in [lines.ts](https://github.com/palantir/tslint/blob/master/src/test/lines.ts#L85-L86)
In this case error starts on next character after `}` (in second case) and this character is line break - which doesn't have length and can't be marked with regular `~`.

mohsen1:
@JKillian I think this is a valid use case

src/rules/importSpacingRule.ts
```diff
@@ -0,0 +1,124 @@
...
+        ruleName: "import-spacing",
+        description: "Ensures proper spacing between import statement keywords",
+        optionsDescription: "",
+        options: {},
+        optionExamples: [],
+        type: "style",
+        typescriptOnly: false,
+    };
+
+    public static ADD_SPACE_AFTER_IMPORT = "Add space after import";
```
mohsen1: Please feel free to correct my error wordings here.

src/rules/importSpacingRule.ts
```diff
@@ -0,0 +1,124 @@
...
+    public static NO_LINE_BREAKS = "No line breaks are allowed in import declaration";
+
+    public apply(sourceFile: ts.SourceFile): Lint.RuleFailure[] {
+        const comparisonWalker = new ImportStatementWalker(sourceFile, this.getOptions());
+        return this.applyWithWalker(comparisonWalker);
+    }
+}
+
+class ImportStatementWalker extends Lint.RuleWalker {
+    private static IMPORT_KEYWORD_LENGTH = 6; // "import".length;
```
nchen63: you can just define it as `"import".length` and omit the comment. It is probably even more clear to not define this variable and just just `"import".length` everywhere

src/rules/importSpacingRule.ts
```diff
@@ -0,0 +1,124 @@
...
+        super.visitNamespaceImport(node);
+    }
+
+    private checkModuleWithSideEffect(node: ts.ImportDeclaration) {
+        const moduleSpecifierStart = node.moduleSpecifier.getStart();
+        const nodeStart = node.getStart();
+        if ((nodeStart + ImportStatementWalker.IMPORT_KEYWORD_LENGTH + 1) < moduleSpecifierStart) {
+            this.addFailure(this.createFailure(nodeStart, moduleSpecifierStart - nodeStart, Rule.TOO_MANY_SPACES_AFTER_IMPORT));
+        }
+        if ((nodeStart + ImportStatementWalker.IMPORT_KEYWORD_LENGTH) === moduleSpecifierStart) {
```
nchen63: `else if`

src/rules/importSpacingRule.ts
```diff
@@ -0,0 +1,124 @@
...
+            } else if (importClauseStart > (importKeywordEnd + 1)) {
+                this.addFailure(this.createFailure(nodeStart, importClauseStart - nodeStart, Rule.TOO_MANY_SPACES_AFTER_IMPORT));
+            }
+
+            const fromString = node.getText().substring(importClauseEnd - nodeStart, moduleSpecifierStart - nodeStart);
+
+            if (/from$/.test(fromString)) {
+                this.addFailure(this.createFailure(importClauseEnd, fromString.length, Rule.ADD_SPACE_AFTER_FROM));
+            }
+            if (/from\s{2,}$/.test(fromString)) {
```
nchen63: `else if`

src/rules/importSpacingRule.ts
```diff
@@ -0,0 +1,124 @@
...
+                this.addFailure(this.createFailure(importClauseEnd, fromString.length, Rule.TOO_MANY_SPACES_AFTER_FROM));
+            }
+            if (/^\s{2,}from/.test(fromString)) {
+                this.addFailure(this.createFailure(importClauseEnd, fromString.length, Rule.TOO_MANY_SPACES_BEFORE_FROM));
+            }
+            if (/^from/.test(fromString)) {
```
nchen63: `else if`

src/rules/importSpacingRule.ts
mohsen1: Isn't it more performance friendly to do this?

nchen63: The perf difference is not measurable. See the jsperf run https://jsperf.com/str-length

nchen63:
@mohsen1 thanks!

adidahiya:
The banning of line breaks in import statements is problematic here. We need to prescribe a way to make this rule play nicely with max-line-length out of the box. Right now we don't have an option for the line length rule to skip import statements (https://github.com/palantir/tslint/issues/1872#issuecomment-267085885). IMO I would prefer to _keep_ the max-line-length rule in effect for import statements and allow multiline import statements in the default TSLint config.

```ts
import {
    RuleWalker,
    RuleFailure,
    ...
} from "tslint";
```

mohsen1:
It does allow line breaks in destructuring import clause. although it won't let a very long module specifier like this:

```ts
import * as veryVeryVeryVeryVeryVeryVeryVeryVeryVeryLongName from './very-very-very-very-very-very-very-very-very-long-name';
``` 

But a line break might not solve the issue if the specifier itself is longer than max lenght.

adidahiya:
Ah, cool, sorry I missed that. This implementation looks reasonable then.

Issue #1911, opened 2016-12-21
Autofix object-literal-key-quotes


mohsen1:
TSLint 4.1.1

It should be possible to add quotes to properties that miss it.

mohsen1:
I looked into doing this but I couldn't figure out how this rule can know the quotation preference to apply the fix. Other than that fix should be trivial.

JoshuaKGoldberg:
This was resolved! 🎉

PR #1953, opened 2016-12-31
Add fix for object-literal-key-quotes

andy-hanson:
#### PR checklist

- [X] Addresses an existing issue: #1911
  - [X] Includes tests
- [ ] Documentation update

Added fixes for `object-literal-key-quotes` (except in `consistent` mode, where a fix would be ambiguous).

This always uses double quotes. Might it be possible to peer into the setting for `quotemark` and choose quotes accordingly?

### DIFF
--- a/src/rules/objectLiteralKeyQuotesRule.ts
+++ b/src/rules/objectLiteralKeyQuotesRule.ts
@@ -97,6 +97,7 @@ class ObjectLiteralKeyQuotesWalker extends Lint.RuleWalker {
                 break;
             case "consistent":
                 if (quotesAreInconsistent(properties)) {
+                    // No fix -- don't know if they would want to add quotes or remove them.
                     this.addFailureAt(node.getStart(), 1, Rule.INCONSISTENT_PROPERTY);
                 }
                 break;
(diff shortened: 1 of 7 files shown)


mohsen1:
I'm interested to know how rules can know about other rule options

nchen63:
I don't think we want to encourage rules looking at the options of a different rule. I'm fine with the fix using double quotes since most people use double quotes, and the quotemark fixer is likely to come after object-literal-key-quotes and fix the quote type anyways.

nchen63:
@andy-hanson thanks!

mohsen1:
I think allowing quote preference as an option would be nice. Might work on a pr today

mohsen1:
Actually if quote rule fix come after this fix it will fix it. 

I suggest making sure the recommended config has the right order of these two rules

ajafff:
> the quotemark fixer is likely to come after object-literal-key-quotes and fix the quote type anyways.

@nchen63 how is that possible? They operate on the same AST so the order doesn't matter, as the `quotemark` rule wont see the quotes added by `object-literal-key-quotes` until the next run.

With `--fix` linting is run twice - second run without fixing -, if there were fixes. The quotes would only be reported as lint errors but are not fixed.

nchen63:
@ajafff The AST is reloaded each time fixes are made. See linter.ts, line 121.

If it didn't reload, then 2 different fixes would incorrectly be applied since fixes are position based and fix 1 could later offsets to be different than expected.

PR #1960, opened 2017-01-01
Fix fixer tests of object-literal-key-quotes

ajafff:
#### PR checklist

- [ ] Addresses an existing issue: #0000
- [x] New feature, bugfix, or enhancement
  - [x] Includes tests
- [ ] Documentation update
- [x] Enable CircleCI for your fork (https://circleci.com/add-projects)

#### What changes did you make?

Fix test failures introduces by #1953

#### Is there anything you'd like reviewers to focus on?

(optional)

PR #1961, opened 2017-01-01
Fix only-arrow-functions

ajafff:
#### PR checklist

  - [x] Includes tests
- [ ] Documentation update

* Allow function declarations with `allow-named-functions`
  * you could argue, that one could simply use `allow-declarations`, but function declarations are also _named_
  * maybe this commit could just be replaced with a documentation update, as I was not expecting the current behaviour by reading the docs
* fix error location for `async` or `export`ed functions
* cache lookup of options

~~The ci failure will go away when #1960 get's merged and this branch is rebased~~ ... and it's gone

### DIFF
--- a/src/rules/onlyArrowFunctionsRule.ts
+++ b/src/rules/onlyArrowFunctionsRule.ts
@@ -58,7 +58,7 @@ export class Rule extends Lint.Rules.AbstractRule {
 
 class OnlyArrowFunctionsWalker extends Lint.RuleWalker {
     public visitFunctionDeclaration(node: ts.FunctionDeclaration) {
-        if (!this.hasOption(OPTION_ALLOW_DECLARATIONS)) {
+        if (!this.hasOption(OPTION_ALLOW_DECLARATIONS) && !this.hasOption(OPTION_ALLOW_NAMED_FUNCTIONS)) {
             this.failUnlessExempt(node);
         }
         super.visitFunctionDeclaration(node);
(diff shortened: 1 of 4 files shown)


src/rules/onlyArrowFunctionsRule.ts
```diff
@@ -57,24 +57,44 @@ export class Rule extends Lint.Rules.AbstractRule {
 }
 
 class OnlyArrowFunctionsWalker extends Lint.RuleWalker {
+    private allowDeclarations: boolean;
+    private allowNamedFunctions: boolean;
+
+    constructor(sourceFile: ts.SourceFile, options: Lint.IOptions) {
+        super(sourceFile, options);
+        this.allowDeclarations = options.ruleArguments.indexOf(OPTION_ALLOW_DECLARATIONS) !== -1;
```
andy-hanson: Why not use `this.hasOption` here?

ajafff: `options` is already in scope, so why not use it.

I don't have a strong opinion on this one

src/rules/onlyArrowFunctionsRule.ts
```diff
@@ -57,24 +57,44 @@ export class Rule extends Lint.Rules.AbstractRule {
...
     }
 
     private failUnlessExempt(node: ts.FunctionLikeDeclaration) {
         if (!functionIsExempt(node)) {
-            this.addFailureAt(node.getStart(), "function".length, Rule.FAILURE_STRING);
+            this.addFailureAtNode(this.getFunctionKeyword(node), Rule.FAILURE_STRING);
+        }
+    }
+
+    private getFunctionKeyword(node: ts.FunctionLikeDeclaration): ts.Node {
```
andy-hanson: One-liner as of #1962

ajafff: thanks, I changed it to use the new helper function

PR #1962, opened 2017-01-02
Add `childOfKind` helper function

andy-hanson:
#### PR checklist

  - [ ] Includes tests
- [ ] Documentation update

Added a helper function for a common need.
Note that this function may be unperformant since it calls `getChildren`. But it's no worse than before. (At least it's better than `.filter(...)[0]`.)

### DIFF
--- a/src/language/utils.ts
+++ b/src/language/utils.ts
@@ -117,6 +117,24 @@ export function getBindingElementVariableDeclaration(node: ts.BindingElement): t
     return <ts.VariableDeclaration> currentParent;
 }
 
+/** Shim of Array.find */
+function find<T>(a: T[], predicate: (value: T) => boolean): T | undefined {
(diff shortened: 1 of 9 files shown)


src/language/utils.ts
```diff
@@ -117,6 +117,24 @@ export function getBindingElementVariableDeclaration(node: ts.BindingElement): t
     return <ts.VariableDeclaration> currentParent;
 }
 
+/** Shim of Array.find */
```
adidahiya: We don't need this shim -- `Array.prototype.find` is supported on Node v4+, and we only support 4.2+ (see [circle.yml](https://github.com/palantir/tslint/blob/a7cbdef1008e4193653275493c746ae1861d1f76/circle.yml#L7)).

PR #1966, opened 2017-01-02
Rewrite eoflineRule

ajafff:
#### PR checklist

  - [x] Includes tests
- [ ] Documentation update

Add tests.
Allow disabling the rule.
Use SourceFile's text instead of EndOfFileToken.

I'd like to get feedback, if checking for `\n` is a problem on systems where only `\r` is used (macOS?).

The original code was also checking for `\n`, but only if the line before EOF contained only comments or whitespace.


nchen63:
mac uses `/n` and windows uses `\r\n` so I think we're good here

PR #1972, opened 2017-01-02
Add `prefer-arrow-shorthand-return` rule

andy-hanson:
#### PR checklist

  - [X] Includes tests
- [X] Documentation update

Added the `expression-valued-functions` rule, which suggests to change `() => { return x; }` to `() => x`.


src/rules/expressionValuedFunctionsRule.ts
```diff
@@ -0,0 +1,102 @@
...
+            const expr = getSimpleReturnExpression(node.body as ts.Block);
+            if (expr !== undefined) {
+                this.addFailureAtNode(expr, Rule.FAILURE_STRING, this.fix(node, node.body as ts.Block, expr));
+            }
+        }
+
+        super.visitArrowFunction(node);
+    }
+
+    private fix(arrowFunction: ts.FunctionLikeDeclaration, body: ts.Block, expr: ts.Expression): Lint.Fix | undefined {
```
nchen63: rename to `createArrowFunctionFix`

src/rules/expressionValuedFunctionsRule.ts
```diff
@@ -0,0 +1,102 @@
...
+    public static FAILURE_STRING =
+        "Arrow function body may be converted just this expression, with no `{ return ... }`.";
+
+    public apply(sourceFile: ts.SourceFile): Lint.RuleFailure[] {
+        return this.applyWithWalker(new Walker(sourceFile, this.getOptions()));
+    }
+}
+
+class Walker extends Lint.RuleWalker {
+    public visitArrowFunction(node: ts.FunctionLikeDeclaration) {
```
ajafff: You are only handling `ts.ArrowFunction`s here, so you can use it as your parameter's type. Same in the `fix` method below.
You can then omit the check for `node.body &&` in the condition below, as it will always be defined.

src/rules/expressionValuedFunctionsRule.ts
```diff
@@ -0,0 +1,102 @@
...
+ */
+
+import * as ts from "typescript";
+
+import * as Lint from "../index";
+
+export class Rule extends Lint.Rules.AbstractRule {
+    /* tslint:disable:object-literal-sort-keys */
+    public static metadata: Lint.IRuleMetadata = {
+        ruleName: "expression-valued-functions",
```
ajafff: I don't really like this name. To me it's not clear what this rule does without looking at the description. I'm not a native speaker though.
I'd rename it to `prefer-arrow-shorthand-return` or similar ...

src/rules/expressionValuedFunctionsRule.ts
```diff
@@ -0,0 +1,102 @@
...
+        description: "Suggests to convert `() => { return x; }` to `() => x`.",
+        optionsDescription: "Not configurable.",
```
ajafff: Can you add an option to exclude multiline arrow-functions.
Otherwise it would conflict with https://www.npmjs.com/package/vrsource-tslint-rules#multiline-arrow
It's not a core rule, but I like to use it to increase readability.

src/rules/expressionValuedFunctionsRule.ts
```diff
@@ -0,0 +1,102 @@
...
+            // " {"
+            deleteFromTo(arrow.end, openBrace.end),
+            // "return "
+            deleteFromTo(statement.getStart(), expr.getStart()),
+            // " }" (may include semicolon)
+            deleteFromTo(expr.end, closeBrace.end),
+        ]);
+
+        function hasComments(start: number, end: number): boolean {
+            return !isAllWhitespace(text, start, end);
```
ajafff: Can you use ts.get{Leading,Trailing}CommentRanges like in https://github.com/palantir/tslint/blob/master/src/rules/noEmptyRule.ts#L51-L52 instead of implementing it on your own?
If you feel like it, you could add a function to `utils.js`, e.g. `hasCommentBetween(sourceFile, fromNode, toNode)`

src/rules/expressionValuedFunctionsRule.ts
adidahiya: yeah, I like `prefer-arrow-shorthand-return` as well

src/rules/expressionValuedFunctionsRule.ts
adidahiya: I think this should skip multiline arrow functions by default; if you want to lint those too, you can add that as an option

src/rules/preferArrowShorthandReturnRule.ts
```diff
@@ -0,0 +1,105 @@
...
+
+import * as ts from "typescript";
+
+import * as Lint from "../index";
+
+const OPTION_MULTILINE = "multiline";
+
+export class Rule extends Lint.Rules.AbstractRule {
+    /* tslint:disable:object-literal-sort-keys */
+    public static metadata: Lint.IRuleMetadata = {
```
adidahiya: can you please add `hasFix: true` to the metadata?

src/rules/preferArrowShorthandReturnRule.ts
```diff
@@ -0,0 +1,105 @@
...
+        ruleName: "prefer-arrow-shorthand-return",
```
adidahiya: sorry for changing this up again -- I actually think we can simplify the name further and put "shorthand" at the end to be consistent with `object-literal-shorthand`. how about "arrow-return-shorthand"?

andy-hanson:
#2004 fixes tests.

nchen63:
@andy-hanson thanks!

Martin-Wegner:
> no-floating-promises requires type checking

This is missing in the rules documentation.

PR #1977, opened 2017-01-03
Performance improvements pt. 1

ajafff:
#### PR checklist

- [x] ~~New feature, bugfix, or~~ enhancement
  - [ ] Includes tests
- [ ] Documentation update

This is a first attempt to improve performance of core rules. It's basically #1747 ... search for `\.get((Full)?Text|Start|(LeadingTrivia)?Width|(First|Last)Token|Child(ren|Count))\(\)|\.getChildAt\(\d+\)` and pass SourceFile as parameter.

To keep the diff somewhat small, I chose to submit the changes in chunks as I progress. So here we go from `alignRule` until `noDefaultExportRule`.

Of course this is only the low hanging fruit. When I find the time, I will look at the rules in more detail.

### DIFF
--- a/src/language/utils.ts
+++ b/src/language/utils.ts
@@ -54,11 +54,16 @@ export function createCompilerOptions(): ts.CompilerOptions {
 }
 
 export function doesIntersect(failure: RuleFailure, disabledIntervals: IDisabledInterval[]) {
-    return disabledIntervals.some((interval) => {
-        const maxStart = Math.max(interval.startPosition, failure.getStartPosition().getPosition());
-        const minEnd = Math.min(interval.endPosition, failure.getEndPosition().getPosition());
-        return maxStart <= minEnd;
-    });
+    const failureStart = failure.getStartPosition().getPosition();
+    const failureEnd = failure.getEndPosition().getPosition();
+    for (const interval of disabledIntervals) {
+        const maxStart = Math.max(interval.startPosition, failureStart);
+        const minEnd = Math.min(interval.endPosition, failureEnd);
+        if (maxStart <= minEnd) {
+            return true;
+        }
+    }
+    return false;
 }
 
 export function scanAllTokens(scanner: ts.Scanner, callback: (s: ts.Scanner) => void) {
@@ -81,9 +86,12 @@ export function hasModifier(modifiers: ts.ModifiersArray | undefined, ...modifie
         return false;
     }
 
-    return modifiers.some((m) => {
-        return modifierKinds.some((k) => m.kind === k);
-    });
+    for (const modifier of modifiers) {
+        if (modifierKinds.indexOf(modifier.kind) !== -1) {
+            return true;
+        }
+    }
+    return false;
 }
 
 /**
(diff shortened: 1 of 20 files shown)


ajafff:
As a side note: Maybe we could create a tslint rule for that? Maybe in a separate project as this would not be relevant for end users of tslint?

src/language/utils.ts
```diff
@@ -81,9 +86,12 @@ export function hasModifier(modifiers: ts.ModifiersArray | undefined, ...modifie
         return false;
     }
 
-    return modifiers.some((m) => {
-        return modifierKinds.some((k) => m.kind === k);
-    });
+    for (const modifier of modifiers) {
```
nchen63: why was this changed?

src/language/utils.ts
```diff
@@ -54,11 +54,16 @@ export function createCompilerOptions(): ts.CompilerOptions {
 }
 
 export function doesIntersect(failure: RuleFailure, disabledIntervals: IDisabledInterval[]) {
-    return disabledIntervals.some((interval) => {
-        const maxStart = Math.max(interval.startPosition, failure.getStartPosition().getPosition());
-        const minEnd = Math.min(interval.endPosition, failure.getEndPosition().getPosition());
-        return maxStart <= minEnd;
-    });
+    const failureStart = failure.getStartPosition().getPosition();
```
nchen63: why change?

src/language/utils.ts
```diff
@@ -117,22 +125,17 @@ export function getBindingElementVariableDeclaration(node: ts.BindingElement): t
...
-    return undefined;
-}
-
 /**
  * Finds a child of a given node with a given kind.
  * Note: This uses `node.getChildren()`, which does extra parsing work to include tokens.
  */
-export function childOfKind(node: ts.Node, kind: ts.SyntaxKind): ts.Node | undefined {
-    return find(node.getChildren(), (child) => child.kind === kind);
+export function childOfKind(node: ts.Node, kind: ts.SyntaxKind, sourceFile?: ts.SourceFile): ts.Node | undefined {
```
nchen63: why change?

nchen63:
I'm wary of making perf improvements without a performance profiler indicating a problem or a benchmark improvement.

ericanderson:
I can't speak to whether or not there is a performance issue that __needs__ to be fixed, but for loops seem to nearly always be faster than method invocations.

A simple test: http://jsperf.com/array-some-vs-loop

ajafff:
@ericanderson you're right at the moment, but eventually there will be no difference, hopefully.

However, that specific commit was not intended to be included in this or any other PR, so I reverted it.

ajafff:
@nchen63 You're right, I should have done the profiling before changing everything.
I just profiled some pretty big libraries including `TypeScript` with only `one-line`-Rule enabled, which has the most usages of `node.getChildren` and friends. The result was crushing this PR:
Out of 24s runtime only 16ms were spent in `getSourceFileOfNode`.

I'll close this PR and maybe you can even close #1747.

PR #1978, opened 2017-01-03
Enable more rules in codebase

adidahiya:
See commit messages


mohsen1:
Also add import-spacing and typeof-compare to base?

adidahiya:
@mohsen1 I only want to enable rules that have been released -- we use the latest TSLint release to lint our source code, not the current binary in `./bin/tslint`

src/rules/importBlacklistRule.ts
```diff
@@ -76,7 +76,7 @@ class NoRequireFullLibraryWalker extends Lint.RuleWalker {
     }
 
     public visitImportEqualsDeclaration(node: ts.ImportEqualsDeclaration) {
-        let moduleReference = <ts.ExternalModuleReference> node.moduleReference;
+        const moduleReference = <ts.ExternalModuleReference> node.moduleReference;
```
nchen63: were these not caught by the auto fixer?

adidahiya: you mean the angle bracket assertion? this is an outdated diff; I did that in the other PR

PR #1981, opened 2017-01-04
Update tslint:latest config

adidahiya:

Enabled additional rules in the `tslint:latest` config.

### DIFF
--- a/src/configs/latest.ts
+++ b/src/configs/latest.ts
@@ -38,6 +38,7 @@ export const rules = {
     // added in v4.3
     "import-spacing": true,
     "space-before-function-paren": [true, "never"],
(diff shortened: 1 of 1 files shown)


src/configs/latest.ts
```diff
@@ -15,8 +15,32 @@
...
+    // added in v4.2
+    "callable-types": true,
+    "interface-over-type-literal": true,
+    "no-empty-interface": true,
+    "no-string-throw": true,
+
+    // added in v4.3
+    "import-spacing": true,
+    "space-before-function-paren": [true, "never"],
+    "unified-signatures": true,
```
mohsen1: `compare-typeof` too?

Issue #1971, opened 2017-01-02
docs: Clarify supported environments


ajafff:
The docs do not state, what environments are supported by tslint.
I stumbled across this while developing my own rules. It's not clear if I can safely use ES6 features like `Map` or `Set`.

I also see a lot of shims for ES6 features pop up in the codebase (Ref: #1970)

My thoughts on this:
`tslint` runs in a `node` environment. The oldest supported version of node is v4 (currently 4.7.x).
According to [node.green](http://node.green), features like `Map` and `Set` and all those pretty `Array.prototype` methods can be used safely.

So maybe we should specify the correct libs in tsc options or state which is the minimum requirement to run tslint.

nchen63:
Agree, I don't see why we couldn't move to ES6

adidahiya:
+1, the minimum node version is 4.2 (that's the oldest we test on Circle CI)

JoshuaKGoldberg:
A simple entry in `README.md` mentioning the minimum support being `es2015`, as per this issue and `src/tsconfig.json`, would be great.

JoshuaKGoldberg:
☠️ *TSLint's time has come!* ☠️ 

TSLint is no longer accepting most feature requests per #4534. See **[typescript-eslint.io](https://typescript-eslint.io)** for the new, shiny way to lint your TypeScript code with **ESLint**. ✨ 

It was a pleasure open sourcing with you all!

JoshuaKGoldberg:
🤖 Beep boop! 👉  **TSLint is deprecated** 👈 _(#4534)_  and you should **switch to [typescript-eslint](https://typescript-eslint.io)**! 🤖 

🔒 This issue is being locked to prevent further unnecessary discussions. Thank you! 👋

PR #1984, opened 2017-01-05
Set `--lib es6`, and use `Map`/`Set` methods where possible.

andy-hanson:
#### PR checklist

- [X] Addresses an existing issue: #1971
  - [ ] Includes tests
- [ ] Documentation update

Set the `"lib"` option to `["es6", "dom"]` ("dom" typings used by underscore).
**EDIT**: Removed underscore; it's only used by 1 function, `camelize`, that's a one-liner. [Their source](https://github.com/epeli/underscore.string/blob/master/camelize.js), [mine](https://github.com/palantir/tslint/pull/1984/files#diff-e1495d267619047a7cca5cfe8f692729R46).

Used `Map` and `Set` in cases where their absence appeared to have been compensated for.

It's more efficient to lookup in a Map only once, so avoid this pattern:
```ts
if (map.has(key)) {
	use(map.get(key))
}
```

Prefer:
```ts
const got = map.get(key);
if (got !== undefined) {
	use(got);
}
```


src/rules/noMagicNumbersRule.ts
```diff
@@ -68,25 +68,21 @@ export class Rule extends Lint.Rules.AbstractRule {
...
 
     public visitNode(node: ts.Node) {
         const isUnary = this.isUnaryNumericExpression(node);
-        if (node.kind === ts.SyntaxKind.NumericLiteral && node.parent !== undefined && !Rule.ALLOWED_NODES[node.parent.kind] || isUnary) {
-            const text = node.getText();
-            if (!this.allowed[text]) {
-                this.addFailureAtNode(node, Rule.FAILURE_STRING);
-            }
+        const isNumber = node.kind === ts.SyntaxKind.NumericLiteral && !Rule.ALLOWED_NODES.has(node.parent!.kind);
+        const isMagicNumber = (isNumber || isUnary) && !this.allowed.has(node.getText());
```
andy-hanson: Seems like the `isUnary` case should also check for `ALLOWED_NODES`, but this is what it did before.

src/rules/preferForOfRule.ts
```diff
@@ -47,17 +47,15 @@ interface IIncrementorState {
...
 // a map of incrementors and whether or not they are only used to index into an array reference in the for loop
-interface IncrementorMap {
-    [name: string]: IIncrementorState;
-}
+type IncrementorMap = Map<string, IIncrementorState>;
 
 class PreferForOfWalker extends Lint.BlockScopeAwareRuleWalker<{}, IncrementorMap> {
     public createScope() {
-        return {};
+        return new Map();
```
ajafff: this method could just return `void` (need to update type parameters). `scope` is never used anywhere, only `blockScope` is needed.

src/language/walker/skippableTokenAwareRuleWalker.ts
```diff
@@ -21,11 +21,10 @@ import {IOptions} from "../rule/rule";
 import {RuleWalker} from "./ruleWalker";
 
 export class SkippableTokenAwareRuleWalker extends RuleWalker {
-    protected tokensToSkipStartEndMap: {[start: number]: number};
+    private tokensToSkipStartEndMap = new Map<number, number>();
 
     constructor(sourceFile: ts.SourceFile, options: IOptions) {
```
nchen63: don't need constructor anymore

src/rules/noMagicNumbersRule.ts
nchen63: Agree. Opened issue #2005

adidahiya:
this is an awesome PR, thanks @andy-hanson

PR #1986, opened 2017-01-05
Fix false positive in `prefer-for-of` when array is in parentheses

### LINKED ISSUE #1931: prefer-for-of rule false positive
### Bug Report

- __TSLint version__: `4.2.0`

```ts
const a = [1, 2];
const b = [2, 3];
for (let i = 0; i < a.length; i++) {
    console.log(a[i] + b[i]);
}
```

```json
{
    "extends": "tslint:latest",
    "rules": {
        "forin": false,
        "interface-name": [
            false
        ],
        "max-line-length": [
            false
        ],
        "no-var-requires": false,
        "no-console": [
            false
        ],
        "no-string-literal": false,
        "no-reference": false,
        "ordered-imports": [
            false
        ],
        "object-literal-sort-keys": false,
        "variable-name": [
            true,
            "ban-keywords"
        ],
        "no-bitwise": false,
        "member-access": false,
        "arrow-parens": false,
        "array-type": [true, "array"],
        "max-classes-per-file": [false]
    }
}
```

#### Actual behavior

Expected a 'for-of' loop instead of a 'for' loop with this simple iteration

#### Expected behavior

No error.
Because the code cannot be transformed to for...of.

nchen63:
fixes #1931


src/rules/preferForOfRule.ts
```diff
@@ -17,7 +17,7 @@
 
 import * as ts from "typescript";
 import * as Lint from "../index";
-import { isAssignment } from "../language/utils";
+import { isAssignment, unwrapParentheses } from "../language/utils";
```
ajafff: don't know if this is a requirement, but all other rules use the functions from the `Lint` namespace. for consistency you could use `Lint.isAssignment` and `Lint.unwrapParentheses` here

adidahiya: Good point, but that was mainly done for expediency when transitioning the codebase from namespaces to modules -- it was easiest to reference things on the `Lint` object. Now that we have modules, it would be nice to slowly move to more granular imports.

Issue #628, opened 2015-09-01
whitespace rule enhancements


rcchen:
Create some sort of rule for maximum spacing between characters in a single line I think. For example,

``` ts
public foo { // one space, all good

public foo  { // two spaces, this should be flagged
```

adidahiya:
This should be an extension of the current `whitespace` rule.

glen-84:
@adidahiya It's not only for braces. Right now this is valid:

```js
        if     (1)     {
            let x;
        }    else                 {
            let x;
        }
```

That looks quite awful. Perhaps each `check-*` should have a `spaces` integer which controls the number of spaces that must be used.

However, I would imagine that in 99% of cases, a single space is used/required, so another option would be to just add a `single-spaces` "global" option to enforce a single space in all cases except *before* an equals sign (as people often align assignments).

In a perfect world, it would be able to keep track of groups of assignments and report on misalignment, but that may not be simple.

myitcv:
@rcchen @glen-84 rather than enforcing spacing rules (indeed formatting rules in general) by hand, you might be interested in [`tsfmt`](https://github.com/vvakame/typescript-formatter). It automatically formats code for you.

unional:
Another enhancement: `anonymous-function`
```ts
function() {}

// vs
function () {}
```

unional:
Another enhancement: `anonymous-function`
```ts
function() {}

// vs
function () {}
```

glen-84:
@myitcv Thanks for the suggestion, but I'd prefer to use linting in this case.

JoshuaKGoldberg:
Closing per #4534. Thanks for the discussion folks!

PR #2002, opened 2017-01-06
Added "check-preblock" option to whitespace rule

cameron-mcateer:
Also added fixes for inserting spaces and updated tests

#### PR checklist

  - [x] Includes tests
- [ ] Documentation update

I added the option to verify there is whitespace immediately before the opening brace of a block node. I also added fixes for the missing whitespace characters.

### DIFF
--- a/src/rules/whitespaceRule.ts
+++ b/src/rules/whitespaceRule.ts
@@ -26,6 +26,7 @@ const OPTION_MODULE = "check-module";
 const OPTION_SEPARATOR = "check-separator";
 const OPTION_TYPE = "check-type";
 const OPTION_TYPECAST = "check-typecast";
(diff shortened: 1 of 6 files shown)


adidahiya:
nice, this looks like a good rule option.

how does it handle plain blocks?

```ts
{
    const foo = 123;
    // code that just wants to be encapsulated in a block scope
}
```

cameron-mcateer:
@adidahiya Good question! I added a couple examples to the test to show that they still pass.

glen-84:
This seems to be covered by the `one-line` rule (with the `check-whitespace` option).

Unfortunately, neither of these rules check the *number* of spaces (see #628).

Issue #1987, opened 2017-01-05
Rule request: minimal-namespace-depth


JoshuaKGoldberg:
#### TypeScript code being linted

```ts
// File one
namespace A.B.C1 {
    export const D = 7;
}

// File two
namespace A.B.C2 {
    // Could also be referenced by C1.D
    const copy = A.B.C1.D;
}
```

#### Expected behavior

The `A.B.C1.D` reference is unnecessary long, which leads to bloated source files and bloated output code. It would be beneficial for folks still in `namespace` land to have a rule that blocks using unnecessarily prefixed `namespace` paths.

JoshuaKGoldberg:
@andy-hanson does #2008 close this issue?

andy-hanson:
Yes, I mentioned this actually.

PR #2008, opened 2017-01-08
Add `no-unnecessary-qualifier` rule

andy-hanson:
#### PR checklist

- [X] Addresses an existing issue: #1987
  - [X] Includes tests
- [X] Documentation update

Added the `no-unnecessary-qualifier` rule, which warns for `A.x` if just `x` would have the same meaning.

### DIFF
--- a/src/language/rule/rule.ts
+++ b/src/language/rule/rule.ts
@@ -111,6 +111,10 @@ export class Replacement {
         return replacements.reduce((text, r) => r.apply(text), content);
     }
 
+    public static deleteFromTo(start: number, end: number): Replacement {
+        return new Replacement(start, end - start, "");
+    }
+
     constructor(private innerStart: number, private innerLength: number, private innerText: string) {
     }
 
--- a/src/language/rule/rule.ts
+++ b/src/language/rule/rule.ts
@@ -111,10 +111,6 @@ export class Replacement {
         return replacements.reduce((text, r) => r.apply(text), content);
     }
 
-    public static deleteFromTo(start: number, end: number): Replacement {
-        return new Replacement(start, end - start, "");
-    }
-
     constructor(private innerStart: number, private innerLength: number, private innerText: string) {
     }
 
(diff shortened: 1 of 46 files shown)


src/language/rule/rule.ts
```diff
@@ -111,6 +111,10 @@ export class Replacement {
         return replacements.reduce((text, r) => r.apply(text), content);
     }
 
+    public static deleteFromTo(start: number, end: number): Replacement {
```
nchen63: use the one in ruleWalker.ts

nchen63:
@andy-hanson  @adidahiya how about naming this `shortest-qualifier`?

andy-hanson:
The rule is about notifying you that a certain node in your AST can be removed if it's unnecessary. It doesn't measure the length of anything.
If you had a rule that warned on `!!x` (where `x` is of type boolean), you wouldn't call it `shortest-boolean-expression`, you'd call it `no-unnecessary-bang-bang`, right?
Open to name changes, but `shortest` implies that it compares multiple possible qualifiers for length.

nchen63:
My concern is that `no-unnecessary-qualifier` sounds like a binary rule, like either have a qualifier or don't have a qualifier. For example, if you're in the `a.b.c` namespace, it would only warn if the qualifier exactly matches `a.b.c`. Maybe it's because I don't view `a` as a qualifier for `b.c` (but should?) I don't feel that strongly about changing the name either way.

andy-hanson:
In my case, I'm just trying to eliminate single qualifiers -- `X.foo` in namespace `X`.
But in #1987 there were multiple qualifiers and a few, but not all, could be removed.

I can see how `no-unnecessary-qualifier` sounds like it does nothing for the latter case. But `shortest-qualifier` sounds like it does nothing for the former case.

Will think about this later.

ajafff:
This should work for enum too. At least the name implies that. Don't know if it does already, but there is no test for it. 

```typescript
enum Foo {
    Bar, 
    Baz = Foo.Bar // could be written as Baz = Bar
}
```

nchen63:
@andy-hanson thanks! name is ok

PR #2009, opened 2017-01-08
Add JSX text nodes into skip list of walker (fixes #2007)

### LINKED ISSUE #2007: comment-format TSX false positive
### Bug Report

- __TSLint version__: `4.3.1`
- __Running TSLint via__: CLI / Visual Studio Code

```tsx
const a = (
	<a href="https://github.com/">
		https://github.com/
	</a>
);
```

```json
{
}
```

#### Actual behavior

```
test.tsx[3, 11]: comment must start with a space
```

#### Expected behavior

Should be OK.

IllusionMH:
#### PR checklist

- [x] Addresses an existing issue: #2007
- [x] ~New feature,~ bugfix, ~or enhancement~
  - [x] Includes tests
- [ ] Documentation update

This PR adds `JsxText` nodes into skip map in `SkippableTokenAwareRuleWalker`.

1. This change should definitely skip `JsxText` nodes in `EnableDisableRulesWalker`, `CommentWalker` and `JsdocWalker`, but I'm unsure about `WhitespaceWalker` and `NoTrailingWhitespaceWalker`.

2. If `JsxText` node consist of whitespace and "comment" it isn't added to skip map since `getStart()` and `getEnd()` return same value. See [test case](https://github.com/palantir/tslint/pull/2009/files#diff-a6ab2cec0ac51f38a677150e1b954946R32)

### DIFF
--- a/src/rules/noTrailingWhitespaceRule.ts
+++ b/src/rules/noTrailingWhitespaceRule.ts
@@ -41,6 +41,11 @@ export class Rule extends Lint.Rules.AbstractRule {
 }
 
 class NoTrailingWhitespaceWalker extends Lint.SkippableTokenAwareRuleWalker {
+    // prevents skipping of JsxText nodes
+    protected visitJsxText(node: ts.JsxText) {
+        this.walkChildren(node);
+    }
+
     public visitSourceFile(node: ts.SourceFile) {
         super.visitSourceFile(node);
         let lastSeenWasWhitespace = false;
--- a/src/language/walker/skippableTokenAwareRuleWalker.ts
+++ b/src/language/walker/skippableTokenAwareRuleWalker.ts
@@ -37,6 +37,11 @@ export class SkippableTokenAwareRuleWalker extends RuleWalker {
         super.visitTemplateExpression(node);
     }
 
+    protected visitJsxText(node: ts.JsxText) {
+        this.addTokenToSkipFromNode(node);
+        super.visitJsxText(node);
+    }
+
     protected addTokenToSkipFromNode(node: ts.Node) {
         const start = node.getStart();
         const end = node.getEnd();
--- a/src/language/walker/syntaxWalker.ts
+++ b/src/language/walker/syntaxWalker.ts
@@ -206,6 +206,10 @@ export class SyntaxWalker {
         this.walkChildren(node);
     }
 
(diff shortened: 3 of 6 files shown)


ajafff:
> I'm unsure about WhitespaceWalker

`WhitespaceWalker` skips Jsx(Selfclosing)Element anyways, so skipping JsxText may not be a big deal here. I don't know why they are skipped though. Maybe to solve a similar problem like this PR?

IllusionMH:
`WhitespaceWalker` most likely won't be affected by this change, but `NoTrailingWhitespaceWalker` might have some logic here and will require separate implementation.

Still there is a question: if there is need to support text nodes that consist of whitespaces and text that matches comment format?

src/rules/noTrailingWhitespaceRule.ts
```diff
@@ -41,6 +41,11 @@ export class Rule extends Lint.Rules.AbstractRule {
 }
 
 class NoTrailingWhitespaceWalker extends Lint.SkippableTokenAwareRuleWalker {
+    // prevents skipping of JsxText nodes
+    protected visitJsxText(node: ts.JsxText) {
+        this.walkChildren(node);
```
IllusionMH: Is there other way to restore `super.super.visitJsxText` functionality?

ajafff:
Maybe it would be sufficient to just pass the correct `ts.LanguageVariant` to `ts.createScanner`. Currently on `LanguageVariant.Standard` is used, so the scanner doesn't know how to parse jsx. In this case `LanguageVariant.JSX` is needed.
The easiest way would be to use `sourceFile.languageVariant`.

IllusionMH:
@ajafff Thnaks for suggestion. I tried to use `sourceFile.languageVariant` as well as `LanguageVariant.JSX` in scanner initialization, and haven't found any difference at first glance.
I'll try to dive deep into TS scanner difference between language variants and other ways to handle text nodes that has only comment text.

src/rules/noTrailingWhitespaceRule.ts
nchen63: how about adding a constructor to `SkippableTokenAwareRuleWalker` with a parameter `skipJsx`?

IllusionMH:
Fixed in #2036.

nchen63:
superseded by #2036

Issue #1910, opened 2016-12-20
Rule Suggestion: disallow explicit boolean comparisons


chrismbarr:
Novice programmers might have something like this
```ts
let foo = true;
// other code...

if (foo === true) {
  //do something...
}
```

or maybe 

```ts
if (foo !== false) {
  //do something...
}
```
(I have seen both of these in real code!)

When `foo` is a boolean, this is unnecessary and could result in a linting violation.  However if `foo` can be something else like `foo: true | number | null` then these explicit checks should be allowed.

perhaps this could be added as an option to the existing [`strict-boolean-expressions`](https://palantir.github.io/tslint/rules/strict-boolean-expressions/) rule?

PR #2013, opened 2017-01-10
Add `no-boolean-literal-compare` rule

andy-hanson:
#### PR checklist

- [X] Addresses an existing issue: #1910
  - [X] Includes tests
- [X] Documentation update

Added the `no-boolean-compare` rule, which warns for `x === true`.


src/language/rule/rule.ts
```diff
@@ -111,6 +111,10 @@ export class Replacement {
         return replacements.reduce((text, r) => r.apply(text), content);
     }
 
+    public static deleteFromTo(start: number, end: number): Replacement {
```
nchen63: can you use `RuleWalker.deleteText(start: number, length: number)` instead?

src/rules/noBooleanCompareRule.ts
```diff
@@ -0,0 +1,114 @@
...
+    const rightValue = booleanFromExpression(right);
+    if (rightValue !== undefined) {
+        return { negate: rightValue !== operator, expression: left };
+    }
+    return undefined;
+}
+
+function operatorKind(operatorToken: ts.BinaryOperatorToken): boolean | undefined {
+    switch (operatorToken.kind) {
+        case ts.SyntaxKind.EqualsEqualsToken: case ts.SyntaxKind.EqualsEqualsEqualsToken:
```
nchen63: cases on separate lines

test/rules/no-boolean-compare/test.ts.lint
```diff
@@ -0,0 +1,30 @@
+declare const x: boolean;
```
nchen63: add cases for more complex expressions like functions, result of binary boolean expression, and negating a boolean

adidahiya:
how about a more explicit name to clarify that this only lints for literals? `no-boolean-literal-compare`

src/rules/noBooleanLiteralCompareRule.ts
```diff
@@ -62,17 +62,32 @@ class Walker extends Lint.ProgramAwareRuleWalker {
...
         }
 
         this.addFailureAtNode(expression, Rule.FAILURE_STRING(negate), this.createFix(...replacements));
     }
 }
 
+function needsParenthesesForNegate(node: ts.Expression) {
+    switch (node.kind) {
+        case ts.SyntaxKind.AsExpression:
+        case ts.SyntaxKind.BinaryExpression:
```
nchen63: is there a more reliable way to check? 

ternary?

andy-hanson: I can't think of a better way to do this, unfortunately.
A ternary operator wouldn't happen here since it `===` binds more tightly, so the ternary would have to be parenthesized already anyway.

nchen63:
thanks, @andy-hanson

andy-hanson:
Remember to close #1910

Issue #1950, opened 2016-12-30
How to handle invalid config in custom rule


ajafff:
## Question / Request for docs / Feature request

I haven't found anything about this in the docs:
How should implementers of custom rules inform the user about invalid rule configuration?

It seems tslint core rules don't check if all config values are valid. If an invalid config option is specified, it is handled as if it was not provided and some default is taken.
If your editor supports validating `tslint.json` with jsonschema, at least core rule's config will be checked. But that doesn't seem to be as easy for custom rules.

__What implemeters could do:__
(only works if an instance of the rule is created, excluding rules that are disabled with a rule flag)
* throw an error
  * cli will exit and display the error message
  * other tools like vscode addon will crash with no meaningful output
* log a warning to console and continue with defaults (and pollute formatted output)
* ignore and maybe crash later if config is not in the expected format

__What tslint could do:__

* validate rule config according to `Rule.metadata` (related to #889)
* provide an api to log these warnings while not polluting formatted output (see #1840)
* call some optional static method - for example `Rule.validateConfig(config)` - to check configuration for that rule once on startup

JoshuaKGoldberg:
### 💀 _**It's time!**_ 💀

**TSLint is being deprecated** and no longer accepting pull requests for major new changes or features. See #4534. 😱

If you'd like to see this change implemented, you have two choices:
* **Recommended:** Check if this is available in [ESLint](https://eslint.org) + [typescript-eslint](https://typescript-eslint.io) ✅
* _Not Recommended: Fork TSLint locally_ 🤷‍♂️

👋 It was a pleasure open sourcing with you!

_If you believe this message was posted here in error, please comment so we can re-open the issue!_

JoshuaKGoldberg:
🤖 Beep boop! 👉  **TSLint is deprecated** 👈 _(#4534)_  and you should **switch to [typescript-eslint](https://typescript-eslint.io)**! 🤖 

🔒 This issue is being locked to prevent further unnecessary discussions. Thank you! 👋

Issue #889, opened 2015-12-20
Validate tslint.json Rule Options


andy-hanson:
Ignores "max-line-length": 5

Here's my tslint.json:

```
{
    "rules": {
        "max-line-length": 5
    }
}
```

This doesn't result in any errors because I should have written `[true, 5]`.

I'd prefer `5` alone to just work, but if not it would be nice to throw errors for bad rule options.

JKillian:
Actually working on that as we speak. I went through added metadata to all the branches [on this branch](https://github.com/palantir/tslint/tree/next-rule-metadata), and it should be possible to check if options are valid based off of that data. 

In the meantime, rules should all have [example configurations](http://palantir.github.io/tslint/rules/max-line-length/) documented. (Website still a WIP.) The `schema` part refers to the format of data that comes after the `[true,` part and is identical to the config metadata each rule now has. This will all be better documented someday.

JKillian:
Once #1311 merges, it should be fairly easy to implement this

felixfbecker:
It would be great to have this

mesaugat:
Is it possible to validate the tslint.json file from the CLI? 

Maybe something like this: `tslint --validate tslint.json`

I think `tslint -c tslint.json index.js` does validate the configuration before running the linter, but a separate `--validate` flag could be helpful.

JoshuaKGoldberg:
Wondering also if something like https://github.com/gristlabs/ts-interface-checker could be used to lessen the amount of work TSLint has to do... I could have sworn someone was Tweeting about a similar library recently but can't find it.

ReggaePanda:
a `tslint --validate` can be very very helpful!

JoshuaKGoldberg:
### 💀 _**It's time!**_ 💀

**TSLint is being deprecated** and no longer accepting pull requests for major new changes or features. See #4534. 😱

If you'd like to see this change implemented, you have two choices:
* **Recommended:** Check if this is available in [ESLint](https://eslint.org) + [typescript-eslint](https://typescript-eslint.io) ✅
* _Not Recommended: Fork TSLint locally_ 🤷‍♂️

👋 It was a pleasure open sourcing with you!

_If you believe this message was posted here in error, please comment so we can re-open the issue!_

JoshuaKGoldberg:
[...]

PR #2019, opened 2017-01-11
Clean up quotemarkRule.ts

andy-hanson:
#### PR checklist

  - [ ] Includes tests
- [ ] Documentation update

Cleaned up `quotemarkRule.ts`.

`isEnabled()` is still ugly. We need a solution that allows us to declaratively describe the options and get back the parsed options, without calling `hasOptions` over and over again.

### DIFF
--- a/src/rules/quotemarkRule.ts
+++ b/src/rules/quotemarkRule.ts
@@ -19,11 +19,6 @@ import * as ts from "typescript";
 
 import * as Lint from "../index";
 
-enum QuoteMark {
-    SINGLE_QUOTES,
-    DOUBLE_QUOTES,
-}
-
 export class Rule extends Lint.Rules.AbstractRule {
     /* tslint:disable:object-literal-sort-keys */
     public static metadata: Lint.IRuleMetadata = {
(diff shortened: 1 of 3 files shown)


src/rules/quotemarkRule.ts
```diff
@@ -19,7 +19,7 @@ import * as ts from "typescript";
 
 import * as Lint from "../index";
 
-enum QuoteMark {
+const enum QuoteMark {
```
mohsen1: What this does?

andy-hanson: https://www.typescriptlang.org/docs/handbook/enums.html

adidahiya:
I don't think calling `hasOption` over and over is so bad actually. Does it have a noticeable performance impact?

andy-hanson:
I don't think it will matter for performance (though note that it does occur for every source file rather than once at configuration time). I'm more concerned that there is both `isEnabled` and `constructor` duplicating some of the same work. And if validation is ever to be done (#1950 #889), that would make for triplication if some refactoring isn't done.

PR #2028, opened 2017-01-13
Add `prefer-method-signature` rule

andy-hanson:
#### PR checklist

  - [X] Includes tests
- [X] Documentation update

Added the `prefer-method-signature` rule, which recommends to use `foo(): void` over `foo: () => void`.

It looks like these will be semantically equivalent in TS2.2. (And were in 2.0.) https://github.com/Microsoft/TypeScript/issues/13148#issuecomment-270204165

### DIFF
--- a/src/rules/preferMethodSignatureRule.ts
+++ b/src/rules/preferMethodSignatureRule.ts
@@ -23,7 +23,8 @@ export class Rule extends Lint.Rules.AbstractRule {
     /* tslint:disable:object-literal-sort-keys */
     public static metadata: Lint.IRuleMetadata = {
         ruleName: "prefer-method-signature",
-        description: "Prefer `foo(): void` over `foo: () => void`.",
+        description: "Prefer `foo(): void` over `foo: () => void` in interfaces and types.",
+        hasFix: true,
         optionsDescription: "Not configurable.",
         options: null,
         optionExamples: ["true"],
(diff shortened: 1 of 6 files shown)


adidahiya:
How does this interact with `member-ordering`? I believe that rule treats these two different syntaxes differently w/r/t/ ordering.

andy-hanson:
I guess the fix wouldn't fully fix it in that case, because the method would then have to be manually re-ordered.

src/rules/preferMethodSignatureRule.ts
```diff
@@ -0,0 +1,58 @@
...
+
+import * as ts from "typescript";
+
+import * as Lint from "../index";
+
+export class Rule extends Lint.Rules.AbstractRule {
+    /* tslint:disable:object-literal-sort-keys */
+    public static metadata: Lint.IRuleMetadata = {
+        ruleName: "prefer-method-signature",
+        description: "Prefer `foo(): void` over `foo: () => void`.",
```
nchen63: append "in interface and types"

src/rules/preferMethodSignatureRule.ts
```diff
@@ -0,0 +1,58 @@
...
+        optionsDescription: "Not configurable.",
```
nchen63: add `hasFix: true,`

nchen63:
@andy-hanson thanks!

PR #2029, opened 2017-01-13
Wordsmith the error message for no-inferrable-types a bit

mprobst:
In particular, call out that only trivially inferred types of literals
should be left out, not every type that's inferrable.

#### PR checklist

- [x] Documentation update

Wordsmith the error message for no-inferrable-types to make it easier to understand for users why this is considered lint, and when they should use types that could be inferred.


mprobst:
@adidahiya thanks! Somehow the CircleCI thing is stuck for me. Can you merge?

Issue #1555, opened 2016-09-15
Warn enums with possible 0 values used in logical operations


radykov:
I would like a rule that warns against enums with possible 0 values being used in logical operations. 

This is mainly an issue for enums with default values, as typescript (1.8.10) will give the first enum a default value of `0`. People might not realise that their enums as being treated the same as `0` in logical operations.

A problem with this is that assume you have
`enum MyEnum { A, B, C}`
`myObj.enumValue = MyEnum.A;`

Later on you perform the following, it's common to assume the if statement evaluates to `true`.
`if (myObj.enumValue)`

However`myObj.enumValue` evaluates to `MyEnum.A` which evaluates to `0`. `if(0)` is false.

While some people will be expecting this behaviour, my assumption is it might cause more problems than not.

adidahiya:
looks like this is a related but more specific pattern than https://github.com/palantir/tslint/issues/253; maybe it would be incorporated as an option into that rule

adidahiya:
strict-boolean-expressions should handle this.

PR #2033, opened 2017-01-14
strict-boolean-expressions: Add 'allow-null-union', 'allow-undefined-union', 'allow-string', and 'allow-number' options

andy-hanson:
#### PR checklist

  - [X] Includes tests
- [X] Documentation update

Added new options to `strict-boolean-expressions` to make it not quite as strict.
TypeScript allows *anything* in an if-statement, so there's a lot of room in between that and requiring that everything be a boolean.

This rule wouldn't allow you to use `string | number` as a condition even if both `allow-string` and `allow-number` are on. I think that's fine.
It will allow `number | undefined` if `allow-undefined` is on, since that contains `undefined` in the union. But it's probably dangerous if the user just wanted to test for `undefined`. Thoughts?


andy-hanson:
Added an `allow-mix` option to handle the case mentioned above. No more `number | undefined` without the option on.

test/rules/strict-boolean-expressions/allow-mix/test.ts.lint
```diff
@@ -0,0 +1,20 @@
+declare function get<T>(): T;
+
+if (get<string | number>()) {}
+    ~~~~~~~~~~~~~~~~~~~~~~ [If statement condition may be falsy if it is 0. Expected a boolean, undefined-union, or string.]
```
nchen63: This error is unclear. How about `This type is not allowed in the 'if' condition because it could be a number. Allowed types are boolean, undefined-union, or string`. You might have to modify it to make the string easier to generate

test/rules/strict-boolean-expressions/allow-string/test.ts.lint
```diff
@@ -0,0 +1,14 @@
+declare function get<T>(): T;
+
+if (get<string>()) {}
+
+if (get<boolean | string>()) {}
+    ~~~~~~~~~~~~~~~~~~~~~~~ [If statement condition mixes multiple kinds of falsy values. Expected a boolean or string.]
```
nchen63: `This type is not allowed in the 'if' condition because it unions one or more disallowed truthy/falsy types. Allowed types are boolean or string

andy-hanson: Changed "one or more" to "more than one" since it allows having exactly one falsy type.

usuyama:
I like the idea of adding more options to 'strict-boolean-expressions'. Thank you for working on this.

I haven't started yet, but I was planning to add a new option (ex. no-enum-boolean-coercion) to 'strict-boolean-expressions' because our team wants to enable the rule only for enum values first. It is risky because enum values start from 0 by default. On the other hand, we want to allow number-boolean-coercion if we're using enum as flags. (Ref: https://basarat.gitbooks.io/typescript/content/docs/enums.html#enums-as-flags )

Ref: https://github.com/palantir/tslint/issues/1555

Did you consider having options like:
- no-string-boolean-coercion
- no-number-boolean-coercion
- no-undefined-boolean-coercion
so that we can enable the rules step by step? For example, our team wants to check only enum-boolean-coercion first.

andy-hanson:
`strict-boolean-expressions` as it is already forbids everything, including enums, numbers, `RegExp`s, etc. This PR adds flags to *allow* more. So a `no-string-boolean-coercion` flag would be unnecessary, since if you don't want strings, you can just *not* set the `allow-string` option.

Adding an `allow-flags-enum` option *might* be doable, but it would be tricky to detect which ones are flags enums. We might consider an enum a flags enum if all of its members are either powers of two, or are unions of the other members.

nchen63:
nice, thanks @andy-hanson!

PR #2034, opened 2017-01-14
Add test to circleci for environments with typescript@2.0.10

nchen63:


circle.yml
```diff
@@ -4,10 +4,11 @@ general:
       - gh-pages
 dependencies:
   pre:
-    - case $CIRCLE_NODE_INDEX in 0) nvm use 4.2 ;; 1) nvm use 5.7 ;; 2) nvm use 6.1 ;; esac
+    - case $CIRCLE_NODE_INDEX in 0) nvm use 4.2 ;; 1) nvm use 5.7 ;; [2-3]) nvm use 6.1 ;; esac
 test:
   override:
-    - npm run verify
+    - case $CIRCLE_NODE_INDEX in [0-2]) npm run verify ;; 3) npm run clean && npm run compile && npm i typescript@2.0.10 && npm run test ;; esac:
```
adidahiya: how about making `npm run clean && npm run compile && npm i typescript@2.0.10 && npm run test` a new NPM script called `verify-with-ts2.0` to make this more readable?

nchen63: The reason I didn't do that is because it can mess up someone's dev environment without them realizing it

PR #2037, opened 2017-01-14
Add `prefer-function-over-method` rule

andy-hanson:
#### PR checklist

  - [X] Includes tests
- [X] Documentation update

Added the `no-unused-this` rule, which warns for methods that don't use `this` (they could be functions).

Also, once Microsoft/TypeScript#13217 is in, I'll add an exception for `override`, since that's a legitimate case where you would need to write a method that might not use `this`.


ajafff:
Just a suggestion:
You could detect methods that only calls itself recursive.

```typescript
class Foo {
    bar() { this.bar(); }
    ~~~ [lorem ipsum]
}
```

If you want to go the easy way, you could "borrow" my implementation for this rule: https://github.com/ajafff/tslint-consistent-codestyle/blob/master/rules/preferStaticMethodRule.ts

andy-hanson:
@ajaff Thanks! Have you considered turning some of your rules into pull requests here?
`no-collapsible-if` and `no-unnecessary-else` were some I was considering working on.

ajafff:
On second thought this name of this rule is a bit misleading.
For me `no-unused-this` would be something like the following:

```typescript
// function is not using this, but declares it in parameter list
function doStuff(this: Foo, bar: Bar) {
                 ~~~~~~~~~ [unused this]
    return bar;
}
```

andy-hanson:
How about `methods-must-use-this`?

nchen63:
That sort of makes it sound like the rule is suggesting that `this` should be added to the method.

nchen63:
@andy-hanson thanks!

Issue #1983, opened 2017-01-05
Rule request: no-duplicate-super


JoshuaKGoldberg:
#### TypeScript code being linted

```ts
class Foo { }

class Bar extends Foo {
    constructor() {
        super();
        super();
    }
}
```

#### Actual behavior

No problems - JavaScript and TypeScript allow this.

#### Expected behavior

Calling `super()` twice is probably not the intended behavior. It's more likely to be a copy&paste error or improper source control merge.

JoshuaKGoldberg:
FWIW, this was also just PRd as a reference error into the ECMA 262 spec. https://github.com/tc39/ecma262/pull/762

bterlson:
To be clear, it was always a reference error per 262. tc39/ecma262#762 merely proposes that the reference error should occur prior to invoking the super constructor (since it seems surprising that the second super half "works" in its current form).

andy-hanson:
Given that this code fails at runtime, have you tried making an issue on TypeScript?

JoshuaKGoldberg:
@andy-hanson I suggested it to @bterlson, who pointed it that the following behavior is hard to catch:

```typescript
class Foo {
    constructor() {
        if (Math.random() > .99) super();
        super();
    }
}

Getting in the business of determining when a `super()` call is redundant isn't really something the language team should be taking on. The very obvious case is something a linter could do.

gruckion:
> Given that TypeScript accepts this code, which then fails at runtime, have you tried making an issue on TypeScript?

I request that this rule is re-evaluated, given that in 2019 ESLint is replacing TSLink. It does not seem suitable to add this within the typescript subset of rules when it is applicable to standard javascript too.

adidahiya:
If it is applicable to standard javascript, you should file the request in an ESLint-related repo, not here.

PR #2038, opened 2017-01-14
Add `no-duplicate-super`

andy-hanson:
#### PR checklist

- [X] Addresses an existing issue: #1983
  - [X] Includes tests
- [X] Documentation update

Added the `no-duplicate-super` rule, which warns if `super()` appears twice in a constructor.


test/rules/no-duplicate-super/test.ts.lint
```diff
@@ -0,0 +1,40 @@
...
+}
+
+class D {
+    // Unfortunately, we warn here too.
+    constructor() {
+        if (cond()) {
+            super(true);
+        } else {
+            super(false);
+            ~~~~~ [0]
```
ajafff: I think this one should be doable. This could result in many false positives in real world code.

The following would be more challenging and I don't know if we want to support code like that:
```typescript
constructor() {
  if (cond)
    super(true);
  if (!cond)
    super(false);
}
```

And then there is this nasty but totally valid constructor:
```typescript
constructor() {
  if(cond)
    return super(foo);
  if(someOtherCondition)
    return super(bar);
  return super(baz);

Comment (andy-hanson) [inline test/rules/no-duplicate-super/test.ts.lint]: What if the `if` has a `return` instead of `else`? What if there's a `switch` statement with `break`?
TypeScript does control flow analysis, but as far as I can tell, there's no API for accessing it. But I would hate to add a bunch of exceptions that duplicate work done better by TypeScript itself.

File: test/rules/no-duplicate-super/test.ts.lint [hunk 2]
```diff
@@ -0,0 +1,40 @@
...
+        }
+    }
+}
+
+[0]: 'super()' is used twice. It must be called only once.
```
ajafff: would reword to something like: `multiple calls to super() found. It must be called only once.`

nchen63:
I don't think this should go in until it can use control flow analysis. There's going to be a lot of false positives.

andy-hanson:
Added some minimal control flow analysis. Tell me if there are any test cases I missed.

andy-hanson:
Not sure why the tests are failing. It only happens in one of the 4 containers.

nchen63:
Container 3 tests against typescript 2.0.10. I don't know why there aren't any failures in the log, though.

andy-hanson:
I rebased and it works now.

Issue #1912, opened 2016-12-21
member-ordering: Allow category grouping


DanielSWolf:
The `member-ordering` rule gives me a set of member categories (`public-static-field`, `protected-instance-method`...) and allows me to arrange them in any order. The problem with this approach is that I have to specify a *total order* -- that is, I can't have "gray areas" where I can mix members belonging to multiple categories.

There is at least one case where this is problematic: fields depending on one another. Here's an example:

```js
class Foo {
  private static implementationDetail = 42;
  public static bar = Foo.implementationDetail * 2;
}
```

In this case, I *have* to specify the private field `implementationDetail` before the public field `bar`. Otherwise, the resulting JavaScript code will try to initialize `bar` before `implementationDetail`, meaning that `bar` will be set to `undefined * 2`.

In practice, I may have public fields depending on private fields (as in this example), private fields depending on public fields, or even both. So there isn't a *total order* that can work.

What I'd love to see is a way to *group* categories, allowing me to mix members from all categories within a group. The result could look like this:

```js
"member-ordering": [true, { "order": [
    ["public-static-field", "protected-static-field", "private-static-field"],
    "public-static-method",
    "protected-static-method",
    "private-static-method",
    ["public-instance-field", "protected-instance-field", "private-instance-field"],
    "constructor",
    "public-instance-method",
    "protected-instance-method",
    "private-instance-method",
]}]
```

aciccarello:
I think this was resolved

billfranklin:
as of what version?

aciccarello:
@billfranklin Sorry, I should have added more information.

The [rule documentation](https://palantir.github.io/tslint/rules/member-ordering/) demonstrates how to create your own categories. These should allow mixing order within the category.

> You can also make your own categories by using an object instead of a string:
> ```
> {
>     "name": "static non-private",
>     "kinds": [
>         "public-static-field",
>         "protected-static-field",
>         "public-static-method",
>         "protected-static-method"
>     ]
> }
> ```

billfranklin:
@aciccarello I don't think that does what you think it does.  I believe what we are all wanting is the same functionality that ordered-imports has....with the caveat that groups of kinds would be used instead of groups of imports....but since a blank line exists already for most, perhaps a blank line and comment is required?
_from ordered-imports:_
"Groups of imports are delineated by blank lines. You can use these to group imports however you like, e.g. by first- vs. third-party or thematically."

_what we want:_
[...]

PR #2041, opened 2017-01-15
member-ordering: Support categories

andy-hanson:
#### PR checklist

- [X] Addresses an existing issue: #1912
  - [X] Includes tests
- [X] Documentation update

Changed member-ordering to allow to use categories rather than just a single option.
This also lets us get rid of the "old" options as they can be described in terms of new presets using categories.


nchen63:
@andy-hanson I get a whole bunch of errors when linting tslint src files using existing tslint.json on develop

```
(member-ordering) src/rules/whitespaceRule.ts[217, 5]: Declaration of public member not allowed after declaration of private member. Instead, this should come at the beginning of the class/inter
face.
(member-ordering) src/runner.ts[110, 5]: Declaration of public member not allowed after declaration of private member. Instead, this should come at the beginning of the class/interface.
(member-ordering) src/runner.ts[112, 5]: Declaration of public member not allowed after declaration of private member. Instead, this should come at the beginning of the class/interface.
```

PR #2046, opened 2017-01-15
Add `strict-type-predicates` rule

andy-hanson:
#### PR checklist

  - [X] Includes tests
- [X] Documentation update

Added the `strict-type-predicates` rule, which warns for `typeof x === "string"` where `x` is always (or never) a string, and for `x === undefined` where `x` is always (or never) `undefined`.

Can't do `instanceof` checks yet as the type checker does not expose assignability checks. If it did, the `getTypePredicateForKind` function could be removed.


test/rules/strict-type-predicates/test.ts.lint
```diff
@@ -0,0 +1,154 @@
...
+    get<number | undefined>() === null;
+    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~ [F]
+
+    get<number | null>() === null;
+
+    get<number | null>() === undefined;
+    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~ [F]
+
+    get<number | undefined>() === undefined;
+    get<string | undefined>() == null;
```
ajafff: Could this rule also check for non strict comparisons to `null` or `undefined` if only one of them is possible?
Example:
```typescript
get<string|undefined>() == null;
~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~ [use === undefined instead]

get<string|undefined|null>() == null; // still allowed
```

test/rules/strict-type-predicates/test.ts.lint
```diff
@@ -0,0 +1,154 @@
...
+    get<string | null>() == undefined;
+    get<string | null>() == null;
+    get<string | undefined>() == undefined;
```
ajafff: here the rule could suggest to use strict equality comparison

ajafff: Please add a test for the following:

```typescript
get<null|undefined> == null;
~~~~~~~~~~~~~~~~~~~~~~~~~~~~ [T]
get<null|undefined> != undefined;
~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~ [F]
```

andy-hanson:
Would this rule make `typeof-compare` unnecessary?

mohsen1:
@andy-hanson if it catches typos like `typeof foo === 'Object'` it should...

nchen63:
sorry, can you merge with master?

PR #2062, opened 2017-01-17
Rewrite and fix EnableDisableRulesWalker

ajafff:
#### PR checklist

  - [x] Includes tests
- [x] Documentation update

* Add minor clarification to line switches documentation.
* Split up methods into maintainable chunks of code
* Fix bugs to make behaviour as described in the docs:
  * `enable-line` and `disable-line` switched from the start of the line only to the end of the comment. That is correct for single line comments, but multiline comments can be followed by other statements.
  * `enable-next-line` and `disable-next-line` switched from the start of the comment instead of the start of the next line.
  * handling of multiline comment end
    * `/*tslint:disable foo bar*/` removed `bar*/` resulting in swiching only `foo`
    * `/*tslint:disable foo*/` removed `foo*/` resulting in switching **all rules**
    * `/*tslint:disable:foo*/` tried to switch rule `foo*/`, which does not exist -> nothing changed
* add tests for everything stated above

Don't know if this is a breaking change, because it used to be this way for a while, although not documented.

Furthermore this changes the API of `EnableDisableRulesWalker` which is no longer a subclass of `RuleWalker`. This class is exported by the package, so it *could* be used by others. Don't know if care or if we even want to support this, because technically every `Rule` could be used from outside the project.
If we consider this an implementation detail, we should remove the export in `index.ts` for the next major version. This should also be done for all other "internal" functions and classes.

### DIFF
--- a/src/enableDisableRules.ts
+++ b/src/enableDisableRules.ts
@@ -86,28 +86,20 @@ export class EnableDisableRulesWalker {
     }
 
     private handleComment(commentText: string, pos: TokenPosition) {
-        const match = /^\s*tslint:/.exec(commentText);
-        if (match !== null) {
-            // remove everything up to the colon from text
-            commentText = commentText.substr(match[0].length);
-            this.handlePossibleTslintSwitch(commentText, pos);
-        }
-    }
-
-    private handlePossibleTslintSwitch(commentText: string, pos: TokenPosition) {
-        // regex is: start of string followed by either "enable" or "disable"
+        // regex is: start of string followed by any amount of whitespace
+        // followed by tslint and colon
+        // followed by either "enable" or "disable"
         // followed optionally by -line or -next-line
         // followed by either colon, whitespace or end of string
-        const enableOrDisableMatch = /^(enable|disable)(?:-(line|next-line))?(:|\s|$)/.exec(commentText);
-
-        if (enableOrDisableMatch !== null) {
+        const match = /^\s*tslint:(enable|disable)(?:-(line|next-line))?(:|\s|$)/.exec(commentText);
+        if (match !== null) {
             // remove everything matched by the previous regex to get only the specified rules
             // split at whitespaces
             // filter empty items coming from whitespaces at start, at end or empty list
-            let rulesList = commentText.substr(enableOrDisableMatch[0].length)
+            let rulesList = commentText.substr(match[0].length)
                                        .split(/\s+/)
                                        .filter((rule) => !!rule);
-            if (rulesList.length === 0 && enableOrDisableMatch[3] === ":") {
+            if (rulesList.length === 0 && match[3] === ":") {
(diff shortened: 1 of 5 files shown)


src/enableDisableRules.ts
```diff
@@ -18,20 +18,20 @@
...
-import {RuleWalker} from "./language/walker/ruleWalker";
 import {IEnableDisablePosition} from "./ruleLoader";
 
-export class EnableDisableRulesWalker extends RuleWalker {
-    public enableDisableRuleMap: {[rulename: string]: IEnableDisablePosition[]} = {};
-
-    constructor(sourceFile: ts.SourceFile, options: IOptions, rules: {[name: string]: any}) {
-        super(sourceFile, options);
+export class EnableDisableRulesWalker {
+    private enableDisableRuleMap: {[rulename: string]: IEnableDisablePosition[]};
```
andy-hanson: Use a `Map`?

src/enableDisableRules.ts
```diff
@@ -41,116 +41,114 @@ export class EnableDisableRulesWalker extends RuleWalker {
...
             });
         }
     }
 
-    private getLatestRuleState(ruleName: string): boolean {
-        const ruleStateMap = this.enableDisableRuleMap[ruleName];
-
-        return ruleStateMap[ruleStateMap.length - 1].isEnabled;
+    private handleComment(commentText: string, pos: TokenPosition) {
+        const match = /^\s*tslint:/.exec(commentText);
```
andy-hanson: How about `/^\s*tslint:((enable|disable)(-(line|next-line))?):([\w-]*)/` and use the capture groups?

src/enableDisableRules.ts
```diff
@@ -18,20 +18,20 @@
...
 import {forEachComment, TokenPosition} from "./language/utils";
-import {RuleWalker} from "./language/walker/ruleWalker";
 import {IEnableDisablePosition} from "./ruleLoader";
 
-export class EnableDisableRulesWalker extends RuleWalker {
-    public enableDisableRuleMap: {[rulename: string]: IEnableDisablePosition[]} = {};
-
-    constructor(sourceFile: ts.SourceFile, options: IOptions, rules: {[name: string]: any}) {
-        super(sourceFile, options);
+export class EnableDisableRulesWalker {
```
andy-hanson: Since there's no `walk()` this could use a name change.

src/enableDisableRules.ts
ajafff: I wanted to change that, but there are many exported functions that expect this to be an object. So this would require a larger rewrite.

src/enableDisableRules.ts
ajafff: Sounds reasonable. Open for suggestions

nchen63:
Also, I don't think anyone is using this class so I don't think it's breaking

Issue #1465, opened 2016-08-08
Rule suggestion: only throw Error and subclasses


danvk:
I'd like a rule to flag this as invalid:

``` typescript
throw "An error occurred.";
```

in favor of:

``` typescript
throw new Error("An error occurred.");
```

There are good reasons to throw Errors, rather than strings. See http://www.devthought.com/2011/12/22/a-string-is-not-an-error/

adidahiya:
sounds good, we could have `no-throw-literal` like eslint: http://eslint.org/docs/rules/no-throw-literal

JoshuaKGoldberg:
Perfect use case for TypeScript over raw JavaScript... With the type checker this should probably be something like `only-throw-errors`.

```typescript
class Foo { }
const foo = new Foo();
throw foo(); // should be an error
```

> Due to the limits of static analysis, this rule cannot guarantee that you will only throw Error objects.

andy-hanson:
Should be fixed by #2073.

adidahiya:
well, #2073 did not merge, but #1878 (`no-string-throw`) did

PR #1878, opened 2016-12-15
add rule to detect throwing of string literals

atscott:
#### PR checklist

- [ ] Addresses an existing issue: #0000
- [x] New feature, bugfix, or enhancement
  - [x] Includes tests
- [x] Documentation update

#### What changes did you make?

I've added a rule that detects when a throw statement has a string literal, which does not produce a proper stack trace. The suggested fix is to change `throw 'string'` to `throw new Error('string')`. I didn't see an existing issue for this, but can create one if needed.

PR #2073, opened 2017-01-19
Added only-throw-errors rule

JoshuaKGoldberg:
#### PR checklist

- [x] Addresses an existing issue: #1465
  - [x] Includes tests
- [ ] Documentation update
- [x] Enable CircleCI for your fork (https://circleci.com/add-projects) *(I think)*

If something is being thrown whose type is known, its type and base
types must contain the `"Error"` class.

Not feeling great about what I named the `utils` functions.

### DIFF
--- a/src/language/utils.ts
+++ b/src/language/utils.ts
@@ -193,6 +193,34 @@ export function isObjectFlagSet(objectType: ts.ObjectType, flagToCheck: ts.Objec
 }
 
 /**
+ * @returns true if type is or extends from a class of the given name
+ */
+export function isTypeOfClassName(type: ts.Type, className: string): boolean {
+    if (isTypeClassName(type, className)) {
+        return true;
+    }
+
+    const baseTypes = type.getBaseTypes();
+
+    if (baseTypes) {
+        for (const baseType of type.getBaseTypes()) {
+            if (isTypeClassName(baseType, className)) {
+                return true;
+            }
+        }
+    }
+
+    return false;
+}
+
+/**
+ * @returns true if type is a class of the given name
+ */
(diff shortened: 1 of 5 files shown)


JoshuaKGoldberg:
Having problems with the `e` in `catch (e)`, otherwise works...

adidahiya:
we already have [`no-string-throw`](https://github.com/palantir/tslint/pull/1878)... this needs to be reconciled with that rule. is there a significant advantage to using the type checker here?

JoshuaKGoldberg:
Oh, missed that rule! The comments from you, @andy-hanson and @nchen63 are relevant.

This is a situation where the rules do basically the same thing, but one is more powerful because it can use the type checker. There are four options I can think of:
1. Only have the weaker, easier-to-use rule *(current)*
2. Have both rules
3. Only have the stronger, harder-to-use rule
4. Allow for rule logic that only gets activated

Option 4 is interesting, and could be really useful in situations like this. It'd give the best of both worlds without requiring extra user configuration. Think I should file a separate issue?

andy-hanson:
`"unknown"` is the type you get when there's a compile error that prevents you from getting the type.
If I copy-paste that code into VSCode I get correct information on hover, so maybe there's a compile error somewhere else that effects that? Try copying your test code into a '.ts' file and seeing what errors show up.

nchen63:
I think option 4 would be ideal but the details might get messy since this would be the first dual rule. Also, the name `no-string-throw` isn't quite right and changing it would be a breaking change.

JoshuaKGoldberg:
Moved issue to new thread.

Issue #2064, opened 2017-01-18
Allow --test against a glob pattern of directories


JoshuaKGoldberg:
Instead of `yarn test:rules` to run them all, or manually modifying `ruleTestRunner.js` to change what's running, it'd be nice to quickly run `yarn test:rules -- no-floating-promises/*`.

It could be done pretty easily in `ruleTestRunner.ts` with:

```typescript
// Abridged for ease of reading
const paths = process.argv[2] || "**";
glob.sync(`test/rules/${paths}/tslint.json`)
```

mohsen1:
You can use tslint CLI to test a specific rule. See `--test` flag in help. If you are making changes to tslint itself, you can compile and use the new compiled CLI

JoshuaKGoldberg:
> making changes to tslint

@mohsen1 yes, that's what I'm doing now. I'd like to sporadically run `test:rules` on *specific* files without having to recompile the CLI.

JoshuaKGoldberg:
To be a little more elaborate - `--test` only runs against one directory. 

```cmd
C:\Code\tslint>tslint --test test/rules/no-floating-promises/**
C:\Code\tslint\lib\configuration.js:114
            throw new Error("Could not find config file at: " + path.resolve(suppliedConfigFilePath));
            ^

Error: Could not find config file at: C:\Code\tslint\test\rules\no-floating-promises\**\tslint.json
```

It would be nice to have a little more control and be able to test multiple directories. Perhaps letting `--test` run against multiple directories would accomplish the same thing...

nchen63:
I have a slight preference for `--test` that works against more than 1 directory

JoshuaKGoldberg:
That makes sense, they're basically the same. Updated the title.

PR #2079, opened 2017-01-19
Changed --test to take multiple directories

JoshuaKGoldberg:
#### PR checklist

- [x] Addresses an existing issue: #2064
  - [ ] Includes tests
- [x] Documentation update

Used an extra `glob.sync` to allow for wildcards / general glob patterns from `--test`.

A couple of issues I think should be addressed before merge:
- [ ] No tests added. I didn't see much coverage for `runTest` other than the CLI's general exit code testing. I can add exit similar testing for `runTests`, but is there a more direct testing method you'd like instead?
- [ ] Can part of `ruleTestRunner.ts` be replaced with `runTests`?

### DIFF
--- a/README.md
+++ b/README.md
@@ -192,10 +192,11 @@ tslint accepts the following command-line options:
     option is set.
 
 --test:
-    Runs tslint on the specified directory and checks if tslint's output matches
-    the expected output in .lint files. Automatically loads the tslint.json file in the
-    specified directory as the configuration file for the tests. See the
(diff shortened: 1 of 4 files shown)


JoshuaKGoldberg:
Thanks for the quick merge @nchen63! About the extra points in the PR - I'll look into doing both of those as described unless directed otherwise.

nchen63:
Yep, those are both fine. Make sure all changes to ruleTestsRunner are backwards compatible

PR #2084, opened 2017-01-20
Fail build if CircleCI parallelism < 4

nchen63:
We run using 4 nodes so that each node can test a different environment. Since forks default to 1 node, we don't get full coverage from PRs unless parallelism is set to 4 or more

Environments:
- node 4.2
- node 5.7
- node 6.1
- TypeScript 2.0.10

### DIFF
--- a/scripts/assertMinCircleNodes.js
+++ b/scripts/assertMinCircleNodes.js
@@ -1,0 +1,8 @@
+var requiredNodes = 4;
+var nodes = parseInt(process.argv[2], 10);
+if (requiredNodes != null && requiredNodes > nodes) {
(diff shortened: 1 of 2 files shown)


scripts/assertMinCircleNodes.js
```diff
@@ -0,0 +1,8 @@
+var requiredNodes = 4;
+var nodes = parseInt(process.argv[2], 10);
+if (requiredNodes != null && requiredNodes > nodes) {
```
IllusionMH: Shouldn't it be `requiredNodes >= nodes`?

Issue #259, opened 2015-01-15
New Rule: disallow use of an unbound class method in a callback


markelliot:
I'd like to detect and prevent patterns such as:

``` ts
    public foo() { /* body */ }
    public bar() {
        registerCallback(this.foo); // should be registerCallback(() => { this.foo(); });
                                    // or registerCallback(this.foo.bind(this));
    }
```

JKillian:
:+1:

myitcv:
:+1: we had a discussion about this just last week.

Suspect we need #680 in order to do this properly?

adidahiya:
it seems possible to do syntactically, without type information, but it's not an optimal solution:

1. in one pass over the file AST, keep track of all the class methods for each class in the file.
2. in a second pass, look for `CallExpression` nodes within instance methods and scan their arguments. if one of the arguments references another instance method without using the fat arrow or `.bind` syntax, report a failure

^ this would probably fail with class hierarchies, though, because you wouldn't know which `super...` values are methods and which are members if the parent class is declared outside this file.

so yeah, looks like full type information would be needed to do this properly

adidahiya:
Note: this rule should apply to JSX as well:

```ts
// bad
public render() {
    return <form onSubmit={this.handleSubmit}>
}

private handleSubmit(e: React.SyntheticEvent) {
    ...
}
```

PR #2089, opened 2017-01-21
Add `no-unbound-method` rule

andy-hanson:
#### PR checklist

- [X] Addresses an existing issue: #259
  - [X] Includes tests
- [X] Documentation update

Added the `no-unbound-method` rule, which forbids the use of `x.foo` where `foo` is a method on a class or interface, but allows `y => x.foo(y)`.

### DIFF
--- a/src/rules/noUnboundMethodRule.ts
+++ b/src/rules/noUnboundMethodRule.ts
@@ -1,0 +1,75 @@
+/**
+ * @license
+ * Copyright 2017 Palantir Technologies, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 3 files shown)


src/rules/noUnboundMethodRule.ts
```diff
@@ -0,0 +1,75 @@
...
+        optionsDescription: "Not configurable.",
+        options: null,
+        optionExamples: ["true"],
+        type: "functionality",
+        typescriptOnly: true,
+        requiresTypeInfo: true,
+    };
+    /* tslint:enable:object-literal-sort-keys */
+
+    public static FAILURE_STRING = "'this' may be unbound when a method is used as a property without being invoked.";
```
nchen63: How about `Avoid referencing unbound class methods which may cause unintentional scoping of 'this'"

andy-hanson: Removed "class" from that since it may be on an interface.

nchen63:
nice, thanks!

adidahiya:
Nothing like closing a 2 year old feature request 😃 

Can we add tests for JSX usage? As per https://github.com/palantir/tslint/issues/259#issuecomment-156529024 and https://github.com/palantir/tslint-react/issues/63

Issue #1131, opened 2016-04-15
Investigate TSLint Performance


JKillian:
No hard data, but running TSLint on a project's worth of TypeScript files has been feeling a little slow to me as of late. Would be worth looking into and benchmarking to see if there's a particular code chokepoint or rule that is slowing things down. Also would be neat to see how performance changes as the TypeScript language services change.

myitcv:
@JKillian anecdotally, I've noticed this as well.

JKillian:
I generated a quick flamegraph based off of running TSLint on its own src with its own configuration. The first point of note is that `no-unused-variable` and right after it, `no-use-before-declare` are taking up about the same runtime as all the other enabled rules combined. In this case, the recursive nature of walking the AST makes it a little hard to break down the flamegraph at a more granular level than this.

However, from recall and confirmed by a quick grep, these two rules are the only two rules which use special `ts.LanguageServices` APIs: they both use the `getDocumentHighlights` method to find other locations a variable is used.

![image](https://cloud.githubusercontent.com/assets/1400151/14619939/182fbf7e-0588-11e6-8b03-a2999ffdfb9b.png)

JKillian:
And indeed, it does seem that most of the runtime is spent in `getDocumentHighlights`:

![image](https://cloud.githubusercontent.com/assets/1400151/14620829/11109ca4-058d-11e6-91e1-5a395c7e66ec.png)


Here's a breakdown of one of those `getDocumentHighlights` calls:
![image](https://cloud.githubusercontent.com/assets/1400151/14620779/ba22f946-058c-11e6-9eaa-0104dc9dc790.png)

lowkay:
So thinking about this it makes sense it's slow right? Typescript language services are not really designed with the use case, these rules check *every* variable is used. I haven't checked, but I'm pretty sure language services do not cache variable usage, so essentially each variable check is traversing the subset of the AST in which the variable can be in scope.

The solution is to not use language services in cases where all variables are inspected. Instead there needs to be a block scope aware Walker that accumulates variables.

Basically build up a cache, but because of 'use before declare' that will need to process the file and then match up unresolved variables at the end.

Or my premise is totally wrong :P

lowkay:
Anyhow, I was going to rewrite no unused variables in this way to see how complex it is... doing so should also fix the fact it doesn't detect unused imports in all situations... just wanted to see if anybody has more of an insight on this than me...

[...]

PR #2093, opened 2017-01-21
Introduce AbstractWalker for performance

ajafff:
#### PR checklist

- [x] Addresses an existing issue: #1131 
- [x] New feature or enhancement

Implement `AbstractWalker` as announced in https://github.com/palantir/tslint/issues/1900#issuecomment-272658525
As always: all names are available for discussion.
When extending `AbstractWalker` one has to implement the `walk` method to only handle the nodes needed for this rule.
This is done to avoid the poor performance of `SyntaxWalker`. For example `SyntaxWalker.visitNode` currently accounts for 25% of the total runtime - that means the method itself takes that long, not including the methods it calls along the way.

I took the time to rewrite most of the rather simple rules to use the new AbstractWalker. That's not included in this PR, but can be viewed in https://github.com/ajafff/tslint/commit/e6754c998071eeda4afc061cc8c437a16aaf976f (plus some minor unrelated changes)
If you approve this PR, I will submit separate PRs for the rules I changed for easier review.

Perf numbers for linting `tslint`'s sources:
Total time of the actual linting (time spent inside `Linter.applyRule`) excluding any IO.
master: ~9.0 sec
https://github.com/ajafff/tslint/commit/e6754c998071eeda4afc061cc8c437a16aaf976f: ~5.5 sec

//cc @JKillian @andy-hanson @jmlopez-rod @JoshuaKGoldberg as they seem to care for performance, too


andy-hanson:
Thanks for working on this. Excited to see the performance gains. But I think there's still one too many classes.

You have:

```ts
class Rule {
    ...
         return this.applyWithWalker(new NullWalker(sourceFile, this.getOptions()));
    }
}

class NullWalker extends Lint.AbstractWalker {
            if (node.kind === ts.SyntaxKind.NullKeyword && !Lint.someAncestor(node.parent!, isTypeNode)) {
                return this.addFailureAtNode(node, Rule.FAILURE_STRING);
            }
        };
        return ts.forEachChild(sourceFile, cb);
     }
}
```

But the second class only implements one method. An abstract class with a *single* method is a Java-esque workaround for a lack of ordinary functions. This is JavaScript, we just use a function! All we need is an ordinary function that adds failures to a context and returns nothing.

```ts
class Rule {
    ...
    public apply(sourceFile: ts.SourceFile) {
        return this.applyWithWalk(walk);
    }
}

function walk(ctx: Lint.WalkContext): void {
    ts.forEachChild(ctx.sourceFile, recur);

    function recur(node: ts.Node) {
        if (node.kind === ts.SyntaxKind.NullKeyword && !Lint.someAncestor(node.parent!, isTypeNode)) {
            ctx.addFailureAtNode(node, Rule.FAILURE_STRING);
        }
        ts.forEachChild(node, recur);
    }
}
```

(Note that a `WalkContext` would just be an `AbstractWalker` with the abstract method taken out to make it a concrete class.)

src/language/walker/walker.ts
```diff
@@ -17,10 +17,59 @@
...
+
+    public getSourceFile() {
+        return this.sourceFile;
+    }
+
+    public getFailures() {
+        return this.failures;
+    }
+
+    public hasOption(option: any) {
```
andy-hanson: I think we should encourage parsing options up-front rather than calling `hasOption` deep within a loop. Since options are extraneous to this class' responsibility of adding failures, couldn't this all (the `options` property and `hasOptions`) be removed from here? Options could simply be passed to `walk` as an ordinary argument -- or parsed first, then passed as structured data.

src/language/walker/walker.ts
```diff
@@ -17,10 +17,59 @@
...
+    constructor(public readonly sourceFile: ts.SourceFile, options: IOptions) {
+        this.failures = [];
+        this.options = options.ruleArguments;
+        this.limit = sourceFile.getFullWidth();
+        this.ruleName = options.ruleName;
+    }
+
+    public abstract walk(sourceFile: ts.SourceFile): void;
+
+    public getSourceFile() {
```
andy-hanson: Can't we just expose the `sourceFile` property?

src/language/walker/walker.ts
ajafff: I like that idea. I have the following in mind:
```typescript
class AbstractWalker<T> {
    ...
    constructor(public readonly sourceFile: ts.SourceFile, public readonly options: T) {
        ...
    }
}
```
This way you have the choice to parse the options in `Rule` or inside the constructor of your specific walker implementation ... or you can opt out completely by setting the type parameter to `any`;

src/language/walker/walker.ts
ajafff: `sourceFile` is exposed as public property. This methods only needs to be there to implement the `IWalker` interface for compatibility with `RuleWalker`

ajafff:
@andy-hanson Regarding the `WalkContext` thing:
Are you suggesting to remove all Walker classes and implement everything as functions, even if there are other methods now? That would be very awkward to pass the context to every function that could basically just be an instance method and access `this`.

Or do you mean replacing the Walker class with a simple function only where no instance methods are needed?

```typescript
// can be used as the context for a walk function
class WalkContext {
    addFailure() {...}
    getFailures() {...}
    ... // everything else that is on AbstractWalker except walk and getSourceFile
}

// will be used as Walker if the logic is more complex and involves instance methods in the concrete implementation
abstract class AbstractWalker extends WalkContext implements IWalker {
    abstract walk(sourceFile);
    getSourceFile() {...} // so satisfy the interface
}
```

andy-hanson:
I don't think using a class for walking is necessary for any rule. A class where all methods but one are private is just as suspect as an interface with only one method. Private methods can be changed to just function declarations inside the scope of `walk`. No need for `this.` all over.

ajafff:
> Private methods can be changed to just function declarations inside the scope of walk. No need for this. all over.

@andy-hanson I understand your point, but I would not recommend going that route (in this particular case).

First of all this class should and will be extended to migrate things like `ScopeAwareRuleWalker`. Currently there are already some rules extending the walker of another rule. Sure, that's all possible with closures, but not very maintainable IMO.

Second, I wanted this to be somewhat compatible with `RuleWalker` to make migration easier. I'm not a fan of refactoring just for the sake of it.

Third, performance: closures are not at all your friend in the hot path.
Some interesting reads on this matter:
* [performance recommendations by Google](https://developers.google.com/speed/articles/optimizing-javascript)
* [a blog post / rant by trevnorris about callbacks, closures and performance](http://blog.trevnorris.com/2013/08/long-live-callbacks.html) (also read the followup)
* [a rather technical explanation how closures work in V8](http://mrale.ph/blog/2012/09/23/grokking-v8-closures-for-fun.html)
* [a (short) summary of the two items above](http://neversaw.us/2013/09/04/on-the-performance-of-closures-in-v8/)

I'm following the performance work done in node.js very closely. They try to avoid closures by wrapping callbacks and the variables it would close over in a class. As of node v7.x they even use standalone function declarations and use `.bind()` to provide context to the function, because this is [slightly faster in recent V8](http://benediktmeurer.de/2016/01/14/optimizing-bound-functions-further/) (v5.4 IIRC).

andy-hanson:
It wouldn't be too hard to write an `AbstractWalker` in terms of a `WalkContext` (just delegate or inherit from it), so why not support both?

ajafff:
I don't have a strong opinion whether one should prefer classes over closures or the other way around. At the end of the day it's only personal preference which style you choose. That's not my decision ... I'm just a random guy on the internet :laughing:

I made some changes based on your feedback. Now you can do both. Note that `hasOption` is also gone and a parsed `options` object *should* be passed to `AbstractWalker`, `WalkContext` or `Rule.applyWithFunction`.
To showcase this I refactored two existing rules to either extend `AbstractWalker` or use a closure with `WalkContext`.
Looking forward for your feedback.

src/rules/noMagicNumbersRule.ts
```diff
@@ -63,44 +61,31 @@ export class Rule extends Lint.Rules.AbstractRule {
...
+                       (node as ts.PrefixUnaryExpression).operator === ts.SyntaxKind.MinusToken) {
+                this.checkNumericLiteral(node, "-" + ((node as ts.PrefixUnaryExpression).operand as ts.NumericLiteral).text);
+            } else {
+                ts.forEachChild(node, cb);
             }
-        } else {
-            super.visitNode(node);
-        }
+        };
+        return ts.forEachChild(sourceFile, cb);
```
andy-hanson: `ts.forEachChild` actually does return a value (assuming the callback does), so this was a confusing at first.

src/language/rule/rule.ts
```diff
@@ -265,3 +278,37 @@ export class RuleFailure {
...
+export class WalkContext<T> {
+    private limit: number;
+    private failures: RuleFailure[];
+
+    constructor(public readonly sourceFile: ts.SourceFile, public readonly ruleName: string, public readonly options: T) {
+        this.failures = [];
+        this.limit = sourceFile.getFullWidth();
+    }
+
+    public getFailures() {
```
andy-hanson: What's the point of making it private then?

andy-hanson:
Other than that code looks good.

src/language/rule/rule.ts
jmlopez-rod: The failures array could just be declared as `public readonly failures: RuleFailure[] = []` and let the AbstractWalker implement the `getFailures` method.

src/rules/noMagicNumbersRule.ts
ajafff: That's why the callback function here has the return type annotation `: void`.

The return is actually there on purpose to serve [V8's tail call optimizer](https://github.com/v8/v8/blob/master/src/compiler/tail-call-optimization.cc). AND once ES6 proper tail calls get shipped (currently only available behind `--harmony_tailcalls` flag) we get the ability to handle arbitrarily deep nested trees for free.

ajafff:
@andy-hanson @jmlopez-rod I moved `getFailures()` to `AbstractWalker` and made `failures` public readonly in `WalkContext` based on your feedback.

I also made some changes in `AbstractRule` to get the ruleName and ruleArguments easier.

`WalkContext` and therefore `AbstractWalker` now have a fourth parameter `languageService` which is optional. This enables users of `Rule.applyWithFunction` to use the language service, program and type checker.
Because it is optional one always needs to check for `undefined` or use a non-null assertion (`!`) to use it. I currently have no idea how to circumvent that.
One possible solution is to make the parameter required, although I don't think it is actually needed very often.
Another very hacky solution would be the following:
```diff

 export class WalkContext<T> {
+    public readonly languageService: ts.LanguageService;
     constructor(public readonly sourceFile: ts.SourceFile,
                 public readonly ruleName: string,
                 public readonly options: T,
-                public readonly languageService?: ts.LanguageService) {
+                languageService?: ts.LanguageService) {
+        if (languageService !== undefined) {
+            this.languageService = languageService;
+        }
```

jmlopez-rod:
Does the solution you presented above actually work? if we don't pass the language service then its property won't ever get initialized. I would think typescript would complain about something like that. I think it is fine though, we can make it into a runtime error if we need the languageService and it wasn't provided. Otherwise if this is very popular (not to me right now) we could create a `WalkContextWithLanguageService<T>` class that extends `WalkContext<T>`?

andy-hanson:
TypeScript doesn't check for uninitialized properties. Maybe you want a lint rule :smiley:.
Anyway, agree that a `WalkContextWithLanguageService` would be preferrable rather than to have it optional on `WalkContext`. Since `languageService` is mostly used for the `typeChecker`, there should be a `TypedContext` providing `typeChecker` directly as a readonly property. But that all can wait.

ajafff:
Ok, removed the languageService parameter property for now. As @andy-hanson suggested there should be a `TypedContext` added in a followup PR.

Seems like this is ready for now, awaiting final approval.

I would rather wait some time to document this in case changes are required shortly after landing.

Short summary why those changes were made:
* Add `WalkContext` and `Rule.applyWithFunction`: for simple rules you no longer need to declare a new class extending one of the base walkers. Instead you only declare a single function. (For more involved logic I would still recommend a class to keep it maintainable and avoid possible performance penalty)
* Make `options` generic: encourage implementers to preparse their options to avoid constantly calling `hasOption()`
* Implement `AbstractWalker` on top of `WalkContext`:
  * abstract method `walk` forces implementers to do the walking as needed for this rule, which yields a significant performance gain
  * can be used more or less as drop in replacement for `RuleWalker`
  * omits all those unnecessary class methods like `createReplacement` and friends: those are now static methods in `Replacement`

src/language/rule/rule.ts
```diff
@@ -265,3 +278,33 @@ export class RuleFailure {
         return new RuleFailurePosition(position, lineAndCharacter);
     }
 }
+
+export class WalkContext<T> {
+    public readonly failures: RuleFailure[];
+    private limit: number;
```
nchen63: rename to `sourceLastIndex` or similar

src/language/rule/rule.ts
```diff
@@ -265,3 +278,33 @@ export class RuleFailure {
         return new RuleFailurePosition(position, lineAndCharacter);
     }
 }
+
+export class WalkContext<T> {
```
nchen63: belongs in separate file

src/rules/noMagicNumbersRule.ts
nchen63: there is a return value, but why return it from `walk`?

src/language/rule/rule.ts
```diff
@@ -265,3 +278,37 @@ export class RuleFailure {
...
+    public getFailures() {
+        return this.failures;
+    }
+
+    /** Add a failure with any arbitrary span. Prefer `addFailureAtNode` if possible. */
+    public addFailureAt(start: number, width: number, failure: string, fix?: Fix) {
+        this.addFailure(start, start + width, failure, fix);
+    }
+
+    public addFailure(start: number, end: number, failure: string, fix?: Fix) {
```
JKillian: Perhaps we should use `addFailureFromStartToEnd` so it mirrors the methods in `RuleWalker`? I know it's a bit of an awkward name...

src/rules/noMagicNumbersRule.ts
```diff
@@ -63,44 +61,31 @@ export class Rule extends Lint.Rules.AbstractRule {
...
-        const num = getLiteralNumber(node);
-        if (num !== undefined) {
-            if (!Rule.ALLOWED_NODES.has(node.parent!.kind) && !this.allowed.has(num)) {
-                this.addFailureAtNode(node, Rule.FAILURE_STRING);
+class NoMagicNumbersWalker extends Lint.AbstractWalker<Set<string>> {
+    public walk(sourceFile: ts.SourceFile) {
+        const cb = (node: ts.Node): void => {
+            if (node.kind === ts.SyntaxKind.NumericLiteral) {
+                this.checkNumericLiteral(node, (node as ts.NumericLiteral).text);
+            } else if (node.kind === ts.SyntaxKind.PrefixUnaryExpression &&
```
JKillian: unimportant, but I do think this is a nice use case for a type guard function so that the upcoming type assertion can be removed

src/language/rule/abstractRule.ts
```diff
@@ -57,11 +57,17 @@ export abstract class AbstractRule implements IRule {
 
     public abstract apply(sourceFile: ts.SourceFile, languageService: ts.LanguageService): RuleFailure[];
 
-    public applyWithWalker(walker: IWalker): RuleFailure[] {
+    protected applyWithWalker(walker: IWalker): RuleFailure[] {
         walker.walk(walker.getSourceFile());
         return this.filterFailures(walker.getFailures());
     }
 
+    protected applyWithFunction<T>(sourceFile: ts.SourceFile, options: T, walkFn: (ctx: WalkContext<T>) => void) {
```
JKillian: maybe `options` should be a third optional param? Since it's sometimes `undefined`?

src/rules/noMagicNumbersRule.ts
```diff
@@ -63,44 +61,31 @@ export class Rule extends Lint.Rules.AbstractRule {
     public static DEFAULT_ALLOWED = [ -1, 0, 1 ];
 
     public apply(sourceFile: ts.SourceFile): Lint.RuleFailure[] {
-        return this.applyWithWalker(new NoMagicNumbersWalker(sourceFile, this.getOptions()));
+        const options = this.getOptions();
+        const ruleArguments = options.ruleArguments;
+        const allowedNumbers = ruleArguments.length > 0 ? ruleArguments : Rule.DEFAULT_ALLOWED;
+        return this.applyWithWalker(new NoMagicNumbersWalker(sourceFile, options.ruleName, new Set(allowedNumbers.map(String))));
```
JKillian: I like this concept of parsing rule options as part of the rule instead of part of the rule implementation - seems like a better separation of concerns to me.

src/rules/noMagicNumbersRule.ts
```diff
@@ -63,44 +61,29 @@ export class Rule extends Lint.Rules.AbstractRule {
     public static DEFAULT_ALLOWED = [ -1, 0, 1 ];
 
     public apply(sourceFile: ts.SourceFile): Lint.RuleFailure[] {
-        return this.applyWithWalker(new NoMagicNumbersWalker(sourceFile, this.getOptions()));
+        const allowedNumbers = this.ruleArguments.length > 0 ? this.ruleArguments : Rule.DEFAULT_ALLOWED;
```
JKillian: I like this concept of parsing rule options as part of the rule instead of part of the rule implementation - seems like a better separation of concerns to me.

JKillian:
Nice work @ajafff and everyone involved in reviewing! I really like the performance results we're seeing from these changes, should be a major improvement to lint times for all users.

src/rules/noMagicNumbersRule.ts
ajafff: All return statements in this method return `void`. As explained above, it is much faster to use return for calls in tail position when it does not return anything.

src/language/rule/rule.ts
ajafff: I would prefer to leave it as it is now. This new implementation does not need to keep the mistakes made in `RuleWalker`.
When migrating an existing rule, you will have *some* conflicts anyway because of the other missing methods

src/rules/noMagicNumbersRule.ts
ajafff: I agree, but I don't think it's a good idea to scatter the declarations of typeguard functions through the whole codebase. Currently it is done this way which leads to many duplicated small functions. Maybe we should add them to a new module `src/language/typeguard.ts` as needed.

But that's not directly related to this PR and can be considered when migrating other exisiting rules.

ajafff:
@nchen63 @JKillian I think I addressed all of your review comments.

src/rules/noMagicNumbersRule.ts
JKillian: Agreed, no need to change for this PR. Also saw [the discriminated union issue](https://github.com/Microsoft/TypeScript/issues/13634) you created, that would be a neat solution as well.

src/language/rule/rule.ts
JKillian: Assuming all rules will eventually be migrated over to the new style, I'm in agreement. The type system will find incorrect usages though, and the new naming is nicer.

If some rules are going to stay as extensions of `RuleWalker` though, then it'll get confusing to have the different methods existing for perpetuity.

ajafff: Sure, we should try to migrate all existing rules, even `ScopeAwareRuleWalker` and subclasses thereof. At least that's my naive plan.

I don't know if or when we can eventually deprecate and remove `RuleWalker` for implementers of custom rules. In the end it's your decision if you want to keep it for backwards compatibility.

Regarding the method name I don't have a strong opinion. I will change it if you want.

src/language/rule/rule.ts
nchen63: can you put this class in a separate file?

src/language/rule/abstractRule.ts
```diff
@@ -37,33 +37,39 @@ export abstract class AbstractRule implements IRule {
...
-    public getOptions(): IOptions {
-        return this.options;
+    public isEnabled(): boolean {
+        return AbstractRule.isRuleEnabled(this.value);
     }
 
     public abstract apply(sourceFile: ts.SourceFile, languageService: ts.LanguageService): RuleFailure[];
 
-    public applyWithWalker(walker: IWalker): RuleFailure[] {
+    protected applyWithWalker(walker: IWalker): RuleFailure[] {
```
nchen63: can you make this public?

ajafff:
@nchen63 Your latest review comments were on an outdated diff. They are already fixed in a more recent commit.

As of now this PR is only waiting for a final decision whether to rename `AbstractWalker#addFailure` to `addFailureFromStartToEnd`. https://github.com/palantir/tslint/pull/2093#discussion_r98704451

src/language/rule/rule.ts
JKillian: You wrote the code and have thought about it more than me - happy to defer to your preference and leave it as is on this one.

nchen63:
@ajafff thanks!

jmlopez-rod:
This is awesome @ajafff , can't wait till I can start using this.

JoshuaKGoldberg:
^ yeah, this is awesome. Really excited for the perf boost!

JKillian:
of course, we'll need to convert the rules to really start seeing the benefits of this 😄 

@nchen63 might be worth making a wiki page or issue with a list of rules that need to be converted in order of priority?

JKillian:
Also might be worth addressing this issue first before doing a lot of conversion: https://github.com/palantir/tslint/pull/2093#discussion_r98706381

jmlopez-rod:
@JKillian It would be nice to see this list of rules that are up to date so that we may refer to them as examples. A guide to writing and updating the rules would be welcomed as this could make developers mindful of possible performance pitfalls.

Issue #1921, opened 2016-12-23
completed-docs false positive when entire doc comment consists of only tags


buu700:
### Bug Report

- __TSLint version__: `4.2.0`
- __TypeScript version__: `2.1.4`
- __Running TSLint via__: CLI

#### TypeScript code being linted

```ts
/**
 * @ignore
 * @see foo
 */
private static bar () : number {
	return 4;
}
```

with `tslint.json` configuration:

```json
{
	"rules": {
		"completed-docs": true
	}
}
```

#### Actual behavior

`Methods must have documentation.`

#### Expected behavior

No error, as with TS 2.0.10 and tslint 4.1.1.

Rationale: we currently have `/** @ignore */` on all private properties and methods (which will be the case until we have some extra bandwidth to spend on documenting all that stuff), and in some cases where there isn't anything more to add on top of a class description we have public properties whose docs are just `/** @see $ClassName */`.

JoshuaKGoldberg:
A good solution to this could be changing the rule config to allow for conditions that mark a construct as not needing to be documented. Some conditions:
* Not on the whitelist of types (what's currently implemented)
* Has tags but no textual description (project may or may not consider that acceptable)
* Has a tag that indicates it shouldn't be documented, such as `@ignore` or `@inheritdoc`

Proposal: extend the rule options to have global and per-construct settings.

* If settings is the string `"all"`, enable it for every possible construct.
* If settings is an object, apply the `IConstructDocsSettings` logic to every possible constructs.
* If settings is an array, for each member:
    1. If it's a string `"all"`: enable all constructs
    2. If it's a string: enable it for that construct
    3. If it's an object, for each [key, value]:
        * The key is either `"all"` for all constructors or the name of an individual construct
        * The value is either `false` to disable, `true` to enable completely, or an `IConstructDocsSettings`

```typescript
interface IConstructDocsSettings {
    /**
     * Tags that indicate a block is ok.
     */
    "ok-tags"?: boolean;

    /**
     * Having tags without text is ok.
     */
    "tags-without-text"?: boolean;
}
```

Thoughts?

battmanz:
Just to add another usage example. I tend to use [`@inheritdoc`](http://usejsdoc.org/tags-inheritdoc.html) when implementing an interface or abstract class. However, I'm currently getting the `"Documentation must exist for public methods"` warning when doing so.

```typescript
/**
 * An abstract class to illustrate my point.
 */
abstract class Foo {
  /**
[...]

PR #2095, opened 2017-01-22
Added location, type, and privacy to completed-docs

### LINKED ISSUE #1713: Add more lintable locations to completed-docs rule
As per the list by HamletDRC in https://github.com/Microsoft/tslint-microsoft-contrib/issues/162, a more complete list of things that could have rules:

* modules
* exported classes
* all classes
* exported interfaces
* interfaces
* interface members
* exported variables
* functions
* exported functions
* public member variables
* all member variables
* public member methods
* all member methods

JoshuaKGoldberg:
Also adds interfaces, enums, types, and a few other doc-able contructs.

Fixes #1713.
Lays the groundwork for #1921.


src/rules/completedDocsRule.ts
```diff
@@ -19,100 +19,415 @@ import * as ts from "typescript";
...
+export type RequirementDescriptor = IBlockRequirementDescriptor | IClassRequirementDescriptor;
+
+export interface IRequirementDescriptors {
+    [type: string /* DocType */]: RequirementDescriptor;
+}
+
+export type All = typeof Rule.ALL;
+
+export type DocType = All
+    | typeof Rule.ARGUMENT_CLASSES
```
nchen63: isn't `typeof Rule.ARGUMENT_CLASSES` "string"?

JoshuaKGoldberg: What version of TypeScript are you on? For me (VS Code, TypeScript 2.1.X) `typeof Rule.ARGUMENT_CLASSES` correctly resolves to the `"classes"` literal.

Historically, before support for discriminated literal types, I liked having this explicit typing - both to inform the user and for future-proofing the type declarations.

nchen63: I'm using VS Code, TS 2.1.4. So you're saying that `const a: Visibility = "s";` causes a compilation error for you?

nchen63: Did you meant to use `const` instead of `public static`? In that case, I see the string literal type.

JoshuaKGoldberg: Oh, you're totally right. Thanks for the catch! I've switched them to `const`.

nchen63:
I think you need to rebase

JoshuaKGoldberg:
Rebase should be good to go.

nchen63:
I think this may be considered a breaking change if someone already specifies "all" because they'll see new errors. Thoughts?

JoshuaKGoldberg:
Agreed. Changed it to use the current default, so that the new locations are opt-in. For the next TSLint release we could consider expanding default coverage.

FDiskas:
:running_shirt_with_sash: when it will be released?

adidahiya:
@FDiskas it's available since v4.5. you can find out by yourself by clicking through to the squashed commit:

image

and look at the tags list:

image

Viktor-Bredihin:
what's the status? I still can't get it work

JoshuaKGoldberg:
@viktor-bredihin what is your tslint.json config, and what are you trying to lint? Can you post a minimum reproduction we can take a look at? What problems are you seeing?

Viktor-Bredihin:
@JoshuaKGoldberg 
I'm trying to make this rule work '"completed-docs": true', it's not showing any errors in webstorm, however works fine when I use tslint in the console

JoshuaKGoldberg:
Ah - that's not an issue with this rule. That's an issue with the Webstorm TSLint extension.

`completed-docs` is a "type checked" rule ([link to docs](https://palantir.github.io/tslint/usage/type-checking)). That means it uses more expensive APIs from TypeScript. Many extensions don't support this kind of rule for performance reasons.

PR #2096, opened 2017-01-22
Added tests for wildcard --test

JoshuaKGoldberg:
Followup to #2079.


nchen63:
@JoshuaKGoldberg thanks for the followup PR

PR #2100, opened 2017-01-22
Add `no-let-undefined` rule

andy-hanson:
#### PR checklist

  - [X] Includes tests
- [X] Documentation update

Added the `no-let-undefined` rule, which forbids `let x = undefined;`, preferring `let x;`.

### DIFF
--- a/src/rules/noLetUndefinedRule.ts
+++ b/src/rules/noLetUndefinedRule.ts
@@ -1,0 +1,56 @@
+/**
+ * @license
+ * Copyright 2017 Palantir Technologies, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import * as ts from "typescript";
+
+import * as Lint from "../index";
+
+export class Rule extends Lint.Rules.AbstractRule {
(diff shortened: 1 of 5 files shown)


src/rules/noLetUndefinedRule.ts
```diff
@@ -0,0 +1,56 @@
...
+import * as ts from "typescript";
+
+import * as Lint from "../index";
+
+export class Rule extends Lint.Rules.AbstractRule {
+    /* tslint:disable:object-literal-sort-keys */
+    public static metadata: Lint.IRuleMetadata = {
+        ruleName: "no-let-undefined",
+        description: "Forbids a 'let' statement to be initialized to 'undefined'.",
+        optionsDescription: "Not configurable.",
```
nchen63: hasFix: true

test/rules/no-let-undefined/test.ts.lint
```diff
@@ -0,0 +1,9 @@
+let x: string | undefined = undefined;
```
nchen63: add arg initializer test?
add `const x = undefined;` which shouldn't fail lint

ajafff:
IMO this rule is a bit too specialized.
What about `var`? Why not check parameters and destructuring? Then it could also be renamed to `no-unnecessary-initializer`

andy-hanson:
Yeah, this could work for `var`.
It can't be extended to parameters, though: `function f(x: string | undefined = undefined)` will allow `f()`, but `function x(x: string | undefined)` will not. (Besides, one should use `x?: string` anyway.

For destructuring you mean to warn on this?

```ts
declare function f(): { x?: number };
const { x = undefined } = f(); // Could be `const { x } = f()`
```

ajafff:
Didn't know there is this difference for parameters. So the checking would be limited to optional parameters only ...

Yes, that's what I had in mind for destructuring. All of the following examples should result in a failure:
```typescript
let {foo = undefined} = someObject; // doesn't need initializer, always defaults to undefined
let [foo = undefined] = someArray; // same as above

let {foo = undefined} = {}; // is valid only if there is an initializer, but this is either by mistake or really bad practice to declare variables
let [foo = undefined] = []; // same as above 
```

Issue #973, opened 2016-02-17
Add "alphabetize" option to member-ordering rule


adidahiya:
Enabling the `"alphabetize"` option would enforce alphabetic ordering of properties and methods (as separate groups).

glen-84:
Should this have been closed by #2101?

PR #2101, opened 2017-01-22
member-ordering: Add 'alphabetize' option

andy-hanson:
#### PR checklist

- [X] Addresses an existing issue: #973
  - [X] Includes tests
- [X] Documentation update

Added the 'alphabetize' option to member-ordering, which enforces alphabetical ordering within each category.

This uses `<` for string comparison. I think `.localeCompare` would be troublesome if people in different locales want linting to succeed on all of their machines.

### DIFF
--- a/src/rules/memberOrderingRule.ts
+++ b/src/rules/memberOrderingRule.ts
@@ -251,7 +251,7 @@ export class MemberOrderingWalker extends Lint.RuleWalker {
                     }
 
                     const curName = nameString(member.name);
-                    if (prevName !== undefined && curName < prevName) {
+                    if (prevName !== undefined && caseInsensitiveLess(curName, prevName)) {
                         this.addFailureAtNode(member.name,
                             Rule.FAILURE_STRING_ALPHABETIZE(this.findLowerName(members, rank, curName), curName));
                     } else {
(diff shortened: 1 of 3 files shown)


src/rules/memberOrderingRule.ts
```diff
@@ -309,12 +321,41 @@ export class MemberOrderingWalker extends Lint.RuleWalker {
...
                 this.addFailureAtNode(member, errorLine1);
             } else {
+                if (this.opts.alphabetize && member.name) {
+                    if (rank !== prevRank) {
+                        // No alphabetical ordering between different ranks
+                        prevName = undefined;
+                    }
+
+                    const curName = nameString(member.name);
+                    if (prevName !== undefined && curName < prevName) {
```
nchen63: I think `<` takes case into account, but I don't think it should

nchen63: ah, never mind, just saw your other comment

nchen63: For example, in `ordered-imports`, we use `<` and `>`, but it calls out case-insensitive, lowercase-first and lowercase-last. I think we could just default to case insensitive and call `.toLowerCase()` before comparison

nchen63:
@andy-hanson please rebase

nchen63:
@andy-hanson thanks!

ackvf:
How to use the option? It's not apparent from the documentation.

PR #2102, opened 2017-01-22
Add `await-promise` rule

andy-hanson:
#### PR checklist

  - [X] Includes tests
- [X] Documentation update

Added the `await-Promise` rule, which ensures that anything in an `await` expression could actually be a Promise. Prevents needlessly awaiting synchronously available values.

This is complementary to  #1632: That ensures that Promises are awaited, while this ensures that awaiteds are Promises.

### DIFF
--- a/src/rules/awaitPromiseRule.ts
+++ b/src/rules/awaitPromiseRule.ts
@@ -1,0 +1,74 @@
+/**
+ * @license
+ * Copyright 2017 Palantir Technologies, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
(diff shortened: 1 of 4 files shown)


src/rules/awaitPromiseRule.ts
```diff
@@ -0,0 +1,74 @@
...
+ * limitations under the License.
+ */
+
+import * as ts from "typescript";
+import * as Lint from "../index";
+
+export class Rule extends Lint.Rules.TypedRule {
+    /* tslint:disable:object-literal-sort-keys */
+    public static metadata: Lint.IRuleMetadata = {
+        ruleName: "strict-boolean-expressions",
```
IllusionMH: Looks like wrong `ruleName`. Should be `await-promise` instead of `strict-boolean-expressions`.

mohsen1: Good catch! 

Can we have a script that uses a template for creating a rule please?

PR #2106, opened 2017-01-22
Rename `no-let-undefined` to `no-unnecessary-initializer`

andy-hanson:
* And handle 'var', destructuring, and parameter initializers

#### PR checklist

  - [X] Includes tests
- [X] Documentation update

Handled more things in `no-let-undefined`, so changing the name too. See comments on #2100.


test/rules/no-unnecessary-initializer/test.ts.lint
```diff
@@ -0,0 +1,23 @@
...
+    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~ [0]
+
+for (let c: number | undefined = undefined; c < 2; c++) {}
+         ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~ [0]
+
+let d;
+
+const e = undefined;
+
+function f(x: string | undefined = undefined) {}
```
ajafff: It should not recommend making a parameter optional if it is followed by a required parameter:
```typescript
function f(foo: string|undefined = undefined, bar: number) {} // no error, parameter foo cannot be optional

function f2(foo: string|undefined = undefined, bar: number = 1) {} // foo can be optional
            ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~ [1]
```

test/rules/no-unnecessary-initializer/test.ts.lint
```diff
@@ -0,0 +1,24 @@
...
+    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~ [0]
+
+for (let c: number | undefined = undefined; c < 2; c++) {}
+         ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~ [0]
+
+let d;
+
+const e = undefined;
+
+function f(x: string | undefined = undefined, bar: number) {}
```
ajafff: On second thought, on this function the `= undefined` should lead to failure [0].
If parameter `bar` is required, you need to pass a value for `x`. So is makes no sense to set a default of `undefined`, because the default value is only used when the passed value for that parameter is `undefined`.

nchen63:
@andy-hanson thanks!

PR #2113, opened 2017-01-24
Fix typo

andy-hanson:
Mistake from #2102

### DIFF
--- a/src/rules/awaitPromiseRule.ts
+++ b/src/rules/awaitPromiseRule.ts
@@ -21,7 +21,7 @@ import * as Lint from "../index";
 export class Rule extends Lint.Rules.TypedRule {
     /* tslint:disable:object-literal-sort-keys */
     public static metadata: Lint.IRuleMetadata = {
-        ruleName: "strict-boolean-expressions",
+        ruleName: "await-promise",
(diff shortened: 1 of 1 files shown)


andy-hanson:
`node_modules/@types/underscore` has been in the CircleCI cache for weeks now. Any idea how I can make it go away?

IllusionMH:
> The Rebuild without cache button in the UI will disable the cache for a single build so that you could debug any problems caused by your dependencies.
> If such rebuild successfully passes the dependency phase, it will save a new cache which will be used by future builds.
> 
In theory - should work fine after first successful `Rebuild without cache`.

> Normally you will not have to clear the cache permanently, but if you feel that’s what you need, you can just remove the necessary parts of the cache anywhere in your `circle.yml`’s dependencies section, before the cache is saved:
> ```
>dependencies:
>   post:
>     - rm -r ~/.gradle
> ```
Or as alternative cache can be removed during build process for some time (even if it will slowdown builds) or switch from `npm install` to `yarn install` (I suppose it will use different cache, however not sure if problem not in npm cache itself or with cached by Circle Ci `node_modules` folder)


Quotes from [How cache works](https://circleci.com/docs/how-cache-works/)

PR #2117, opened 2017-01-25
Add `match-default-export-name` rule

andy-hanson:
#### PR checklist

  - [X] Includes tests
- [X] Documentation update

Added the `match-default-export-name` rule, which requires that default imports have the same name as the default exports they reference.
This doesn't check named exports, since they require explicit `import { foo as bar }` and are unlikely to be name-changed by mistake.
This also doesn't check `import =` and `export =`, since `export =` comes from declaration files which probably don't use the name you want, and if it's an `@types` module you can't simply change it.


adidahiya:
very cool! this programming error was one of my main motivations for the [no-default-export](https://palantir.github.io/tslint/rules/no-default-export/) rule.

src/rules/matchDefaultExportNameRule.ts
```diff
@@ -0,0 +1,69 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import * as ts from "typescript";
+
+import * as Lint from "../index";
+
+export class Rule extends Lint.Rules.AbstractRule {
```
nchen63: use `TypedRule` and `applyWithProgram`

test/rules/match-default-export-name/test.ts.lint
```diff
@@ -0,0 +1,3 @@
+import b from "./named";
+       ~ [Default export 'a' is imported as 'b'.]
+import anyName from "./anonymous";
```
andy-hanson: That's not legal syntax anyway.

nchen63:
@andy-hanson thanks!

PR #1957, opened 2016-12-31
Rewrite member-ordering rule

andy-hanson:
#### PR checklist

- [X] Addresses an existing issue: #1954
- [X] New feature, bugfix, or enhancement
  - [X] Includes tests
- [ ] Documentation update
- [X] Enable CircleCI for your fork (https://circleci.com/add-projects)

#### What changes did you make?

Major rewrite of member-ordering. This is similar to #1874 in that it removes instance state from the class and uses "manual" iteration instead of doing it through the Walker.

#### Is there anything you'd like reviewers to focus on?

Discovered #1956 while working on this.
Wouldn't it be possible to specify the "old" options in terms of the "new" style? The error messages might change a bit, but it could error at the same places.

PR #2123, opened 2017-01-26
 Support multiple old-style options.

andy-hanson:
Re: https://github.com/palantir/tslint/pull/2041#issuecomment-275024439
Unreverts #2118 (restores #1957 and #2041).
New changes are at https://github.com/palantir/tslint/pull/2123/commits/81266d62ce63260271d1d84bd577af352e46a7df .

I was not aware that the old-style options for `member-ordering` supported more than one option at once. Had to update the code to reflect that.

Why don't we always use `./bin/tslint` to lint tslint instead of the one in node_modules?


adidahiya:
> Why don't we always use ./bin/tslint to lint tslint instead of the one in node_modules?

This can produce complex issues -- if there's a bug in TSLint, it may not lint properly.

andy-hanson:
That was the idea -- it would fail immediately, rather than after the next publish to NPM.

PR #2126, opened 2017-01-26
Perf improvements, cache/reuse to reduce i/o

nchen63:
running tslint on src/:

old: ~6.4s
new: ~5.7s


ajafff:
While you are already changing ruleLoader.ts: can you change `loadRules` to skip loading and instantiating rules that are disabled in tslint.json. That should be as simple as inserting `if (!(ruleName in enableDisableRuleMap)) continue;` somewhere around line 45

ajafff:
Is it possible to avoid `findRule` completely for cached rules?
The PR as it is now only caches `require`. That's sufficient for core rules, but for third party rules all rule directories will be searched before the cache can be used. And that's done for every SourceFile although the absolute path of the rule should not change (correct me if I'm wrong).

nchen63:
@ajafff my concern is that using a single path for the Rule key is more straightforward than creating a new key from an array of paths for caching `findRule`. I can add another check in `findRule` to avoid repeatedly hitting `loadRule`

src/ruleLoader.ts
```diff
@@ -18,13 +18,15 @@
...
-import {camelize, dedent} from "./utils";
+import { getRelativePath } from "./configuration";
+import { AbstractRule } from "./language/rule/abstractRule";
+import { IDisabledInterval, IRule } from "./language/rule/rule";
+import { arrayify, camelize, dedent } from "./utils";
 
 const moduleDirectory = path.dirname(module.filename);
 const CORE_RULES_DIRECTORY = path.resolve(moduleDirectory, ".", "rules");
 const shownDeprecations: string[] = [];
+const cachedRules: { [fullPath: string]: typeof AbstractRule | null } = {}; // null indicates that the rule was not found
```
andy-hanson: Why not use a `Map`?

nchen63: habit, I guess. updated

PR #2137, opened 2017-01-26
Show time taken for each rule test if > 100ms

nchen63:
Some rules are running very slowly, this PR adds timing to the rules so we're aware (though there might not be much we can do about it)

|rule|time
|-|-|
|await-promise| 769 ms
|completed-docs/functions| 603 ms
|no-boolean-literal-compare| 492 ms
|no-floating-promises/promises| 509 ms
|no-for-in-array| 502 ms
|no-inferred-empty-object-type| 497 ms
|no-unbound-method| 505 ms
|no-unnecessary-qualifier| 793 ms
|no-unsafe-any| 508 ms
|no-unused-variable/type-checked| 505 ms
|no-void-expression| 570 ms

### DIFF
--- a/src/test.ts
+++ b/src/test.ts
@@ -42,6 +42,7 @@ export interface TestResult {
             markupFromMarkup: string;
         },
     };
+    timeMs?: number;
 }
 
 export function runTests(pattern: string, rulesDirectory?: string | string[]): TestResult[] {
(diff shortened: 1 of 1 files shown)


src/test.ts
```diff
@@ -74,6 +75,7 @@ export function runTest(testDirectory: string, rulesDirectory?: string | string[
     const results: TestResult = { directory: testDirectory, results: {} };
 
     for (const fileToLint of filesToLint) {
+        const startTime = new Date().getTime();
```
ajafff: Since we are running in a node environment, what about using `process.hrtime()`?

ajafff:
I don't see how these numbers are meaningful at all.

First of all these are only tests. The time depends on how involved your tests are.

All rules you listed above seem to be using the type checker. The numbers you are measuring are mostly the overhead of typescript creating a program, locating and parsing `lib.d.ts` and the creation of a type checker.

That could lead implementers to the conclusion that their rule is slow or needs to be tuned, just because it requires type checking

nchen63:
@ajafff You're right. Will look into other ways of measuring perf of a rule

PR #2142, opened 2017-01-27
no-floating-promises: Fix metadata

andy-hanson:
This rule uses the type checker.
Re: https://github.com/palantir/tslint/pull/1972#issuecomment-275384846


nchen63:
@andy-hanson thanks

Issue #2128, opened 2017-01-26
`no-unbound-method` errors even for non-problematic references


ethanresnick:
### Bug Report

- __TSLint version__: `4.4.1`
- __TypeScript version__: `2.1.5`
- __Running TSLint via__: cli

#### TypeScript code being linted

```ts
// Check if an optional method exists on obj (or its prototype chain)
function a(obj) {
  return obj && obj.method !== undefined;
}
```

with `tslint.json` configuration:

```json
{"no-unbound-method": true}
```

#### Actual behavior
no-unbound-method gives an error: `Avoid referencing unbound methods which may cause unintentional scoping of 'this'`.

#### Expected behavior

Storing a reference to the unbound method should give an error, but simply referencing it without capturing that reference should not.

rhysd:
I think `.apply` and `.call` are also exceptional case for `no-unbound-method` rule because they binds value to `this`.

```typescript
const a = [0];
Array.prototype.push.apply(a, [1]); // This line causes an error
console.log(a);
```

nchen63:
cc: @andy-hanson

nchen63:
fixed

PR #2143, opened 2017-01-27
Add more exceptions to 'no-unbound-method'

andy-hanson:
#### PR checklist

- [X] Addresses an existing issue:  #2128
  - [X] Includes tests
- [ ] Documentation update

* Allow property access.
* Allow binary expression (except `||`).


IllusionMH:
Interesting case with `forEach`. It and other array methods that take `this` reference as last parameter  provide similar functionality.

However as I can see from typings in `lib.d.ts` it is not possible to distinguish such methods from other :(

nchen63:
@andy-hanson thanks!

PR #2144, opened 2017-01-27
object-literal-key-quotes: No need to quote a float if its .toString() is the same.

andy-hanson:
#### PR checklist

  - [X] Includes tests
- [ ] Documentation update

A property `{ "1.23": null }` does not need quotes since `{ 1.23: null }` is the same object. But the values `{ ".123": null }` and `{ .123: null }` differ. Try it out: `node -p '({ 1.23: null })'`.


nchen63:
@andy-hanson thanks!

PR #2145, opened 2017-01-27
Add no-string-jsx rule

mohsen1:
#### PR checklist

  - [x] Includes tests
- [x] Documentation update

Adds no-string-jsx rule for projects that use translation functions to enforce use of those functions

### DIFF
--- a/src/rules/noStringJsx.ts
+++ b/src/rules/noStringJsx.ts
@@ -1,0 +1,56 @@
+/**
+ * @license
+ * Copyright 2017 Palantir Technologies, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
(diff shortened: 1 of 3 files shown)


src/rules/noStringJsx.ts
```diff
@@ -0,0 +1,56 @@
...
+    public visitJsxElement(node: ts.JsxElement) {
+        for (const child of node.children) {
+            if (child.kind === ts.SyntaxKind.JsxText) {
+                this.addFailure(this.createFailure(node.getStart(), node.getWidth(), Rule.ERROR_MESSAGE));
+            }
+        }
+    }
+
+    public visitJsxExpression(node: ts.JsxExpression) {
+        if (node.expression && node.expression.kind === ts.SyntaxKind.StringLiteral) {
```
IllusionMH: Shouldn't it also check for string concatenation and template literals?

src/rules/noStringJsx.ts
```diff
@@ -0,0 +1,56 @@
...
+
+    public apply(sourceFile: ts.SourceFile): Lint.RuleFailure[] {
+        return this.applyWithWalker(new NoStringJSXWalker(sourceFile, this.getOptions()));
+    }
+}
+
+class NoStringJSXWalker extends Lint.RuleWalker {
+    public visitJsxElement(node: ts.JsxElement) {
+        for (const child of node.children) {
+            if (child.kind === ts.SyntaxKind.JsxText) {
```
IllusionMH: May be it is worth to add `visitJsxText` method to `syntaxWalker`? ([working example](https://github.com/palantir/tslint/pull/2009/files#diff-956e8e0193a5fd2b0905b8cb88c46466R209) from closed PR)
Might be useful for next rules too.

mohsen1:
Moved to https://github.com/palantir/tslint-react/pull/65

Any ideas about the failure there?

Issue #2125, opened 2017-01-26
Ability to not overwrite source file when using `--fix`


ianks:
Currently, running `--fix` replaces the file in place. In certain cases, this causes problem with integrating with other editors, etc. For example, https://github.com/sbdchd/neoformat likes to have a copy of the changed file so it can perform diffs on the current buffer. Would you consider adding an option which allow `--fix` to emit the changes to stdout optionally?

nchen63:
I'm not sure if this would belong in the this repo, but it would be pretty simple to write a custom formatter to write out the fixes in whatever format you want. The text replacements are available in the API even if --fix isn't used.

See https://palantir.github.io/tslint/develop/custom-formatters/

Each `RuleFailure` contains fix.innerReplacements which contains the offset and replacement texts. You just have to take that data and output a string in whatever format works for you.

ianks:
@nchen63 how would this solve the problem of files being overwritten in place?

nchen63:
don't call with --fix so nothing is overwritten
formatter outputs fix info

ianks:
That sounds reasonable; however, I do not see a public API for getting the variables I would need to apply these changes. 

https://github.com/palantir/tslint/blob/master/src/linter.ts#L101

ianks:
Thanks for the info @nchen63!

I created an MVP here: https://github.com/ianks/tslint-fix-formatter

It might be nice to include this in core, since I imagine it would be useful to many users (VIM users, at least). I will leave that up to the team though.

Cheers!

adidahiya:
@ianks sure, we'd accept a PR for a fix formatter.

ianks:
Continuing off of #2149. Making a copy of the file in a temp for is actually what neoformat attempts to do, however the imports , etc are broken at that point. Overall, that mechanism is a bit too error-prone and complicated to use in most cases.

JoshuaKGoldberg:
Seems like this is a subset of #1760. Closing for housekeeping, but if I've misinterpreted and this is actually a separate issue, please do comment here!

PR #2149, opened 2017-01-28
(feature): FixFormatter: new formatter for printing fixed files

ianks:
#### PR checklist
  - [x] Includes tests

This PR adds a FixFormatter which take a single file, and prints the fixed contents of said file. The goal is to be able to integrate with editor plugins such as [neoformat](https://github.com/sbdchd/neoformat), which rely on the  the formatted text being printed to stdout.

Currently, this formatter will only work when passing in a single file. I think from a practicality standpoint, this makes a lot of sense as the output will always be merged into a single stream (stdout, some file, etc), where the end user probably will not want to merge all of the files together.

Happy to any suggestions and improve this in any way you see fit!

-Ian


palantirtech:
Thanks for your interest in [palantir/tslint](https://github.com/palantir/tslint), @ianks! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

ianks:
@nchen63 why was this closed?

nchen63:
The approach of using a formatter just isn't feasible given my reasons above.

ianks:
Ok, so what would be the best way to get this feature implemented? Should we continue the discussion in https://github.com/palantir/tslint/issues/2125?

nchen63:
sure, let me know if my idea of copying the file doesn't work

PR #2155, opened 2017-01-30
Add `no-import-side-effect` rule

### LINKED ISSUE #2116: Warn for import statements with side-effects
Lets suppose we have the modules `P`, `C1` and `C2`. `P` loads `C1` and `C2` lazily. Module `C1` has the following imports:

```ts
import { Observable } from 'rxjs/Observable';
import 'rxjs/add/observable/of';
```

And `C2` has:

```ts
import { Observable } from 'rxjs/Observable';
import 'rxjs/add/operator/mapTo';
```

Now lets suppose in `tsconfig.json`'s files property we have listed the entry files for `P`, `C1` and `C2`. Since the three modules won't be bundled together, we can get a runtime error in the following cases:

- User navigates to `P`, which loads `C1` which uses `mapTo` (referenced in `C2`).
- User navigates to `P`, which loads `C2` which uses `Observable.of` (referenced in `C1`).

Behind the hood, rxjs dynamically adds the `mapTo` method to the prototype of `Observable`, and `of` to the constructor function.

Although it is possible to apply type checking as a way for verification of correctness in these cases, it's hard to combine the proper set of files to be compiled together in a real-life project (i.e. in this case `P` together with `C1` and `P` together with `C2`).

In this case, it might be a good idea to provide warning to the user in case of side-effect import statements (original source https://github.com/angular/angular-cli/issues/3904).

We can have a rule which if finds:

```ts
import 'path-to-module';
```

Produces a warning `import with explicit side-effect`, and has a description: `Avoid import statements with side-effect`.

Do you think such rule will be a good fit for the set of core rules of tslint? I'd love to help with a PR.

// cc @hansl @filipesilva

mgechev:
Fix #2116

#### PR checklist

- [x] Addresses an existing issue: #2116, related to https://github.com/angular/angular-cli/issues/3904
  - [x] Includes tests
- [x] Documentation update

Introduced a rule which warns when used imports with side-effects. For more details here https://github.com/palantir/tslint/issues/2116.

Have I added the required documentation.

### DIFF
--- a/src/rules/noImportSideEffectRule.ts
+++ b/src/rules/noImportSideEffectRule.ts
@@ -1,0 +1,78 @@
+/**
+ * @license
+ * Copyright 2017 Palantir Technologies, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import * as ts from "typescript";
+
+import * as Lint from "tslint";
+
+const OPTION_IGNORE_MODULE = "ignore-module";
+
+export class Rule extends Lint.Rules.AbstractRule {
+    public static metadata: Lint.IRuleMetadata = {
+        description: "Avoid import statements with side-effect.",
+        optionExamples: ["true", `[true, { "${OPTION_IGNORE_MODULE}": "(\\.html|\\.css)$" }]`],
+        options: {
+            items: {
+                properties: {
+                    "ignore-module": {
+                        type: "string",
+                    },
+                },
+                type: "object",
+            },
+            maxLength: 1,
+            minLength: 0,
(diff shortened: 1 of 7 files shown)


palantirtech:
Thanks for your interest in [palantir/tslint](https://github.com/palantir/tslint), @mgechev! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

test/rules/no-import-side-effect/ignore-pattern/tslint.json
```diff
@@ -0,0 +1,21 @@
+{
```
nchen63: file path should be `ignore-module` not `ignore-pattern`

src/rules/noImportSideEffectRule.ts
```diff
@@ -0,0 +1,78 @@
+/**
+ * @license
+ * Copyright 2015 Palantir Technologies, Inc.
```
nchen63: 2017

mgechev:
@nchen63 thanks for the review! I addressed your comments.

nchen63:
@mgechev thanks!

johnpapa:
I think this topic begs the question of where and when do we import these? We could import each operator and function for rxjs in every place it is used. Upside is you get exact behavior, downside is you have tons and tons of import statements.

Or we could import at the root of a module that uses the features. Upside is reduced imports but still have the behavior we want within a module.

Or we could import at a root module only. This is a bit more risky if a sub module could also be used elsewhere, as its imports wouldnt travel.

cc @wardbell

PR #2164, opened 2017-02-01
Improve error messages for no-any and object-literal-shorthand.

mprobst:
Users often use `any` when they could equally well use `{}`, because
they are not aware of the empty type.

Users also often don't know what object shorthand is - giving them a
hint as to what syntax is expected makes it easier to fix the lint
warning.

#### PR checklist

  - [x] Includes tests

Improve error messages for no-any and object-literal-shorthand.

### DIFF
--- a/src/rules/noAnyRule.ts
+++ b/src/rules/noAnyRule.ts
@@ -34,7 +34,9 @@ export class Rule extends Lint.Rules.AbstractRule {
     };
     /* tslint:enable:object-literal-sort-keys */
 
-    public static FAILURE_STRING = "Type declaration of 'any' is forbidden (use the empty type '{}'?)";
+    public static FAILURE_STRING = "Type declaration of 'any' loses type-safety. " +
+        "Consider replacing it with a more precise type, the empty type ('{}'), " +
+        "or suppress this occurrence.";
 
     public apply(sourceFile: ts.SourceFile): Lint.RuleFailure[] {
         return this.applyWithWalker(new NoAnyWalker(sourceFile, this.getOptions()));
--- a/src/rules/noAnyRule.ts
+++ b/src/rules/noAnyRule.ts
@@ -34,7 +34,7 @@ export class Rule extends Lint.Rules.AbstractRule {
     };
     /* tslint:enable:object-literal-sort-keys */
 
-    public static FAILURE_STRING = "Type declaration of 'any' is forbidden";
+    public static FAILURE_STRING = "Type declaration of 'any' is forbidden (use the empty type '{}'?)";
 
     public apply(sourceFile: ts.SourceFile): Lint.RuleFailure[] {
         return this.applyWithWalker(new NoAnyWalker(sourceFile, this.getOptions()));
--- a/src/rules/objectLiteralShorthandRule.ts
+++ b/src/rules/objectLiteralShorthandRule.ts
@@ -32,8 +32,8 @@ export class Rule extends Lint.Rules.AbstractRule {
     };
     /* tslint:enable:object-literal-sort-keys */
 
-    public static LONGHAND_PROPERTY = "Expected property shorthand in object literal ('{foo, bar}').";
-    public static LONGHAND_METHOD = "Expected method shorthand in object literal ('{foo() {...}}').";
+    public static LONGHAND_PROPERTY = "Expected property shorthand in object literal ";
+    public static LONGHAND_METHOD = "Expected method shorthand in object literal ";
 
     public apply(sourceFile: ts.SourceFile): Lint.RuleFailure[] {
         const objectLiteralShorthandWalker = new ObjectLiteralShorthandWalker(sourceFile, this.getOptions());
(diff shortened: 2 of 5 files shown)


mprobst:
Lots of lints warnings in tslint itself (though `yarn verify` passes locally for me) - probably not worth fixing in this PR? The tests affected by this change pass locally.

nchen63:
can you rebase?

mprobst:
@nchen63 done.

mprobst:
@nchen63 not sure how to fix the build failure about requiring 4x parallelism. I've made that setting in the UI as suggested by the error, but the build still seems to fail?

IllusionMH:
@mprobst have you tried "Try build" on `shorthand-err` branch after you set parallelism to 4? https://circleci.com/gh/mprobst/tslint/edit#parallel-builds
As I remember it started to work after few attempts for me

mprobst:
@IllusionMH ah, it appears if you "Rebuild" a failed build, it'll use the same parallelism settings. Which is not that helpful if the build failed because of too little parallelism...

In any case, seems fixed now.

src/rules/objectLiteralShorthandRule.ts
```diff
@@ -31,8 +31,8 @@ export class Rule extends Lint.Rules.AbstractRule {
     };
     /* tslint:enable:object-literal-sort-keys */
 
-    public static LONGHAND_PROPERTY = "Expected property shorthand in object literal.";
-    public static LONGHAND_METHOD = "Expected method shorthand in object literal.";
+    public static LONGHAND_PROPERTY = "Expected property shorthand in object literal ('{foo, bar}').";
```
alexeagle: I doubt all users will understand this, because it's not obvious that foo is a property name (it also looks syntactically like destructuring). Easy fix is print `{prop1, prop2}` instead so the names are self-documenting.
Ideally we should point to the https://palantir.github.io/tslint/rules/object-literal-shorthand/ page, and make that have more information explaining the syntax for newbies (they may not be aware of the shorthand syntax at all)

src/rules/objectLiteralShorthandRule.ts
```diff
@@ -31,8 +31,8 @@ export class Rule extends Lint.Rules.AbstractRule {
...
+    public static LONGHAND_METHOD = "Expected method shorthand in object literal ('{foo() {...}}').";
```
alexeagle: this would be a great check to add a suggested fix

src/rules/noAnyRule.ts
```diff
@@ -33,7 +33,7 @@ export class Rule extends Lint.Rules.AbstractRule {
     };
     /* tslint:enable:object-literal-sort-keys */
 
-    public static FAILURE_STRING = "Type declaration of 'any' is forbidden";
+    public static FAILURE_STRING = "Type declaration of 'any' is forbidden (use the empty type '{}'?)";
```
alexeagle: I think users will still stumble here. Linking to the doc would be best (see other comment)

Wordsmith attempt:
`Type declaration of 'any' loses type-safety. Consider replacing it with a more precise type, the empty type ('{}'), or suppress this occurrence`

Separately, it would be nice if the end of a tslint failure report pointed you to the docs for suppression https://palantir.github.io/tslint/usage/rule-flags/ - nearly every user I've coached was unaware of those.

ajafff: > Separately, it would be nice if the end of a tslint failure report pointed you to the docs for suppression https://palantir.github.io/tslint/usage/rule-flags/ - nearly every user I've coached was unaware of those.

I don't think bloating the failure message even more adds any more value to it. Most of the time you see it as a tooltip in your editor and cannot even click the link.

Some (most?) IDEs with tslint integration show you a "quickfix" option at rule failures that allows you to automatically disable that rule for that line. At least that's what vscode, IntelliJ and friends do.

alexeagle: We have a lot of novice users, who are mailing out their PR for review, and their command-line (or even an email) reports that they have to stop and go back to fix something. For these users, there is plenty of space for more text, and this is the right UX to also explain to a novice user what they did wrong and how to fix it.

IMO the core problem here is we have one string field which must fit into a tooltip, in an IDE environment where there is other help, and also be used in the command-line interface where the context is missing.

Maybe on the command line we could also print the Description field? Though I know most descriptions were not written with this in mind. Maybe add a new field "LONG_FAILURE_STRING" or similar? @JKillian ?

src/rules/objectLiteralShorthandRule.ts
mprobst: Done.

src/rules/objectLiteralShorthandRule.ts
mprobst: I have a different PR for that, though only for the shorthand property - the functions are tricky to find in the AST.

src/rules/noAnyRule.ts
mprobst: I went with yours @alexeagle, @ajafff @nchen63 let me know what you think.

nchen63: I think the updated text has the right amount of verbosity

nchen63:
@mprobst thanks!

Issue #968, opened 2016-02-12
Eror locations for typedef rule incorrect


JKillian:
Error locations should arguably be beneath the variable that needs a typedef instead of where the typedef could possibly start.

```
var NoTypesFn = function (
             ~             [expected variable-declaration: 'NoTypesFn' to have a typedef]
    a,
     ~ [expected parameter: 'a' to have a typedef]
    b) {
     ~   [expected call-signature to have a typedef]
     ~   [expected parameter: 'b' to have a typedef]
    var c = a + b,
         ~         [expected variable-declaration: 'c' to have a typedef]
        d = a - b;
         ~         [expected variable-declaration: 'd' to have a typedef]
```

adidahiya:
Does that code block show the current or the desired behavior? Agreed that it should highlight the variable itself to avoid confusion.

JKillian:
Yep, the code block shows current behavior.

PR #2166, opened 2017-02-01
Fix error location for typedef rule (#968)

fhemery:
#### PR checklist

- [x] Addresses an existing issue: #968
- [x] Includes tests
- [ ] Documentation update - N/A

I modified the typedef rule so that error location is adjusted, and updated the tests accordingly.

I have chosen to modify as well the behaviour for the arrow operator and the function calls, in an attempt to be homogeneous. I think you should have a close look at this behaviour change in particular. Feel free to reject the PR if behaviour is not correct, I'll patch it.
Please note that this is my first contribution to open source, so any comment is welcome !

### DIFF
--- a/src/rules/typedefRule.ts
+++ b/src/rules/typedefRule.ts
@@ -87,7 +87,7 @@ class TypedefWalker extends Lint.RuleWalker {
             && node.parent.kind !== ts.SyntaxKind.CallExpression
             && !isTypedPropertyDeclaration(node.parent)) {
 
-            this.checkTypeAnnotation("arrow-call-signature", location, node.type, node.name);
+            this.checkTypeAnnotation("arrow-call-signature", node.getStart(), node.type, node.name);
         }
         super.visitArrowFunction(node);
     }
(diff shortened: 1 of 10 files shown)


palantirtech:
Thanks for your interest in [palantir/tslint](https://github.com/palantir/tslint), @fhemery! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

adidahiya:
@fhemery thanks for taking the time to make this contribution!

I don't think we should accept this change, however. The original error location for this rule was intentional because it pointed to the place where you add a typedef (e.g. the location of the `:` in a `: boolean`) typedef. Why do you want to change it?

fhemery:
Your comment on the 12th of February says that the variable itself should be highlighted. So there are three options from my point of view :
* I misinterpreted it totally, and I should highlight the whole variable instead of the beginning of the variable only
* I did it correctly for the variables, but shouldn't have made the change for the function calls, in which case I can amend the commit to remove the function part
* The behaviour is no longer needed, and I should try another bug (any suggestion ?)

Could you let me know which option is correct ? 
Thanks in advance,
Frederic

nchen63:
@fhemery I don't think you misinterpreted the original issue. 

If you updated the fix to include the entire name of the variable/function/destructuring object we'd accept the PR. Also, could you change the error to say "type definition" instead of "typedef"?

nchen63:
@fhemery also, please run `npm run verify` to make sure your project lints and runs tests

adidahiya:
sorry, this PR is stale. please open a new one if you wish to pursue this change.

Issue #623, opened 2015-08-30
Request: rule for requiring JSDoc comments for all functions


maxmantz:
Having a rule to enforce using JSDoc style comments for every function would help a lot with code documentation. This rule could be further refined by only needing to provide comments for public functions, only commenting once per occurrence - for example an exception to not comment a class implementation when it was already commented in an interface, etc.

adidahiya:
Sure, that sounds like a reasonable request. Would you want to require type annotations on parameters in those jsdoc comments? Or just look for a /** block comment right before the function declaration? I'm strongly inclined towards the latter since TS already has type annotations.

maxmantz:
I don't think that type annotations on parameters in those comments are necessary, as you've pointed out. However needing to describe the parameters would be another useful requirement for the rule. Another useful rule could be that double-slash quotes are required on (public) properties of a class. Here is an example of what I have in mind:

    interface IDemoClass {
    // I am an optional requirement
    property1: string;

    /**
     * This comment is required
     * @param {string} input This description is an optional requirement
     */
    doSomething(input: string): void;
    }

    class DemoClass implements IDemoClass {
    
    // I am also optional requirement, since the interface has already provided a comment
    property1: string;

    // I am also optional, since the property is private
    private property2: number;

    /**
     * A constructor comment is required
     * @param {string} prop1 Describing the parameters
     * @param {string} prop2 is a useful (optional) requirement
     */
    constructor(prop1: string, prop2: number) {
        this.property1 = prop1;
        this.property2 = prop2;
    }

    // no comment required, since it's been commented in the interface
    doSomething(input: string): void {
        // do something...
    }

    /**
     * This comment is required
     * @returns this is optional
     */
    doSomethingPublicly(): number {
        return 123;
    }

    /**
     * An optional requirement for private functions
     * @param {string} input optional as well
     * @returns optional requirement
     */
    private doSomethingPrivately(input: string): string {
        return 'foo';
    }
    }

The types of the params in my JSDoc comments have been automagically generated by my IDE, but they should be optional as well.

nchen63:
fixed by #2095

PR #2184, opened 2017-02-07
[New Rule] Adds support for enforcing JSDoc

orta:
#### PR checklist

- [x] Addresses an existing issue: #623
- [x] New feature
  - [x] Includes tests
- [ ] Documentation update

Hey there, I had been wanting to add a [feature](https://github.com/eslint/eslint/pull/7867) to ESLint for a while, but we moved to TypeScript, and so I actually needed it in TSLint also. 

This is close to an exact port of the "require-jsdoc" port from ESLint, but with less rigid defaults. It will default to only raising errors to `export`ed classes/functions. You can make it apply to all functions though, too.

- I've only done JavaScript for ~6 months TS for 1 month, there are things I've probably not taken into account.
- Should this take into account TS' private / public functions?
- Not sure what I have to do re:docs

--

Sidenotes:

* I updated the settings for VS Code - the JSON shape has changed, and it would show the JS files instead of just TS files in search
* I added a quick feature to the ruleTestRunner so you could run a subset of tests with ` node ./build/test/ruleTestRunner.js "test/rules/require-jsdoc/*/tslint.json"` - 30 seconds  was too long.

### DIFF
--- a/src/rules/requireJsdocRule.ts
+++ b/src/rules/requireJsdocRule.ts
@@ -1,0 +1,135 @@
+/**
+ * @license
+ * Copyright 2017 Palantir Technologies, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import * as ts from "typescript";
+import * as Lint from "../index";
+
+export class Rule extends Lint.Rules.TypedRule {
+    /* tslint:disable:object-literal-sort-keys */
+    public static metadata: Lint.IRuleMetadata = {
+        ruleName: "require-jsdoc",
+        description: "Require JSDoc comments.",
+        hasFix: false,
+        optionsDescription: Lint.Utils.dedent`
+            Ensure your project is documented. Defaults to only show errors 
+            on undocumented classes or functions which are exported.
+
+            You can opt out of checking for a specific set of declarations in
+            the options.
+            `,
+        options: {
+            type: "object",
+            properties: {
+                exportedOnly: {
+                    type: "boolean",
+                },
+                functionDeclaration: {
+                    type: "boolean",
+                },
+                classDeclaration: {
+                    type: "boolean",
+                },
+                methodDefinition: {
+                    type: "boolean",
+                },
+            },
+            additionalProperties: false,
+        },
+        optionExamples: ['[true, {"exportedOnly": false, "classDeclaration": false}]',
+                         '[true, {"exportedOnly": false, "classDeclaration": false, functionDeclaration: false, methodDefinition: false}]'],
+
+        type: "style",
+        typescriptOnly: false,
+    };
+    /* tslint:enable:object-literal-sort-keys */
+
+    public static FAILURE_STRING = "Missing JSDoc comment.";
+
+    public applyWithProgram(sourceFile: ts.SourceFile, langSvc: ts.LanguageService): Lint.RuleFailure[] {
+        const noForInArrayWalker = new RequireJSDocWalker(sourceFile, this.getOptions(), langSvc.getProgram());
+        return this.applyWithWalker(noForInArrayWalker);
+    }
+}
+
+class RequireJSDocWalker extends Lint.ProgramAwareRuleWalker {
+    public visitClassDeclaration(node: ts.ClassDeclaration) {
+        const settings = this.getRequireJSDocOptions();
+        if (settings.classDeclaration) {
+            this.checkJSDoc(node);
+        }
+        super.visitClassDeclaration(node);
+    }
+
+    public visitFunctionExpression(node: ts.FunctionExpression) {
+        const settings = this.getRequireJSDocOptions();
+        if (settings.functionDeclaration) {
+            this.checkJSDoc(node);
+        }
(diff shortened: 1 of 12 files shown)


palantirtech:
Thanks for your interest in [palantir/tslint](https://github.com/palantir/tslint), @orta! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

src/rules/requireJsdocRule.ts
```diff
@@ -0,0 +1,135 @@
...
+    private getRequireJSDocOptions() {
+        const DEFAULT_OPTIONS = {
+            classDeclaration: true,
+            exportedOnly: true,
+            functionDeclaration: true,
+            methodDefinition: true,
+        };
+
+        const ruleOptions = this.getOptions();
+        return Object.assign(DEFAULT_OPTIONS, ruleOptions && ruleOptions[0]);
```
ajafff: you could do this once in the constructor and assign the result to a property

src/rules/requireJsdocRule.ts
```diff
@@ -0,0 +1,135 @@
...
+                if (settings.exportedOnly && !this.isExportedNode(node)) {
+                    return;
+                }
+                this.addFailureAtNode(node.name, Rule.FAILURE_STRING);
+            }
+        }
+    }
+
+    private isExportedNode(node: ts.Node): boolean {
+        if (Lint.hasModifier(node.modifiers, ts.SyntaxKind.ExportKeyword)) { return true; }
```
ajafff: that doesn't seem to be the correct way to check for exports.
consider the following code:
```ts
export function exportedFunction() { // <-- JSDoc required
    function nestedFunction() {} // <-- JSDoc required
}
```
maybe a better approach would be
```ts
// if you use a discriminated union as parameter this switch will be exhaustive
switch(node.kind) {
    case ts.SyntaxKind.FunctionDeclaration:
    case ts.SyntaxKind.ClassDeclaration:
        return Lint.hasModifier(node.modifiers, ts.SyntaxKind.ExportKeyword);
    case ts.SyntaxKind.MethodDeclaration:
        return Lint.hasModifier(node.parent!.modifiers, ts.SyntaxKind.ExportKeyword);
    case ts.SyntaxKind.FunctionExpression:
        return node.parent!.kind === ts.SyntaxKind.VariableDeclaration && Lint.hasModifier(node.parent!.parent!.parent!.modifiers, ts.SyntaxKind.ExportKeyword);
        // .parent!.parent! is the VariableDeclarationList, .parent!.parent!.parent! is the VariableStatement, which can be exported
}
```

src/rules/requireJsdocRule.ts
```diff
@@ -0,0 +1,135 @@
...
+    private checkJSDoc(node: ts.FunctionDeclaration | ts.FunctionExpression | ts.ClassDeclaration | ts.MethodDeclaration) {
+        if (node.name) {
+            const checker = this.getTypeChecker();
+            const symbol = checker.getSymbolAtLocation(node.name);
+            const hasDocs = symbol.getDocumentationComment().length > 0;
+
+            if (!hasDocs ) {
+                const settings = this.getRequireJSDocOptions();
+                // console.log(settings);
+                if (settings.exportedOnly && !this.isExportedNode(node)) {
```
ajafff: you should check if you want to enforce JSDoc on this node *before* you use the typechecker

test/ruleTestRunner.ts
```diff
@@ -25,7 +25,12 @@ console.log();
 console.log(colors.underline("Testing Lint Rules:"));
 /* tslint:enable:no-console */
 
-const testDirectories = glob.sync("test/rules/**/tslint.json").map(path.dirname);
+// Support running a subset of lint rules:
```
ajafff: is this change necessary?

you can run rule tests by simply using `./bin/tslint --test some/glob*pattern`

test/rules/require-jsdoc/exports-only-classes-only/test.ts.lint
```diff
@@ -0,0 +1,17 @@
+/** I am documented */
+export class Documented {
+      
+}
+
+export class Undocumented {
+             ~~~~~~~~~~~~   [Missing JSDoc comment.]
```
ajafff: please add some tests for nested functions

src/rules/requireJsdocRule.ts
```diff
@@ -0,0 +1,135 @@
...
+    public static metadata: Lint.IRuleMetadata = {
+        ruleName: "require-jsdoc",
+        description: "Require JSDoc comments.",
+        hasFix: false,
+        optionsDescription: Lint.Utils.dedent`
+            Ensure your project is documented. Defaults to only show errors 
+            on undocumented classes or functions which are exported.
+
+            You can opt out of checking for a specific set of declarations in
+            the options.
```
ajafff: this should describe how to use the `options` object

src/rules/requireJsdocRule.ts
```diff
@@ -0,0 +1,135 @@
...
+                methodDefinition: {
+                    type: "boolean",
+                },
+            },
+            additionalProperties: false,
+        },
+        optionExamples: ['[true, {"exportedOnly": false, "classDeclaration": false}]',
+                         '[true, {"exportedOnly": false, "classDeclaration": false, functionDeclaration: false, methodDefinition: false}]'],
+
+        type: "style",
```
ajafff: requiresTypeInfo: true

test/ruleTestRunner.ts
orta: Ah yeah, I did not know about this feature! useful 💯

nchen63:
I think this duplicates behavior from `completed-docs` https://github.com/palantir/tslint/pull/2095

orta:
Oh yeah -  @nchen63 it does, and that's much more comprehensive 👍 

@ajafff - thanks for the review ( it was useful, and I got pretty far with your changes, but not much point continuing ) 

I'm going to close this. :1st_place_medal:

Issue #2091, opened 2017-01-21
Rules should be able to opt into --type-check


JoshuaKGoldberg:
Spinning discussion out of #2073: it would be useful to have rules that run in the default mode but can also benefit from type checking.

For example, `no-string-throw` isn't able to use the type checker to detect non-`Error` objects that are thrown. The discussion in #1878 indicated a preference for making it available by default (since that catches most usages). 

Right now there are two rules types: `TypedRule` and `AbstractRule` (or something like that). I propose one of two options:
1. A third type, `OptionallyTypedRule` be added. It should implement both `apply` and `applyWithProgram`.
2. `TypedRule` be allowed to optionally implement `apply` as well as `applyWithProgram`.

I have a slight preference for 1. Later on, it could be reasonable to deprecate all both `AbstractRule` and `TypedRule` and rename `OptionallyTypedRule` to just `Rule`.

ajafff:
Isn't that already possible?
A `TypedRule` simply throws an exception in `apply`. If you extend `TypedRule` and implement both `apply` and `applyWithProgram`, the former will be called if you run tslint without `--program` and the latter will be called if you run tslint with that flag.

JoshuaKGoldberg:
Super interesting, I didn't even think to look for that. Thanks @ajafff!

Why bother keeping both a `TypedRule` and its parent `AbstractRule`? Feels like it would make sense to just have one class, `Rule`, that can have `apply` and/or `applyWithProgram`?

ajafff:
Both of them are abstract to guide the developer which method hast to be implemented. I think it makes sense to keep both of them. 
Maybe it would make sense to also add a `OptionallyTypedRule` that has both methods `apply` and `applyWithProgram` as abstract members. That would make the intent of the rule clearer.

PR #2188, opened 2017-02-07
Don't die when a rule requiring type checking but doesn't have it

### LINKED ISSUE #2169: Rules that require --type-check should not throw as breaks extends
### Bug Report

- __TSLint version__: `4.4.2`
- __TypeScript version__: `2.1.5`
- __Running TSLint via__: CLI / VS Code

```ts
(include if relevant)
```

```
{
   "rulesDirectory": ["node_modules/tslint-microsoft-contrib"],
   "extends": "tslint-microsoft-contrib/recommended_ruleset.js",
   "rules": {
      "no-for-in-array": false
   }
}
```

#### Actual behavior

The throw is annoying anyway as breaks TSLint with little visibility of why in VSCode extension

But more importantly it breaks "extends". TS throws an exception for "no-for-in-array" defined in base recommend_ruleset as --type-check is not enabled even though we tried to disable that check in the derived config.

#### Expected behavior

Just a warning

nchen63:
fixes #2169

### DIFF
--- a/src/error.ts
+++ b/src/error.ts
@@ -15,6 +15,8 @@
  * limitations under the License.
  */
 
+const shownWarnings = new Set<string>();
+
 /**
  * Generic error typing for EcmaScript errors
  * Define `Error` here to avoid using `Error` from @types/node.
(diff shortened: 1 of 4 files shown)


ajafff:
@nchen63 this is not the best solution for the problem. Before this change it was possible to write a Rule that would work both with and without a TypeChecker, see https://github.com/palantir/tslint/issues/2091#issuecomment-274450190. This PR broke that behaviour.

Maybe it would have been better to just wrap the call to `rule.applyWithProgram()` in `try ... catch` and print exceptions as warning.
When going that route it would also make sense to do the same for `rule.apply()`. This way linting still works even if a rule crashes.

Issue #2172, opened 2017-02-03
strict-boolean-expressions silently requires strictNullChecks TS option


usuyama:
### Bug Report

- __TSLint version__: `4.4.2`

#### Description

Some options (`allow-null-union`, `allow-undefined-union`, `allow-mix`) of [strict-boolean-expressions ](https://palantir.github.io/tslint/rules/strict-boolean-expressions/) assume strictNullChecks TypeScript compiler option, but it is not documented.

The related part of code is [here ](https://github.com/palantir/tslint/blob/master/src/rules/strictBooleanExpressionsRule.ts#L279). Only string, number, Boolean, and enum are handled as NOT always true or false. For example, object type is treated as always truthy, but it's not true without strictNullChecks . This code path is used even without the mentioned arguments, but somehow [the default test ](https://github.com/palantir/tslint/blob/master/test/rules/strict-boolean-expressions/default/test.ts.lint)passes, too.

The tests specify strictNullChecks option (Ex. https://github.com/palantir/tslint/tree/master/test/rules/strict-boolean-expressions/allow-null-union )

#### Actual behavior

strictNullChecks  option is assumed for these arguments `allow-null-union`, `allow-undefined-union`, `allow-mix`, but not documented.

#### Expected behavior

Should be documented. Or, nice if we can lift the assumption.

nchen63:
cc: @andy-hanson

nchen63:
it looks like you can check compiler options in program.getCompilationSettings().strictNullChecks

andy-hanson:
How should a rule report a configuration error like that?

usuyama:
For the sake of #2200, we need to lift the requirement of strictNullChecks, instead of just throwing error. Please check the discussion in #2200.

PR #2200, opened 2017-02-12
New Rule: no-enum-boolean-expressions

usuyama:
#### PR checklist

- [x] Addresses an existing issue: #1555
  - [x] Includes tests
- [x] Documentation update

A new rule ` no-enum-boolean-expressions`, which disallows the use of Enum in boolean expressions.

I tried to make it an option of [strict-boolean-expressions](https://palantir.github.io/tslint/rules/strict-boolean-expressions/), but it wasn't straight forward because [strict-boolean-expressions](https://palantir.github.io/tslint/rules/strict-boolean-expressions/) assumes strictNullChecks TypeScript option. For example, Enum types are always truthy with that assumption. So, I made a new rule from strictBooleanExpressionsRule.ts (as of 6dcf3f06e9d79de1437b238dc89d2822964f4b02).

As mentioned above, please note that the diff from strictBooleanExpressionsRule.ts (as of 6dcf3f06e9d79de1437b238dc89d2822964f4b02) is small.

### DIFF
--- a/src/rules/noEnumBooleanExpressionsRule.ts
+++ b/src/rules/noEnumBooleanExpressionsRule.ts
@@ -1,0 +1,158 @@
+/**
+ * @license
+ * Copyright 2016 Palantir Technologies, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import * as ts from "typescript";
+import * as Lint from "../index";
+import * as utils from "../language/utils";
+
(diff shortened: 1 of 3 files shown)


nchen63:
strict-boolean-expressions disallows enums by default. would fixing https://github.com/palantir/tslint/issues/2172 meet the needs of this pr?

usuyama:
@nchen63 In addition to relaxing the requirement of strictNullChecks option in [strict-boolean-expressions](https://palantir.github.io/tslint/rules/strict-boolean-expressions/), we need another option to support the scenario. **By default, strict-boolean-expressions checks everything and that was too much for our project. We want to check only enum in boolean expressions.**

[strict-boolean-expressions](https://palantir.github.io/tslint/rules/strict-boolean-expressions/) has several options like `allow-number`, `allow-string`, `allow-union-undefined`. For checking only enum, perhaps we can add `allow-object`, too, then using all the options (`allow-object`, `allow-number`, `allow-string`, `allow-null-union`, `allow-undefined-union`, `allow-mix`) would be similar to what we want (#1555)

Or, we could add an option like `disallow-only-enum`, which cannot be used with `allow-*` options, but it's not consistent.

Thoughts?

adidahiya:
> For checking only enum, perhaps we can add allow-object, too, then using all the options (allow-object, allow-number, allow-string, allow-null-union, allow-undefined-union, allow-mix) would be similar to what we want

I like this approach for consistency's sake, let's do that

andy-hanson:
Now that #2373 is in, this is no longer needed, right?

PR #2209, opened 2017-02-16
Add `no-public-modifiers` rule

mikechamberlain:
#### PR checklist

  - [ ] Includes tests
- [ ] Documentation update

Added rule to prohibit explicit public modifiers. I prefer a more minimalist approach to code.

(optional)

### DIFF
--- a/src/rules/noPublicModifiersRule.ts
+++ b/src/rules/noPublicModifiersRule.ts
@@ -1,0 +1,116 @@
+/**
+ * @license
+ * Copyright 2013 Palantir Technologies, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import * as ts from "typescript";
+
+import * as Lint from "tslint";
+
+export class Rule extends Lint.Rules.AbstractRule {
+    /* tslint:disable:object-literal-sort-keys */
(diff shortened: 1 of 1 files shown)


palantirtech:
Thanks for your interest in [palantir/tslint](https://github.com/palantir/tslint), @mikechamberlain! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

ajafff:
Can you instead add an option to the existing `member-access` rule to avoid conflicting rules?
The code of both rules seems to be the same anyway.

mikechamberlain:
Hi @ajafff, I'm not sure how this use case could be merged with the existing `member-access` rule. 

Member access is about _requiring_ visibility accessors:

> Requires explicit visibility declarations for class members

But this rule is about _prohibiting_ them (the `public` ones, at least). I can't think of a sane way to add an option into the existing `member-access` rule that would not introduce the possibility of conflict. What do you suggest?

mikechamberlain:
Also, can anyone assist on the failing circleci tests? It's not clear to me from the description what I've done wrong.

ajafff:
CI is failing because of `src/rules/noPublicModifiersRule.ts[41, 6]: Unnecessary semicolon`.
You can run `yarn lint` in your local repository to check this before pushing.


> Member access is about requiring visibility accessors

Well, yes and no. It's about enforcing consistent access modifiers. You're right that right now it requires them everywhere, but that can be changed.

To add this as an option to `member-access` you could add the following at this line: https://github.com/palantir/tslint/blob/master/src/rules/memberAccessRule.ts#L99

```ts
if (this.hasOption('ban-public')) { // option name is for discussion
    if (Lint.hasModifier(node.modifiers, ts.SyntaxKind.PublicKeyword)) {
        this.addFailureAt(...);
    }
    return; // no further checks required
}
```

nchen63:
I agree with @ajafff, this should go in `member-access`. The `ban-public` suggestion sounds good

mikechamberlain:
I agree with everything proposed here. Unfortunately I will have to leave this to someone else as I don't have the inclination to figure it out myself.

adidahiya:
superseded by #2247

Issue #2180, opened 2017-02-06
Use a separate package for typeguards and other utilities


ajafff:
As a followup of https://github.com/palantir/tslint/pull/2093#discussion_r98706381 we want / need to add typeguards to the codebase to keep code readable.

While moving all typeguards to a newly created `typeguard.ts` I realized, that it is really awkward to use `Lint.isStringLiteral` etc. 
tslint core is not the right place to provide typeguard functions for every rule, even custom rules.

The idea was to create a separate project which maintains typeguards and AST utilities for everyone, not just tslint and custom rules packages. This way tslint core could remain just a runtime for rules instead of growing to a bulk of utility functions.

Now there is this repository: https://github.com/ajafff/tsutils
I haven't released a package on npm yet, because I wanted to get your feedback first.
Of course I can add you and others as collaborators so this can be maintained even if I don't have time.

nchen63:
Sounds like a good idea. Maybe the super node for discriminated unions can go there too if https://github.com/Microsoft/TypeScript/issues/13634 isn't implemented

PR #2217, opened 2017-02-17
Use tsutils package for utilities

ajafff:
#### PR checklist

- [x] Closes: #2180
- [x] enhancement

Use the package `tsutils` for typeguards and other utility functions.

> The idea was to create a separate project which maintains typeguards and AST utilities for everyone, not just tslint and custom rules packages. This way tslint core could remain just a runtime for rules instead of growing to a bulk of utility functions.


nchen63:
@ajafff thanks!

Issue #785, opened 2015-11-07
switch-case-fall-through check failed on blocked case clauses


falsandtru:
``` ts
switch (0) {
  case 0: {
    break;
  }
  default:
}
```

JKillian:
Thanks for the report @falsandtru. For a possible workaround, I'd try using this style for now:

```typescript
switch (0) {
  case 0: {
    //whatever code here
  } break;
  default:
}
```

falsandtru:
Thanks for your response. But I cannot solve some cases.

```ts
switch (n) {
  case 0: {
    let m = n;
    // ...
    return n * m;
  }
  default:
}
```

Would you fix this issue?

JKillian:
@falsandtru Yes, this should definitely be fixed.
Also, @Think7 also encountered this problem over in #898.

nsgundy:
Just ran into the same issue, so I am all for having it fixed :)

Thanks for making tslint :+1: 

JSHint used to have the same problem, see https://github.com/jshint/jshint/issues/858.

JKillian:
Wanted to mention that the newest versions of the TS compiler now have a [--noFallthroughCasesInSwitch](https://github.com/Microsoft/TypeScript/wiki/What's-new-in-TypeScript#case-clause-fall-throughs) flag which could lead to this rule being deprecated soon.

falsandtru:
I agree, this rule will be needless.

JKillian:
Although a TSLint rule offers the advantage that you can selectively disable it, whereas as far as I know you can't do that with the compiler flag.

falsandtru:
Then TSLint would be better to change its default(example) config to disable.

0815fox:
Yeah, actually the `--noFallthroughCasesInSwitch` - Option is even more dangerous, because you will not notice, that it does not fall through even though you expected it to. I would prefer to have this tslint option fixed and not use `--noFallthroughCasesInSwitch`.

ufukomer:
Fix this pls!
+1

kuldarim:
+1

Jan539:
Same happens with if-else-statements where both have a return.

rvalimaki:
@Jan539 

I might misunderstood, but if you have an if-else statement both having a return, you shouldn't have. Since if your "if" already returns, there is no need for "else". Rest of the code is "else", and thus "else" should be omitted.

Matmo10:
I also get issues with this rule in the following scenario:

```
[...]

Issue #1538, opened 2016-09-02
no-switch-case-fall-through thrown on nested switch statements where all cases of inner switch have return statements


sumitkmr:
### Bug Report
- **TSLint version**: `3.2.0`
- **TypeScript version**: `1.8.10`
- **Running TSLint via**: Node.js API
#### TypeScript code being linted

``` ts
function foo (a: number, b: number): number {
    switch (a) {
        case 1:
            switch (b) {
                case 2:
                    return 1;
                default:
                    return 2;
            }
        default:
            return 3;
    }
}
```

with `tslint.json` configuration:

``` json
(include if relevant)
```
#### Actual behavior

Throws no-switch-case-fall-through at line before default case of the outer loop.
#### Expected behavior

No error.

Buuhuu:
Same for 
```
function(a: number, b:number) {
    switch (a) {
        case 1:
            if (b > 10) {
                return;
            } else {
                return;
            }
        case 2:
            return;
        default:
            return;
    }
}
```

michaeljota:
It does not event need to be nested to throw. 

```ts
function(a: number, b:number): void {
    switch (a) {
        case 1: {
            return;
        } // Throws
        case 2:
            return;
        default:
            return;
    }
}
```

raveclassic:
Same here - `no-switch-case-fall-through` doesn't recognize `break/return` inside case blocks `case ... { ... }`.

Issue #1033, opened 2016-03-12
no-switch-case-fall-through: Expect break before default


unional:
Hi, I was expecting `no-switch-case-fall-through` to catch issues like this:

``` ts
switch (prop) {
  case 1:
    // do stuff...
    // forget to break....
  case 2:
    // do stuff...
    break;
  default:
    // do stuff...
    break;
```

But it also catches these:

``` ts
switch (prop) {
  case 1:
  case 2:
    // do stuff...
    break;
  case 3:
  default:
    // do stuff...
    break;
}
```

I think that (especially the `default` case) is a pretty common and valid pattern. What do you think?

JKillian:
You can add a special `/* falls through */` comment, which will tell the rule that you want to allow a fall through to happen. I think 

```
  case 3:
  default:
```

should be allowed though, seems like a bug in the rule potentially.

HolgerJeromin:
I think this is a bug as the comment
https://github.com/palantir/tslint/blob/e5b5ff16677dd00e580839fb1a9ad7602ab114e2/src/rules/noSwitchCaseFallThroughRule.ts#L40
mention it as a valid case.
And perhaps the [test](https://github.com/palantir/tslint/blob/e5b5ff16677dd00e580839fb1a9ad7602ab114e2/test/rules/no-switch-case-fall-through/test.ts.lint#L30) is wrong, too.

Issue #950, opened 2016-01-29
no-switch-case-fall-through does not detect conditionals


domoritz:
tslint wants me to add a break but ts doesn't allow it. 

``` ts
switch (x) {
  case a:
    if (foo) {
      break;
    } else {
      break;
    } 
    // tslint expects a break here but ts doesn't allow it because it is unreachable code
  case b:
    break;
}
```

joskuijpers:
+1

ajafff:
#1815 brought me here

I would like to fix this issue.
But before I do, I'd like to know if this rule is still necessary and supported, since the compiler also has the ability to check for switch-case-fallthrough.
So is this worth the effort or will this rule be deprecated soon?

domoritz:
I think it is still relevant because the compiler check is implemented in a way that it is less flexible. But I suggest that you find a few test cases and see whether they raise warnings/errors in tslint vs tsc.

PR #2218, opened 2017-02-17
Rewrite and bugfix no-switch-case-fallthrough

ajafff:
#### PR checklist

- [x] Fixes: #785, #950, #1033, #1538
  - [x] Includes tests

[bugfix] handle break, throw, continue and return nested in block, if-else and switch
[bugfix] allow empty case clauses before default clause
[enhancement] allow single line comment `// falls through`

Depends on #2217

### DIFF
--- a/package.json
+++ b/package.json
@@ -43,6 +43,7 @@
     "glob": "^7.1.1",
     "optimist": "~0.6.0",
     "resolve": "^1.1.7",
+    "tsutils": "^1.0.0",
     "update-notifier": "^2.0.0"
   },
   "peerDependencies": {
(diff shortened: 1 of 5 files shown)


nchen63:
@ajafff thanks!

ajafff:
Github didn't automatically closed the other open issues mentioned above: #950, #1033, #1538

glen-84:
The `// falls through` comment check should probably be case-insensitive and allow additional text.

* `// falls through`
* `// Falls through`
* `// Falls through.`
* `// Falls through -- for reasons.`

ajafff:
@glen-84 makes sense. I opened #2983 to address this.

Issue #2068, opened 2017-01-18
Support destructuring parameter of prefer-const


hberntsen:
ESLint [allows](http://eslint.org/docs/rules/prefer-const) you to ignore `prefer-const` when destructuring. I'd like to see that in TSLint as well.

Issue #2195, opened 2017-02-09
prefer-const false positive in for loop


Avol-V:
### Bug Report

- __TSLint version__: `4.4.2`
- __TypeScript version__: `2.1.6`
- __Running TSLint via__: Visual Studio Code

#### TypeScript code being linted

```ts
for ( let i = 0, n = 10; i < n; i++ )
{
	console.log( i );
}
```

with `tslint.json` configuration:

```json
"prefer-const": true,
```

#### Actual behavior

```
[tslint] Identifier 'n' is never reassigned; use 'const' instead of 'let'. (prefer-const)
```

#### Expected behavior

I can't do this in `for` loop.

nchen63:
can you pull out `n` from the loop?

Avol-V:
It's just a simplified example but with common using of `for`, for my opinion. The main idea there is that I need this variables to be only inside loop block, and I can do this by defining it in initialisation section of `for`. If I create my variable outside the loop, it will be available in whole block later. If I create my variable inside the loop, it will be created on every iteration.

ajafff:
I get your point. This also bugged me several times.
A real world example:
```ts
for (let i = 1, l = arr.length; i<l; ++i)
    // do stuff
```

To summarize: this is more a feature request than a bug report.
The request is for an option to ignore VariableDeclarationLists where at least one variable is reassigned.
Am I right @Avol-V?

Avol-V:
@ajafff yes it can be an option for any list, but I think it much better to ignore just `for` loops, like in `one-variable-per-declaration` rule.

ajafff:
I'm currently working on a rewrite of this rule and will make it configurable while I'm already there. You can expect a PR this week.

PR #2219, opened 2017-02-17
Rewrite prefer-const

ajafff:
#### PR checklist

- [x] Fixes: #2068, #2195
  - [x] Includes tests
- [x] Documentation update

[enhancement] show warnings for `var`
[bugfix] correctly handle variables shadowed by parameters and catched exceptions
[bugfix] don't warn if one variable in a for loop initializer can not be const
[bugfix] handle more cases in destructuring
[bugfix] only fix initialized variables
[new-rule-option] `{destructuring: "all"}` to only warn if all variables in a destructuring can be const

Tests fail with typescript@2.0 because one test contains syntax (spread) which is not supported in that version of the language. In real life this will not result in a problem.
How to handle the ci failure?


nchen63:
I don't think we should special case variable declaration in `for` loops. This makes sense:

```
for (const x of y) { ... }
```

The issue with #2195 actually didn't really have to do with `for` loops so much as declaring 2 things with the same `let`.

ajafff:
@nchen63 I also thought about adding an option for multiple variables declared in the same `let`. But that can simply be fixed by using separate VaraibleStatements.

The special handling is only for `for`-loops, not `for ... of` or `for ... in`;
The reason is that for loops are the only place where you cannot split a VariableDeclarationList into multiple VariableStatements without moving a variable to another scope.

```ts
for (let foo of bar); // still fails
for (let foo in bar); // still fails
for (let foo=bar; someCondition;); // still fails because foo is not reassigned
let i = 0;
for (let len=foo.length; i<len;++i); // still fails because len is not reassigned

// this is the only case that does not fail anymore
// Otherwise you would need to move variable len out of the loop initializer to the parent scope, which could leed to unexpected behaviour if it shadows another variable
for (let i=0, len=foo.length; i<len; ++i); // no fail for len because i is reassigned
```

Btw: eslint also works that way.

ajafff:
I made some further changes to correctly handle reassigned variables in parameter initializer.
Additionally fixes will now only be applied if all variables of a VariableDeclarationList are initialized to avoid compiler errors.

ajafff:
Regarding the failing tests with older versions of typescirpt:

#2253 will eventually have the same problem and so will all future rules based on new language features or syntax.
I think we need a way to tell the testRunner to ignore certain tests for certain versions of typescript.

ajafff:
The failing tests are now fixed. Should be ready to merge.

nchen63:
@ajafff thanks!

Issue #2202, opened 2017-02-13
Autofixing incorrectly handles strings with \n


WiseBird:
### Bug Report

- __TSLint version__: `4.4.2`
- __TypeScript version__: `2.1.4`
- __Running TSLint via__: CLI

Auto fixing quotes " => ' in
str.split("\n");

#### Actual behavior

modelingResultStr.split('
');

#### Expected behavior

str.split('\n');

PR #2224, opened 2017-02-20
Preserve newlines and tabs when autofixing quote marks

moritzg91:
#### PR checklist

- [x] Addresses an existing issue: #2202
- [x] Includes tests
- [x] Documentation update

Simple change in quotemarkRule::visitStringLiteral. Search the escaped text for \n and \t and replace with escaped versions. Also added a string containing tab/newline to the test cases for the quotemark rule.

First PR to this project, so I'm sorry if I missed something obvious

### DIFF
--- a/src/rules/quotemarkRule.ts
+++ b/src/rules/quotemarkRule.ts
@@ -82,11 +82,10 @@ class QuotemarkWalker extends Lint.RuleWalker {
 
     public visitStringLiteral(node: ts.StringLiteral) {
         const expectedQuoteMark = node.parent!.kind === ts.SyntaxKind.JsxAttribute ? this.jsxQuoteMark : this.quoteMark;
-        const actualQuoteMark = node.getText()[0];
+        const text = node.getText();
+        const actualQuoteMark = text[0];
         if (actualQuoteMark !== expectedQuoteMark && !(this.avoidEscape && node.text.includes(expectedQuoteMark))) {
-            let escapedText = node.text.replace(new RegExp(expectedQuoteMark, "g"), `\\${expectedQuoteMark}`);
-            // replace newlines and tabs with their escaped versions
-            escapedText = escapedText.replace(new RegExp("\n", "g"), "\\n").replace(new RegExp("\t", "g"), "\\t");
+            const escapedText = text.slice(1, -1).replace(new RegExp(expectedQuoteMark, "g"), `\\${expectedQuoteMark}`);
             const newText = expectedQuoteMark + escapedText + expectedQuoteMark;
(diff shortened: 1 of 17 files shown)


palantirtech:
Thanks for your interest in [palantir/tslint](https://github.com/palantir/tslint), @moritzg91! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

src/rules/quotemarkRule.ts
```diff
@@ -84,7 +84,9 @@ class QuotemarkWalker extends Lint.RuleWalker {
         const expectedQuoteMark = node.parent!.kind === ts.SyntaxKind.JsxAttribute ? this.jsxQuoteMark : this.quoteMark;
         const actualQuoteMark = node.getText()[0];
         if (actualQuoteMark !== expectedQuoteMark && !(this.avoidEscape && node.text.includes(expectedQuoteMark))) {
-            const escapedText = node.text.replace(new RegExp(expectedQuoteMark, "g"), `\\${expectedQuoteMark}`);
+            let escapedText = node.text.replace(new RegExp(expectedQuoteMark, "g"), `\\${expectedQuoteMark}`);
```
nchen63: try getting `node.getText()`, which will return a string with escaped characters, so you don't have to do the second replace. Otherwise you'd need to add a more replacement characters like "\\" and "\r".

moritzg91: makes sense, updated to use getText.

jmlopez-rod:
I'm using tslint 4.5.1 but the problem with the escaped characters still seems to be present.

I'm having this

```ts
 { code: "var {\nx,y\n} = y", options: ["never"] },
```

turned into

```ts
  { code: 'var {
x,y
} = y', options: ["never"] },
```

nchen63:
can't reproduce on 4.5.1. if it's still an issue, please open a ticket

jmlopez-rod:
I should have checked using the command line, @nchen63 you are correct. The issue I'm seeing is with my IDE. I'm using

```
IntelliJ IDEA 2017.1 EAP
Build #IU-171.3224.1, built on February 14, 2017
IntelliJ IDEA EAP User
Expiration date: March 16, 2017
JRE: 1.8.0_112-release-b702 x86_64
JVM: OpenJDK 64-Bit Server VM by JetBrains s.r.o
Mac OS X 10.12
```

It has a feature to fix the issues (uses the output provided by tslint) but it seems that this is the culprit and not tslint.

screen shot 2017-03-14 at 1 06 09 pm

I'll update and see if they fixed it, otherwise I'll report the issue.

PR #2225, opened 2017-02-21
Add fixer for commentFormatRule

QuentinFchx:
#### PR checklist

  - [x] Includes tests
- [ ] Documentation update

Add a fixer for `commentFormat` rule

I quickly tried to fix lower/upper case but I don't really know how replacement stack is applied (fifo/lifo?). I can add a fix in this PR if necessary

### DIFF
--- a/src/rules/commentFormatRule.ts
+++ b/src/rules/commentFormatRule.ts
@@ -36,6 +36,7 @@ export class Rule extends Lint.Rules.AbstractRule {
     public static metadata: Lint.IRuleMetadata = {
         ruleName: "comment-format",
         description: "Enforces formatting rules for single-line comments.",
+        hasFix: true,
         rationale: "Helps maintain a consistent, readable style in your codebase.",
         optionsDescription: Lint.Utils.dedent`
             Three arguments may be optionally provided:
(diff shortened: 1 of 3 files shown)


palantirtech:
Thanks for your interest in [palantir/tslint](https://github.com/palantir/tslint), @QuentinFchx! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

ajafff:
> I don't really know how replacement stack is applied

Replacements are sorted by their start position in descending order. Overlapping replacements are not really possible.
In your case if you want to add one space and change the first character to upper/lower, you would have the same start index. Unfortunately `Array.prototype.sort()` does not guarantee that the order is preserved.

I don't have a good idea how to fix both.

nchen63:
To handle the case for a comment with multiple errors, you could emit one error instead of multiple errors and generate a string that describes multiple problems. The fix would then fix all of the problems at once.

adidahiya:
Closing for now, feel free to push changes & leave a comment here if you'd like us to re-open the PR.

Issue #1905, opened 2016-12-18
Multiline trailing comma rule flags type parameters 2


mizunashi-mana:
### Bug Report

- __TSLint version__: `4.1.1`
- __TypeScript version__: `2.1.4`
- __Running TSLint via__: gulp-tslint

#### TypeScript code being linted

```ts
var foo = <
  A,
  B
>(a: A, b: B) => 42;
function foo(a: [
  number,
  string
]): void {
  return 42; 
}
```

with `tslint.json` configuration:

```json
{
    "trailing-comma": [true, {"multiline": "always", "singleline": "never"}]
}
```

#### Actual behavior

```typescript
var foo = <
  A,
  B
   ~ [Missing trailing comma]
>(a: A, b: B) => 42;
function foo(a: [
  number,
  string
  ~ [Missing trailing comma]
]): void {
  return 42; 
}
```

This is invalid typescript code.  Related: #1769 

#### Expected behavior

```typescript
var foo = <
  A,
  B
>(a: A, b: B) => 42;
function foo(a: [
  number,
  string
]): void {
  return 42; 
}
```

nchen63:
also fomr #1769, this fails:

```
function foo(a: [
  number,
  string
  ~ [Missing trailing comma] // expecting no warnings
]): void {
  return 42; 
}
```

Issue #1810, opened 2016-12-01
trailing-comma rule should be smarter about interfaces


adidahiya:
Interfaces can be written with semicolon delimiters, comma delimiters, or no delimiter at all.

Currently, TSLint will require the following trailing comma if you don't use any delimiter, and this is awkward:

```ts
interface Post {
  title: string
  age: number,
}
```

__Proposal__: enhance the rule logic to only require the trailing comma if the other fields have trailing commas.

reported by @schickling and @Zjaaspoer in #960

IllusionMH:
@adidahiya I'm trying to get failing tests to reproduce this issue, but still no errors on `master` branch

I'm altering files in `test/rules/trailing-comma/multiline-always`
`tslint.json`
```json
{
  "rules": {
    "trailing-comma": [true, {"multiline": "always"}]
  }
}
```
`test.ts.lint`
```ts
interface Post {
  title: string
  age: number
}

interface Post2 {
  title: string,
  age: number
}
```

Am I missing something?

nchen63:
this may be a auto-fixer bug

franekp:
Also, please consider cases such as:
```
type EventEmitterMap = {
    [K in keyof DocumentEventMap]: EventEmitter<K>
} & {
    onfoo: EventEmitter<{foo: Foo}>
    onbar: EventEmitter<{bar: Bar}>
}
```

w0rp:
It would be nice to require trailing commas for array literals, object literals, function declarations and function calls, and type literals, but _also_ require that no commas are ever used at the end of lines for `interface`, where the commas are redundant.

Levino:
What is the status here?

JoshuaKGoldberg:
This should have been fixed by #2236. 🎉

Closing for housekeeping purposes, but please do file a new issue or comment on #2285 if that's not the case!

PR #2236, opened 2017-02-24
Rewrite and simplify trailing-comma

ajafff:
#### PR checklist

- [x] Addresses an existing issue: #1905
  - [x] Includes tests
- [x] Documentation update

[bugfix] No longer enforce trailing commas in type parameters and tuple types. Fixes: #1905
[enhancement] New checks for CallSignature and NamedExports
Removed buggy check for interfaces.

This removes the buggy check for interfaces (Ref #1810). I could not find any smart logic to find out if trailing comma is desired or not. That maybe needs a new config option. Let me know if this is needed to get this merged.


test/rules/trailing-comma/singleline-never/test.ts.lint
```diff
@@ -114,7 +114,6 @@ var [ddOne, ddTwo] = dd;
 var a: [string, number] = null;
 
 var a: [string, number,] = null;
-                      ~          [Unnecessary trailing comma]
```
nchen63: shouldn't this still be flagged?

test/rules/trailing-comma/singleline-never/test.ts.lint
```diff
@@ -376,8 +375,7 @@ class Test<A, B, C> {
     }
 
     foo5<
-        T extends Test<[A, A], [A, A,], A>,
-                                    ~       [Unnecessary trailing comma]
```
nchen63: why modify this test case? (other places too)

src/rules/trailingCommaRule.ts
```diff
@@ -19,13 +19,18 @@ import * as ts from "typescript";
 
 import * as Lint from "../index";
 
+interface Options {
+    multiline: string;
```
nchen63: make this of type `"always" | "never"`?

src/rules/trailingCommaRule.ts
```diff
@@ -62,205 +67,106 @@ export class Rule extends Lint.Rules.AbstractRule {
...
     }
+}
 
-    public visitArrowFunction(node: ts.ArrowFunction) {
-        this.lintChildNodeWithIndex(node, 1);
-        super.visitArrowFunction(node);
+class TrailingCommaWalker extends Lint.AbstractWalker<Options> {
+    public walk(sourceFile: ts.SourceFile) {
+        const cb = (node: ts.Node): void => {
+            switch (node.kind) {
```
nchen63: nice, so much cleaner

src/rules/trailingCommaRule.ts
```diff
@@ -62,205 +67,106 @@ export class Rule extends Lint.Rules.AbstractRule {
...
-                const hasSemicolon = grandChildren.some((grandChild) =>
-                    Lint.childOfKind(grandChild, ts.SyntaxKind.SemicolonToken) !== undefined);
-
-                if (!hasSemicolon) {
-                    const endLineOfClosingElement = this.getSourceFile().getLineAndCharacterOfPosition(children[i + 2].getEnd()).line;
-                    this.lintChildNodeWithIndex(children[i + 1], grandChildren.length - 1, endLineOfClosingElement);
-                }
+        const sourceText = this.sourceFile.text;
+        for (const member of members) {
+            // PropertSignature in TypeLiteral can end with semicolon or comma. If one ends with a semicolon, don't check for trailing comma
```
nchen63: `PropertySignature`

src/rules/trailingCommaRule.ts
```diff
@@ -62,205 +67,106 @@ export class Rule extends Lint.Rules.AbstractRule {
...
-
-        for (let i = 0; i < children.length - 2; i++) {
-            syntaxListWrapperTokens.forEach(([openToken, closeToken]) => {
-                if (children[i].kind === openToken &&
-                    children[i + 1].kind === ts.SyntaxKind.SyntaxList &&
-                    children[i + 2].kind === closeToken) {
-                    this.lintChildNodeWithIndex(node, i + 1);
-                }
-            });
+    private checkListWithEndToken(node: ts.Node, list: ts.NodeArray<ts.Node>, closeTokenKind: ts.SyntaxKind) {
```
nchen63: you don't really need `closeTokenKind` as a parameter since it's always called with`CloseParenToken`, but fine as-is

test/rules/trailing-comma/singleline-never/test.ts.lint
ajafff: Actually the trailing comma in tuply *types* was never valid:
http://www.typescriptlang.org/play/#src=var%20a%3A%20%5Bstring%2C%20number%2C%5D%20%3D%20null%3B
The previous implementation of this rule with option "always" even enforced the trailing comma here, which would give you compile errors. I removed the check, because typescript will show errors for unnecessary (invalid) trailing commas in this case.

test/rules/trailing-comma/singleline-never/test.ts.lint
ajafff: see comment above

src/rules/trailingCommaRule.ts
ajafff: I'd prefer to leave it as is.
Maybe someday type parameters can have trailing commas. Then this method could also handle `GreaterThenToken`.

nchen63:
@ajafff thanks!

PR #2201, opened 2017-02-12
add fix function to noConsecutiveBlankLines

YuichiNukiyama:
#### PR checklist

- [x] Addresses an existing issue: #2185
- [x] New feature, bugfix, or enhancement
  - [x] Includes tests
- [x] Documentation update

#### What changes did you make?

add fix function to noConsecutiveBlankLines

PR #2239, opened 2017-02-25
Fix no-consecutive-blank-lines

ajafff:
#### PR checklist

- [x] bugfix, or enhancement

#### Overview of change:
Since landing #2201 the tests for `no-consecutive-blank-lines` fail.
This PR fixes the tests and also simplifies the whole rule logic.
* Correctly apply fixes at EOF.
* Fix tests.
* Simplify collecting consecutive blank lines.

The error position changes when allowing multiple blank lines. Before it was always at the second line. Now the error is at the first line, that exeeds the limit.


nchen63:
@ajafff thanks!

Issue #772, opened 2015-11-05
New rule: "prefer-template" to disallow string concatenation


adidahiya:
favor template strings instead

adidahiya:
eslint calls it `prefer-template` http://eslint.org/docs/rules/prefer-template

apexskier:
I think this can be closed

PR #2234, opened 2017-02-24
Upgrade to TypeScript 2.2

andy-ms:
* Need to provide a type argument to `Promise` or there are compile errors.

#### PR checklist

- [ ] Addresses an existing issue: #0000
- [ ] New feature, bugfix, or enhancement
  - [ ] Includes tests
- [ ] Documentation update

#### What changes did you make?

Upgraded the compiler version.

PR #2243, opened 2017-02-26
Add `prefer-template` rule

andy-hanson:
#### PR checklist

- [X] Addresses an existing issue: #772
  - [X] Includes tests
- [X] Documentation update

#### Overview of change:

Added the `prefer-template` rule, which warns on string literal concatenation.

#### CHANGELOG.md entry:

[new-rule] `prefer-template`


andy-hanson:
Test failures fixed by #2234.

andy-hanson:
IMO it's always better to use a template (except for simple cases like `x + "!"`).

```ts
const str = "a" +
	someVeryLongFunction(someVeryLongArgument()) +
	"b";
```

should be:

```ts
const x = someVeryLongFunction(someVeryLongArgument());
const str = `a${x}b`;
```

If you like complex nested expressions, you can still do this:

```ts
const str = `a${
	someVeryLongFunction(someVeryLongArgument())
}b`;
```

ajafff:
You're absolutely right for the cases you mentioned.
I meant this:
```ts
const str = "I'm a really really really long text " +
            "that needs to be split into several lines " +
            "to keep it readable. "+
            "But I shall not contain any line breaks. " + 
            "How would you write this as a template string?";
```

andy-hanson:
Actually, it already would never warn for concatenation of string literal + string literal. Added a warning to use a template literal if the strings *did* contain newlines.

test/rules/prefer-template/default/test.ts.lint
```diff
@@ -0,0 +1,32 @@
...
+x`a` + y; // OK, can't simplify something with a tag
+
+"a" + "b"; // OK to concatenate strings without newlines.
+
+"a\n" + "b";
+~~~~~~~~~~~ [1]
+
+"a" + "\nb";
+~~~~~~~~~~~ [1]
+
```
nchen63: also add test cases for 1+1, 1+1+"a", and x+y

andy-hanson: :+1:

nchen63:
@andy-hanson thanks!

zakhenry:
@ajafff not sure which is cleaner, but in your example you can always do
```ts
const str = `I'm a really really really long text\
that needs to be split into several lines\
to keep it readable.\
But I shall not contain any line breaks.\
How would you write this as a template string?"`;
```

PR #662, opened 2015-09-16
VSCode uses .vscode instead of .settings since 0.8.0

DickvdBrink:
Just moved the files to the new folder (was done automatically by opening it in VSCode)

VSCode seems to changed the line-ending or something... not sure why it changed; can fix it if you want me to :)

PR #2246, opened 2017-02-26
Remove `.settings` directory

andy-hanson:
Presumably should have been part of #662?


DickvdBrink:
I left it in PR 662 because it is not a VSCode file, I thought it was maybe Eclipse or something.
Not sure if it is required though ;)

andy-hanson:
Ah, it says something about Eclipse on the last line.
There's also a [`.project`](https://github.com/palantir/tslint/blob/master/.project) file.
These haven't been touched in a while -- does Eclipse still use a custom build system for TypeScript? The file has gotten out-of-date; it even specifies `compiler.noImplicitAny=false`.

Issue #2228, opened 2017-02-22
Verbose should be the default formatter, and prose should be abandoned


gotofritz:
The only difference between prose and verbose is that the latter also has the rule name at the beginning of the error message. I can't think of a case when someone would NOT want the rule name - ESlint for example does it by default. To me prose seems completely useless and should be dumped in favour of verbose. Thank you.

adidahiya:
I'm all for it, but it's a small break in case users are parsing the output of the default formatter. Will tackle it in 6.0.

JoshuaKGoldberg:
Removing the `Type: Breaking Change` label per #4811. Now accepting PRs!

PR #2247, opened 2017-02-26
member-access: Clean up and add "no-public" option.

andy-hanson:
#### PR checklist

  - [X] Includes tests
- [X] Documentation update

#### Overview of change:

Added a `no-public` option to `member-access`.
Also removed documentation about `constructor` and `get/set` only being public, because this appears to not be true any more.

#### CHANGELOG.md entry:

[new-rule-option] `member-access`: `no-public`


src/rules/memberAccessRule.ts
```diff
@@ -15,127 +15,123 @@
...
-        if (this.hasOption("check-constructor")) {
-            // constructor is only allowed to have public or nothing, but the compiler will catch this
-            this.validateVisibilityModifiers(node);
+        const options = this.getOptions().ruleArguments;
+        const noPublic = options.indexOf(OPTION_NO_PUBLIC) !== -1;
+        let checkAccessor = options.indexOf(OPTION_CHECK_ACCESSOR) !== -1;
+        let checkConstructor = options.indexOf(OPTION_CHECK_CONSTRUCTOR) !== -1;
+        if (noPublic) {
+            if (checkAccessor || checkConstructor) {
+                throw new Error("If 'no-public' is present, it should be the only option.");
```
nchen63: Config error should state rule name like:

 "`member-access` rule: the `no-public` option cannot be combined with other options"

src/rules/memberAccessRule.ts
```diff
@@ -15,127 +15,123 @@
...
-    public visitConstructorDeclaration(node: ts.ConstructorDeclaration) {
-        if (this.hasOption("check-constructor")) {
-            // constructor is only allowed to have public or nothing, but the compiler will catch this
-            this.validateVisibilityModifiers(node);
+        const options = this.getOptions().ruleArguments;
+        const noPublic = options.indexOf(OPTION_NO_PUBLIC) !== -1;
+        let checkAccessor = options.indexOf(OPTION_CHECK_ACCESSOR) !== -1;
+        let checkConstructor = options.indexOf(OPTION_CHECK_CONSTRUCTOR) !== -1;
+        if (noPublic) {
+            if (checkAccessor || checkConstructor) {
```
nchen63: I can compile a private constructor and private getter

src/rules/memberAccessRule.ts
andy-hanson: I disagree; rules shouldn't have to state their name. It should just be the default to include the name of the failing rule in the output. See #2228.

src/rules/memberAccessRule.ts
andy-hanson: This checks that when `no-public` is set, `check-accessor` and `check-constructor` aren't set, because someone who wants to ban the `public` keyword presumably wants to ban it everywhere, so we always check everything, so other options aren't necessary.

nchen63:
@andy-hanson thanks!

aluanhaddad:
So happy to have this rule. Thank you!

CKGrafico:
Probably we didn't understand well the rule but for me this is to don't use 'public' because is default value in TS.
When we use `"member-access": [true, "no-public"]` we have this error:

`The class method 'kendoDatePicker' must be marked either 'private', 'public', or 'protected'`

ajafff:
@CKGrafico do you use tslint 5.0.0 or greater?

PR #2251, opened 2017-02-26
Add `return-undefined` rule

andy-hanson:
#### PR checklist

  - [X] Includes tests
- [X] Documentation update

#### Overview of change:

Adds the `return-undefined` rule, which specifies when to prefer `return;` vs `return undefined;`.

#### CHANGELOG.md entry:

[new-rule] `return-undefined`


src/rules/returnUndefinedRule.ts
```diff
@@ -0,0 +1,135 @@
...
+        check(node);
+        return ts.forEachChild(node, cb);
+    }
+
+    function check(node: ts.Node): void {
+        if (!u.isReturnStatement(node)) {
+            return;
+        }
+
+        const returnKindFromExpression = !node.expression
```
nchen63: rewrite as `if` statements

src/rules/returnUndefinedRule.ts
```diff
@@ -0,0 +1,135 @@
...
+            ? ReturnKind.Void
+            : u.isIdentifier(node.expression) && node.expression.text === "undefined"
+            ? ReturnKind.Value
+            : undefined;
+
+        if (returnKindFromExpression === undefined) {
+            return;
+        }
+
+        const functionReturningFrom = Lint.ancestorWhere(node, isFunctionLike)! as FunctionLike;
```
nchen63: Should handle undefined return value properly. This may crash TSLint later when linting invalid code.

nchen63:
@andy-hanson broke master when merged. Looks like it needs to converted to use `Program`

PR #2253, opened 2017-02-27
Add `use-default-type-parameter` rule

andy-hanson:
#### PR checklist

  - [X] Includes tests
- [X] Documentation update

#### Overview of change:

Added the `use-default-type-parameter` rule, which encourages the use of defaults for type parameters.
Note: Tests will of course only work with `typescript@next` (to be `typescript@2.3` when it releases). How should we handle this so that CircleCI doesn't fail? Maybe the TSLint testing tool could manage installations of TypeScript, and allow settings for rules that only work in newer versions?

#### CHANGELOG.md entry:

[new-rule] `use-default-type-parameter`


adidahiya:
Here's how I think we should proceed with this:

1. wrap up all the commits we want for the v4.5 release
2. merge `master` branch into `next`
3. submit this PR against `next`, which uses `typescript@next`

andy-hanson:
Actually, now that we support specifying a typescript version for a test (#2323), does this still need to be targeted to `next`?

adidahiya:
@andy-hanson is right; this can stay on `master` now

andy-hanson:
Note: Needed to change typescript devDependency to `"next"`.

adidahiya:
bump @andy-hanson, I think we're unblocked here now

andy-hanson:
Still need to finish the 2.3 upgrade first (#2544)

src/rules/useDefaultTypeParameterRule.ts
```diff
@@ -0,0 +1,129 @@
...
+    if (sym === undefined || sym.declarations === undefined) {
+        return undefined;
+    }
+
+    for (const decl of sym.declarations) {
+        switch (decl.kind) {
+            case ts.SyntaxKind.ClassDeclaration:
+            case ts.SyntaxKind.ClassExpression:
+            case ts.SyntaxKind.TypeAliasDeclaration:
+                return (decl as ts.ClassLikeDeclaration | ts.TypeAliasDeclaration).typeParameters;
```
adidahiya: hmm... how about this

```ts
import { isClassLikeDeclaration, isTypeAliasDeclaration } from "tsutils";

for (const decl of sym.declarations) {
    if (isClassLikeDeclaration(decl) || isTypeAliasDeclaration(decl)) {
        return decl.typeParameters;
    }
}
```

test/rules/await-promise/custom-promise/tslint.json
```diff
@@ -1,7 +1,4 @@
 {
-  "linterOptions": {
-    "typeCheck": true
-  },
```
adidahiya: this change is unrelated, please revert...

test/rules/use-default-type-parameter/test.ts.lint
```diff
@@ -0,0 +1,22 @@
...
+
+class C<T = number> {}
+function h(c: C<number>) {}
+                ~~~~~~ [0]
+new C<number>();
+      ~~~~~~ [0]
+class D extends C<number> {}
+                  ~~~~~~ [0]
+
+[0]: This is the default value for this type parameter.
```
adidahiya: can we add some tests for interfaces too?

Issue #2058, opened 2017-01-17
No-unused-expression wrongly lints a comma-separated expression.


Aarilight:
### Bug Report

- __TSLint version__: `4.3.1`
- __TypeScript version__: `2.1.5`
- __Running TSLint via__: Visual Studio Code

#### TypeScript code being linted

```ts
let x: number, y: number;
x = 1, y = 2;
```

with `tslint.json` configuration:

```json
{
    "rules": {
		"no-unused-expression": true
	}
}
```

#### Actual behavior

Line 2 lints with `expected an assignment or function call`

#### Expected behavior

Line 2  contains two assignments, so it should not be linted.

andy-hanson:
Why not just use `x = 1; y = 2;`?

Aarilight:
Doesn't work on one line if statements.

Issue #1839, opened 2016-12-10
no-unused-new: should not be triggered


eamodio:
### Bug Report

- __TSLint version__: `4.0.2`
- __TypeScript version__: `2.1.4`
- __Running TSLint via__:  VSCode

#### TypeScript code being linted

```ts
class Foo {
    value: string;
 }
class Bar {
    foo: Foo;
 }

function setFoo(bar: Bar) {
    bar.foo || (bar.foo = new Foo()); // lint error here (no-unused-new)
    bar.foo.value = 'hi';
}

let bar = new Bar();
setFoo(bar);
```

with `tslint.json` configuration:

```json
{
    "rules": {
        "arrow-parens": false,
        "class-name": true,
        "comment-format": [
            false
        ],
        "curly": false,
        "indent": [
            true,
            "spaces"
        ],
        "new-parens": true,
        "no-duplicate-variable": true,
        "no-eval": true,
        "no-for-in-array": false,
        "no-internal-module": true,
        "no-reference": true,
        "no-trailing-whitespace": true,
        "no-unsafe-finally": true,
        "no-unused-expression": false,
        "no-unused-new": true,
        "no-var-keyword": true,
        "no-var-requires": true,
        "object-literal-key-quotes": [
            true,
            "as-needed"
        ],
        "one-line": [
            true,
            "check-open-brace",
            "check-whitespace"
        ],
        "one-variable-per-declaration": [
            true,
            "ignore-for-loop"
        ],
        "ordered-imports": [
            false,
            {
                "named-imports-order": "case-insensitive"
            }
        ],
        "quotemark": [
            true,
            "single",
            "avoid-escape"
        ],
        "radix": true,
        "semicolon": [
            true,
            "always"
        ],
        "trailing-comma": [
            true,
            {
                "multiline": "never",
                "singleline": "never"
            }
        ],
        "triple-equals": [
            true,
            "allow-null-check"
        ],
        "typedef-whitespace": [
            true,
            {
                "call-signature": "nospace",
                "index-signature": "nospace",
                "parameter": "nospace",
                "property-declaration": "nospace",
                "variable-declaration": "nospace"
            }
        ],
        "use-isnan": true,
        "variable-name": [
            true,
            "allow-leading-underscore",
            "allow-pascal-case",
            "ban-keywords",
            "check-format"
        ],
        "whitespace": [
            true,
            "check-branch",
            "check-decl",
            "check-operator",
            "check-module",
            "check-separator",
            "check-type"
[...]

Issue #2244, opened 2017-02-26
Deprecate no-unused-new


andy-hanson:
Making an issue for my comment https://github.com/palantir/tslint/issues/1839#issuecomment-266313995 .
I don't think `no-unused-new` deserves to be its own rule since we have `no-unused-expression`.
This would also allow us to clean up the implementation of `no-unused-expression`, since `no-unused-new` currently inherits from it.

andy-hanson:
Should be fixed by #2269

PR #2269, opened 2017-02-28
Rewrite no-unused-expression and remove no-unused-new

### LINKED ISSUE #2293: no-unused-expression warns on tagged templates
### Bug Report

- __TSLint version__: 4.4.2
- __TypeScript version__: 2.2.1
- __Running TSLint via__: VSCode

```ts
function foo (strings: TemplateStringsArray, ...values: any[]) {}

foo`a template string`;
```

```json
{
  "extends": "tslint:recommended",
  "rules": {
    "no-unused-expression": true
  },
  "rulesDirectory": [
  ]
}

```

#### Actual behavior
TSLint gives a warning `expected an assignment or function call (no-unused-expression)`

#### Expected behavior
TSLint should recognize the tagged template as the equivalent of a function call.

ajafff:
#### PR checklist

- [x] Addresses an existing issue: #2244
  - [x] Includes tests
- [x] Documentation update

#### Overview of change:

[removed-rule] `no-unused-new` removed, with logic moved into `no-unused-expression`
[new-rule-option] `no-unused-expression` adds option `allow-new` to allow `new` without using the new object
[new-rule-option] `no-unused-expression` adds option `allow-tagged-template` to allow tagged templates for side effects
Fixes #2293 
[bugfix] `no-unused-expression` allow comma separated assignments
Fixes: #2058
[enhancement] `no-unused-expression` allow indirect eval `(0, eval)("");`
[enhancement] `no-unused-expression` checking for unused new can now use option `allow-fast-null-checks`
Fixes: #1839
[enhancement] `no-unused-expression` find unused comma separated expressions in all locations of the code
[enhancement] `no-unused-expression` find unused expressions inside void expression

As @andy-hanson mentioned in #2244 `no-unused-new` is not worth its own rule. This PR integrates it in `no-unused-expression` and removes the rule.


ajafff:
I added special handling for tagged templates to fix #2293 behind a new config option to explicitly allow them. For me that was the best way to go as I don't want to encourage anyone to use tag functions for side effects. That's just my opinion. I can still remove the option and make it the default.

`no-unused-new` is now completely removed. Let me know if we need a temporary solution to deprecate it properly

adidahiya:
> For me that was the best way to go as I don't want to encourage anyone to use tag functions for side effects. That's just my opinion.

I agree, sgtm

> no-unused-new is now completely removed. Let me know if we need a temporary solution to deprecate it properly

Nah, I think this is fine since it's going into 5.0

PR #2275, opened 2017-02-28
Improve the UI for the rules documentation

chrismbarr:
### PR checklist

- [x] enhancement
  - [ ] Includes tests
- [x] Documentation update

### Overview of change:
This PR make some UI changes to the the list of rules, and the individual rule detail pages.

- The rules on the rules list page are now clickable, not just the rule name
- Alternating background colors + hover colors
- "feature" badges for each rule in the list, and larger versions on the actual rule details page
  - yellow badge with document icon when `typescriptOnly` is true
  - green badge with wrench icon when `hasFix` is true
  - red badge with info icon when `requiresTypeInfo` is true

**Here's what the changes look like!**
tslint-docs

Eventually it would be really nice to have some interactive rule filters on that page.  To be able to filter the rules list down to only show the rules that require type info, or could be auto-fixed would be awesome!  I wanted to keep this PR minimal & focused though.

### DIFF
--- a/docs/_includes/rule_list.html
+++ b/docs/_includes/rule_list.html
@@ -1,6 +1,21 @@
-<ul>
+<ul class="rules-list">
  {% assign rules = site.data.rules | where: "type",include.ruleType | sort: "name" %}
  {% for rule in rules %}
-  <li><a href="{{rule.ruleName}}">{{rule.ruleName}}</a> - {{rule.description | markdownify | remove:"<p>" | remove: "</p>"}}</li>
+  <li>
+      <a href="{{rule.ruleName}}">
+        <div class="rule-features">
+            {% if rule.typescriptOnly %}
+                <span class="feature feature-sm feature-ts-only" title="This rule cannot be run against JavaScript files">TS Only</span>
+            {% endif %}
+            {% if rule.hasFix %}
+                <span class="feature feature-sm feature-fixer" title="This rule has the ability to auto-fix violations">Has Fixer</span>
+            {% endif %}
+            {% if rule.requiresTypeInfo %}
+                <span class="feature feature-sm feature-requires-type-info" title="This rule requires type information to run">Requires Type Info</span>
+            {% endif %}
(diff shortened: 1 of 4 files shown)


chrismbarr:
This shows that CircleCI is still running the tests, but clicking through to the details shows that they all passed!  perhaps this was related to the Amazon S3 outage yesterday?

nchen63:
For some reason, the info icon doesn't render for me (Chrome/OSX)

screen shot 2017-03-01 at 12 25 40 pm

chrismbarr:
@nchen63 Hmm, that's odd, it's just a standard unicode symbol.  Here's a test page for that symbol, does it render here?

http://www.fileformat.info/info/unicode/char/1f6c8/browsertest.htm

What about this one?  This could be switched out: http://www.fileformat.info/info/unicode/char/2139/browsertest.htm

adidahiya:
@chrismbarr your first link doesn't work for me, but the [second one](http://www.fileformat.info/info/unicode/char/2139/browsertest.htm) does

chrismbarr:
Ok, change made.  Here's what it now should look like

image

nchen63:
The circle is weirdly stretched out on my browser

screen shot 2017-03-02 at 3 52 26 pm

I pushed a commit to your branch that should fix the issue. Can you see if it looks ok on your side?

chrismbarr:
Ah good idea, that's a better solution to fix the width to the width to `1em;`  I had to move that up to apply to the redular sized badges as well, you added them to just the smaller ones that appeared in the rules list.

The larger ones appear on the actual rules page, and they looked like this:
2017-03-03 07_40_22-rule_ promise-function-async

But I've fixed that now!

I tested this on Windows 10 in Chrome, Firefox, Edge, IE11, IE10, and IE9 - all look good.  Edge & all IE's render the icons a but different using the solid color non-emoji look to them, but they work.

nchen63:
@chrismbarr thanks!

Issue #2278, opened 2017-03-01
Handle overriding just rule severity or just options


nchen63:
with #1738, we have support for warning/error. We also need to handle changing just the severity level or just the options when extending a config file

PR #2284, opened 2017-03-02
Allow more granular composition of rule configuration

nchen63:
#### PR checklist

- [x] Addresses an existing issue: #2278
  - [x] Includes tests
- [ ] Documentation update

#### Overview of change:

Can override just severity or just options
Refactor to parse/upgrade/standardize config up front

#### CHANGELOG.md entry:

[no-log] this follows #1738


src/language/rule/abstractRule.ts
```diff
@@ -17,38 +17,24 @@
...
-        } else {
-            this.ruleArguments = [];
-        }
-
-        if (value.options) {
-            this.ruleArguments = arrayify(value.options);
-        }
-
-        this.ruleSeverity = getRuleSeverity(value);
+    constructor(private options: IOptions) {
```
adidahiya: this is another API break for 5.0 too, right? let's make sure to document the upgrade steps for custom rules

adidahiya:
should we update the tslint:recommended and tslint:latest to use the new config syntax?

src/language/rule/abstractRule.ts
nchen63: I don't think this should affect custom rules. Nobody should be calling the constructor except for ruleLoader in tslint itself.

ajafff: But you can still override the constructor in a base class

PR #2287, opened 2017-03-02
Support dev versions of typescript

andy-ms:
This would help us to use the latest `tslint` alongside an install of `typescript@next`. (Microsoft/TypeScript#14391). CC @vladima.
Previously there was a `next` tag of `tslint` but it doesn't seem to have been updated in a while, and I don't think it's necessary if we can just do this?

If this is accepted, I will also make the same PR to `tsutils`.

### DIFF
--- a/package.json
+++ b/package.json
@@ -47,7 +47,7 @@
     "update-notifier": "^2.0.0"
   },
   "peerDependencies": {
-    "typescript": ">=2.0.0"
+    "typescript": ">=2.0.0 || >=2.0.0-dev || >=2.1.0-dev || >=2.2.0-dev || >=2.3.0-dev"
   },
   "devDependencies": {
     "@types/babel-code-frame": "^6.20.0",


adidahiya:
I'm leaning towards keeping with our current approach with the `next` branch in this repo so that we can compile TSLint with `typescript@next` and ensure that it works in CI before publishing a release. With this change as it is, there are no guarantees that `tslint@latest` will work with `typescript@>=2.3.0-dev`

andy-ms:
Can you add `typescript@next` to your `circle.yaml` like you do for `typescript@2.0.10`?

adidahiya:
Hm, fair point. @nchen63 let's look into that

nchen63:
sounds good, will update circle

ajafff:
> I will also make the same PR to tsutils

@andy-ms done in https://github.com/ajafff/tsutils/commit/f7bfadf9119c54ff0fe11a460990ba8031928680

PR #2288, opened 2017-03-02
Update circle.yml to test against typescript@next

nchen63:
followup to #2287


andy-ms:
CircleCI has a maximum of 4 containers.

nchen63:
Yep, looks like you get 4 containers for open source projects and need to pay to get more. I can combine one of the older node versions with typescript@next

PR #2294, opened 2017-03-04
Merge release/4.5.x into next

adidahiya:


adidahiya:
Node v4 is failing in appveyor because of peer deps. We need a new version of tsutils with https://github.com/ajafff/tsutils/commit/f7bfadf9119c54ff0fe11a460990ba8031928680, but I'm not going to block on that here.

PR #2298, opened 2017-03-05
Restore old behaviour of text wrap in vscode

ajafff:
#### Overview of change:

#2250 changed how vscode wrapped long lines. After that change lines were wrapped at the end of the viewport. This PR restores the old behaviour of wrapping at the specified column.

I also added `wrappingIndent: none`. Wrapped lines will begin in column 0, which makes it easier to distinguish from regular indented lines.

#### CHANGELOG.md entry:
[no-log]


adidahiya:
@andy-hanson @nchen63 is this ok?

PR #2300, opened 2017-03-06
Added an abstract OptionallyTypedRule

### LINKED ISSUE #2091: Rules should be able to opt into --type-check

JoshuaKGoldberg:
Like `AbstractRule`, it provides abstract `apply*` members that must be
implemented.

*(errata: switched the order of checks in `linter.ts` because it felt
quicker having a property check before the function call)*

Also moved `isTypedRule` logic to a separate function, as it's no longer
unique to the `TypedRule` class. An alternative I'd considered but
rejected would be to make a member `isTyped(): this is IRule` on
`IRule`, but that would require classes not inheriting from
`AbstractRule` to also define their own `isTyped()`.

Resolves #2091.

### DIFF
--- a/src/language/rule/optionallyTypedRule.ts
+++ b/src/language/rule/optionallyTypedRule.ts
@@ -20,12 +20,7 @@ import * as ts from "typescript";
 import {AbstractRule} from "./abstractRule";
 import {ITypedRule, RuleFailure} from "./rule";
 
-export abstract class TypedRule extends AbstractRule implements ITypedRule {
-
-    public apply(): RuleFailure[] {
-        // if no program is given to the linter, throw an error
-        throw new Error(`The '${this.ruleName}' rule requires type checking`);
-    }
+export abstract class OptionallyTypedRule extends AbstractRule implements ITypedRule {
 
     public abstract applyWithProgram(sourceFile: ts.SourceFile, languageService: ts.LanguageService): RuleFailure[];
 }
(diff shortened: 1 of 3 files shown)


src/language/rule/optionallyTypedRule.ts
```diff
@@ -0,0 +1,28 @@
...
+ */
+
+import * as ts from "typescript";
+
+import {AbstractRule} from "./abstractRule";
+import {ITypedRule, RuleFailure} from "./rule";
+
+export abstract class OptionallyTypedRule extends AbstractRule implements ITypedRule {
+
+    public abstract apply(): RuleFailure[];
```
ajafff: this one is inherited by AbstractRule and does not need to be specified here again

nchen63:
@JoshuaKGoldberg thanks!

Issue #2305, opened 2017-03-06
`prefer-function-over-method` should ignore abstract methods


sbj42:
`prefer-function-over-method` seems to warn "Class method does not use 'this'. Use a function instead." for abstract class methods.  Abstract class methods do not have implementations, and so of course they don't use 'this'.  `prefer-function-over-method` should ignore all abstract methods.

TSLint version: 4.4.2
TypeScript version: 2.2.1
Running TSLint via: CLI

The following code:
```ts
export abstract class PreferFunction {
    public abstract abstractMethod();
}
```

With the following tslint.json:
```json
{
    "extends": "tslint:latest",
    "rules": {
        "prefer-function-over-method": true
    }
}
```

Produces the following output:
```
example3.ts[2, 21]: Class method does not use 'this'. Use a function instead.
```

ajafff:
You're right. It should ignore all methods without a body, including
overload signatures.

Am 06.03.2017 19:46 schrieb "James Clark" <notifications@github.com>:

PR #2307, opened 2017-03-07
`prefer-function-over-method` should ignore abstract methods

sbj42:
#### PR checklist

- [x] Addresses an existing issue: #2305
  - [x] Includes tests
- [ ] Documentation update

#### Overview of change:

The `prefer-function-over-method` rule was warning for abstract methods.  This change adds a test case for that, and resolves it.  The rule already checks for the `static` modifier, so it was simple enough to also check for the `abstract` modifier.

#### CHANGELOG.md entry:

[bugfix] `prefer-function-over-method` now ignores abstract methods


nchen63:
@sbj42 thanks!

Issue #2255, opened 2017-02-27
`use-isnan` fails when use NaN in `||` operator


timocov:
### Bug Report

- __TSLint version__: 4.3.0
- __TypeScript version__: 2.2.1
- __Running TSLint via__: CLI

#### TypeScript code being linted

```ts
declare var a: number | undefined;
console.log(a || NaN);
```

with `tslint.json` configuration:

```json
{
  "rules": {
    "use-isnan": true
  }
}
```

#### Actual behavior
`Found an invalid comparison for NaN: a || NaN`

#### Expected behavior
No errors

ajafff:
But that's not a comparison.
Your code does basically the following: If a is truthy (`a !== 0 && a !== undefined && !isNan(a)`) log a, else log NaN.
There's nothing wrong in this piece of code and nothing the `use-isnan`-Rule could detect.

sbj42:
I believe @timocov would agree with you, that `use-isnan` should not complain about that code, but it does. I was able to reproduce it by adding the following lines to `rules/use-isnan/test.ts.lint`:

```ts
// no violation for logical operators
x = 0 || NaN;
x = 0 && NaN;
```

Then ran the tests and got:

```
// no violation for logical operators
x = 0 || NaN;
    ~~~~~~~~  [Found an invalid comparison for NaN: 0 || NaN]
x = 0 && NaN;
    ~~~~~~~~  [Found an invalid comparison for NaN: 0 && NaN]
```

The implementation of `use-isnan` has an exception only for the simple assignment operator, but TypeScript has many other operators which are not "comparison" operators.  Many of them would be very silly to use with NaN, but nevertheless the rule does not correctly limit its scope.

ajafff:
Ah, sorry. I misread the opening post.

@sbj42 you already analyzed the problem. Fixing this should not be that difficult. Would you like to fix it or should I go for it?

sbj42:
I'd like to fix it.  I'll give it a try later today.

sbj42:
I believe this is now resolved.

PR #2317, opened 2017-03-08
`use-isnan` should only apply to comparison operators

sbj42:
#### PR checklist

- [x] Addresses an existing issue: #2255
  - [x] Includes tests
- [ ] Documentation update

#### Overview of change:

The `use-isnan` rule was warning about all binary operators except for the simple assignment operator `=`.  This was problematic because `x || NaN` is a reasonable thing to do.  This change ensures that it only applies to comparison operators.  Instead of checking for the assignment exception, the rule now checks against a complete list of comparison operators.

Most binary operators would be silly to use with a literal `NaN`.  I'm not sure whether it would have been better to just add more exceptions (specifically for `||` and `&&`).  The technique in this PR has the risk that if a new comparison operator is later added to the language (e.g. `<=>`) then the rule will need to be updated.  But it seemed better to me to limit the scope of the rule completely.

#### CHANGELOG.md entry:

[bugfix] `use-isnan` now applies only to comparison operators

### DIFF
--- a/src/rules/useIsnanRule.ts
+++ b/src/rules/useIsnanRule.ts
@@ -56,13 +56,13 @@ class UseIsnanRuleWalker extends Lint.RuleWalker {
     }
 
     private isComparisonOperator(operator: ts.BinaryOperator) {
-        return (operator == ts.SyntaxKind.LessThanToken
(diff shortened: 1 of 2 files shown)


nchen63:
@sbj42 thanks for the contribution

Issue #2291, opened 2017-03-03
completed-docs does not report any errors


Rukenshia:
### Bug Report

- __TSLint version__: `4.5.1`
- __TypeScript version__: `2.0.2`
- __Running TSLint via__: CLI

#### TypeScript code being linted

```ts
export function x() {
    return 1;
}
```

with `tslint.json` configuration:

```json
{
  "rules": {
    "completed-docs": true
  }
}
```

#### Actual behavior
```
$ node_modules/.bin/tslint --type-check src/test.ts
Warning: The 'completed-docs' rule requires type checking
```

#### Expected behavior
Shouldn't I get a message about documentation missing? Sadly https://palantir.github.io/tslint/rules/completed-docs/ just displays a white page for me. Maybe I'm just using or imagining it wrong but ideally, I'd like to enforce everything being documented.

ajafff:
For `--type-check` to work you also need to pass `--project ./path/to/tsconfig.json` as stated in the CLI documentation. Also the file you want to lint needs to be part of you project (included in `tsconfig.json`)

But maybe that should also be printed as error or warning ...

Rukenshia:
Okay, that solved it. In the actual script I am running I just passed `-p tsconfig.json`, but tslint needs `--project`. I assumed I could pass it as `-p` like with tsc.

adibbleWLT:
I am having a similar issue using tslint with Angular 4, but the above does not resolve this for me. It was working fine with Angular 2, but after upgrading this one rule does not get respected. All other lint settings, as far as currently noted, run as expected. Setting the rule as: "completed-docs": [ true, "classes", "methods", "functions" ], or even "completed-docs": true, does not cause a lint error even when I delete all documentation comments from all classes/methods/functions in a file. The file is definitely being linted as other rules do get flagged if they fail there. 

Note, the linting configuration settings did not change in the upgrade process. One thing that did change is that --project tsconfig.json is no longer passed to the command (and trying to do so gets an error that "The option '--project' is not registered with the lint command."), but instead in the "lint" section of the .angular-cli.json file. This is correctly set however, and I can change settings in the tsconfig file to affirm that the changes are being reflected in the linting output. The lint command being run is ""lint": "ng lint --type-check".

Issue #2087, opened 2017-01-20
Exit code always 1 with type check errors and `--force` active


lepus2589:
### Bug Report

- __TSLint version__: `4.3.1`
- __TypeScript version__: `2.1.5`
- __Running TSLint via__: CLI

#### TypeScript code being linted

```ts
(x: Number): void => {
    // x is not used in here
}
```

with `tslint.json` configuration: irrelevant

#### Actual behavior

Type checking is activated via the `--type-check` option as well as the `--force` switch. `noUnusedParameters` is active in the `tsconfig.json`, which is given to TSLint with the `--project` argument. The Typescript compiler will of course flag the parameter of the above function as unused.
The file [runner.js](https://github.com/palantir/tslint/blob/master/src/runner.ts#L174) prints the error message (`'x' is declared but never used.`) using an Error. It is not caught anywhere, so the exit code is 1 despite `--force` being used and the (valid) error message is followed by a stack trace. 
Is this the intended behaviour?

#### Expected behavior

I would expect TSLint to print the error messages emitted by the Typescript compiler as linting errors, without an ugly stack trace containing TSLint source files. If not that, then at least when using the `--force` flag, the exit code should be 0, as stated in the documentation.

calidion:
any progress on this?
tslint with errors but still returns 1

adidahiya:
@calidion the issue is marked as closed & fixed. if you're still seeing it you should probably open a new bug report. note that the `--type-check` option is deprecated.

PR #2322, opened 2017-03-09
Extend cli options

ajafff:
#### PR checklist

- [x] Addresses an existing issue: #2087
  - [x] Includes tests
- [x] Documentation update

#### Overview of change:
[bugfix] exit with 0 on type check errors when `--force` is specified
Fixes: #2087
[bugfix] `--test` now correctly strips single quotes from patterns on windows
[enhancement] `--test` can handle multiple paths at once
^useful when shell expands globs
[enhancement] added `-p` as shorthand for `--project` to be consistent with `tsc`
Kind of fixes: #2291
[enhancement] print error when `--type-check` is used without `--project`
[enhancement] don't print stack trace on type check error

### DIFF
--- a/src/runner.ts
+++ b/src/runner.ts
@@ -112,35 +112,30 @@ export class Runner {
     public run(onComplete: (status: number) => void) {
         if (this.options.version) {
             this.outputStream.write(Linter.VERSION + "\n");
-            onComplete(0);
-            return;
+            return onComplete(0);
         }
 
         if (this.options.init) {
             if (fs.existsSync(CONFIG_FILENAME)) {
                 console.error(`Cannot generate ${CONFIG_FILENAME}: file already exists`);
-                onComplete(1);
-                return;
+                return onComplete(1);
             }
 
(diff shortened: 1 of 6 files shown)


src/runner.ts
```diff
@@ -218,23 +216,22 @@ export class Runner {
...
                     // MPEG transport streams use the '.ts' file extension. They use 0x47 as the frame
                     // separator, repeating every 188 bytes. It is unlikely to find that pattern in
                     // TypeScript source, so tslint ignores files with the specific pattern.
                     console.warn(`${file}: ignoring MPEG transport stream`);
-                    return;
+                    fs.closeSync(fd);
+                    continue;
                 }
-            } finally {
+            } catch (e) {
```
nchen63: why is this in a catch instead of a finally? The file descriptor should be closed regardless of what happens. Line 231 can be removed too

ajafff: good catch, I thought finally would not be executed when leaving the try block with `return` or `continue`. Will revert that change

nchen63:
@ajafff thanks!

PR #2331, opened 2017-03-12
Add `prefer-switch` rule

andy-hanson:
#### PR checklist

  - [X] Includes tests
- [X] Documentation update

#### Overview of change:

Added the `prefer-switch` rule, which warns on repeated use of `===` on the same variable in an if statement or if-elseif chain.

#### CHANGELOG.md entry:

[new-rule] `prefer-switch`


src/rules/preferSwitchRule.ts
```diff
@@ -0,0 +1,130 @@
...
+
+type SimpleExpression = ts.PropertyAccessEntityNameExpression | ts.Identifier | ts.NumericLiteral | ts.StringLiteral;
+
+function isSimple(node: ts.Node): node is SimpleExpression {
+    switch (node.kind) {
+        case ts.SyntaxKind.PropertyAccessExpression:
+            return isSimple((node as ts.PropertyAccessExpression).expression);
+        case ts.SyntaxKind.Identifier:
+        case ts.SyntaxKind.NumericLiteral:
+        case ts.SyntaxKind.StringLiteral:
```
ajafff: should include `ts.SyntaxKind.ThisKeyword`

src/rules/preferSwitchRule.ts
```diff
@@ -0,0 +1,130 @@
...
+        let minCases = 2;
+        if (options.length) {
+            const obj = options[0];
+            minCases = obj[OPTION_MIN_CASES];
+        }
+        return this.applyWithFunction(sourceFile, (ctx) => walk(ctx, minCases));
+    }
+}
+
+function walk(ctx: Lint.WalkContext<void>, minCases: number): void {
```
ajafff: Why pass `minCases` as parameter? That's what `ctx.options` is for

andy-hanson: I don't see the point of `ctx.options` when we're just calling a function.

adidahiya:
I forsee a semantic merge conflict since this hasn't been added to `tslint:all`. I'll merge after you make that change @andy-hanson

PR #2333, opened 2017-03-12
Move @types to devDependencies

andy-hanson:
`diff` and `semver` are only used for testing, although `test.ts` is in `src`. Should they be `devDependencies` too?


nchen63:
@andy-hanson thanks!

PR #2334, opened 2017-03-12
Improve Curly Rule to allow no brackets same-line statements

devrelm:
#### PR checklist

- [x] Addresses an existing issue: #822
  - [x] Includes tests
- [ ] Documentation update

#### Overview of change:

In #822, there was a request to improve the Curly rule with the ability to have single-line unbraced statements.

I had originally opened PR #1565 to fix this, which added the options per [my comment on that issue](https://github.com/palantir/tslint/issues/822#issuecomment-247307510). The requested updates to that pull request, plus some merge conflicts from TSLint 4.0 refactoring in the meantime made it more useful to create this new PR instead of updating the original.

This PR is much more focused than the first and keeps closer to the original request (#822). I still tried to leave the code easy to extend in case more options are added later, but this should hopefully come off as a much more subtle and evolutionary change compared to my prior PR's near-rewrite.

Please read the prior pull request (#1565) for context on design decisions.

#### CHANGELOG.md entry:

[new-rule-option] `curly-rule` added options `always` and `ignore-same-line`


devrelm:
One question re: documentation. I updated the documentation at the top of `curlyRule.ts`, but are there any scripts that I have to run manually to update other static doc files (e.g. for the website)?

devrelm:
It looks like the tests are failing, though they're running fine locally for me.

Could it perhaps be a node version issue? I'm running 6.2.2.

devrelm:
Yep, looks like it was a node version issue. I was trying to be fancy with `includes` -- I switched it to `indexOf` and everything looks good :)

nchen63:
docs are no longer checked in and are generated for each release

src/rules/curlyRule.ts
```diff
@@ -34,9 +37,26 @@ export class Rule extends Lint.Rules.AbstractRule {
...
-        options: null,
-        optionExamples: ["true"],
+        optionsDescription: Lint.Utils.dedent`
+            The rule may be set to \`true\`, or to any combination of these arguments:
+
+            * \`"${OPTION_ALWAYS}"\` requires braces for all control-flow statements (same as default)
+            * \`"${OPTION_IGNORE_SAME_LINE}"\` skips checking braces for control-flow statements
+            that are on one line and start on the same line as their control-flow keyword
+
+            Note: \`"${OPTION_ALWAYS}"\` will be overriden by any "ignore" options that are also set.
```
nchen63: We can safely remove `always` since it's not actually used

devrelm: @nchen63 Do you mean just from the docs or overall as an option in general?

devrelm: @nchen63 I went ahead and pulled out the entire option.

PR #2336, opened 2017-03-13
Removed unused test utils

### LINKED ISSUE #2337: Test utils used only in eofRuleTests
In `test/utils.ts`, the following functions are only used by `eofRuleTests.ts`: `getRule`, `getFormatter`, `applyRuleOnFile`, `createFailure`. `eofRuleTests.ts` could probably be simplified so that these utilities aren't necessary.

andy-ms:
Fixes #2337.
Discovered that these utils weren't used.
Also removing `eofLineRuleTests.ts` as that rule has regular tests.


nchen63:
@andy-ms thanks!

PR #2341, opened 2017-03-14
[Breaking change] Have `no-var-requires` check at all depth levels

andy-hanson:
#### PR checklist

  - [X] Includes tests
- [ ] Documentation update

#### Overview of change:

Removed an undocumented feature where `no-var-require` would ignore all `require` statements not at the top level.
I don't see why that would be a desirable feature. It is mentioned in https://github.com/palantir/tslint/commit/18bbae82aea308b40cfc2f7901d98a2e339f4cd3 as "in case user defines their own require". If we want to allow shadowing of 'require', we can make this an optionally-typed rule and detect shadowed declarations using the type checker.


andy-hanson:
Actually, this would make this identical to the `no-require-imports` rule. It seems like one of them should be deprecated anyway though?

adidahiya:
> Actually, this would make this identical to the no-require-imports rule. It seems like one of them should be deprecated anyway though?

No, they're slightly different because `no-var-requires` still allows `import foo = require("foo")` syntax, while `no-require-imports` bans it.

andy-hanson:
Updated, but a better idea might be to deprecate this rule in favor of `no-require-imports` which would now be doing basically the same thing.

JoshuaKGoldberg:
@andy-hanson what do you recommend should happen with this PR? Do you think it should go in? Should the `no-require-imports` rule consume `no-var-requires`? Both?

JoshuaKGoldberg:
Given #4534 and the lack of activity on this PR, I'll go ahead and close it for housekeeping. Anybody: feel free to comment here or in a new issue if you'd like to discuss further!

Issue #2338, opened 2017-03-13
`unified-signatures` erroneously claims overloads can be combined


pelotom:
### Bug Report

- __TSLint version__: 4.5.1
- __TypeScript version__: 2.2.1
- __Running TSLint via__: CLI, VSCode

#### TypeScript code being linted

```ts
function foo(...xs: number[]): void
function foo(xs: number[], y: string): void
function foo(...args: any[]) {
  // ...
}
```

#### Actual behavior
```
function foo(xs: number[], y: string): void
                           ^^^^^^^^^
These overloads can be combined into one signature with an optional parameter.
```

#### Expected behavior

There should be no warning here, because these signatures cannot be further combined. The first signature uses varargs, but the second cannot due to the extra `y` parameter.

PR #2342, opened 2017-03-14
Fix for rest parameters in `unified-signatures`

sbj42:
#### PR checklist

- [x] Addresses an existing issue: #2338
  - [x] Includes tests

#### Overview of change:

`signaturesDifferByOptionalOrRestParameter()` previously considered two signatures similar if the common parameters had the same type, but it did not consider the `dotDotDotToken` sigil when it dd that comparison.  This change has the function check the last parameter of the shorter signature to see if it is a rest parameter.  If so, it may not be possible to combine the two signatures.

This change makes tslint more lenient about a particular scenario:
```ts
function rest2(...xs: number[]): void;
function rest2(xs: number[], y: number): void;
function rest2(...args: any[]) {}
```

tslint previously correctly noted that those signatures could be combined.  After this change it does not warn about them.  Reviewer should consider whether fixing the false warning of #2338 in exchange for failing to warn in this alternative case is a net improvement.


nchen63:
looks good. thanks

PR #2344, opened 2017-03-14
Edit tslint.json

ajafff:
#### Overview of change:
The use of switch statements increased recently. The current `tslint.json` doesn't make it convenient to use. Therefore I enabled `no-switch-case-fall-through` and disabled `switch-default` for this project.

#### CHANGELOG.md entry:
[no-log] internal


nchen63:
I think you edited the wrong tslint.json file

ajafff:
That explains why there were no lint errors :laughing:

Should be fixed now

nchen63:
@ajafff thanks!

PR #2346, opened 2017-03-15
Rewrite no-unsafe-finally

andy-hanson:
#### PR checklist

  - [ ] Includes tests
- [ ] Documentation update

#### Overview of change:

Just a rewrite. Should be simpler now.

### DIFF
--- a/src/rules/noUnsafeFinallyRule.ts
+++ b/src/rules/noUnsafeFinallyRule.ts
@@ -15,7 +15,6 @@
  * limitations under the License.
  */
 
-import * as utils from "tsutils";
 import * as ts from "typescript";
 import * as Lint from "../index";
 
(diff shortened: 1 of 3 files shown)


src/rules/noUnsafeFinallyRule.ts
```diff
@@ -40,171 +39,111 @@ export class Rule extends Lint.Rules.AbstractRule {
...
-function isContinueBoundary(scope: IFinallyScope, node: ts.ContinueStatement): boolean {
-    return node.label ? scope.labels.indexOf(node.label.text) >= 0 : scope.isContinueBoundary;
+        node = parent;
+    }
 }
 
-function isBreakBoundary(scope: IFinallyScope, node: ts.BreakStatement): boolean {
-    return node.label ? scope.labels.indexOf(node.label.text) >= 0 : scope.isBreakBoundary;
+type JumpStatement = ts.BreakStatement | ts.ContinueStatement | ts.ThrowStatement | ts.ReturnStatement;
+function showKind(node: JumpStatement): string {
```
adidahiya: how about naming this `printJumpKind`

src/rules/noUnsafeFinallyRule.ts
```diff
@@ -40,171 +39,111 @@ export class Rule extends Lint.Rules.AbstractRule {
...
-        const scopes = this.getAllScopes();
-
-        let currentScope = this.getCurrentScope();
-        let depth = this.getCurrentDepth();
-
-        while (currentScope) {
-            if (isControlFlowBoundary(currentScope, node)) {
-                return false;
+function jumpIsLocalToFinallyBlock(jump: JumpStatement): boolean {
+    const isBreakContinue = jump.kind === ts.SyntaxKind.BreakStatement || jump.kind === ts.SyntaxKind.ContinueStatement;
```
adidahiya: `isBreakOrContinue`

src/rules/noUnsafeFinallyRule.ts
```diff
@@ -40,171 +39,111 @@ export class Rule extends Lint.Rules.AbstractRule {
...
-
-function isContinueBoundary(scope: IFinallyScope, node: ts.ContinueStatement): boolean {
-    return node.label ? scope.labels.indexOf(node.label.text) >= 0 : scope.isContinueBoundary;
+        node = parent;
+    }
 }
 
-function isBreakBoundary(scope: IFinallyScope, node: ts.BreakStatement): boolean {
-    return node.label ? scope.labels.indexOf(node.label.text) >= 0 : scope.isBreakBoundary;
+type JumpStatement = ts.BreakStatement | ts.ContinueStatement | ts.ThrowStatement | ts.ReturnStatement;
```
adidahiya: let's move this type to just before `function walk` since it's used above too

src/rules/noUnsafeFinallyRule.ts
```diff
@@ -40,171 +39,111 @@ export class Rule extends Lint.Rules.AbstractRule {
...
-            super.visitThrowStatement(node);
-            return;
+function walk(ctx: Lint.WalkContext<void>): void {
+    let inFinally = false;
+    ts.forEachChild(ctx.sourceFile, function cb(node: ts.Node): void {
+        switch (node.kind) {
+            case ts.SyntaxKind.TryStatement:
+                const { tryBlock, catchClause, finallyBlock } = node as ts.TryStatement;
+                ts.forEachChild(tryBlock, cb);
+                if (catchClause) {
```
adidahiya: try to avoid implicit boolean coercion, check `catchClause !== undefined` instead. same a few lines down

adidahiya: yeah, I'd like to enable it, I remember having some issues when trying to use it elsewhere but I haven't put in the time to investigate it fully.

src/rules/noUnsafeFinallyRule.ts
```diff
@@ -40,171 +39,112 @@ export class Rule extends Lint.Rules.AbstractRule {
...
+            case ts.SyntaxKind.ThrowStatement:
+            case ts.SyntaxKind.ReturnStatement:
+                if (inFinally && !jumpIsLocalToFinallyBlock(node as JumpStatement)) {
+                    ctx.addFailureAtNode(node, Rule.FAILURE_STRING(printJumpKind(node as JumpStatement)));
+                }
+                ts.forEachChild(node, cb);
+                break;
+
+            default:
+                ts.forEachChild(node, cb);
```
ajafff:
```ts
if (isFunctionScopeBoundary(node)) {
    const old = inFinally;
    inFinally = false;
    ts.forEachChild(node, cb);
    inFinally = old;
    break;
} else {
}
```

src/rules/noUnsafeFinallyRule.ts
```diff
@@ -40,171 +39,112 @@ export class Rule extends Lint.Rules.AbstractRule {
...
-        while (currentScope) {
-            if (isControlFlowBoundary(currentScope, node)) {
-                return false;
+function jumpIsLocalToFinallyBlock(jump: JumpStatement): boolean {
+    const isBreakOrContinue = jump.kind === ts.SyntaxKind.BreakStatement || jump.kind === ts.SyntaxKind.ContinueStatement;
+    const label = isBreakOrContinue ? (jump as ts.BreakStatement | ts.ContinueStatement).label : undefined;
+
+    let node: ts.Node = jump;
+    // This should only be called inside a finally block, so we'll eventually reach the TryStatement case and return.
+    while (true) {
```
ajafff: Avoid property access on undefined by ending the loop at SourceFile.
You should also stop at function scope boundary.
```ts
while(node.kind !== ts.SyntaxKind.SourceFile || isFunctionScopeBoundary(node))
```

src/rules/noUnsafeFinallyRule.ts
```diff
@@ -40,171 +39,112 @@ export class Rule extends Lint.Rules.AbstractRule {
...
+                }
+                break;
             }
 
-            if (currentScope.isFinallyBlock) {
+            case ts.SyntaxKind.ClassDeclaration:
+            case ts.SyntaxKind.ClassExpression:
+            case ts.SyntaxKind.FunctionDeclaration:
+            case ts.SyntaxKind.FunctionExpression:
+            case ts.SyntaxKind.ArrowFunction:
```
ajafff: this does not include all function scope boundaries. use `isFunctionScopeBoundary(parent)` instead, see comment above

src/rules/noUnsafeFinallyRule.ts
```diff
@@ -40,171 +39,112 @@ export class Rule extends Lint.Rules.AbstractRule {
...
-
-        let currentScope = this.getCurrentScope();
-        let depth = this.getCurrentDepth();
-
-        while (currentScope) {
-            if (isControlFlowBoundary(currentScope, node)) {
-                return false;
+function jumpIsLocalToFinallyBlock(jump: JumpStatement): boolean {
+    const isBreakOrContinue = jump.kind === ts.SyntaxKind.BreakStatement || jump.kind === ts.SyntaxKind.ContinueStatement;
+    const label = isBreakOrContinue ? (jump as ts.BreakStatement | ts.ContinueStatement).label : undefined;
```
ajafff: nit: `ts.BreakOrContinueStatement`

src/rules/noUnsafeFinallyRule.ts
```diff
@@ -40,171 +39,112 @@ export class Rule extends Lint.Rules.AbstractRule {
...
+
+            case ts.SyntaxKind.BreakStatement:
+            case ts.SyntaxKind.ContinueStatement:
+            case ts.SyntaxKind.ThrowStatement:
+            case ts.SyntaxKind.ReturnStatement:
+                if (inFinally && !jumpIsLocalToFinallyBlock(node as JumpStatement)) {
+                    ctx.addFailureAtNode(node, Rule.FAILURE_STRING(printJumpKind(node as JumpStatement)));
+                }
+                ts.forEachChild(node, cb);
+                break;
```
ajafff: nit: `return ts.forEachChild(...)`

src/rules/noUnsafeFinallyRule.ts
```diff
@@ -40,171 +39,112 @@ export class Rule extends Lint.Rules.AbstractRule {
...
             }
 
-            if (currentScope.isFinallyBlock) {
+            case ts.SyntaxKind.ClassDeclaration:
+            case ts.SyntaxKind.ClassExpression:
+            case ts.SyntaxKind.FunctionDeclaration:
+            case ts.SyntaxKind.FunctionExpression:
+            case ts.SyntaxKind.ArrowFunction:
                 return true;
```
ajafff: shouldn't this return `false`?

src/rules/noUnsafeFinallyRule.ts
andy-hanson: We'll always exit this loop, as we only get here if we saw a TryStatement above.

src/rules/noUnsafeFinallyRule.ts
andy-hanson: Done

src/rules/noUnsafeFinallyRule.ts
andy-hanson: If you're climbing parents and see a function *before* you reach the try statement, that means there is a function inside the finally block. Jump statements can't leave a function, so the jump is definitely local to that finally block (because it's local to the function and the function is inside the finally block).

src/rules/noUnsafeFinallyRule.ts
andy-hanson: I don't think that's necessary. `jumpIsLocalToFinallyBlock` checks for function scopes.

src/rules/noUnsafeFinallyRule.ts
```diff
@@ -40,171 +41,110 @@ export class Rule extends Lint.Rules.AbstractRule {
...
+            case ts.SyntaxKind.ContinueStatement:
+            case ts.SyntaxKind.ThrowStatement:
+            case ts.SyntaxKind.ReturnStatement:
+                if (inFinally && !jumpIsLocalToFinallyBlock(node as JumpStatement)) {
+                    ctx.addFailureAtNode(node, Rule.FAILURE_STRING(printJumpKind(node as JumpStatement)));
+                }
+                // falls through
+
+            default:
+                return ts.forEachChild(node, cb);
```
ajafff: you're right, you don't *need* to check for function boundaries here, but it would avoid doing unnecessary time spent in `jumpIsLocalToFinallyBlock` and would make the code clearer.
You can of course avoid doing the additional check if `inFinally === false`.
```ts
if (inFinally && isFunctionScopeBoundary(node)) {
    inFinally = false;
    ts.forEachChild(node, cb);
    inFinally = true;
} else {
}
```

andy-hanson: `jumpIsLocalToFinallyBlock` is only activated if a jump statement appears in a finally block, which with 99% certainty won't happen if people are linting with this rule. (Who on earth would put a loop with a break statement inside of a finally block?) So I don't think it's worth doing more work in the walker to try to save time in `jumpIsLocalToFinallyBlock`.

src/rules/noUnsafeFinallyRule.ts
```diff
@@ -40,171 +41,110 @@ export class Rule extends Lint.Rules.AbstractRule {
...
-
-            if (currentScope.isFinallyBlock) {
-                return true;
+function jumpIsLocalToFinallyBlock(jump: JumpStatement): boolean {
+    const isBreakOrContinue = utils.isBreakOrContinueStatement(jump);
+    const label = isBreakOrContinue ? (jump as ts.BreakOrContinueStatement).label : undefined;
+
+    let node: ts.Node = jump;
+    // This should only be called inside a finally block, so we'll eventually reach the TryStatement case and return.
+    while (true) {
```
andy-hanson: Added a test case. It doesn't crash.

PR #2348, opened 2017-03-15
Clean up prefer-for-of rule

andy-hanson:
#### PR checklist

  - [ ] Includes tests
- [ ] Documentation update

#### Overview of change:

Don't unnecessarily use a BlockScopeAwareRuleWalker, just store a stack of for-loop scopes.

### DIFF
--- a/src/rules/preferForOfRule.ts
+++ b/src/rules/preferForOfRule.ts
@@ -15,6 +15,7 @@
  * limitations under the License.
  */
 
+import * as utils from "tsutils";
 import * as ts from "typescript";
 import * as Lint from "../index";
 import { isAssignment, unwrapParentheses } from "../language/utils";
@@ -42,7 +43,7 @@ export class Rule extends Lint.Rules.AbstractRule {
 
 interface IncrementorState {
     indexVariableName: string;
-    arrayToken: ts.Identifier;
+    arrayExpr: ts.Expression;
     onlyArrayReadAccess: boolean;
 }
 
@@ -64,19 +65,15 @@ function walk(ctx: Lint.WalkContext<void>): void {
 
     function visitForStatement(node: ts.ForStatement): void {
         const arrayNodeInfo = getForLoopHeaderInfo(node);
-        if (!node.incrementor || !arrayNodeInfo) {
+        if (!arrayNodeInfo) {
             return ts.forEachChild(node, cb);
         }
 
-        const { indexVariable, arrayToken } = arrayNodeInfo;
+        const { indexVariable, arrayExpr } = arrayNodeInfo;
         const indexVariableName = indexVariable.text;
 
         // store `for` loop state
-        const state: IncrementorState = {
-            indexVariableName,
-            arrayToken: arrayToken as ts.Identifier,
-            onlyArrayReadAccess: true,
-        };
+        const state: IncrementorState = { indexVariableName, arrayExpr, onlyArrayReadAccess: true };
         scopes.push(state);
         ts.forEachChild(node.statement, cb);
         scopes.pop();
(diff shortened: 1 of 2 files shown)


src/rules/preferForOfRule.ts
```diff
@@ -36,168 +36,177 @@ export class Rule extends Lint.Rules.AbstractRule {
...
+
+}
+
+// returns the iterator and array of a `for` loop if the `for` loop is basic.
+function getForLoopHeaderInfo(forLoop: ts.ForStatement): { indexVariable: ts.Identifier, arrayToken: ts.Expression } | undefined {
+    let indexVariableName: string | undefined;
+    let indexVariable: ts.Identifier | undefined;
+
+    // assign `indexVariableName` if initializer is simple and starts at 0
+    if (forLoop.initializer && forLoop.initializer.kind === ts.SyntaxKind.VariableDeclarationList) {
```
ajafff:
```ts
if (forLoop.initializer && isVariableDeclarationList(forLoop.initializer) && forLoop.initializer.declarations.length === 1) {
    const declaration = forLoop.initializer.declarations[0];
    if (declaration.name.kind === ts.SyntaxKind.Identifier &&
        declaration.initializer !== undefined && isNumericLiteral(declaration.initializer) && declaration.initializer.text === "0") {
        indexVariable = declaration.name;
        indexVariableName = indexVariable.text;
    }
}
```

src/rules/preferForOfRule.ts
```diff
@@ -36,168 +36,177 @@ export class Rule extends Lint.Rules.AbstractRule {
...
-        // ensure `for` condition
-        if (indexVariableName == null
-            || forLoop.condition == null
-            || forLoop.condition.kind !== ts.SyntaxKind.BinaryExpression
-            || forLoop.condition.getChildAt(0).getText() !== indexVariableName
-            || forLoop.condition.getChildAt(1).getText() !== "<") {
+    // ensure `for` condition
+    if (!indexVariableName
+        || !forLoop.condition
+        || forLoop.condition.kind !== ts.SyntaxKind.BinaryExpression
```
ajafff:
```ts
!isBinaryExpression(forLoop.condition) ||
!isVariable(forLoop.condition.left, indexVariableName) || // see my other comments regarding indexVariableName
forLoop.condition.operatorToken.kind !== ts.SyntaxKind.LessThanToken ||
// condition from line 159
!isPropertyAccessExpression(forLoop.condition.right) ||
forLoop.condition.right.name.text !== "length"
```

src/rules/preferForOfRule.ts
```diff
@@ -36,168 +36,177 @@ export class Rule extends Lint.Rules.AbstractRule {
...
-            if (incrementorState.onlyArrayReadAccess) {
-                this.addFailureFromStartToEnd(node.getStart(), incrementorState.forLoopEndPosition, Rule.FAILURE_STRING);
-            }
+        const { indexVariable, arrayToken } = arrayNodeInfo;
+        const indexVariableName = indexVariable.text;
+
+        // store `for` loop state
+        const state: IncrementorState = {
+            indexVariableName,
+            arrayToken: arrayToken as ts.Identifier,
```
ajafff: `arrayToken` is not guaranteed to be `ts.Identifier`

src/rules/preferForOfRule.ts
```diff
@@ -36,168 +36,177 @@ export class Rule extends Lint.Rules.AbstractRule {
...
-                    }
+function updateIncrementorState(node: ts.Identifier, state: IncrementorState): void {
+    // check if iterator is used for something other than reading data from array
+    if (node.parent!.kind === ts.SyntaxKind.ElementAccessExpression) {
+        const elementAccess = node.parent as ts.ElementAccessExpression;
+        const arrayIdentifier = unwrapParentheses(elementAccess.expression) as ts.Identifier;
+        if (state.arrayToken.getText() !== arrayIdentifier.getText()) {
+            // iterator used in array other than one iterated over
+            state.onlyArrayReadAccess = false;
+        } else if (elementAccess.parent && isAssignment(elementAccess.parent)) {
```
ajafff: `elementAccess.parent` is never undefined, so you can simply assert it

ajafff:
AFAICT the rule will bail out if you use an identifier with the same name as the index variable **somewhere** inside the loop body. That's no blocker for now as the old implementation behaved the same I guess.
```ts
for (let i = 0; i < arr.length; ++i) {
    {
        let i; // i is not used in element access, the rule bails out
    }
    class Foo {
        private i; // same as above
        constructor(i: number) {} // same as above
    }
    let foo: { i: number } = { i: 1 }; // same as above for both identifiers
}
```

Issue #2281, opened 2017-03-01
radix lint unintentionally matches any function named parseInt


evmar:
### Bug Report

- __TSLint version__: 4.4.2
- __TypeScript version__: 2.1.6
- __Running TSLint via__: cli

#### TypeScript code being linted

```ts
namespace my {
  function parseInt(x: string) { ... }
}
my.parseInt('a');
```

with `tslint.json` configuration:

```json
    "radix": true,
```

#### Actual behavior
[Lint: tslint] Missing radix parameter

#### Expected behavior
Don't warn on parseInt unless it's the builtin one.

alexsapps:
Sorry if obvious but just to be sure, it should work in this form too:
```
import {parseInt} from ...;
parseInt('a');
```

adidahiya:
This is nontrivial because checking if it's the `parseInt` from lib.d.ts will require using the type checker. We could change the behavior to be act differently if this rule is run with type checking enabled, but for the most part we value fast syntactic checks that are accurate _most_ of the time over slow semantic checks that are accurate 100% of the time.

Related: https://github.com/palantir/tslint/issues/2091

evmar:
I think syntactically you could just catch whether parseInt is a property access or not to cover 99% of cases.  It's pretty confusing to import a parseInt that shadows the global one.  I think this came up in some code where a parser class had `.parseString` and `.parseInt` and so on.

ajafff:
#2352 fixes the problem with property access / method call

Shadowing `parseInt` with another function will still result in a failure (like the import example above). But that seems to be overkill for this rule. Other rules like `no-eval` also don't track shadowed indentifiers.

PR #2352, opened 2017-03-15
Bugfix for radix rule and property access

ajafff:
#### PR checklist

- [x] Addresses an existing issue: #2281 
  - [x] Includes tests
- [ ] Documentation update

#### Overview of change:
[bugfix] don't warn for missing radix on method calls
[enhancement] added check for global.parseInt and window.parseInt

### DIFF
--- a/src/rules/radixRule.ts
+++ b/src/rules/radixRule.ts
@@ -15,6 +15,7 @@
  * limitations under the License.
  */
 
+import { isCallExpression, isIdentifier, isPropertyAccessExpression } from "tsutils";
 import * as ts from "typescript";
 
 import * as Lint from "../index";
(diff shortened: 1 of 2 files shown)


andy-hanson:
I wouldn't mark this as fixing #2281 since fixing that would require making this an OptionallyTypedRule.

ajafff:
You are right, that it does not fix all cases.
I wouldn't make it an OptionallyTypedRule, because it would not fail in CLI but fail in vscode, which is really annoying.

Also other rules like this (no-eval, no-construct, array-type, ...) don't check if the identifier is shadowed. That's what line switches are made for.

If the current behavior is really a problem (awating more feedback on this), we could use some code from `no-shadowed-variable` (I'm currently working on a reusable utility)

PR #2354, opened 2017-03-16
Rewrite no-duplicate-variable

andy-hanson:
#### PR checklist

  - [ ] Includes tests
- [ ] Documentation update

#### Overview of change:

Just refactored the implementation.

### DIFF
--- a/src/rules/noDuplicateVariableRule.ts
+++ b/src/rules/noDuplicateVariableRule.ts
@@ -15,6 +15,7 @@
  * limitations under the License.
  */
 
+import * as utils from "tsutils";
 import * as ts from "typescript";
 
 import * as Lint from "../index";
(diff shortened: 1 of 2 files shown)


src/rules/noDuplicateVariableRule.ts
```diff
@@ -38,73 +39,56 @@ export class Rule extends Lint.Rules.AbstractRule {
...
+function forEachBoundIdentifier(name: ts.BindingName, action: (id: ts.Identifier) => void): void {
+    switch (name.kind) {
+        case ts.SyntaxKind.Identifier:
+            return action(name);
+        case ts.SyntaxKind.ObjectBindingPattern:
+            for (const e of name.elements) {
+                forEachBoundIdentifier(e.name, action);
+            }
+            break;
+        case ts.SyntaxKind.ArrayBindingPattern:
```
ajafff: you can make this case clause the `default` clause and handle both binding patterns to avoid duplicated code since both blocks are very similar. you can then drop the `case ObjectBindingPattern`

src/rules/noDuplicateVariableRule.ts
```diff
@@ -38,73 +39,56 @@ export class Rule extends Lint.Rules.AbstractRule {
...
-    public visitVariableDeclaration(node: ts.VariableDeclaration) {
-        const isSingleVariable = node.name.kind === ts.SyntaxKind.Identifier;
-
-        // destructuring is handled by this.visitBindingElement()
-        if (isSingleVariable && !Lint.isBlockScopedVariable(node)) {
-            this.handleSingleVariableIdentifier(node.name as ts.Identifier);
+        if (utils.isVariableDeclaration(node) && !utils.isBlockScopedVariableDeclaration(node)) {
+            const scope = scopes[scopes.length - 1];
+            forEachBoundIdentifier(node.name, (id) => {
+                const { text } = id;
```
ajafff: destructure the parameter directly
```ts
({text}) => { ... }
```

src/rules/noDuplicateVariableRule.ts
```diff
@@ -38,73 +39,56 @@ export class Rule extends Lint.Rules.AbstractRule {
...
-        const isBlockScoped = Lint.isBlockScopedBindingElement(node);
-
-        // duplicate-variable errors for block-scoped vars are caught by tsc
-        if (isSingleVariable && !isBlockScoped) {
-            this.handleSingleVariableIdentifier(node.name as ts.Identifier);
+function walk(ctx: Lint.WalkContext<void>): void {
+    const scopes: Array<Set<string>> = [new Set()];
+    return ts.forEachChild(ctx.sourceFile, function cb(node: ts.Node): void {
+        if (utils.isFunctionScopeBoundary(node)) {
+            scopes.push(new Set());
```
ajafff: you don't need a stack. just replace `scopes` with `scope`. Save it in a variable here, assign the new Scope and restore it 2 lines below.

src/rules/noDuplicateVariableRule.ts
andy-hanson: It's jused as a node below.

andy-hanson:
There's a comment in the tests saying that parameters should be handled by `no-shadowed-variable`.

PR #2355, opened 2017-03-16
Refactor variable-name rule

andy-hanson:
#### PR checklist

  - [ ] Includes tests
- [ ] Documentation update

#### Overview of change:

Just a refactor.

### DIFF
--- a/src/rules/variableNameRule.ts
+++ b/src/rules/variableNameRule.ts
@@ -21,8 +21,9 @@ import * as ts from "typescript";
 
 import * as Lint from "../index";
 
-const BANNED_KEYWORDS = new Set(["any", "Number", "number", "String", "string", "Boolean", "boolean", "Undefined", "undefined"]);
-const bannedKeywordsStr = Array.from(BANNED_KEYWORDS).map((kw) => `\`${kw}\``).join(", ");
+const BANNED_KEYWORDS = ["any", "Number", "number", "String", "string", "Boolean", "boolean", "Undefined", "undefined"];
+const bannedKeywordsSet = new Set(BANNED_KEYWORDS);
+const bannedKeywordsStr = BANNED_KEYWORDS.map((kw) => `\`${kw}\``).join(", ");
 
 const OPTION_LEADING_UNDERSCORE = "allow-leading-underscore";
 const OPTION_TRAILING_UNDERSCORE = "allow-trailing-underscore";
@@ -67,7 +68,7 @@ export class Rule extends Lint.Rules.AbstractRule {
     public static KEYWORD_FAILURE = "variable name clashes with keyword/type";
 
     public apply(sourceFile: ts.SourceFile): Lint.RuleFailure[] {
-        return this.applyWithFunction(sourceFile, (ctx) => walk(ctx, parseOptions(this.getOptions().ruleArguments)));
+        return this.applyWithFunction<Options>(sourceFile, walk, parseOptions(this.ruleArguments));
     }
 }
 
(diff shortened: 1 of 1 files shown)


src/rules/variableNameRule.ts
```diff
@@ -15,11 +15,14 @@
...
 
+// tslint:disable object-literal-sort-keys
+
 import * as ts from "typescript";
 
 import * as Lint from "../index";
 
-const BANNED_KEYWORDS = ["any", "Number", "number", "String", "string", "Boolean", "boolean", "Undefined", "undefined"];
+const BANNED_KEYWORDS = new Set(["any", "Number", "number", "String", "string", "Boolean", "boolean", "Undefined", "undefined"]);
+const bannedKeywordsStr = Array.from(BANNED_KEYWORDS).map((kw) => `\`${kw}\``).join(", ");
```
ajafff: why initialize a Set with an array and the convert it back to an array? Wouldn't it make more sense to just use the same array for both?
Also IIRC a set does not guarantee any order, which could separate `Number` and `number` in the output for example.

src/rules/variableNameRule.ts
```diff
@@ -60,130 +62,136 @@ export class Rule extends Lint.Rules.AbstractRule {
...
+}
 
-    private isCamelCase(name: string) {
-        const firstCharacter = name.charAt(0);
-        const lastCharacter = name.charAt(name.length - 1);
-        const middle = name.substr(1, name.length - 2);
+function isCamelCase(name: string, options: Options): boolean {
+    const firstCharacter = name.charAt(0);
+    const lastCharacter = name.charAt(name.length - 1);
+    const middle = name.substr(1, name.length - 2);
```
ajafff: could be simplified to `name.slice(1, -1)`

src/rules/variableNameRule.ts
```diff
@@ -60,130 +62,136 @@ export class Rule extends Lint.Rules.AbstractRule {
...
+            return false;
     }
+}
 
-    private isCamelCase(name: string) {
-        const firstCharacter = name.charAt(0);
-        const lastCharacter = name.charAt(name.length - 1);
-        const middle = name.substr(1, name.length - 2);
+function isCamelCase(name: string, options: Options): boolean {
+    const firstCharacter = name.charAt(0);
```
ajafff: `str.chatAt(pos)` vs `str[pos]` is your personal preference?
I prefer the latter, because it is shorter. But that's totally nice to have.

src/rules/variableNameRule.ts
```diff
@@ -60,130 +62,136 @@ export class Rule extends Lint.Rules.AbstractRule {
...
-            return false;
-        }
-        if (!this.hasOption(OPTION_TRAILING_UNDERSCORE) && lastCharacter === "_") {
-            return false;
-        }
-        if (!this.hasOption(OPTION_ALLOW_PASCAL_CASE) && !isLowerCase(firstCharacter)) {
-            return false;
-        }
-        return middle.indexOf("_") === -1;
+    if (name.length <= 0) {
```
ajafff: I know that check was there before, but when would this ever be true?
Edit: function declarations without a name have an identifier with zero width, but that is invalid syntax anyway.

src/rules/variableNameRule.ts
```diff
@@ -60,130 +62,136 @@ export class Rule extends Lint.Rules.AbstractRule {
...
     };
-    /* tslint:enable:object-literal-sort-keys */
 
     public static FORMAT_FAILURE = "variable name must be in camelcase or uppercase";
     public static KEYWORD_FAILURE = "variable name clashes with keyword/type";
 
     public apply(sourceFile: ts.SourceFile): Lint.RuleFailure[] {
-        const variableNameWalker = new VariableNameWalker(sourceFile, this.getOptions());
-        return this.applyWithWalker(variableNameWalker);
+        return this.applyWithFunction(sourceFile, (ctx) => walk(ctx, parseOptions(this.getOptions().ruleArguments)));
```
ajafff: no need for `this.getOptions()`, use `this.ruleArguments`

ajafff: and as always, why not pass options as third parameter to `applyWithFunction` and access it via `ctx.options`?
You would save a closure and a (unnecessary) function call.

PR #1845, opened 2016-12-11
Simplify code for adding failures

andy-hanson:
#### PR checklist

- [ ] Addresses an existing issue: #0000
- [ ] New feature, bugfix, or enhancement
  - [ ] Includes tests
- [ ] Documentation update
- [X] Code improvement

#### What changes did you make?

This greatly simplifies how to add new rule failures, using just a few helper functions.
Only 2 rules remain that directly call `createFailure`; these two keep their own list of failures and do complex things with them.  All other rules were just creating a failure and immediately adding it, so those steps can be combined.

PR #2356, opened 2017-03-16
[Breaking change] Remove `createFix`, and just use the Replacement(s) as the fix.

andy-hanson:
#### PR checklist

  - [ ] Includes tests
- [ ] Documentation update

#### Overview of change:

The `Fix` class is not really necessary. It's easier to just pass parameters to `addFailure`.
This could be viewed as a continuation of #1845. So now instead of `this.addFailure(this.createFailure(..., this.createFix(foo)))`, you just `this.addFailure(..., foo)`.

#### CHANGELOG.md entry:

[api] Remove `createFix`. Replacements should be passed directly into `addFailure`.

### DIFF
--- a/src/language/rule/rule.ts
+++ b/src/language/rule/rule.ts
@@ -17,6 +17,7 @@
 
 import * as ts from "typescript";
 
+import {arrayify, flatMap} from "../../utils";
 import {IWalker} from "../walker";
 
 export interface IRuleMetadata {
@@ -132,14 +133,18 @@ export function isTypedRule(rule: IRule): rule is ITypedRule {
 }
 
 export class Replacement {
+    public static applyFixes(content: string, fixes: Fix[]): string {
+        return this.applyAll(content, flatMap(fixes, arrayify));
+    }
+
     public static applyAll(content: string, replacements: Replacement[]) {
         // sort in reverse so that diffs are properly applied
         replacements.sort((a, b) => b.end - a.end);
         return replacements.reduce((text, r) => r.apply(text), content);
     }
 
-    public static replaceNode(node: ts.Node, text: string): Replacement {
-        return this.replaceFromTo(node.getStart(), node.getEnd(), text);
+    public static replaceNode(node: ts.Node, text: string, sourceFile?: ts.SourceFile): Replacement {
+        return this.replaceFromTo(node.getStart(sourceFile), node.getEnd(), text);
     }
 
     public static replaceFromTo(start: number, end: number, text: string) {
@@ -158,23 +163,10 @@ export class Replacement {
         return new Replacement(start, 0, text);
     }
 
-    constructor(private innerStart: number, private innerLength: number, private innerText: string) {
-    }
-
-    get start() {
-        return this.innerStart;
-    }
-
-    get length() {
-        return this.innerLength;
-    }
+    constructor(readonly start: number, readonly length: number, readonly text: string) {}
 
     get end() {
-        return this.innerStart + this.innerLength;
-    }
-
-    get text() {
-        return this.innerText;
+        return this.start + this.length;
     }
 
     public apply(content: string) {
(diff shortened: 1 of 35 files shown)


src/language/rule/rule.ts
```diff
@@ -188,19 +192,10 @@ export class Fix {
...
     }
 
-    constructor(private innerRuleName: string, private innerReplacements: Replacement[]) {
-    }
-
-    get ruleName() {
-        return this.innerRuleName;
-    }
-
-    get replacements() {
```
andy-hanson: `get innerReplacements() { return this.innerInnerReplacements; }` 😜

src/language/rule/rule.ts
```diff
@@ -138,6 +138,10 @@ export class Replacement {
         return replacements.reduce((text, r) => r.apply(text), content);
     }
 
+    public static replaceNode(node: ts.Node, text: string): Replacement {
```
ajafff: please add an optional `sourceFile` parameter, that will be passed to `node.getStart` (for performance)

src/language/rule/rule.ts
```diff
@@ -240,19 +235,21 @@ export class RuleFailure {
...
                 private ruleName: string,
-                private fix?: Fix) {
+                fix?: Replacement | Replacement[]) {
 
         this.fileName = sourceFile.fileName;
         this.startPosition = this.createFailurePosition(start);
         this.endPosition = this.createFailurePosition(end);
         this.rawLines = sourceFile.text;
         this.ruleSeverity = "error";
+        this.fix = fix ? new Fix(Array.isArray(fix) ? fix : [fix]) : undefined;
```
ajafff: use `arrayify`?

src/language/walker/ruleWalker.ts
```diff
@@ -65,7 +65,7 @@ export class RuleWalker extends SyntaxWalker implements IWalker {
     }
 
     /** @deprecated Prefer `addFailureAt` and its variants. */
-    public createFailure(start: number, width: number, failure: string, fix?: Fix): RuleFailure {
+    public createFailure(start: number, width: number, failure: string, fix?: Replacement | Replacement[]): RuleFailure {
```
ajafff: I would prefer a rest parameter like `createFix` had

src/language/walker/walkContext.ts
```diff
@@ -17,31 +17,27 @@
...
     }
 
-    public addFailure(start: number, end: number, failure: string, fix?: Fix) {
+    public addFailure(start: number, end: number, failure: string, fix?: Replacement | Replacement[]) {
         const fileLength = this.sourceFile.end;
-        this.failures.push(
-            new RuleFailure(this.sourceFile, Math.min(start, fileLength), Math.min(end, fileLength), failure, this.ruleName, fix),
-        );
+        start = Math.min(start, fileLength);
+        end = Math.min(end, fileLength);
```
ajafff: Since you are already changing the style of the code, you could inline `this.sourceFile.end` here and remove `fileLength`

src/language/walker/ruleWalker.ts
andy-hanson: I tried that at first, but that can make life difficult if you have a function returning optional fixes. You would need to use `const fix = optionalFix(); ctx.addFailureAtNode(node, failure, ...(fix ? fix : []))`, instead of just `ctx.addFailureAtNode(node, failure, optionalFix())`.

src/language/rule/rule.ts
```diff
@@ -240,19 +212,21 @@ export class RuleFailure {
...
     private ruleSeverity: RuleSeverity;
+    private fix?: Fix;
 
     constructor(private sourceFile: ts.SourceFile,
                 start: number,
                 end: number,
                 private failure: string,
                 private ruleName: string,
-                private fix?: Fix) {
+                fix?: Fix) {
```
adidahiya: why this change?

src/rules/alignRule.ts
```diff
@@ -108,13 +108,9 @@ class AlignWalker extends Lint.RuleWalker {
...
             if (curPos.line !== prevPos.line && curPos.character !== alignToColumn) {
                 const diff = alignToColumn - curPos.character;
-                let fix;
-                if (0 < diff) {
-                    fix = this.createFix(this.appendText(node.getStart(), " ".repeat(diff)));
-                } else {
-                    fix = this.createFix(this.deleteText(node.getStart() + diff, -diff));
-                }
-                this.addFailureAtNode(node, kind + Rule.FAILURE_STRING_SUFFIX, fix);
+                this.addFailureAtNode(node, kind + Rule.FAILURE_STRING_SUFFIX, 0 < diff
```
adidahiya: IMO the inlined `fix` variable is hard to read, can you save it a variable before calling this method?

andy-hanson:
#2403 merged, so closing.

PR #2358, opened 2017-03-16
resolve rulesDirectory - fixes #2163

### LINKED ISSUE #2163: Feature Request: use node module resolution semantics for rulesDirectory
`extends` uses node module resolution semantics to find files and you can specify relative paths to files, too. I'd need the same behaviour for `rulesDirectory`. E.g. I want to use `tslint-react/rules` which is sometimes in `./node_modules/tslint-react/rules` and sometimes in `../node_modules/tslint-react/rules` in my case. (I build a generic build tool configuration similar to `create-react-app`, so I don't always have full control about the projects where my configuration is used..)

donaldpipowitch:
#### PR checklist

- [x] Addresses an existing issue: #2163

#### Overview of change:

When someone uses a `rulesDirectory` we try to resolve it with Node resolve logic first, before falling back to the old algorithm.

#### CHANGELOG.md entry:

[enhancement] `rulesDirectory` can now be resolved with [`dir-resolve`](https://www.npmjs.com/package/dir-resolve) (#2163)

### DIFF
--- a/src/configuration.ts
+++ b/src/configuration.ts
@@ -277,6 +277,12 @@ export function getRelativePath(directory?: string | null, relativeTo?: string) 
     return undefined;
 }
 
+// check if directory should be used as path or if it should be resolved like a module
+// matches if directory starts with '/', './', '../', 'node_modules/' or equals '.' or '..'
+export function useAsPath(directory: string) {
+    return /^(?:\.?\.?(?:\/|$)|node_modules\/)/.test(directory);
+}
+
 /**
  * @param directories A path(s) to a directory of custom rules
  * @param relativeTo A path that directories provided are relative to.
@@ -285,17 +291,25 @@ export function getRelativePath(directory?: string | null, relativeTo?: string) 
  * @return An array of absolute paths to directories potentially containing rules
  */
 export function getRulesDirectories(directories?: string | string[], relativeTo?: string): string[] {
-    const rulesDirectories = arrayify(directories)
-        .map((dir) => getRelativePath(dir, relativeTo))
+    return arrayify(directories)
+        .map((dir) => {
+            if (!useAsPath(dir)) {
+                try {
+                    return path.dirname(resolve.sync(dir, { basedir: relativeTo }));
+                } catch (err) {
+                    // swallow error and fallback to using directory as path
+                }
+            }
+
+            const absolutePath = getRelativePath(dir, relativeTo);
+            if (absolutePath != null) {
+                if (!fs.existsSync(absolutePath)) {
+                    throw new Error(`Could not find custom rule directory: ${dir}`);
+                }
+            }
+            return absolutePath;
+        })
         .filter((dir) => dir !== undefined) as string[];
-
-    for (const directory of rulesDirectories) {
-        if (directory != null && !fs.existsSync(directory)) {
-            throw new Error(`Could not find custom rule directory: ${directory}`);
-        }
-    }
-
-    return rulesDirectories;
 }
 
 /**
--- a/test/config/tslint-custom-rules-with-package-fallback.json
+++ b/test/config/tslint-custom-rules-with-package-fallback.json
@@ -1,3 +1,3 @@
 {
-  "rulesDirectory": ["../external/tslint-test-custom-rules/rules"]
+  "rulesDirectory": ["relative-rules-directory"]
 }
--- a/test/config/tslint-custom-rules-with-package-fallback.json
+++ b/test/config/tslint-custom-rules-with-package-fallback.json
@@ -1,0 +1,3 @@
+{
+  "rulesDirectory": ["../external/tslint-test-custom-rules/rules"]
+}
--- a/test/config/tslint-custom-rules-with-package.json
+++ b/test/config/tslint-custom-rules-with-package.json
@@ -1,0 +1,3 @@
+{
+  "rulesDirectory": ["tslint-test-custom-rules/rules"]
+}
--- a/test/configurationTests.ts
+++ b/test/configurationTests.ts
@@ -16,6 +16,7 @@
 
 import { assert } from "chai";
 import * as fs from "fs";
+import * as path from "path";
 
 import {
     convertRuleOptions,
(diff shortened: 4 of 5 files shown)


ajafff:
Would be great if this could work with the `main` field in `package.json`. Then you would only need to specify the package name in `rulesDirectory` and the `package.json` points to the right subdirectory.

This may conflict with the module resolution that is done for finding configs, though.

donaldpipowitch:
I'm not sure if this would be a good addition. AFAIK `main` should point to a file, not a directory.

donaldpipowitch:
Who could review this? @nchen63 or @ajafff? Anything I can do to help? :)

ajafff:
Sorry for the delayed response.
Changes LGTM, but my comment from above still stands.

> I'm not sure if this would be a good addition. AFAIK main should point to a file, not a directory.

I get your point, but then there is this commit by @JKillian which suggests adding an (empty) `index.js` to resolve to the correct directory.
https://github.com/buzinas/tslint-eslint-rules/commit/aa17d2e83a08a84b8635c9d205d77cf295a66492

That's how the configuration directory is resolved.
If `rulesDirectory` is resolved in the same way, rules and configs would need to be located in the same directory. But that's not too bad, I think.

As a bonus that wouldn't add another dependency, because we can simply use https://www.npmjs.com/package/resolve for both.

ajafff:
Additions to my last comment:

> rules and configs would need to be located in the same directory

that's not entirely true. you can still use `package/rules` although the rules directory also needs an index.js

@donaldpipowitch you should also check the path first before resolving as package. When the path starts with `/`, `./` or `../` or equals `.` or `..` you don't resolve but use it as a path.

donaldpipowitch:
Cool. Thank you both for the good feedback. I'll try to make the changes in the next days :)

donaldpipowitch:
I updated my PR.

The logic for resolving the directory is now inside `configuration.ts:getRulesDirectories` and not `ruleLoader.ts:loadCachedRule`. It looks like this one is the better place to do this...? `getRulesDirectories` is called before `loadCachedRule`.

> I get your point, but then there is this commit by @JKillian which suggests adding an (empty) index.js to resolve to the correct directory.
> As a bonus that wouldn't add another dependency, because we can simply use https://www.npmjs.com/package/resolve for both.
> @donaldpipowitch you should also check the path first before resolving as package. When the path starts with /, ./ or ../ or equals . or .. you don't resolve but use it as a path.

I did all that. I tested this with the `  "rulesDirectory": [ "tslint-eslint-rules/dist/rules" ],` and manually adding an empty `index.js` into `"tslint-eslint-rules/dist/rules"`. The rule was loaded correctly, but I got a warning when a linter rule failed (`Warning: this.createFix is not a function` - but it looks like this is a young breaking change https://github.com/palantir/tslint/pull/2403 and not fixed by `tslint-eslint-rule` yet).

> And there need to be tests for this new feature.

Can you give me more guidance to test this? Do I need to mock a 3rd party node module to do that correctly?

ajafff:
> Do I need to mock a 3rd party node module to do that correctly?

There is already. `configurationTests.ts` uses two packages for testing, see https://github.com/palantir/tslint/blob/master/test/config/package.json
You can probably use it for your test, too.

src/configuration.ts
```diff
@@ -277,6 +277,12 @@ export function getRelativePath(directory?: string | null, relativeTo?: string)
     return undefined;
 }
 
+// check if directory should be used as path or if it should be resolved like a module
+export function useAsPath(directory: string) {
+    return directory.startsWith("/") || directory.startsWith("./") || directory.startsWith("../")
```
ajafff: this should probably include paths starting with `node_modules/`

to simplify the check, you can use this regex:
```ts
/^(?:\.?\.?(?:\/|$)|node_modules\/)/.test(directory)
```

src/configuration.ts
```diff
@@ -285,17 +291,26 @@ export function getRelativePath(directory?: string | null, relativeTo?: string)
...
+                if (absolutePath != null) {
+                    if (!fs.existsSync(absolutePath)) {
+                        throw new Error(`Could not find custom rule directory: ${dir}`);
+                    }
+                }
+            } else {
+                try {
+                    absolutePath = resolve.sync(dir, { basedir: relativeTo }).replace("index.js", "");
+                } catch (err) {
+                    throw new Error(`Could not find custom rule directory: ${dir}`);
```
ajafff: instead of throwing here, it should fall back using path as directory

src/configuration.ts
```diff
@@ -285,17 +291,26 @@ export function getRelativePath(directory?: string | null, relativeTo?: string)
...
+            if (useAsPath(dir)) {
+                absolutePath = getRelativePath(dir, relativeTo);
+                if (absolutePath != null) {
+                    if (!fs.existsSync(absolutePath)) {
+                        throw new Error(`Could not find custom rule directory: ${dir}`);
+                    }
+                }
+            } else {
+                try {
+                    absolutePath = resolve.sync(dir, { basedir: relativeTo }).replace("index.js", "");
```
ajafff: just to be safe, because the path *may* include index.js earlier: `.replace(/index.js$/, "")`

src/configuration.ts
```diff
@@ -285,17 +291,31 @@ export function getRelativePath(directory?: string | null, relativeTo?: string)
  * @return An array of absolute paths to directories potentially containing rules
  */
 export function getRulesDirectories(directories?: string | string[], relativeTo?: string): string[] {
-    const rulesDirectories = arrayify(directories)
-        .map((dir) => getRelativePath(dir, relativeTo))
+    return arrayify(directories)
+        .map((dir) => {
+            let absolutePath;
+            if (useAsPath(dir)) {
```
ajafff: to avoid code duplication, consider writing it as follows:

```ts
if (!useAsPath(dir)) {
    try {
        return resolve.sync(...).replace(...);
    } catch (err) {
        // swallow error and fallback to using directory as path
    }
}
const absolutePath = getRelativePath(...);
if (absolutePath != null) {
    ...
}
return absolutePath;
```

donaldpipowitch: :+1:

donaldpipowitch:
Is this fine now?

test/config/tslint-custom-rules-with-package.json
```diff
@@ -0,0 +1,3 @@
+{
+  "rulesDirectory": ["tslint-test-custom-rules/rules"]
```
donaldpipowitch: I tried to add a test, but didn't use `"test/external/tslint-test-custom-rules/rules"`, because it would throw. I'm not sure, this works now as you expected...?

test/config/tslint-custom-rules-with-package-fallback.json
```diff
@@ -0,0 +1,3 @@
+{
+  "rulesDirectory": ["../external/tslint-test-custom-rules/rules"]
```
ajafff: Actually I wanted a test where the path was relative but did not start with `./` or `../`
Consider adding a folder `test/config/relativeRulesDirectory` and change this line to `"rulesDirectory": ["relativeRulesDirectory"]`

test/configurationTests.ts
```diff
@@ -252,6 +252,11 @@ describe("Configuration", () => {
             assert.equal("off", config.rules.get("no-eval")!.ruleSeverity);
         });
 
+        it("resolve rule directory from package", () => {
+            assert.doesNotThrow(() => loadConfigurationFromPath("./test/config/tslint-custom-rules-with-package.json"));
+            assert.doesNotThrow(() => loadConfigurationFromPath("./test/config/tslint-custom-rules-with-package-fallback.json"));
```
ajafff: consider checking if it resolved to the expected `rulesDirectory` in both cases

donaldpipowitch:
@ajafff Thank you for reviewing this :) I updated the tests. I think it looks good now.

PR #2359, opened 2017-03-16
Refactor some rules

ajafff:
#### PR checklist

  - [x] Includes tests
- [ ] Documentation update

#### Overview of change:

Refactored some of the rather simple rules to `WalkContext` or `AbstractWalker`.
`no-trailing-whitespace` is the only rule where the logic changes: now it looks for trailing whitespace in the source text and iff there are lines found, it searches the AST for template spans and comments. 
[rule-change] `no-trailing-whitespace` now checks template strings by default. Use the new options `ignore-template-strings` to restore the old behavior.
[new-rule-option] added `ignore-template-strings` to `no-trailing-whitespace`
[enhancement] fixer of `quotemark` unescapes original quotemark (e.g. `'\''` -> `"'"`)

Refactoring is not done yet, but the perf improvement already looks pretty good: (linting this project with itself):
master: 7.7s
this PR: 6.7s

### DIFF
--- a/src/rules/noInternalModuleRule.ts
+++ b/src/rules/noInternalModuleRule.ts
@@ -15,6 +15,7 @@
  * limitations under the License.
  */
 
+import { getChildOfKind } from "tsutils";
 import * as ts from "typescript";
 
 import * as Lint from "../index";
@@ -36,35 +37,41 @@ export class Rule extends Lint.Rules.AbstractRule {
     public static FAILURE_STRING = "The internal 'module' syntax is deprecated, use the 'namespace' keyword instead.";
 
     public apply(sourceFile: ts.SourceFile): Lint.RuleFailure[] {
-        return this.applyWithWalker(new NoInternalModuleWalker(sourceFile, this.getOptions()));
+        return this.applyWithWalker(new NoInternalModuleWalker(sourceFile, this.ruleName, undefined));
     }
 }
 
-class NoInternalModuleWalker extends Lint.RuleWalker {
-    public visitModuleDeclaration(node: ts.ModuleDeclaration) {
-        if (this.isInternalModuleDeclaration(node)) {
-            const start = this.getStartBeforeModule(node);
-            this.addFailureAt(node.getStart() + start, "module".length, Rule.FAILURE_STRING);
+class NoInternalModuleWalker extends Lint.AbstractWalker<void> {
+    public walk(sourceFile: ts.SourceFile) {
+        return this.checkStatements(sourceFile.statements);
+    }
+
+    private checkStatements(statements: ts.Statement[]) {
+        for (const statement of statements) {
+            if (statement.kind === ts.SyntaxKind.ModuleDeclaration) {
+                this.checkModuleDeclaration(statement as ts.ModuleDeclaration);
+            }
         }
-        super.visitModuleDeclaration(node);
     }
 
-    private isInternalModuleDeclaration(node: ts.ModuleDeclaration) {
-        // an internal module declaration is not a namespace or a nested declaration
-        // for external modules, node.name.kind will be a LiteralExpression instead of Identifier
-        return !Lint.isNodeFlagSet(node, ts.NodeFlags.Namespace)
-            && !Lint.isNestedModuleDeclaration(node)
-            && node.name.kind === ts.SyntaxKind.Identifier
-            && !isGlobalAugmentation(node);
+    private checkModuleDeclaration(node: ts.ModuleDeclaration, nested?: boolean): void {
+        if (!nested &&
+            node.name.kind === ts.SyntaxKind.Identifier &&
+            // augmenting global uses a special syntax that is allowed
+            // see https://github.com/Microsoft/TypeScript/pull/6213
+            node.name.text !== "global") {
+            const moduleKeyword = getChildOfKind(node, ts.SyntaxKind.ModuleKeyword, this.sourceFile);
+            if (moduleKeyword) {
+                this.addFailureAtNode(moduleKeyword, Rule.FAILURE_STRING);
+            }
+        }
+        if (node.body !== undefined) {
+            switch (node.body.kind) {
+                case ts.SyntaxKind.ModuleBlock:
+                    return this.checkStatements(node.body.statements);
+                case ts.SyntaxKind.ModuleDeclaration:
+                    return this.checkModuleDeclaration(node.body, true);
+            }
+        }
     }
-
-    private getStartBeforeModule(node: ts.ModuleDeclaration) {
-        return node.getText().indexOf("module");
-    }
-}
-
-function isGlobalAugmentation(node: ts.ModuleDeclaration) {
-    // augmenting global uses a special syntax that is allowed
-    // see https://github.com/Microsoft/TypeScript/pull/6213
-    return node.name.kind === ts.SyntaxKind.Identifier && node.name.text === "global";
 }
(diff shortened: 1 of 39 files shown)


jmlopez-rod:
@ajafff how do you get the timings? Are you taking averages of runs or just a sample? I'm trying to see what's the best way of measuring performance on my rules so that I can change them to using the abstract walker but I can't really get a hold of a number.

src/rules/noInternalModuleRule.ts
```diff
@@ -36,35 +37,41 @@ export class Rule extends Lint.Rules.AbstractRule {
     public static FAILURE_STRING = "The internal 'module' syntax is deprecated, use the 'namespace' keyword instead.";
 
     public apply(sourceFile: ts.SourceFile): Lint.RuleFailure[] {
-        return this.applyWithWalker(new NoInternalModuleWalker(sourceFile, this.getOptions()));
+        return this.applyWithWalker(new NoInternalModuleWalker(sourceFile, this.ruleName, undefined));
```
andy-hanson: Make `options` an optional parameter instead of using `undefined`

src/rules/noInternalModuleRule.ts
```diff
@@ -36,35 +37,41 @@ export class Rule extends Lint.Rules.AbstractRule {
...
 
-    private isInternalModuleDeclaration(node: ts.ModuleDeclaration) {
-        // an internal module declaration is not a namespace or a nested declaration
-        // for external modules, node.name.kind will be a LiteralExpression instead of Identifier
-        return !Lint.isNodeFlagSet(node, ts.NodeFlags.Namespace)
-            && !Lint.isNestedModuleDeclaration(node)
-            && node.name.kind === ts.SyntaxKind.Identifier
-            && !isGlobalAugmentation(node);
+    private checkStatements(statements: ts.Statement[]) {
+        for (const statement of statements) {
```
andy-hanson: `module` can appear anywhere, so you have to traverse everything.

src/rules/noInternalModuleRule.ts
andy-hanson: Never mind, it doesn't seem like there's any way to do this in a type-safe way.

src/rules/noTrailingWhitespaceRule.ts
```diff
@@ -72,53 +73,49 @@ export class Rule extends Lint.Rules.AbstractRule {
...
-    let lastSeenWasWhitespace = false;
-    let lastSeenWhitespacePosition = 0;
-    utils.forEachTokenWithTrivia(ctx.sourceFile, (fullText, kind, range) => {
-        if (kind === ts.SyntaxKind.NewLineTrivia || kind === ts.SyntaxKind.EndOfFileToken) {
-            if (lastSeenWasWhitespace) {
-                reportFailure(ctx, lastSeenWhitespacePosition, range.pos);
+    const possibleFailures: ts.TextRange[] = [];
+    const sourceFile = ctx.sourceFile;
+    const text = sourceFile.text;
+    for (const range of getLineRanges(sourceFile)) {
```
andy-hanson: Can do it with just a regexp:
```ts
const rgx = /[^\S\n]+(\r?\n|$)/g; // Non-newline spaces, then a newline or EOF
let match: RegExpExecArray | null;
while (match = rgx.exec(text)) {
    const end = text.indexOf("\n", match.index);
    possibleFailures.push({
        pos: match.index,
        end: end === -1 ? text.length : end
    });
}
```

ajafff: Your solution looks nicer at first glance. It seems to include `\r` in the range of the failure, though.

I just added a new property `contentLength` to `getLineRanges` which has already stripped `\r?\n`. The regex is now even simpler.

ajafff:
@jmlopez-rod The easy way is to execute `yarn lint:from-bin` (or something similar like `time bin/tslint ...`) multiple times the get the average. Most of the time the measured times don't differ much. Then build the revision you want to compare to and execute it again.

When you want to measure the time without I/O, you need to use the profiler in node inspector: `node --debug-brk --inspect path/to/tslint`

src/rules/noTrailingWhitespaceRule.ts
andy-hanson: It can be done without `indexOf` because enough information is returned by `exec`. Then it's easier to use `len` instead of `end`, so you end up with:

```ts
function walk(ctx: Lint.WalkContext<IgnoreOption>) {
    const { options, sourceFile } = ctx;
    const possibleFailures = getPossibleFailures(sourceFile.text);
    if (possibleFailures.length === 0) {
        return;
    }

    const excludedRanges = options === IgnoreOption.None
                           ? getTemplateRanges(sourceFile)
                           : getExcludedRanges(sourceFile, options);
    for (const { pos, len } of possibleFailures) {
        if (!excludedRanges.some((range) => range.pos < pos && pos < range.end)) {
            ctx.addFailureAt(pos, len, Rule.FAILURE_STRING, ctx.createFix(Lint.Replacement.deleteText(pos, len)));
        }
    }
}

function getPossibleFailures(text: string): Array<{ pos: number, len: number }> {
    const result = [];
    const rgx = /[^\S\n]+(\r?\n|$)/g;
    let match: RegExpExecArray | null;
    while (match = rgx.exec(text)) {
        const matched = match[0];
        result.push({ pos: match.index, len: matched.endsWith("\n") ? matched.length - 1 : matched.length });
    }
    return result;
}
```

andy-hanson: Also, could this rule have an option to *not* ignore template strings?

src/rules/noTrailingWhitespaceRule.ts
```diff
@@ -72,53 +73,48 @@ export class Rule extends Lint.Rules.AbstractRule {
...
-            if (lastSeenWasWhitespace) {
-                reportFailure(ctx, lastSeenWhitespacePosition, range.pos);
+    const possibleFailures: ts.TextRange[] = [];
+    const sourceFile = ctx.sourceFile;
+    const text = sourceFile.text;
+    for (const line of getLineRanges(sourceFile)) {
+        const match = text.substr(line.pos, line.contentLength).match(/\s+$/);
+        if (match !== null) {
+            possibleFailures.push({
+                end: line.pos + line.contentLength,
```
andy-hanson: This is why I don't like `object-literal-sort-keys`: It doesn't make sense to have the keys in alphabetical order here.

ajafff:
@andy-hanson I changed `no-trailing-whitespace` to check template strings by default and added an option to ignore them.

yarn.lock
```diff
@@ -1039,10 +1039,14 @@ tslint@latest:
     tsutils "^1.1.0"
     update-notifier "^2.0.0"
 
-tsutils@^1.1.0, tsutils@^1.2.2:
+tsutils@^1.1.0:
   version "1.2.2"
   resolved "https://registry.yarnpkg.com/tsutils/-/tsutils-1.2.2.tgz#7e165c601367b9f89200b97ff47d9e38d1a6e4c8"
 
+tsutils@^1.4.0:
```
adidahiya: multiple tsutils again :( 

I've noticed yarn do this often lately when I try to upgrade a package (seems like a bug). I've usually fixed it by deleting all of the entries for a particular package (`tsutils` here) and running `yarn` again; it then resolves them to a single version.

adidahiya:
oh, and yarn.lock has a conflict anyway.

otherwise the PR looks good to go

ajafff:
@adidahiya conflicts in yarn.lock are resolved now

PR #2368, opened 2017-03-18
Add 'encoding' rule

andy-hanson:
#### PR checklist

  - [X] Includes tests
- [X] Documentation update

#### Overview of change:

Added the `encoding` rule, which checks each file's encoding.

This is a tricky one as it performs additional I/O, synchronously. It only reads 3 bytes but there may be lag.
The test runner had to be tweaked so that we would be reading the right file. I tested this rule out (using `npm link`) in both regular and non-type-checked scenarios and it worked, but it's possible that something like gulp-tslint could break with this.


src/test.ts
```diff
@@ -25,6 +25,7 @@ import * as ts from "typescript";
 
 import {Fix} from "./language/rule/rule";
 import * as Linter from "./linter";
+import {readBufferWithDetectedEncoding} from "./rules/encodingRule";
```
nchen63: should go in utils

Issue #2279, opened 2017-03-01
Add default severity level in config


nchen63:
Need to provide a default severity level in the config file in case people want to set most rules to "warning" (we currently default to "error")

Questions to answer:
- does a new default severity just override the old default or does it wipe out all of the inherited custom levels?
  - if it doesn't wipe out inherited severity for each rule, could be annoying to have to override every rule individually
  - if it does wipe out everything, might be difficult to build a nuanced rule set
- should there be a rule level severity `default`?
  - this would allow base ruleset authors to provide flexibility
  - could add unnecessary complexity

Issue #1330, opened 2016-06-22
Identify lines where rules are unnecessarily disabled


adidahiya:
Idea: report errors where a `tslint:disable` comment has no effect (probably as a result of some changes in the linter which make it more lax, or some new configuration options)

JKillian:
Neat idea, seems useful! I think we'd want this to be an option that could be turned on or off

adidahiya:
Yep, I think this could be a boolean flag in `tslint.json`.

adidahiya:
from @andy-hanson: would be best to wait until #2369 is in first

victornoel:
@adidahiya @JKillian @andy-hanson now that #2369 has been merged, is there some plan to support this? :)

adidahiya:
No active work is being done here right now, but it's up for grabs if anyone wants to write a brief proposal and send a PR.

JoshuaKGoldberg:
As mentioned by @reduckted in #4118:

> ESLint has this option. Obviously it's a different tool with different infrastructure, but we might be able to get some inspiration from how it's implemented.

> > ### [--report-unused-disable-directives](https://eslint.org/docs/user-guide/command-line-interface#--report-unused-disable-directives)
> > This option causes ESLint to report directive comments like // eslint-disable-line when no errors would have been reported on that line anyway. This can be useful to prevent future errors from unexpectedly being suppressed, by cleaning up old eslint-disable comments which are no longer applicable.

lukelafountaine:
Is anyone currently working on this? Also, what does the proposal process entail?

JoshuaKGoldberg:
@lukelafountaine nobody's working on this right now; you're more then welcome to! IMO this would be a pretty useful feature.

The definition of a "proposal" is pretty vague, but in general we'd want an explanation of how the feature works from a user's perspective. What are some cases where it would complain? What are some cases where it wouldn't?

Going off of ESLint's `--report-unused-disable-directives` seems like a good start. How would that look for TSLint if so?

adidahiya:
can we make it a core lint rule with an auto fixer so that this change rolls out in `tslint:latest` without users having to change how they use the tslint API?

JoshuaKGoldberg:
From a technical perspective, can we? Is there a way for rules to have access to that info (after all other rules run)?

ajafff:
[...]

PR #2369, opened 2017-03-18
Rewrite enable/disable logic

andy-hanson:
#### PR checklist

  - [X] Includes tests
- [ ] Documentation update

#### Overview of change:

* Move `disabledIntervals` out of rule options. Rules are not responsible for disabling theirselves. Just filter the failures after the rule returns them.
  - While doing this, I removed `filterFailures` entirely, removing code that looked for duplicate failures. It turns out a few rules were returning duplicate failures, so I changed them directly.
* Don't expose `disabledIntervals` at all. Just have a `RuleDisabler` interface and a method for getting one from the disable comments in a source file.
  - Therefore, we can delete `IDisabledInterval`.

There were conflicting tests as to whether explicitly enabling a rule turned off in tslint.json would enable it. I chose that it would. If you don't want that, the `isExplicitlyEnabled` method could be removed.

#### CHANGELOG.md entry:

[api] Remove `disabledIntervals` from Rule options. Rules are no longer responsible for this; their failures will be filtered by tslint.


src/language/rule/rule.ts
```diff
@@ -92,19 +95,12 @@ export interface IOptions {
...
-export interface IDisabledInterval {
-    startPosition: number;
-    endPosition: number;
 }
 
 export interface IRule {
     getOptions(): IOptions;
     isEnabled(): boolean;
     apply(sourceFile: ts.SourceFile): RuleFailure[];
-    applyWithWalker(walker: IWalker): RuleFailure[];
```
andy-hanson: This method was never necessary to have here.

src/ruleLoader.ts
```diff
@@ -43,23 +38,19 @@ export function loadRules(ruleOptionsList: IOptions[],
 
     for (const ruleOptions of ruleOptionsList) {
         const ruleName = ruleOptions.ruleName;
-        const enableDisableRules = enableDisableRuleMap.get(ruleName);
-        if (ruleOptions.ruleSeverity !== "off" || enableDisableRuleMap) {
```
andy-hanson: It's not necessary to check `ruleSeverity` here since `isEnabled()` will do that too.

test/rules/_integration/enable-disable/test.ts.lint
```diff
@@ -78,9 +78,6 @@ var AAAaA = 'test' // previously both `quotemark` and `variable-name` rules were
 var AAAaA = 'test' // tslint:disable-line:quotemark
 var AAAaA = 'test' // ensure that disable-line rule correctly handles previous `enable rule - disable all` switches
 
-// tslint:enable:no-var-keyword
```
andy-hanson: This conflicts with an assertion in `ruleLoaderTests.ts` that it "loads disabled rules if rule in enableDisableRuleMap". (It also now conflicts with `enable-via-comment`.) Which should we choose?

src/ruleLoader.ts
ajafff: technically you are right, but this check was added to do less work for disabled rules. With your change you need to find the rule in rulesDirectory and instantiate it before you know if you will ever need it (for every linted file).

test/rules/_integration/enable-disable/test.ts.lint
ajafff: The behaviour to not load disabled rules was added recently. Even though the rule was loaded, it would have never been executed, because `rule.isEnabled()` always returned false. I don't know why the test was implemented this way...
I think it makes more sense to keep disabled rules disabled. Otherwise you need to guess the `ruleSeverity`

src/ruleLoader.ts
andy-hanson: Isn't the real problem that we re-load every rule for every file? The only SourceFile-specific things here are `ruleDisabler` and `isJs`. Why not eagerly load the rules up-front (including disabled ones and TypeScript-only ones)?

test/rules/_integration/enable-disable/test.ts.lint
andy-hanson: What if someone wants a rule disabled by default but enabled in a particular file?

src/ruleLoader.ts
ajafff: If you use tslint as a library, you can pass different rule configs for every call to `Linter.lint`.
That's why this needs to be done for every file.
On the other hand, I don't know if that makes sense to keep that "feature".

test/rules/_integration/enable-disable/test.ts.lint
ajafff: I though about that too. That's a valid use case. But the problem with the missing severity still stands.
If you could enable/disable a rule and set its severity independently, that wouldn't be a problem.
Or just use a default severity (#2279)

andy-hanson: I have a test case `enable-via-comment` where a rule is turned off, but is enabled in the file and gives errors there. What other test cases would need to work?

andy-hanson: On that topic: All of the custom implementations of `isEnabled()` I see are just turning it off in the case of bad options. We should instead throw an error on bad options. I think just like Rules are no longer responsible for tracking their disabled ranges, they shouldn't be responsible for tracking whether they're enabled or not. So we should remove `isEnabled()` from the `Rule` interface. Would that help solve this problem?

ajafff:
To summarize both inline discussions:
`Linter` needs to be refactored. It should get the rule config only once in the constructor instead of every call to `Linter.lint`. We can then load all rules (even disabled ones to enable them later) upfront to save I/O.
We could / should also validate or parse the options for every rule (related: #1950) instead of doing this partially in some rule's `isEnabled` method. This needs to be done to ensure a rule is properly configured when enabled via comment.

test/rules/_integration/enable-disable/test.ts.lint
ajafff: What I'm actually trying to say: You disable a rule by setting `severity: off`. Now you turn it on via comment. What severity should this rule have? How do I set a specific severity other than the default?

src/rules/maxFileLineCountRule.ts
```diff
@@ -55,17 +54,13 @@ export class Rule extends Lint.Rules.AbstractRule {
...
-            const errorString = Rule.FAILURE_STRING_FACTORY(lineCount, lineLimit);
-            ruleFailures.push(new Lint.RuleFailure(sourceFile, 0, 1, errorString,
-              this.getOptions().ruleName));
+        if (lineCount <= lineLimit) {
+            return [];
         }
-        return ruleFailures;
+
+        return this.applyWithFunction(sourceFile, (ctx) =>
+            ctx.addFailure(0, 1, Rule.FAILURE_STRING(lineCount, lineLimit)));
```
ajafff: you could directly return the failure now that `applyWithFunction` doesn't modify the result anymore

src/rules/maxFileLineCountRule.ts
```diff
@@ -55,17 +54,13 @@ export class Rule extends Lint.Rules.AbstractRule {
...
 
     public apply(sourceFile: ts.SourceFile): Lint.RuleFailure[] {
-        const ruleFailures: Lint.RuleFailure[] = [];
-        const ruleArguments = this.getOptions().ruleArguments;
-        const lineLimit: number = ruleArguments[0];
+        const lineLimit: number = this.ruleArguments[0];
         const lineCount: number = sourceFile.getLineStarts().length;
-        const disabledIntervals = this.getOptions().disabledIntervals;
-
-        if (lineCount > lineLimit && disabledIntervals.length === 0) {
```
ajafff: This changes the behavior of the rule, which should at least be documented.
Before this change the rule would not fail if there were **any** disabledInvervals. Now you would need to have a disable comment in the first line and first column to disable the rule.

You would only get back the old behavior if the failure would span the whole file. I would not suggest doing that.

ajafff:
Related, but no blocker for this PR: when applying fixes, the position of the disabledIntervals may no longer be correct. It needs to be refreshed after each rule that applied fixes.

andy-hanson:
In order to keep this PR small, I've removed `isExplicitlyEnabled`. Changing the Linter API can be done later.

src/ruleLoader.ts
andy-hanson: OK, added the check back.

src/rules/maxFileLineCountRule.ts
andy-hanson: Moved error to the bottom, so a `tslint:disable` anywhere should work. Presumably no one uses `tslint:disable-line` for this rule.

src/rules/maxFileLineCountRule.ts
andy-hanson: I like this better than constructing `new Lint.RuleFailure` directly because I don't have to pass in e.g. `ruleName` explicitly. Could have a `Rule#createFailure` method but I think that would only be used here.

src/rules/eoflineRule.ts
```diff
@@ -41,9 +41,7 @@ export class Rule extends Lint.Rules.AbstractRule {
             return [];
         }
 
-        return this.filterFailures([
-            new Lint.RuleFailure(sourceFile, length, length, Rule.FAILURE_STRING,
-              this.getOptions().ruleName),
-        ]);
+        return this.applyWithFunction(sourceFile, (ctx) =>
```
ajafff: here too

nchen63:
it doesn't look like any unit tests hit enableDisableRules.ts

andy-hanson:
There is a black-box test `test/rules/_integration/enable-disable`.
It would be nice to have code coverage tests though.

andy-hanson:
Test failure fixed by #2691.

src/language/rule/abstractRule.ts
```diff
@@ -61,17 +60,6 @@ export abstract class AbstractRule implements IRule {
     ): RuleFailure[] {
         const ctx = new WalkContext(sourceFile, this.ruleName, options);
         walkFn(ctx);
-        return this.filterFailures(ctx.failures);
-    }
-
-    protected filterFailures(failures: RuleFailure[]): RuleFailure[] {
```
ajafff: breaking change. keep this method, just return the input unchanged and deprecate it

src/language/rule/rule.ts
```diff
@@ -18,7 +18,11 @@
 import * as ts from "typescript";
 
 import {arrayify, flatMap} from "../../utils";
-import {IWalker} from "../walker";
+
+export interface RuleStatic {
```
ajafff: that's a duplicate of `RuleConstructor`?

src/language/rule/rule.ts
```diff
@@ -99,19 +103,12 @@ export interface IOptions {
     ruleArguments: any[];
     ruleSeverity: RuleSeverity;
     ruleName: string;
-    disabledIntervals: IDisabledInterval[];
```
ajafff: this should be kept and deprecated for backwards compatibility

src/language/rule/rule.ts
```diff
@@ -99,19 +103,12 @@ export interface IOptions {
...
-}
-
-export interface IDisabledInterval {
```
ajafff: same as above

src/language/rule/rule.ts
ajafff: should also not be removed for now

src/language/utils.ts
```diff
@@ -19,21 +19,11 @@ import * as path from "path";
...
 import * as ts from "typescript";
 
-import {IDisabledInterval, RuleFailure} from "./rule/rule";
-
 export function getSourceFile(fileName: string, source: string): ts.SourceFile {
     const normalizedName = path.normalize(fileName).replace(/\\/g, "/");
     return ts.createSourceFile(normalizedName, source, ts.ScriptTarget.ES5, /*setParentNodes*/ true);
 }
 
-export function doesIntersect(failure: RuleFailure, disabledIntervals: IDisabledInterval[]) {
```
ajafff: another breaking change

andy-hanson:
@nchen This was originally written during tslint 4.0. What is the schedule for new major tslint releases? It seemed to be much slower in the past but the gap from 4.0 to 5.0 was pretty small. If it will be more than a few months before the next major-version release I can add dummy implementations over the breaking changes as @ajaff mentioned. Otherwise I would be fine with waiting a little longer.

nchen63:
There isn't a major release scheduled. It's usually based on how many breaking changes are queued up

src/language/utils.ts
andy-hanson: I don't think this is public. (#2446)

ajafff:
Everything in language/utils.ts is exported. For example as Lint.hasModifier

Am 06.05.2017 18:56 schrieb "Andy Hanson" <notifications@github.com>:

src/linter.ts
```diff
@@ -158,74 +145,78 @@ class Linter {
...
+
+        // If there were fixes, get the *new* list of failures.
+        return this.getAllFailures(sourceFile, enabledRules);
+    }
+
+    // Only "protected" because a test directly accesses it.
+    // tslint:disable-next-line member-ordering
+    protected applyFixes(sourceFilePath: string, source: string, fixableFailures: RuleFailure[]): string {
+        if (fixableFailures.some((f) => !f.hasFix())) {
+            throw new Error("!");
```
nchen63: needs error text

src/linter.ts
```diff
@@ -158,74 +145,78 @@ class Linter {
...
+    }
+
+    private applyAllFixes(
+            enabledRules: IRule[], fileFailures: RuleFailure[], sourceFile: ts.SourceFile, sourceFileName: string): RuleFailure[] {
+        // When fixing, we need to be careful as a fix in one rule may affect other rules.
+        // So fix each rule separately.
+        let source: string = sourceFile.text;
+
+        for (const rule of enabledRules) {
+            const hasFixes = fileFailures.some((f) => f.hasFix() && f.getRuleName() === rule.getOptions().ruleName);
```
nchen63: not that we should optimize on the --fix code path, but wouldn't it be more efficient to do:

- call rule1
- fix rule1
- call rule2
- fix rule2

instead of 

- call all rules
- call rule1 again if there were fixes
- fix rule1
- call rule2 again if there were fixes
- fix rule2

andy-hanson: Well, the way the enable/disable logic is done here, it's better to handle disables for all rules at once. Otherwise we would have to parse disables after every rule.

nchen63: you _could_ reuse the disable map and reload it when the file is overwritten. I'm ok with how it is now, though

andy-hanson: I would hold off on thatn until #1330 is implemented, since that will need to know all the uses of the disable map.

PR #2374, opened 2017-03-19
Add 'type-literal-delimiter' rule

andy-hanson:
#### PR checklist

  - [X] Includes tests
- [X] Documentation update

#### Overview of change:

Added the `type-literal-delimiter` rule, which prefers `{ x: number, y: string }` over `{ x: number; y: string }`.

#### CHANGELOG.md entry:

[new-rule] `type-literal-delimiter`


test/rules/type-literal-delimiter/test.ts.lint
```diff
@@ -0,0 +1,28 @@
+type T = {
```
nchen63: does it handle space before/after delimiter?

andy-hanson: Not currently, no. I think that would be more appropriate for the `whitespace` rule.

nchen63: I'm asking if it would erroneously flag { a: b , c: d   }

andy-hanson: Oh, gotcha. Added a test.

adidahiya:
wait, this feels pretty inconsistent with how we write out interfaces... users are going to be pretty confused when tslint recommends this format:

```ts
type Foo {
  a: number,
  b: number,
}

interface Bar {
  a: number;
  b: number;
}
```

can we please make the default config prefer semicolons?

andy-hanson:
@adidahiya If so, then the current `trailing-comma` wouldn't handle checking for the last token. Maybe `trailing-comma` should be updated to check for type literal delimiters too, regardless of the kind used?

adidahiya:
@andy-hanson no, I think we can leave trailing-comma as-is. I think _this_ rule (type-literal-delimiter) should enforce the trailing semicolon.

Issue #2330, opened 2017-03-11
kind of undefined in canHaveLeadingTrivia


Pajn:
### Bug Report

- __TSLint version__: 4.5.1
- __TypeScript version__: 2.2.1
- __Running TSLint via__: (pick one) VSCode

#### TypeScript code being linted
No minimal case, but I don't think it's needed here.

#### Actual behavior
```
TypeError: Cannot read property 'kind' of undefined
    at canHaveLeadingTrivia (/home/rasmus/Development/control-flow/ts/node_modules/tsutils/src/utils.js:309:26)
    at /home/rasmus/Development/control-flow/ts/node_modules/tsutils/src/utils.js:283:23
    at iterate (/home/rasmus/Development/control-flow/ts/node_modules/tsutils/src/utils.js:251:20)
    at Array.forEach (native)
    at iterate (/home/rasmus/Development/control-flow/ts/node_modules/tsutils/src/utils.js:253:50)
    at Array.forEach (native)
    at iterate (/home/rasmus/Development/control-flow/ts/node_modules/tsutils/src/utils.js:253:50)
    at forEachToken (/home/rasmus/Development/control-flow/ts/node_modules/tsutils/src/utils.js:254:7)
    at Object.forEachComment (/home/rasmus/Development/control-flow/ts/node_modules/tsutils/src/utils.js:282:12)
    at EnableDisableRulesWalker.getEnableDisableRuleMap (/home/rasmus/Development/control-flow/ts/node_modules/tslint/lib/enableDisableRules.js:42:15)
```

The line `if (parent.parent.kind === ts.SyntaxKind.JsxClosingElement)` in `canHaveLeadingTrivia` must guard for the node not having grandparents, something like `if (parent.parent && parent.parent.kind === ts.SyntaxKind.JsxClosingElement)`

ajafff:
`parent.parent.kind` was actually a bug and should have been `parent.kind`.
I just released `tsutils@1.3.0` which includes this fix.

I couldn't reproduce the error. Can you share a code snipped that could be used as a regression test?

Pajn:
Great!

> Can you share a code snipped that could be used as a regression test?

Sorry, I saw it in the logs in a larger project with format on save when I tried to investigated why format all (and therefore format on save) rarely works however I don't get it now. I might be able to reproduce it by jumping to older commits if there is some way I can see in at least which file it occurred?

ajafff:
That's exactly what I was looking for.
[...]

PR #2386, opened 2017-03-22
Add regression test for invalid JSX

ajafff:
#### PR checklist

- [x] Addresses an existing issue: #2330
  - [x] Includes tests
- [ ] Documentation update

#### Overview of change:
Add regression test to avoid crashes on invalid JSX.
Fixes: #2330 

#### CHANGELOG.md entry:

[no-log]


yarn.lock
```diff
@@ -1039,10 +1039,14 @@ tslint@latest:
     tsutils "^1.1.0"
     update-notifier "^2.0.0"
 
-tsutils@^1.1.0, tsutils@^1.2.2:
+tsutils@^1.1.0:
   version "1.2.2"
   resolved "https://registry.yarnpkg.com/tsutils/-/tsutils-1.2.2.tgz#7e165c601367b9f89200b97ff47d9e38d1a6e4c8"
 
+tsutils@^1.4.0:
```
adidahiya: now we have two copies of tsutils, this looks wrong

test/rules/_integration/react/invalid.tsx.lint
```diff
@@ -0,0 +1,6 @@
+// don't crash on invalid jsx
+// tslint:disable
+const a =
+  <span /
+  >
+  </span>
```
adidahiya: nit: EOF newline

yarn.lock
ajafff: fixed.
this was caused by the dependency on tslint@latest which locked that dependency of tsutils when it was installed.

PR #2387, opened 2017-03-22
Add more align options

ajafff:
#### PR checklist

  - [x] Includes tests
- [x] Documentation update

#### Overview of change:
[new-rule-option] `align` rule added `elements` option, which checks alignment of elements in array literals, array destructuring and tuple types
[new-rule-option] `align` rule added `members` option, which checks alignment of methods and properties of classes, objects, interfaces, type literals and object destructuring

Should I split the new config options into several options that can be enabled independently?
Should I add this to `tslint:latest`?


adidahiya:
> Should I split the new config options into several options that can be enabled independently?

Not sure what you mean? Looks like you added two options, which looks good to me

> Should I add this to tslint:latest?

Yes

ajafff:
> > Should I split the new config options into several options that can be enabled independently?
>
> Not sure what you mean?

I meant splitting `members` into multiple options like `class`, `object`, `interface`, etc.
But we'll see what's the user feedback on this.

Re: tslint:latest
I only added `members` for now.
The `elements` option caused failures on multiline objects in arrays where not aligning elements is intended:
```ts
let foo = [{
    bar,
}, { // open brace of the second object is not aligned with open brace of the first
    baz,
}]
```
How should I handle this one? Also there is no check if the close brace is aligned.

adidahiya:
```ts
let foo = [{
    bar,
}, { // open brace of the second object is not aligned with open brace of the first
    baz,
}]
```

that's actually not how I would typically align those elements; I would choose this, which I assume works with the current rule implementation:

```ts
let foo = [
    {
        bar,
    },
    {
        baz,
    },
]
```

IMO the first alignment is misleading because `bar` and `baz` are not first-level children of `foo` (they are nested)

ajafff:
@adidahiya your example works with the current implementation. IMO it is not very common though.
In this project all array literals containing object literals start with the open brace right after the open bracket.

That's why I didn't enable that option in tslint:latest. Users can still enable it for their projects. I hope you are ok with that.

PR #2394, opened 2017-03-23
Refactor linebreak-style and add fixer

ajafff:
#### PR checklist

  - [x] Includes tests
- [x] Documentation update

#### Overview of change:
[new-fixer] `linebreak-style`
The rule checks the line ending directly in the source text instead of using a scanner. That way we also find failures inside multiline comments and template strings.


src/rules/linebreakStyleRule.ts
```diff
@@ -39,48 +39,34 @@ export class Rule extends Lint.Rules.AbstractRule {
         optionExamples: [`[true, "${OPTION_LINEBREAK_STYLE_LF}"]`, `[true, "${OPTION_LINEBREAK_STYLE_CRLF}"]`],
         type: "maintainability",
         typescriptOnly: false,
+        hasFix: true,
```
adidahiya: nice

PR #2395, opened 2017-03-23
deprecation rule

### LINKED ISSUE #825: Add a warning when using a JSDoc deprecated method
So this was previously suggested in Microsoft/TypeScript#390, but it seems like it should be in tslint, not typescript.

The idea is when a function which has the tag "[@deprecated](http://usejsdoc.org/tags-deprecated.html)" is used it should show a warning.
For example:

``` ts
interface IX {
  /**
   * @deprecated
   */
  function doSomething(): void;
}

class X implements IX {
  function doSomething(): void;

  /**
   * @deprecated
   */
  function doSomethingElse(): void;
}

const x = new X();
x.doSomething(); // Warning: IX.doSomething is deprecated
x.doSomethingElse(); // Warning: X.doSomethingElse is deprecated
```

Same thing should preferably apply to attributes. Getters & setters.

alexeagle:
#### PR checklist

Fixes #825

#### Overview of change:

Warn when @deprecated APIs are used.

Missing lots of test cases.

#### CHANGELOG.md entry:

[new-rule]

### DIFF
--- a/src/rules/deprecationRule.ts
+++ b/src/rules/deprecationRule.ts
@@ -1,0 +1,74 @@
+/**
+ * @license
+ * Copyright 2016 Palantir Technologies, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import * as ts from "typescript";
+import * as Lint from "../index";
+
+export class Rule extends Lint.Rules.TypedRule {
+
+    /* tslint:disable:object-literal-sort-keys */
+    public static metadata: Lint.IRuleMetadata = {
+        ruleName: "deprecation",
+        description: "Warns when deprecated APIs are used.",
+        descriptionDetails: Lint.Utils.dedent``,
+        rationale: Lint.Utils.dedent`
+            Deprecation`,
+        optionsDescription: Lint.Utils.dedent`
+            `,
+        options: null,
+        optionExamples: [],
+        type: "maintainability",
+        typescriptOnly: false,
+    };
+    /* tslint:enable:object-literal-sort-keys */
+    public applyWithProgram(sourceFile: ts.SourceFile, program: ts.Program): Lint.RuleFailure[] {
+        return this.applyWithWalker(new Walker(sourceFile, this.getOptions(), program));
+    }
+}
+
+class Walker extends Lint.ProgramAwareRuleWalker {
+  protected visitIdentifier(node: ts.Identifier) {
+    let decSym = this.getTypeChecker().getSymbolAtLocation(node);
+
+    // tslint:disable-next-line:no-bitwise
+    if (decSym.flags & ts.SymbolFlags.Alias) {
+        decSym = this.getTypeChecker().getAliasedSymbol(decSym);
+    }
+    for (const d of decSym.getDeclarations()) {
+      // Switch to the TS JSDoc parser in the future to avoid false positives here.
+      // For example using '@deprecated' in a true comment.
+      // However, a new TS API would be needed, track at
+      // https://github.com/Microsoft/TypeScript/issues/7393.
+      const commentNode: ts.Node = d;
+
+      if (d.getSourceFile() === node.getSourceFile() && d.pos === node.pos) {
+          continue;
+      }
+
+      const range = ts.getLeadingCommentRanges(commentNode.getFullText(), 0);
+      if (!range) { continue; }
+      for (const {pos, end} of range) {
+        const jsDocText = commentNode.getFullText().substring(pos, end);
+        if (jsDocText.includes("@deprecated")) {
+            this.addFailureAtNode(node, "Is deprecated.");
+        }
+      }
+    }
+
+    super.visitIdentifier(node);
+  }
+}
--- a/src/rules/deprecationRule.ts
+++ b/src/rules/deprecationRule.ts
@@ -34,6 +34,7 @@ export class Rule extends Lint.Rules.TypedRule {
         optionExamples: [],
         type: "maintainability",
         typescriptOnly: false,
+        requiresTypeInfo: true,
     };
     /* tslint:enable:object-literal-sort-keys */
     public applyWithProgram(sourceFile: ts.SourceFile, program: ts.Program): Lint.RuleFailure[] {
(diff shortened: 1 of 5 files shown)


palantirtech:

src/rules/deprecationRule.ts
```diff
@@ -0,0 +1,80 @@
...
+    public applyWithProgram(sourceFile: ts.SourceFile, program: ts.Program): Lint.RuleFailure[] {
+        return this.applyWithWalker(new Walker(sourceFile, this.getOptions(), program));
+    }
+}
+
+class Walker extends Lint.ProgramAwareRuleWalker {
+  // Implementation inspired by angular/tsickle:
+  // https://github.com/angular/tsickle/blob/cad7c180a2155db6f6fb8d22c44151d7e8a9149f/src/decorator-annotator.ts#L42
+  protected visitIdentifier(node: ts.Identifier) {
+    let decSym = this.getTypeChecker().getSymbolAtLocation(node);
```
andy-hanson: May be `undefined`

src/rules/deprecationRule.ts
```diff
@@ -0,0 +1,80 @@
...
+
+    // tslint:disable-next-line:no-bitwise
+    if (decSym.flags & ts.SymbolFlags.Alias) {
+        decSym = this.getTypeChecker().getAliasedSymbol(decSym);
+    }
+    for (const d of decSym.getDeclarations()) {
```
andy-hanson: declarations may be `undefined`

src/rules/deprecationRule.ts
```diff
@@ -0,0 +1,80 @@
...
+      // Switch to the TS JSDoc parser in the future to avoid false positives here.
+      // For example using '@deprecated' in a true comment.
+      // However, a new TS API would be needed, track at
+      // https://github.com/Microsoft/TypeScript/issues/7393.
+      const commentNode: ts.Node = d;
+
+      // Don't warn on the declaration of the @deprecated symbol.
+      if (commentNode.getSourceFile() === node.getSourceFile()
+          && commentNode.getStart() <= node.getStart()
+          && node.getEnd() <= commentNode.getEnd()) {
```
andy-hanson: Did you mean `>=` here? The end of the declaration may come *after* the end of the identifier.

src/rules/deprecationRule.ts
```diff
@@ -0,0 +1,80 @@
...
+    }
+    for (const d of decSym.getDeclarations()) {
+      // Switch to the TS JSDoc parser in the future to avoid false positives here.
+      // For example using '@deprecated' in a true comment.
+      // However, a new TS API would be needed, track at
+      // https://github.com/Microsoft/TypeScript/issues/7393.
+      const commentNode: ts.Node = d;
+
+      // Don't warn on the declaration of the @deprecated symbol.
+      if (commentNode.getSourceFile() === node.getSourceFile()
```
andy-hanson: `getSourceFile()` and `getStart()` are slow, is there another way?

src/rules/deprecationRule.ts
```diff
@@ -0,0 +1,80 @@
...
+          && node.getEnd() <= commentNode.getEnd()) {
+          continue;
+      }
+
+      const range = ts.getLeadingCommentRanges(commentNode.getFullText(), 0);
+      if (!range) { continue; }
+      for (const {pos, end} of range) {
+        const jsDocText = commentNode.getFullText().substring(pos, end);
+        if (jsDocText.includes("@deprecated")) {
+            this.addFailureAtNode(node, node.getText() + " is deprecated.");
```
andy-hanson: Node is an identifier, so use `node.text`

src/rules/deprecationRule.ts
```diff
@@ -0,0 +1,80 @@
...
+      const commentNode: ts.Node = d;
+
+      // Don't warn on the declaration of the @deprecated symbol.
+      if (commentNode.getSourceFile() === node.getSourceFile()
+          && commentNode.getStart() <= node.getStart()
+          && node.getEnd() <= commentNode.getEnd()) {
+          continue;
+      }
+
+      const range = ts.getLeadingCommentRanges(commentNode.getFullText(), 0);
```
andy-hanson: You could just search for `@deprecated` in the range between the `pos` and `getStart()`, instead of using comment ranges.

src/rules/deprecationRule.ts
```diff
@@ -0,0 +1,80 @@
...
+    }
+}
+
+class Walker extends Lint.ProgramAwareRuleWalker {
+  // Implementation inspired by angular/tsickle:
+  // https://github.com/angular/tsickle/blob/cad7c180a2155db6f6fb8d22c44151d7e8a9149f/src/decorator-annotator.ts#L42
+  protected visitIdentifier(node: ts.Identifier) {
+    let decSym = this.getTypeChecker().getSymbolAtLocation(node);
+
+    // tslint:disable-next-line:no-bitwise
```
andy-hanson: Use `Lint.isSymbolFlagSet`.

test/rules/deprecation/test.ts.lint
```diff
@@ -0,0 +1,26 @@
+declare interface D {
```
andy-hanson: Test for imports.

src/rules/deprecationRule.ts
alexeagle: we really need to push TS team to add strictNulls to their API

src/rules/deprecationRule.ts
alexeagle: I changed the order to check pos and getEnd first, which reduces the getSourceFile calls to only the likely declaration site.
Is there a performance benchmark I can run to calculate whether there's still a perf issue? How do you do it?

src/rules/deprecationRule.ts
alexeagle: done

test/rules/deprecation/test.ts.lint
alexeagle: done

src/rules/deprecationRule.ts
alexeagle: @evmar you're the expert in the TS comment AST, any reason not to do this (and in tsickle as well)?

src/rules/deprecationRule.ts
ajafff: the metadata is only used to generate documentation, it doesn't influence the behavior of the rule

The typescript API you are using may return `undefined` even though its signature does not declare that. You already mentioned the problem with strictNullChecks in TS APIs below...

src/rules/deprecationRule.ts
```diff
@@ -0,0 +1,97 @@
...
+    }
+    for (const d of decSym.getDeclarations()) {
+      // Switch to the TS JSDoc parser in the future to avoid false positives here.
+      // For example using '@deprecated' in a true comment.
+      // However, a new TS API would be needed, track at
+      // https://github.com/Microsoft/TypeScript/issues/7393.
+      let commentNode: ts.Node = d;
+
+      if (commentNode.kind === ts.SyntaxKind.VariableDeclaration) {
+          if (!commentNode.parent) { continue; }
```
andy-hanson: Just use the null assertion (`!`) operator here. Parent pointers should be set within lint rules.

src/rules/deprecationRule.ts
```diff
@@ -0,0 +1,97 @@
...
+      // VDList.getFullText will not contain the comment.
+      if (commentNode.kind === ts.SyntaxKind.VariableDeclarationList) {
+        if (!commentNode.parent) { continue; }
+        commentNode = commentNode.parent;
+      }
+
+      // Don't warn on the declaration of the @deprecated symbol.
+      if (commentNode.pos <= node.pos
+          && node.getEnd() <= commentNode.getEnd()
+          && commentNode.getSourceFile() === node.getSourceFile()) {
```
andy-hanson: `node.getSourceFile()` is just `this.sourceFile`.

alexeagle: done

src/rules/deprecationRule.ts
alexeagle: done

src/rules/deprecationRule.ts
alexeagle: have a test now for undefined symbol

alexeagle:
Okay green now, ready to merge?

adidahiya:
thanks @alexeagle!

PR #2403, opened 2017-03-25
[Breaking change] Remove `createFix`, and just use the Replacement(s) as the fix. (Minimal change)

andy-hanson:
#### PR checklist

  - [ ] Includes tests
- [ ] Documentation update

#### Overview of change:

This is #2356, but made to have fewer changes to rule style.

#### CHANGELOG.md entry:

[api] Remove createFix. Replacements should be passed directly into addFailure.

### DIFF
--- a/docs/_data/formatters.json
+++ b/docs/_data/formatters.json
@@ -22,7 +22,7 @@
   {
     "formatterName": "json",
     "description": "Formats errors as simple JSON.",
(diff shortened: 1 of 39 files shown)


adidahiya:
sorry, this has conflicts now (likely because of #2359)

andy-hanson:
Turning tests on #2406 revealed that it did change the output of the JSON formatter.

ajafff:
We should at least mention the changed (JSON) formatted output in the change log. It's hard to guess for embedders to guess the new format.
My suggestion:
``[formatter-change] the `fix` property in `json` output now contains either one replacement or an array of replacements``

Issue #2213, opened 2017-02-17
Rule Request: no-sparse-arrays


Zzzen:
See https://github.com/Microsoft/tslint-microsoft-contrib/commit/6ca2974fb275450e74c8096339e561179e10a8bf for more details.

PR #2407, opened 2017-03-25
Add no-sparse-arrays rule

andy-hanson:
#### PR checklist

- [X] Addresses an existing issue: #2213
  - [X] Includes tests
- [X] Documentation update

#### Overview of change:

Added the `no-sparse-arrays` rule, which warns on `[1, , 3]`.

#### CHANGELOG.md entry:

[new-rule] `no-sparse-arrays`


adidahiya:
why not name this the same as eslint and tslint-microsoft-contrib? `no-sparse-arrays`

src/rules/noArrayLiteralHoleRule.ts
```diff
@@ -0,0 +1,59 @@
...
+function walk(ctx: Lint.WalkContext<void>): void {
+    return ts.forEachChild(ctx.sourceFile, function cb(node: ts.Node): void {
+        if (!utils.isArrayLiteralExpression(node)) {
+            return ts.forEachChild(node, cb);
+        }
+
+        for (const element of node.elements) {
+            if (utils.isOmittedExpression(element)) {
+                // Node has an empty range, so just use range starting at `element.pos`.
+                ctx.addFailureAt(element.pos, 1, Rule.FAILURE_STRING);
```
ajafff: Unfortunately it's not that easy. You need to exclude array destructuring.
So you basically need test cases like:
```ts
[foo, , bar] = [];
[({foo: [, bar]})] = [];
[foo = [, 1]] = []; // should fail
```
Have a look at the tests for prefer-const, which does almost the same.

test/rules/no-array-literal-hole/test.ts.lint
```diff
@@ -0,0 +1,5 @@
+[1, , 3];
+   ~ [Array has a missing element.]
```
ajafff: add a test with trailing comma and one with multiple trailing commas

test/rules/no-array-literal-hole/test.ts.lint
```diff
@@ -0,0 +1,5 @@
...
+
+// Destructuring allowed.
+const [x, , z] = arr;
```
ajafff: That's only destructuring in declarations (ArrayBindingPattern). Destructuring in expression uses ArrayLiteralExpression, see my other comment above.

src/rules/noSparseArraysRule.ts
```diff
@@ -0,0 +1,91 @@
...
+            } else {
+                ts.forEachChild(element, cb);
+            }
+        }
+    });
+}
+
+/** Traverse the LHS of an `=` expression, calling `cb` embedded default value, but ignoring binding patterns. */
+function traverseExpressionsInLHS(node: ts.Node, cb: (node: ts.Expression) => void): void {
+    switch (node.kind) {
```
ajafff: you need to add a case for `ts.SyntaxKind.ParethesizedExpression`

test/rules/no-sparse-arrays/test.ts.lint
```diff
@@ -0,0 +1,22 @@
...
+
+// Destructuring allowed.
+const [foo, , bar] = [];
+const [{foo: [, bar]}] = [];
+const [foo = [, 1]] = [];
+              ~ [0]
+
+[foo, , bar] = [];
+[{foo: [, bar]}] = [];
+[foo = [, 1]] = [];
```
ajafff: can you add another test case like: 
```ts
[foo = bar([,])] = [];
            ~[0]
```
I know this works with your current implementation. Just to make sure we won't break it in the future.

Issue #73, opened 2014-02-05
Feature: ".tslintignore" file to exclude globs


am11:
### Update (2016/06/01)

`--exclude` CLI flag is available since TSLint v3.6. This issue now tracks a feature request for a `.tslintignore` file.
### Original issue

Cc: https://github.com/madskristensen/WebEssentials2013/issues/617

From [JsHint docs](http://www.jshint.com/docs/cli/), there are two flags for exclusion in cli options:

> --exclude
> 
> Allows you to specify directories which you DON'T want to be linted.
> 
> --exclude-path
> 
> Allows you to provide your own .jshintignore file. For example, you can point JSHint to your .gitignore file and use it instead of default .jshintignore.

Please provide us the similar flags, so we can pass .tslintignore file to exclude files and folders when some batch / automated executor script runs lint on it.

See JsHint's implementation of this feature: [cli.js#L182-L208](https://github.com/jshint/jshint/blob/56a4a78dd65a872d93d843621c14d2ebe575cfed/src/cli.js#L182-L208) and its caller; [gather function](https://github.com/jshint/jshint/blob/56a4a78dd65a872d93d843621c14d2ebe575cfed/src/cli.js#L411-L435).

pgrm:
any updates on this?

JKillian:
This seems like a relatively important feature that ought to be prioritized. Wanted to note interest in this feature from a [SO question](http://stackoverflow.com/questions/34578677/how-to-ignore-a-particular-directory-or-file-for-tslint)

adidahiya:
Might be better to support this via tsconfig.json (#858) rather than re-implementing this in a CLI flag...

JKillian:
Hmm, the difficulty with that is that you may want to compile different files than you lint. So I could potentially see using `tsconfig.json` and then also using `--exclude` to not lint a few of the normally included files.

bill-mybiz:
As opposed to CLI, would this not be a natural fit in the tslint.json file and then add an ignore check that parses the "ignorePaths" at https://github.com/palantir/tslint/blob/master/src/tslint.ts#L46?

tsm91:
+1

shaharmor:
+1, a .tslintignore will be great

lukaselmer:
+1 for .tslintignore or add an option to ignore certain paths in tsconfig.json

It could be similar to https://github.com/bbatsov/rubocop#includingexcluding-files

adidahiya:
[...]

Issue #2227, opened 2017-02-22
tslint.json should have exactly the same options as offered on the command line


gotofritz:
### Bug Report

- __TSLint version__: `4.4.1`
- __TypeScript version__: `2.1.6`
- __Running TSLint via__: both angular-cli and node_modules/tslint/bin/tslint using tslint.json and tsconfig.json

When I pass a formatter as a command line option then it works
`node_modules/tslint/bin/tslint --config ./tslint.json --project ./src/tsconfig.json -t verbose`

When I put the formatter in the tslint.json, it doesn't seem to pick it up. 
`node_modules/tslint/bin/tslint --config ./tslint.json --project ./src/tsconfig.json`

I raised a bug about it #2226 , and a contributor said it isn't possible. But most CLI tool offer exactly the same options as arguments or in a config file, it makes no sense to have them different. Normal practice is that either can be used, and if both are presents the command line overrides the config file.

I use tslint via angular-cli and I can't pass any command line options. I would like to be able to use it out of the box. Thanks

JoshuaKGoldberg:
Removing the `Type: Breaking Change` label per #4811. Now accepting PRs!

JoshuaKGoldberg:
☠️ *TSLint's time has come!* ☠️ 

TSLint is no longer accepting most feature requests per #4534. See **[typescript-eslint.io](https://typescript-eslint.io)** for the new, shiny way to lint your TypeScript code with **ESLint**. ✨ 

It was a pleasure open sourcing with you all!

JoshuaKGoldberg:
🤖 Beep boop! 👉  **TSLint is deprecated** 👈 _(#4534)_  and you should **switch to [typescript-eslint](https://typescript-eslint.io)**! 🤖 

🔒 This issue is being locked to prevent further unnecessary discussions. Thank you! 👋

PR #2409, opened 2017-03-25
Add exclude section to configuration files.

mmkal:
This allows users to define default cli options in tslint.json.

#### PR checklist

- [x] Addresses #73 
  - [x] Includes tests
- [x] Documentation update

#### Overview of change:

Adds `cliOptions` object to `IConfigurationFile` interface in `src/configuration.ts`.
Modifies `src/cli.ts` to read from the configuration file (passed in via cli or default `tslint.json`) and use `configuration.cliOptions` as default values for the command line options. 

This would, for example, allow exluding files from linting (the topic of the issue) like this:
```json
{
  "extends": "tslint:latest",
  "cliOptions": {
      "exclude": [
          "bin",
          "lib/*generated.js"
      ]
  }
}
```

suggested tags: [enhancement]


palantirtech:
Thanks for your interest in [palantir/tslint](https://github.com/palantir/tslint), @mmkal! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

adidahiya:
Cool, overall this seems reasonable given all the user feedback around CLI options (mainly summed up in https://github.com/palantir/tslint/issues/2227). It does represent a change of philosophy in what tslint.json does. Previously, the configuration file could only specify _how_ to lint, but now it has the ability to specify _what_ to lint as well. See https://github.com/palantir/tslint/issues/73#issuecomment-236077677 and https://github.com/palantir/tslint/issues/629#issuecomment-261291205. What do you think @JKillian @nchen63 @ajafff @andy-hanson @andy-ms @mgechev @chrismbarr? Any objections?

ajafff:
I'm ok with specifying *what* to lint and making all CLI options available in tslint.json.

There are some questions that need to be resolved:
* move all options to `linterOptions` as @adidahiya suggests OR make them all top level options? Since `linterOptions.typeCheck` is private, the latter would not break any user
* add a new config `include` to be consistent with `tsc`?
* should entries in `exclude` be merged when extending configurations?
* do options passed via CLI merge or override options in the config?

adidahiya:
> move all options to linterOptions as @adidahiya suggests OR make them all top level options? 

not all options are configurable via the CLI. for example, `"extends"` and `"rules"`. So I would prefer to group all the CLI options together into one object and not make them top-level options. semantically, `"extends"` should always be a top-level object. and I wouldn't want to break backcompat just yet by forcing nesting of any existing top-level option (`rules`, `rulesDirectory`).

> add a new config include to be consistent with tsc?

yeah, I would accept that as a PR if it had the same semantics as `tsc`. but I wouldn't block this PR on it.

> should entries in exclude be merged when extending configurations?

hmm, not sure about this one. In general, we don't expect base/intermediate configurations (those you are _extending_) to use `"exclude"` very much (it's just harder to reason about).

> do options passed via CLI merge or override options in the config?

I think they should override the config file's options (this is how  most CLIs work)

mmkal:
> I don't think this exactly solves #73 because tslint doesn't look for additional tslint.json files after it has found a relevant one it can use to lint. I believe the request for a .tslintignore file was made so that users could have multiple such ignore files across their codebase and a single tslint.json (I'm not 100% sure though, you should go back to that thread and confirm).

I didn't go the `.tslintignore` route because it felt clunkier and more obscure. It feels much cleaner to me to be able to open up the `tslint.json` file and from that be able to understand what running `tslint` is going to do. But you're right it doesn't solve it precisely. Maybe if this gets in we could see if anyone thinks `.tslintignore` is still worth developing. I suspect this will cover almost all cases.

> we should name the top-level config key "linterOptions", like this private API we use for testing.

Sure, will push an update soon.

> there are many more CLI options, not just exclude. we need tests for those too. in particular we need to test things like path resolution

I'm unsure what you mean by this. In terms of testing the code I'm adding, I'm almost exclusively relying on existing functionality.

In terms of implementation, I'm loading the configuration object using `findConfigurationPath` and `loadConfigurationFromPath` in `src/configuration.ts`. Then using those values as defaults using `optimist.default(...)`. Is it necessary to test that part? My feeling there was that it would be simply re-testing that `optimist.default(...)` works correctly. Or did you mean I should add some more tests of the configuration merging?

I didn't find any existing tests of `src/cli.ts`, but I could add a new test file?

> rulesDirectory is currently handled specially as a top-level tslint.json key. It should probably move into linterOptions to be consistent (yeah it's a breaking change but we're on the 5.0 track now so we can do that).

Sure - should I deprecate the existing `rulesDirectory`? If so is there an example of something similar being done?

> should entries in exclude be merged when extending configurations?
> > hmm, not sure about this one. In general, we don't expect base/intermediate configurations (those you are extending) to use "exclude" very much (it's just harder to reason about).

I wasn't sure either. But I had a vision of working on a team project and extending an existing `tslint.json` and tearing my hair out trying to figure out why some of my files were being excluded from linting so I made it replace existing values instead.

> do options passed via CLI merge or override options in the config?
> > I think they should override the config file's options (this is how most CLIs work)

Yeah, I was basing this on how `tsc` works. You can set `sourceMap: false` in `tsconfig.json` but override it with `tsc --sourceMap -p .` on the cli for local builds etc.

JKillian:
Haven't reviewed the actual code changes, but I'm fine w/ this in principle if it makes things easier for users. There are some questions about how this affects integration with other tools though:

In the past it was up to the integrating tool to decide what files to lint. If you were using an IDE, your IDE would have to tell TSLint which files to lint. If you're using webpack, this would be controlled by your webpack rule/loader `test` field. If using the CLI, it would be controlled by the command line options you pass to TSLint.

In this PR, presumably the proposed `cliOptions` field would only affect the CLI? Or would IDEs be expected to read/parse it as well? What about a webpack loader?

adidahiya:
@mmkal 

> It feels much cleaner to me to be able to open up the tslint.json file and from that be able to understand what running tslint is going to do. But you're right it doesn't solve it precisely. Maybe if this gets in we could see if anyone thinks .tslintignore is still worth developing. I suspect this will cover almost all cases.

I agree, this is better than `.tslintignore`.

> I didn't find any existing tests of src/cli.ts, but I could add a new test file?

Yeah, never mind the path resolution thing; I just think there should just be a test that verifies that CLI options can override `linterOptions` in the config file by testing the whole workflow with source file fixtures. Add a sample project to lint as a test fixture and run the CLI in a way that checks the new features work as intended.

> should I deprecate the existing rulesDirectory?

Yeah, let's deprecate it with a `console.warn` message directing users to move the option into a `linterOptions` block, but don't remove the feature completely.

> I made it replace existing values instead.

Cool, this makes sense to implement now and wait for feedback if the behavior should be changed later

adidahiya:
@JKillian 

> In this PR, presumably the proposed cliOptions field would only affect the CLI? Or would IDEs be expected to read/parse it as well? What about a webpack loader?

Yes, I would actually expect IDEs and webpack loaders to read these config options (eventually). With `tsc`, we've found that configuring compiler options in one place, declaratively, works well.

adidahiya:
@mmkal ah, actually we do have CLI tests, they're in test/executable/executableTests.ts

JKillian:
> Yes, I would actually expect IDEs and webpack loaders to read these config options (eventually). With tsc, we've found that configuring compiler options in one place, declaratively, works well.

Good point and agreed, `tsconfig.json` does work out quite nicely. Since other tools may be reading this data, it also supports your point that we shouldn't call the key `cliOptions`. What we should do, then is have good utility functions so that 3rd-party tool integration is easy. Nobody will want to have to figure out how to calculate the correct combination of excludes and includes and globs themselves! We'll need to make sure this is a stable API that 3rd-party tools can rely upon.

mmkal:
@adidahiya just to give an update on this... I haven't had a lot of time to look at this since we last spoke, but I managed to look at it a little over the weekend. Moving cliOptions to linterOptions is a little more complicated than I'd originally thought. There are a couple of design decisions to consider, like

* Should I camelCase the options passed in?
* If so, should I allow non-camel-cased options?

Camel casing is more consistent with the rest of tslint.json, but it increases the complexity because the cli options are hyphenated. 

The deprecation of rulesDir and linterOptions.typeCheck is also a little unclear. Presumably I need to support the merging of rules directories from parent configs? If so, how do I reconcile this with the decision above, to overwrite rather than replace properties? Different behavior for different properties? And since you've suggested I deprecate rulesDir, this would mean changing it in a lot of different files since it wouldn't be good to have tslint itself using a deprecated feature. 

Another problem I've seen is that I can't actually seem to make the executableTests run. npm test and npm run verify seem to ignore them. 

Taking all this into account, my solution is no longer looking like a cheap win. As I don't know the codebase well, I would rather not be making big sweeping changes. So I can make two suggestions for what I can do next:

1. Go back to using cliOptions, but don't include rules-dir as that's already implemented.
2. Keep plugging away at linterOptions, after getting answers to the questions above (my preference would be support camelCase and non-camel-case by transforming from one to the other, so that the current implementation of typeCheck can be left alone. This also makes documentation easier as it may be good enough to say "use any cli options in the linterOptions object"). I would still want to exclude rules-dir here to avoid dealing with any complications over inheriting from other configs. If you want to move rules-dir inside linterOptions, it'd have to be done by someone who can give up a bit more time on this project. 
3. Reduce the scope of this work to what it was originally about, and *only* add the "exclude" property to linterOptions. This conveniently avoids the camelCase problem too. Everything else could then be left more or less as it is.

I'd be ok with any one of those three - or any other suggestion you might have. As long as they allow me to only make small code changes - otherwise I won't realistically be able to get them done.

adidahiya:
@mmkal sorry for the delay. Let's just go with the easiest approach, (3), limiting the scope of the work in this PR.

mmkal:
@adidahiya updated and added new functionality to docs.

BorntraegerMarc:
@adidahiya @nchen63 can this PR be merged now and a new release be published? We need this feature desperately...

CaseyPeters:
@adidahiya @mmkal My team is really looking forward to this change!!!!!!

mmkal:
@caseypeters ok, will try to find some time to look at this in the next week

mmkal:
@adidahiya updated - sorry it took so long, I ended up finding a workaround for the project I needed this for, and I didn't have a lot of time to look at this outside of work.

You were right about using __filename not making sense. I moved the check for whether a file is excluded to just before it's linted, since the config for that file is being loaded there already.

adidahiya:
@mmkal I pushed a few small tweaks to your branch.

@ajafff mind taking a look here?

mmkal:
whoops, commented from my work account. Your changes look good to me @adidahiya, thanks!

src/configuration.ts
```diff
@@ -41,11 +41,12 @@ export interface IConfigurationFile {
...
 
     /**
-     * Other linter options, currently for testing. Not publicly supported.
+     * A subset of the CLI options.
      */
-    linterOptions?: {
-        typeCheck?: boolean;
-    };
+    linterOptions?: Partial<{
+        typeCheck: boolean;
```
ajafff: you can remove the `typeCheck`. It's no longer used by the tests

src/runner.ts
```diff
@@ -218,6 +218,14 @@ async function doLinting(
 
     let lastFolder: string | undefined;
     let configFile: IConfigurationFile | undefined;
+    const isFileExcluded = (filepath: string) => {
+        if (configFile === undefined || configFile.linterOptions == null || configFile.linterOptions.exclude == null) {
+            return false;
+        }
+        const fullPath = path.resolve(filepath);
+        return configFile.linterOptions.exclude.some((pattern) => new Minimatch(pattern).match(fullPath));
```
ajafff: The excluded paths should probably be resolved relative to the location of `tslint.json`.
If `/foo/tslint.json` excludes `baz.ts` we should only exclude `/foo/baz.ts` and still lint `/foo/bar/baz.ts`.

That will only work when `filepath` is an absolute path.

test/configurationTests.ts
```diff
@@ -193,6 +193,32 @@ describe("Configuration", () => {
...
+            const expectedConfig = getEmptyConfig();
+            expectedConfig.linterOptions = {
+                exclude: [
+                    "lib",
+                    "bin",
+                ],
+            };
+
+            const actualConfig = extendConfigurationFile(baseConfig, extendingConfig);
+            assertConfigEquals(actualConfig, expectedConfig);
```
ajafff: out of curiosity: what happens if `extendingConfig` contains an empty object as `"linterOptions"`? Would that override the whole object including `"exclude"`?

mmkal: No, only if `exclude` is specified. I added a test for this below.

KSuttle-cng:
Will this be in 5.8?

ajafff:
@KSuttle-cng Yes, this will be part of the next release

KSuttle-cng:
Cool, any timeline for this? I've been trying to get the CLI to respect the `excludes` from the tslint.json, but I have to add `--exclude` for every entry.

ajafff:
There's no release schedule. Maybe I get around to it during the next week.

KSuttle-cng:
Awesome. Thanks @ajafff.

PR #2415, opened 2017-03-27
Added tags logic for completed-docs

### LINKED ISSUE #1921: completed-docs false positive when entire doc comment consists of only tags

JoshuaKGoldberg:
Allows `completed-docs` to list doc tags that mark a node as not requiring a documentation body. Tags can also provide a regexp matcher to validate that their contents are docs-valid.

Fixes #1921.


JoshuaKGoldberg:
:( tests pass locally! Might be a TypeScript version-specific issue? Will take a look soon...

Also, I couldn't find existing examples of how to specify an object mapping strings to strings for the config settings (though I have a vague recollection of looking before). Will do a real scan-through soon also...

nchen63:
Looks like it fails for 2.0.10. I think we can remove support for that in 5.0 so I'll remove the test requirement

JoshuaKGoldberg:
> TS 2.0.1

Great. Does that mean reviewing this PR will be blocked until 5.0?

> JSON schema

Is it ok to wait to standardize these into something other than an `any` type? #2428

ajafff:
> Does that mean reviewing this PR will be blocked until 5.0?

v5.0 is almost ready, see #2426

src/rules/completed-docs/tagExclusion.ts
```diff
@@ -0,0 +1,91 @@
...
+    };
+}
+
+export class TagExclusion extends Exclusion<ITagExclusionDescriptor> {
+    private readonly contentTags = this.descriptor.tags ? this.descriptor.tags.content : {};
+    private readonly existenceTags = new Set(this.descriptor.tags && this.descriptor.tags.existence);
+
+    public excludes(node: ts.Declaration) {
+        const documentationNode = this.getDocumentationNode(node);
+        const tagsWithContents = this.parseTagsWithContents(documentationNode.getFullText());
```
ajafff: if you want to get the jsdoc comment, you can simply do this:

```ts
const [firstChild] = documentationNode.getChildren();
if (firstChild.kind === ts.SyntaxKind.JSDocComment) {
    // firstChild is ts.JSDoc
    // do stuff with the comment
}
```
For more info on `JSDoc`, see https://github.com/Microsoft/TypeScript/blob/master/src/compiler/types.ts#L2065
You also don't need to parse it manually with the regex magic below. Typescript's parser already did that for you

src/rules/completedDocsRule.ts
```diff
@@ -387,24 +357,37 @@ class CompletedDocsWalker extends Lint.ProgramAwareRuleWalker {
...
+        if (node.name === undefined) {
+            return [];
         }
+
+        const symbol = typeChecker.getSymbolAtLocation(node.name);
+        if (symbol === undefined) {
+            return [];
+        }
+
+        return symbol.getDocumentationComment();
```
ajafff: starting from ts@2.1 you don't necessarily need the typechecker for that, see my other comment above

src/rules/completed-docs/tagExclusion.ts
ajafff: actually there may be more JSDocComments for a given node. but all of them precede nodes of other kinds.

src/rules/completedDocsRule.ts
JoshuaKGoldberg: Awesome thanks, I'll update this...

adidahiya:
friendly ping @JoshuaKGoldberg

JoshuaKGoldberg:
Sorry, been busy. :( should be able to fix this up within the week...

src/rules/completed-docs/tagExclusion.ts
JoshuaKGoldberg: Sorry @ajafff, I haven't been able to get that to work. Adding an if check for the node being any of the JSDoc kinds in a new `visitNode` isn't getting triggered. Either I'm missing something obvious or being silly. Are you ok with either taking a look yourself or letting this pass with a followup issue to clean this rule up (switch to the more efficient walker type and use the new APIs instead of type checking)?

JoshuaKGoldberg: Bump @ajafff ?

ajafff: @JoshuaKGoldberg `SyntaxWalker#visitNode` will never visit any JSDoc nodes, that's why I pointed you to `node.getChildren()` which does some magic to include jsdoc as the first entries if available.
However, I don't consider this a blocker for the PR. I'm fine with merging this without that change and tweaking it afterwards.

JoshuaKGoldberg: :+1: thanks. If this goes in I'll file a followup bug to do that and refactor this to the newer, more-performant walker style.

JoshuaKGoldberg:
@ajafff The moon and stars aligned again and I was able to get to this. Could you please take a look?

I still hope to send a followup PR after this to no longer need the type checker (especially considering #2767).

src/rules/completed-docs/blockExclusion.ts
```diff
@@ -0,0 +1,42 @@
+/**
+ * @license
+ * Copyright 2013 Palantir Technologies, Inc.
```
ajafff: 2017

PR #2421, opened 2017-03-28
fix(Linter): apply fix to the correct file

### LINKED ISSUE #2392: Apply fixes depending on the source file used in the failures
### Introduction

I'm working on a tool called [codelyzer](https://github.com/mgechev/codelyzer). It exports a few tslint rules which are Angular specific. Internally the tool resolves the templates and styles of Angular components, parses them and performs analysis over them.

Later, based on this analysis the tool produces diagnostics which not always should be reported in the source file which was initially analyzed by tslint, i.e. tslint may initially perform analysis over `app.component.ts` but the diagnostics can be reported to `app.component.html`. This happens by providing `app.component.html` as source file to the template visitor which extends `RuleWalker` internally.

### Actual behavior

Currently the produced replacements which should be applied to templates or styles are applied to the TypeScript file of the component which references them.

### Expected behavior

The fixes should be applied to the files where the failure is reported, i.e. if the failure is reported for `app.component.html` and the value of `failure.getFileName() === 'app.component.html'` then the fix should be applied to `app.component.html`.

### Fix

The fix will require change in `linter.ts`, something like:

```ts
...
  if (this.options.fix) {
    for (const rule of enabledRules) {
      const ruleFailures = this.applyRule(rule, sourceFile);
      source = fs.readFileSync(fileName, { encoding: "utf-8" });
      const fixsPerFile = this.getFixesPerFile(ruleFailures);
      if (Object.keys(fixsPerFile).length > 0) {
        this.fixes = this.fixes.concat(ruleFailures);
        Object.keys(fixsPerFile)
          .forEach(function (failureFileName) {
            const currentFixes = fixsPerFile[failureFileName];
            const content = fs.readFileSync(failureFileName, { encoding: "utf-8" });
            content = Fix.applyAll(content, currentFixes);
            fs.writeFileSync(failureFileName, content, { encoding: "utf-8" });
            if (fileName === failureFileName) {
              sourceFile = this.getSourceFile(failureFileName, source);
            }
          });
      }
      fileFailures = fileFailures.concat(ruleFailures);
    }
    hasLinterRun = true;
  }
...

private getFixesPerFile(ruleFailures: RuleFailure[]) {
  return ruleFailures
    .reduce((accum, c) => {
      const fn = c.getFileName();
      accum[fn] = accum[fn] || [];
      const fix = c.getFix();
      if (fix) {
        accum[fn].push(fix);
      }
      return accum;
    }, {})
}
```

Let me know what you think. Would love to open a PR.

The final goal is to provide automatic migration for the deprecations introduced by Angular 4.

**The change will not introduce any breaking changes and will not change the current behavior**.

mgechev:
Fix #2392

#### PR checklist

- [ ] Addresses an existing issue: #2392
  - [x] Includes tests
- [ ] Documentation update

#### Overview of change:

The spec that I added extends the `Linter` in order to test the extra
method I added. Since tslint doesn't have any rules which apply fixes in
a file different from the one which is currently being linted, I didn't
come up with additional tests.

The test "mocks" [a codelyzer rule](https://github.com/mgechev/codelyzer/blob/master/src/noAccessMissingMemberRule.ts).

### DIFF
--- a/src/linter.ts
+++ b/src/linter.ts
@@ -104,16 +104,8 @@ class Linter {
         if (this.options.fix) {
             for (const rule of enabledRules) {
                 const ruleFailures = this.applyRule(rule, sourceFile);
-                const fixes = ruleFailures.map((f) => f.getFix()).filter((f): f is Fix => !!f) as Fix[];
-                source = fs.readFileSync(fileName, { encoding: "utf-8" });
-                if (fixes.length > 0) {
-                    this.fixes = this.fixes.concat(ruleFailures);
-                    source = Replacement.applyFixes(source, fixes);
-                    fs.writeFileSync(fileName, source, { encoding: "utf-8" });
-
-                    // reload AST if file is modified
-                    sourceFile = this.getSourceFile(fileName, source);
-                }
+                source = this.applyFixes(fileName, source, ruleFailures);
+                sourceFile = this.getSourceFile(fileName, source);
                 fileFailures = fileFailures.concat(ruleFailures);
             }
             hasLinterRun = true;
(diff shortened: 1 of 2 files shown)


ajafff:
Out of curiosity: you plan to deprecate / drop `no-access-missing-member` and `template-use-public`. Will there still be a use case for your proposed change? Maybe `no-unused-css` I guess?

mgechev:
@ajafff and most importantly "a framework" for automatic migration of deprecations https://github.com/palantir/tslint/issues/2392.

mgechev:
FYI, here's the [fork in action](https://github.com/mgechev/ngmigrate) (still uses version 4 in the demo).

PR #2422, opened 2017-03-28
Remove `no-switch-case-fallthrough` and `no-invalid-this`

### LINKED ISSUE #1861: A few rules are mostly redundant in newer TypeScripts with specific compiler options
Here's the relevant rules of varying redundancy unaddressed in the docs:

- `no-invalid-this` &rarr; `--noImplicitThis` (not a precise equivalent, but it makes all usages explicit)
- `no-switch-case-fallthrough` &rarr; `--noFallthroughCasesInSwitch` (exact equivalent)
- `no-unused-variable` &rarr; `--noUnusedLocals` and `--noUnusedParameters` (exact equivalent between the two)
- `switch-default` is partially redundant with `--noImplicitReturns`, which requires that all code paths are reachable *and* return a value. (It only checks the end, but a final switch-case is rather common.)

nchen63:
fixes #1861 

#### CHANGELOG.md entry:

[removed-rule] `no-switch-case-fallthrough`: use tsc switch `--noFallthroughCasesInSwitch` instead
[removed-rule] `no-invalid-this`: use tsc switch `--noImplicitThis` instead
[recommended-removed] `switch-default` removed from recommended rule set


ajafff:
I'd rather not remove those rules now. Ideally we would wait until typescript adds the ability to disable such errors: https://github.com/Microsoft/TypeScript/issues/11051
`no-unused-variable` was undeprecated for that reason, too.

Keeping those rules around doesn't add much (if any) additional maintenance burden, so I don't see a reason to rush on removing them.

In addition, typescript will eventually start using `no-switch-case-fall-through` in their project, see https://github.com/Microsoft/TypeScript/pull/14886

adidahiya:
fair enough, I didn't realize the TS team is using no-switch-case-fall-through. @nchen63 let's keep the rules around and just remove `switch-default` from `tslint:recommended` in this PR

nchen63:
done in #2424

PR #2426, opened 2017-03-29
Prepare v5.0.0

adidahiya:


ajafff:
You should mention the changed output of the (json) formatter w.r.t fixes / replacements #2403

And probably also note the dropped support for ts 2.0 (even though as of now everything should still work)

Issue #2328, opened 2017-03-11
Run linting logic in spawned threads


JoshuaKGoldberg:
Followup to #389 & #1131. I'm taking a look at this for work because we have a few thousand files to lint. Running them in serial takes a _long_ time. Our options are:

1. Parallelize TSLint.MSBuild _(users of other build systems wouldn't get the benefits, so probably not)_
2. Have TSLint itself run across multiple threads / cores.

Within option 2 I see there being two possibilities:

1. Refactor the internals of `Runner` and `Linter` to be asynchronous
    * I think this would be a nice but making `Linter` in particular would be a breaking API change
    * Do you think you'd want to start to migrate to asynchronous file reading anyway?
2. Create a new `ProcessSplitter` class between `tslint-cli.ts` and `Runner`
    * Don't use it unless `--parallel` is provided and > 1
    * Have it spawn processes equal to `--parallel`
        * It would do its own file globbing to get the full list of files to lint
        * Each process would get `1/<--parallel>` of those files
    * After the processes are done it would collect and sort the results and report them as if they were synchronous

Thoughts?

ajafff:
Some random thoughts:

async processing:
* does async I/O work with `--type-check` or do we get race conditions from that?
* +1 to make I/O in `Runner` async
  * could use `glob-stream` for globbing to start linting as fast as possible (although that's not the bottleneck)
  * use a concurrent transform stream to read the file
  * use a write stream to do the linting
  * Runner should not throw away the `ts.Program` (created when `--project` is passed) when not type checking. We could then avoid the globbing and reuse the SourceFiles instead of reading the files again and creating new SourceFiles
* there should also be an async version of `Configuration` and `RuleLoader`
* making `Linter` async does not seem to be necessary?
  * the only I/O made here is used to apply fixes
  * unrelated: after applying fixes, use `ts.updateSourceFile` instead of `this.getSourceFile`

multiple threads:
* each thread would need to set up typescript, require all rules, cache the paths to the rule, ...
  * don't know if that would slow things down
  * requires much more memory

result cache:
* cache the linting result and only lint that file again if it changed in the meantime
* eslint has an option for this, too

andy-hanson:
Should probably do #2282 first.

JoshuaKGoldberg:
Glad the reaction has been favorable so far. :) I'll start work on some of the low hanging fruit.

[...]

PR #2438, opened 2017-03-30
Converted runner.ts to use Promises

JoshuaKGoldberg:
Moves all of `run`'s logic to a `runAsPromise` that uses async/await.

This uses the `mz` package. I'm curious on your opinions on this: should TSLint do work to wrap the `fs` APIs in Promises, or take on the package as a dependency?

Starts on #2328.


andy-hanson:
* A class with a single method invoked immediately after construction is a code smell. I don't think `Runner` is meant to be public (#2446). You could just use a function `run` returning a Promise.
* `status` could be a const enum.

ajafff:
> I don't think Runner is meant to be public

IIRC @JoshuaKGoldberg was the one who exported it, hence I guess it is used in the wild.

I don't really see a benefit in just using Promises but then awaiting them all of the time. As it is now there will only be overhead with nothing gained.
IMO it would make more sense to run async fs functions in parallel no matter if callback style or promisified.

JoshuaKGoldberg:
> code smell

Interesting. How would you prefer the runner logic be written? 

> public
> 2446

The idea behind separating out `Runner` from the cli logic was that external users might want to be able to use it instead of spinning up a whole Node sub-process. I specifically wanted to experiment with https://github.com/automutate/autotslint. I don't think anybody else uses it, so if it's not (yet/ever) part of the public contract I'm fine with changing `run`.

> `status` could be a const enum.

Agreed! #2182 - since nobody's responded, I think I can send a quick separate PR for this soon.

> overhead 

Even as a prerequisite for #2328?

andy-hanson:
> How would you prefer the runner logic be written?

I wasn't commenting on the way it was implemented but on the way it is exposed. You wouldn't do e.g. `new Adder(1, 2).add()`, so why have `new Runner().run()`?
In terms of implementation, just unindent everything, move the constructor parameters to `run`, and replace `public` with `export function` and `private` with `function` (adding explicit parameters for the members of `this` that were used there).

JoshuaKGoldberg:
I'm not sure I agree with the idea of using plain functions over a class for the runner logic. Right now it has two objects that are used, `options` and `outputStream`, which would have to be passed to (nearly) every function. That itself feels like code smell. What do we do when there are other hooks added, such as performance tracking or a file I/O interface? I see three roads there:
1. Pass a whole bunch of objects around to each function
2. Create a "dependencies" object and pass it to each function
3. Use a class

Is option 1 maintainable? Option 2 is made simpler by using 3 instead.

andy-hanson:
If you find that everything in your program needs to have access to everything else, the solution isn't to put it all in one class! It can be broken down into simple components. See #2572.

JoshuaKGoldberg:
Sure. I'd love to discuss this further but your PR looks better :)

andy-hanson:
Thanks!
But, mine doesn't use Promises, so you might want to leave this open...

JoshuaKGoldberg:
I'll send a new PR when yours is in.

PR #2445, opened 2017-03-31
no-void-expression: Add "ignore-arrow-function-shorthand" option

andy-hanson:
#### PR checklist

- [X] Addresses an existing issue: https://github.com/palantir/tslint/pull/1903#issuecomment-290431073, #2473
  - [X] Includes tests
- [X] Documentation update

#### Overview of change:

Add an option to allow `() => returnsVoid()`.

#### CHANGELOG.md entry:

[new-rule-option] `no-void-expression`: "ignore-arrow-function-shorthand"


src/rules/noVoidExpressionRule.ts
```diff
@@ -36,20 +48,31 @@ export class Rule extends Lint.Rules.TypedRule {
...
+        const allowArrowFunctionShorthand = this.ruleArguments.indexOf(OPTION_ALLOW_ARROW_FUNCTION_SHORTHAND) !== -1;
+        return this.applyWithFunction(sourceFile, (ctx) => walk(ctx, program.getTypeChecker()), allowArrowFunctionShorthand);
     }
 }
 
-class Walker extends Lint.ProgramAwareRuleWalker {
-    public visitNode(node: ts.Node) {
-        if (isPossiblyVoidExpression(node) && node.parent!.kind !== ts.SyntaxKind.ExpressionStatement && this.isVoid(node)) {
-            this.addFailureAtNode(node, Rule.FAILURE_STRING);
+function walk(ctx: Lint.WalkContext<boolean>, checker: ts.TypeChecker): void {
```
adidahiya: a `boolean` walk context looks awkward, let's just make an `IOptions` interface that has one field

src/rules/noVoidExpressionRule.ts
```diff
@@ -20,13 +20,25 @@ import * as ts from "typescript";
 import * as Lint from "../index";
 import { isTypeFlagSet } from "../language/utils";
 
+const OPTION_ALLOW_ARROW_FUNCTION_SHORTHAND = "allow-arrow-function-shorthand";
```
adidahiya: On second thought, can we name this "ignore-arrow-function-shorthand"? I know that we are inconsistent across the codebase in usage of the "allow" and "ignore" prefixes for rule option names (see https://github.com/palantir/tslint/issues/1371), but in this case it does seem more correct to say "ignore linting of arrow function shorthand syntax" because "allow-arrow-function-shorthand" suggests that the rule is checking for a collection of errors, one of which is arrow function shorthand syntax, and this rule option is lenient about that error &mdash; this is not the case; the shorthand syntax is not the error (rather, the implicit non-void return is).

An alternative name that makes sense would be `ignore-implicit-return` (it happens to be that arrow function shorthand is the only way to do that in TS)

Issue #938, opened 2016-01-25
import chai as external module in tests


adidahiya:
`test/assert.ts` is pretty clunky. we should remove it and instead write

``` ts
import { assert } from "chai"
```

in all our test files.

PR #2463, opened 2017-04-01
Explicitly import from chai instead of using a global

andy-hanson:
#### PR checklist

- [X] Addresses an existing issue: #938
  - [ ] Includes tests
- [ ] Documentation update

#### Overview of change:

Use `import { assert } from "chai"` everywhere instead of relying on globals.
This makes it possible to run a single test file using the mocha CLI.


andy-hanson:
Don't forget to close #938.

Issue #2223, opened 2017-02-20
arrow-return-shorthand incorrect


MrAndersen1:
### Bug Report

- __TSLint version__: `4.4.2`
- __TypeScript version__: `2.3.0-dev.20170220`
- __Running TSLint via__: `Visual Studio Code as well as directly on the command-line`

#### TypeScript code being linted

```ts
interface testInterface {
    "something": number;
}
[].map((e): testInterface => { return { "something": 1 }; });
```

#### Actual behavior

'This arrow function body can be simplified by omitting the curly braces and the keyword 'return'. (arrow-return-shorthand)'

#### Expected behavior

no warning as it cannot be simplified by omitting the curly bracers and the keyword 'return'.

MrAndersen1:
Upon reading more about it, however, it turns out you can replace the return and bracers with regular parenthesis ` => ({ "something": 1 })` so only the warning message would need updating for this case.

DethAriel:
Given the comment by @MrAndersen1 I think the warning can be enhanced to handle this type of return statement and suggest to also wrap the thing into brackets

icfantv:
I have to say, though, that wrapping the thing in parenthesis is not intuitive and doesn't look right - even though it may be.

MrAndersen1:
I believe this has been fixed even though it says here that just 3 of 4 tasks are complete.

PR #2466, opened 2017-04-02
arrow-return-shorthand: Improve failure message when return expression is an object literal

andy-hanson:
#### PR checklist

- [X] Addresses an existing issue: #2223
  - [X] Includes tests
- [ ] Documentation update

#### Overview of change:

Changes the failure message to mention that parentheses are needed when the expression is an object literal.

### DIFF
--- a/src/rules/arrowReturnShorthandRule.ts
+++ b/src/rules/arrowReturnShorthandRule.ts
@@ -19,6 +19,7 @@ import * as utils from "tsutils";
 import * as ts from "typescript";
 
 import * as Lint from "../index";
+import { hasCommentAfterPosition } from "../language/utils";
 
 const OPTION_MULTILINE = "multiline";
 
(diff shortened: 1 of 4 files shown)


src/rules/arrowReturnShorthandRule.ts
```diff
@@ -42,59 +43,60 @@ export class Rule extends Lint.Rules.AbstractRule {
...
+        // "return "
+        Lint.Replacement.deleteFromTo(statement.getStart(), expr.getStart()),
+        // " }" (may include semicolon)
+        Lint.Replacement.deleteFromTo(expr.end, closeBrace.end),
+    ];
 
-        function hasComments(node: ts.Node): boolean {
-            return ts.getTrailingCommentRanges(text, node.getEnd()) !== undefined;
-        }
+    function hasComments(node: ts.Node): boolean {
```
ajafff: there is a function called `hasCommentAfterPosition` in `src/language/utils.ts` which does basically the same and will also handle leading trivia of the next token

PR #2475, opened 2017-04-02
Simplify accesses to ruleArguments

andy-hanson:
#### PR checklist

  - [ ] Includes tests
- [ ] Documentation update

#### Overview of change:

Just cleaning up code. No need to go through `getOptions()` to get `ruleArguments`.

### DIFF
--- a/src/rules/cyclomaticComplexityRule.ts
+++ b/src/rules/cyclomaticComplexityRule.ts
@@ -73,7 +73,7 @@ export class Rule extends Lint.Rules.AbstractRule {
     }
 
     private get threshold(): number {
-        const ruleArguments = this.getOptions().ruleArguments;
+        const { ruleArguments } = this;
         if (ruleArguments[0] !== undefined) {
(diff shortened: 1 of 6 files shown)


src/rules/cyclomaticComplexityRule.ts
```diff
@@ -73,7 +73,7 @@ export class Rule extends Lint.Rules.AbstractRule {
     }
 
     private get threshold(): number {
-        const ruleArguments = this.getOptions().ruleArguments;
+        const { ruleArguments } = this;
```
adidahiya: minor: this is the one case where I dislike destructuring. While it may be verbose, I think it's nice to see `this.` for access of things on a class instance. Even with #2037 giving us the ability to guarantee that methods are declared as non-static only when strictly necessary, I'd like to keep the convention of `this.` for all property/method access like this. Would be open to a lint rule that forbids destructuring where the RHS is simply `this`.

PR #2477, opened 2017-04-02
interface-name: Convert to function

andy-hanson:
#### PR checklist

  - [ ] Includes tests
- [ ] Documentation update

#### Overview of change:

Just a refactor.

### DIFF
--- a/src/rules/interfaceNameRule.ts
+++ b/src/rules/interfaceNameRule.ts
@@ -15,9 +15,11 @@
  * limitations under the License.
  */
 
+import * as utils from "tsutils";
 import * as ts from "typescript";
 
 import * as Lint from "../index";
+import { isUpperCase } from "./variableNameRule";
 
 const OPTION_ALWAYS = "always-prefix";
 const OPTION_NEVER = "never-prefix";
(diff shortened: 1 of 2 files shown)


src/rules/interfaceNameRule.ts
```diff
@@ -47,60 +49,26 @@ export class Rule extends Lint.Rules.AbstractRule {
...
+    return ts.forEachChild(ctx.sourceFile, function cb(node: ts.Node): void {
+        if (utils.isInterfaceDeclaration(node)) {
+            const { name } = node;
+            if (never && hasPrefixI(name.text)) {
+                ctx.addFailureAtNode(name, Rule.FAILURE_STRING_NO_PREFIX);
+            } else if (!never && name.text[0] !== "I") {
+                ctx.addFailureAtNode(name, Rule.FAILURE_STRING);
             }
         }
+        return ts.forEachChild(node, cb);
```
ajafff: nit: you could move this line to the else branch, because interfaces cannot occur inside other interfaces

src/rules/interfaceNameRule.ts
```diff
@@ -15,9 +15,11 @@
  * limitations under the License.
  */
 
+import * as utils from "tsutils";
 import * as ts from "typescript";
 
 import * as Lint from "../index";
+import { isUpperCase } from "./variableNameRule";
```
JKillian: I'd recommend moving this to a separate file that is imported by both this rule and variableNameRule. Helps avoid circular dependencies and in a few edge cases, circular dependencies can cause runtime issues to pop up which is never fun. Also, no semantic reason for `isUpperCase` to live in `variableNameRule` imo

(sorry for the drive-by comment after this was merged! Just happened to notice it haha)

Issue #2472, opened 2017-04-02
Unexpected "Expression is always true" error.


vilic:
### Bug Report

- __TSLint version__: 5.0.0
- __TypeScript version__: 2.2.2
- __Running TSLint via__: CLI

#### TypeScript code being linted

```ts
class Foo { }

function foo(p: typeof Foo | object) {
  if (typeof p === 'object') {
    // ...
  }
}
```

With `strict-type-predicates` enabled.

#### Actual behavior

ERROR: Expression is always true.

#### Expected behavior

No error.

andy-hanson:
This depends on Microsoft/TypeScript#13502. `typeof Foo` is assignable to `Function`, but it's not itself a function type.

vilic:
@andy-hanson I am not familiar with that, what do you mean by "function type"? Does `typeof Foo` have a signature of `new (): Foo;`?

andy-hanson:
Actually, this is doable if we look for the construct signature.

PR #2479, opened 2017-04-02
strict-type-predicates: Anything with a construct signature is a function

andy-hanson:
#### PR checklist

- [X] Addresses an existing issue: #2472
  - [X] Includes tests
- [ ] Documentation update

#### Overview of change:

Check for construct signatures in `isFunction`.

#### CHANGELOG.md entry:

[bugfix] `strict-type-predicates`: Check for construct signatures in `isFunction`.


adidahiya:
are you sure this fixes #2472? we should consider adding a test using union types like in #2472.

PR #2481, opened 2017-04-02
Add support for error templates in tests

ajafff:
#### PR checklist

  - [x] Includes tests
- [x] Documentation update

#### Overview of change:
It's pretty unwieldy to write and maintain test for rules that contain a name in the error message (e.g. `prefer-const`, `no-unused-variable`, `no-shadowed-variable`, ...). I added basic template support to error message substitutions which uses node's `util.format()` to replace placeholders.
Syntax is inspired by python: `name % (substitution1[, substitution2, ...])`. I doubt this will accidentally break anyone's tests.

The added test shows best what can be done with this.
As an example I also rewrote the tests of `prefer-const` to use templates.

Do you have a better idea for the syntax?

I'll add documentation once there is consensus regarding the syntax.


nchen63:
Is it common to manually generate the .lint file errors? My workflow is usually to write out the file without the errors, run the test, then copy the generated actual output. This feature seems unnecessarily complex.

ajafff:
> Is it common to manually generate the .lint file errors?

At least I do. And given that @andy-hanson uses message substitutions a lot in his PRs, I think he also does.

Call it test driven development if you like. I want to test what I expect the implementation to do instead of using the output of the initial implementation for future testing.
Your workflow works for setting up the initial test. When the rule changes, e.g. finds more failures or changes its error message, you will also have to edit the test by hand.

src/test/parse.ts
```diff
@@ -114,6 +118,25 @@ export function parseErrorsFromMarkup(text: string): LintError[] {
     return lintErrors;
 }
 
+function substituteMessage(substitutions: Map<string, string>, message: string): string {
```
nchen63: can you rename `substitutions` to `templateMap`? It's confusing because the values that are put into the template are also substitutions. Maybe also rename `message` to something like `templateCall`

src/test/parse.ts
```diff
@@ -114,6 +118,25 @@ export function parseErrorsFromMarkup(text: string): LintError[] {
...
+    const substitution = substitutions.get(message);
+    if (substitution !== undefined) {
+        return substitution;
+    }
+    return formatMessage(substitutions, message);
+}
+
+function formatMessage(substitutions: Map<string, string>, message: string): string {
+    const formatMatch = /^([\w_]+) % \(\s*([^,]+(?:\s*,\s*[^,]+)*)\s*\)$/.exec(message);
```
nchen63: add a comment somewhere about the expected format of `message` and the substitutions since the regex is hard to parse

src/test/parse.ts
```diff
@@ -114,6 +118,25 @@ export function parseErrorsFromMarkup(text: string): LintError[] {
...
+    if (formatMatch !== null) {
+        const base = substitutions.get(formatMatch[1]);
+        if (base !== undefined) {
+            message = format(base, ...(formatMatch[2].trim().split(/\s*,\s*/g)));
```
nchen63: what if the argument has a comma or space? Consider using quotes around each arg

nchen63:
also needs docs. please update https://github.com/palantir/tslint/blob/master/docs/develop/testing-rules/index.md

src/test/parse.ts
ajafff: Re substitutions: It's not really a templateMap, it's a map of substitutions that may happen to contain a template

Re message: It's not necessarily a templateCall. It may be an identifier of a substitution, a "template call" or just plain text

I'd prefer leaving the names that way.

src/test/parse.ts
ajafff: Format arguments now need quotes. The implementation uses typescript's scanner to parse them. Escaping also works.

ajafff:
@nchen63 I updated the implementation and added docs

src/test/parse.ts
```diff
@@ -114,6 +121,74 @@ export function parseErrorsFromMarkup(text: string): LintError[] {
...
+    }
+    return formatMessage(substitutions, message);
+}
+
+/**
+ * Tries to format the message when it has the correct format or returned unchanged:  name % ('substitution1' [, "substitution2" [, ...]])
+ * Where `name` is the name of a message substitution that is used as format string.
+ * If `name` is not found in `substitutions`, `message` is returned unchanged.
+ */
+function formatMessage(substitutions: Map<string, string>, message: string): string {
```
nchen63: The reason I suggested the name change, is that if one were to look at just this function signature, you'd assume that `substitutions: Map<string, string>` is a map between placeholders and replacement values, and that `message` is the template string.

Also, I think it's fine calling these templates since template strings don't necessarily have placeholders

ajafff: That makes sense. I renamed it to `templates` hoping it's less confusing now

ajafff:
@nchen63 @adidahiya you both approved this PR. Can this be merged? I recently missed this feature while I refactored some rules and needed to edit their tests.

adidahiya:
@ajafff thanks for the reminder, sorry for the delay.

PR #2487, opened 2017-04-03
Add no-irregular-whitespace rule

QuentinFchx:
#### PR checklist

  - [x] Includes tests
- [ ] Documentation update

#### Overview of change:

Implementation of `no-irregular-whitespaces` rule from [eslint](http://eslint.org/docs/rules/no-irregular-whitespace)
Especially useful on macOS where pressing `Alt + Space` adds in a non breaking space.

- Shall we consider ignoring comments/strings? If so I'd like advice on how to achieve this through specific walkers 😄
- In the tests, I tried to specify directly unicode characters for invalid whitespaces, but they were escaped when using `node.getText()`. Is there another way?

#### CHANGELOG.md entry:

[new-rule] `no-irregular-whitespaces`


andy-hanson:
You don't need a RuleWalker for this, since you aren't actually traversing the AST.

This could be:

```ts
...
export class Rule {
	...
        return this.applyWithFunction(sourceFile, walk);
    }
}

function walk(ctx: Lint.WalkContext<void>): void {
    IRREGULAR_WHITESPACE.lastIndex = 0;
    let match: RegExpExecArray | null;
    while ((match = IRREGULAR_WHITESPACE.exec(ctx.sourceFile.text)) !== null) {
        const start = match.index;
        const len = match[0].length;
        const fix = new Lint.Replacement(start, len, " "); // Note: probably want to replace vertical whitespace with "\n" instead.
        ctx.addFailureAt(start, len, Rule.FAILURE_STRING, fix);
    }
}

// Inspired by: https://github.com/eslint/eslint/blob/master/lib/rules/no-irregular-whitespace.js
// tslint:disable-next-line max-line-length
const IRREGULAR_WHITESPACE = /[\f\v\u0085\u00A0\ufeff\u00a0\u1680\u180e\u2000\u2001\u2002\u2003\u2004\u2005\u2006\u2007\u2008\u2009\u200a\u200b\u202f\u205f\u3000\u2028\u2029]+/mg;

```

Be sure to include a test with multiple unusual spaces in a row, and a test with vertical whitespace (if that's in the regex there).

andy-hanson:
> In the tests, I tried to specify directly unicode characters for invalid whitespaces, but they were escaped when using node.getText().
As far as I can tell, the space in your `test.ts.lint` is a normal space. I pasted it into the console with `" ".charCodeAt(0)` and got `32` (normal space). Either git is doint something funny or you created the test wrong. You can generate copy-pastable wierd spaces by using e.g. `"\u2000"` in a node console.

QuentinFchx:
I'm still trying to generate a proper test file, do not review now :)
(Unless you can give me clues :D)

test/rules/no-irregular-whitespaces/test.ts.lint
```diff
@@ -0,0 +1,50 @@
...
+let foo;
+   ~     [Irregular whitespace not allowed]
+let　foo;
+   ~     [Irregular whitespace not allowed]
+let
foo;
+   ~     [Irregular whitespace not allowed]
+let
foo;
+   ~     [Irregular whitespace not allowed]
```
ajafff: also add a test case with multiple consecutive whitespaces
and a test with regular whitespace

src/rules/noIrregularWhitespacesRule.ts
```diff
@@ -0,0 +1,59 @@
...
+/* tslint:enable:max-line-length */
+
+function walk(ctx: Lint.WalkContext<void>): void {
+    IRREGULAR_WHITESPACE.lastIndex = 0;
+    let match: RegExpExecArray | null;
+    /* tslint:disable:no-conditional-assignment */
+    while ((match = IRREGULAR_WHITESPACE.exec(ctx.sourceFile.text)) !== null) {
+        const start = match.index;
+        const len = match[0].length;
+        const fix = new Lint.Replacement(start, len, " "); // Note: probably want to replace vertical whitespace with "\n" instead.
```
ajafff: consider using `" ".repeat(len)` to preserve formatting when there are more than one irregular spaces.

src/configs/all.ts
```diff
@@ -132,6 +132,7 @@ export const rules = {
     "max-file-line-count": [true, 1000],
     "max-line-length": [true, 120],
     "no-default-export": true,
+    "no-irregular-whitespaces": true,
```
adidahiya: the eslint rule is called [`no-irregular-whitespace`](http://eslint.org/docs/rules/no-irregular-whitespace). note the lack of trailing "s". please name it exactly the same (it's potentially confusing otherwise).

src/rules/noIrregularWhitespacesRule.ts
```diff
@@ -0,0 +1,59 @@
...
+
+/* Inspired by: https://github.com/eslint/eslint/blob/master/lib/rules/no-irregular-whitespace.js */
+/* tslint:disable:max-line-length */
+export const IRREGULAR_WHITESPACE = /[\u000b\u000c\u0085\ufeff\u00a0\u1680\u180e\u2000\u2001\u2002\u2003\u2004\u2005\u2006\u2007\u2008\u2009\u200a\u200b\u202f\u205f\u3000\u2028\u2029]+/mg;
+/* tslint:enable:max-line-length */
+
+function walk(ctx: Lint.WalkContext<void>): void {
+    IRREGULAR_WHITESPACE.lastIndex = 0;
+    let match: RegExpExecArray | null;
+    /* tslint:disable:no-conditional-assignment */
```
adidahiya: just disable the next line rather than the whole block. use `// tslint:disable-next-line no-conditional-assignment`

src/rules/noIrregularWhitespacesRule.ts
```diff
@@ -0,0 +1,59 @@
...
+    public static FAILURE_STRING = "Irregular whitespace not allowed";
+
+    public apply(sourceFile: ts.SourceFile): Lint.RuleFailure[] {
+        return this.applyWithFunction(sourceFile, walk);
+    }
+}
+
+/* Inspired by: https://github.com/eslint/eslint/blob/master/lib/rules/no-irregular-whitespace.js */
+/* tslint:disable:max-line-length */
+export const IRREGULAR_WHITESPACE = /[\u000b\u000c\u0085\ufeff\u00a0\u1680\u180e\u2000\u2001\u2002\u2003\u2004\u2005\u2006\u2007\u2008\u2009\u200a\u200b\u202f\u205f\u3000\u2028\u2029]+/mg;
```
adidahiya: minor: name this to identify it as a regex. `IRREGULAR_WHITESPACE_REGEX`

test/rules/no-irregular-whitespaces/test.builder.ts
```diff
@@ -0,0 +1,28 @@
...
+import { IRREGULAR_WHITESPACE } from '../../../src/rules/noIrregularWhitespacesRule';
+
+const matches = IRREGULAR_WHITESPACE.source
+    .substring(1, IRREGULAR_WHITESPACE.source.length - 2)
+    .match(/\\\w+/g)
+    .filter(match => ['\\u2029', '\\u2028'].indexOf(match) === -1) // those seems to break the parser
+    .map(match => JSON.parse(`"${match}"`) as string);
+
+function writeFile(path: string, content: string) {
+    fs.unlink(path, () => {
```
adidahiya: why do you need to unlink?

also, use the synchronous versions of these functions. we don't gain anything from async here.

test/rules/no-irregular-whitespaces/test.builder.ts
```diff
@@ -0,0 +1,28 @@
+import * as fs from 'fs';
```
adidahiya: how is this file expected to be run? can you include some documentation please?

QuentinFchx: I might by mistaken, but I did not find a way to directly specify unicode representations of these invalid whitespaces in the tests ([eslint does that](https://github.com/eslint/eslint/blob/master/tests/lib/rules/no-irregular-whitespace.js#L34)).

Since we need to represent these whitespaces anyway, I made this small script to avoid any error when modifying these files. However, I have no idea how this can be included within the test pipeline.
The other solution is to `tsc ./test.builder.ts && node ./test.builder.js` manually but I don't find it satisfying ([ts-node](https://github.com/TypeStrong/ts-node) might be an alternative, not sure if we want to add a devDependency though).

Do you have any suggestion around that?

adidahiya: nah, this is fine. thanks for including a top-level file comment.

Issue #2494, opened 2017-04-04
no-unsafe-any raises errors on import statements that are not typed as "any"


lddubeau:
### Bug Report

- __TSLint version__: 5.0.0
- __TypeScript version__: 2.2.2
- __Running TSLint via__: CLI: `tslint test.ts --type-check --project tsconfig.json`

#### TypeScript code being linted

```ts
import * as ts from "typescript";

// Prevent TypeScript from complaining about ts being unused.
console.log(ts.getCommentRange);
```

with `tslint.json` configuration:

```json
{
    "rules": {
        "no-unsafe-any": true
    }
}
```

I've tried with the following two `tsconfig.json`: 

```json
{
  "compilerOptions": {}
}
```

and:

```json
{
  "compilerOptions": {
    "rootDir": ".",
    "noEmitOnError": true,
    "noImplicitAny": true,
    "noImplicitReturns": true,
    "noImplicitThis": true,
    "skipDefaultLibCheck": true,
    "noUnusedParameters": true,
    "noUnusedLocals": true,
    "skipLibCheck": true,
    "strictNullChecks": true,
    "target": "es5"
  }
}
```

#### Actual behavior

The error:

```
test.ts[1, 21]: Unsafe use of expression of type 'any'.
```

#### Expected behavior

No error. I don't see why the imported symbols are somehow considered to be of type `any`. They obviously aren't since the IDE is able to perform completion that uses TypeScript type information.

amatiasq:
Same issue here: it warns for every expression at the first line of each method

```
ERROR: src/pathfinding/pathfinding.spec.ts[16, 12]: Unsafe use of expression of type 'any'.
ERROR: src/pathfinding/pathfinding.spec.ts[16, 12]: Unsafe use of expression of type 'any'.
ERROR: src/pathfinding/pathfinding.spec.ts[16, 12]: Unsafe use of expression of type 'any'.
ERROR: src/pathfinding/pathfinding.spec.ts[16, 35]: Unsafe use of expression of type 'any'.
ERROR: src/pathfinding/pathfinding.spec.ts[16, 12]: Unsafe use of expression of type 'any'.
ERROR: src/pathfinding/pathfinding.spec.ts[16, 12]: Unsafe use of expression of type 'any'.
ERROR: src/pathfinding/pathfinding.spec.ts[20, 22]: Unsafe use of expression of type 'any'.
[...]

PR #2496, opened 2017-04-05
no-unsafe-any: Allow import declarations

andy-hanson:
#### PR checklist

- [X] Addresses an existing issue: #2494
  - [X] Includes tests
- [ ] Documentation update

#### Overview of change:

Always allow import statements in `no-unsafe-any`.

#### CHANGELOG.md entry:

[bugfix] `no-unsafe-any`: Allow import declarations

### DIFF
--- a/src/rules/noUnsafeAnyRule.ts
+++ b/src/rules/noUnsafeAnyRule.ts
@@ -43,72 +43,185 @@ export class Rule extends Lint.Rules.TypedRule {
     }
 }
 
-// This is marked @internal, but we need it!
-const isExpression: (node: ts.Node) => node is ts.Expression = (ts as any).isExpression;
-
 function walk(ctx: Lint.WalkContext<void>, checker: ts.TypeChecker): void {
-    return ts.forEachChild(ctx.sourceFile, recur);
-    function recur(node: ts.Node): void {
-        if (isExpression(node) && isAny(checker.getTypeAtLocation(node)) && !isAllowedLocation(node, checker)) {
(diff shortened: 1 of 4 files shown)


src/rules/noUnsafeAnyRule.ts
```diff
@@ -91,10 +97,12 @@ function isAllowedLocation(node: ts.Expression, { getContextualType, getTypeAtLo
                     return false;
```
ajafff: you probably need to add a case for `CommaToken`

src/rules/noUnsafeAnyRule.ts
```diff
@@ -47,14 +47,13 @@ export class Rule extends Lint.Rules.TypedRule {
 const isExpression: (node: ts.Node) => node is ts.Expression = (ts as any).isExpression;
 
 function walk(ctx: Lint.WalkContext<void>, checker: ts.TypeChecker): void {
-    return ts.forEachChild(ctx.sourceFile, recur);
-    function recur(node: ts.Node): void {
+    return ts.forEachChild(ctx.sourceFile, function cb(node: ts.Node): void {
         if (isExpression(node) && isAny(checker.getTypeAtLocation(node)) && !isAllowedLocation(node, checker)) {
```
ajafff: I think it would increase performance if you swap the calls to `isAny(...)` and `isAllowedLocation(...)`

andy-hanson:
Did a complete rewrite. I also have a branch testing out applying the rule to tslint itself, but that will have to wait until the fixes are published globally (else `lint:from-global` won't work).

PR #2501, opened 2017-04-05
Tslib

jmlopez-rod:
#### PR checklist

  - [ ] Includes tests
- [ ] Documentation update

#### Overview of change:

The changes to the ts compiler configuration make it so that there is less code duplication. That is, instead of declaring the helper function such as `__extends` in every compiled file now we get it from [tslib](https://github.com/Microsoft/tslib).

For instance, currently we have the following in the `useIsNanRule`:

```js
var __extends = (this && this.__extends) || (function () {
    var extendStatics = Object.setPrototypeOf ||
        ({ __proto__: [] } instanceof Array && function (d, b) { d.__proto__ = b; }) ||
        function (d, b) { for (var p in b) if (b.hasOwnProperty(p)) d[p] = b[p]; };
    return function (d, b) {
        extendStatics(d, b);
        function __() { this.constructor = d; }
        d.prototype = b === null ? Object.create(b) : (__.prototype = b.prototype, new __());
    };
})();
Object.defineProperty(exports, "__esModule", { value: true });
var tsutils_1 = require("tsutils");
var ts = require("typescript");
var Lint = require("../index");
var Rule = (function (_super) {
    __extends(Rule, _super);
    function Rule() {
        return _super !== null && _super.apply(this, arguments) || this;
    }
    Rule.prototype.apply = function (sourceFile) {
        return this.applyWithFunction(sourceFile, walk);
    };
    return Rule;
}(Lint.Rules.AbstractRule));
```
After the change we get

```js
Object.defineProperty(exports, "__esModule", { value: true });
var tslib_1 = require("tslib");
var tsutils_1 = require("tsutils");
var ts = require("typescript");
var Lint = require("../index");
var Rule = (function (_super) {
    tslib_1.__extends(Rule, _super);
    function Rule() {
        return _super !== null && _super.apply(this, arguments) || this;
    }
    Rule.prototype.apply = function (sourceFile) {
        return this.applyWithFunction(sourceFile, walk);
    };
    return Rule;
}(Lint.Rules.AbstractRule));
```

That's quite a few lines that node won't have to read over and over for every file in this project.


src/tsconfig.json
```diff
@@ -7,6 +7,8 @@
         "noUnusedParameters": true,
         "noUnusedLocals": true,
         "strictNullChecks": true,
+        "noEmitHelpers": true,
```
ajafff: this line is not needed, it is set implicitly when using `importHelpers`

ajafff:
I was curious what this really saves, so I just tried it out:
16256 -> 15016 loc

jmlopez-rod:
weird that it failed... seems that it had trouble connecting with yarn 

```
error An unexpected error occurred: "https://registry.yarnpkg.com/typescript: connect ETIMEDOUT 104.16.63.173:443"
```

adidahiya:
I guess i'm not strictly opposed to this... but doesn't this change only really provide tangible improvements if you're bundling your JS modules for use in a browser where bandwidth & parsing time is a concern? Why are you reading the built JS files of a Node package anyway?

jmlopez-rod:
> Why are you reading the built JS files of a Node package anyway?

While I was trying to do update to the new version of tslint I was trying to read the implementation of the methods and I noticed all that duplicated code (too lazy to open the tslint project in another window with the actual ts files).

> but doesn't this change only really provide tangible improvements if you're bundling your JS modules for use in a browser where bandwidth & parsing time is a concern?

That is true, but I would think that less lines for node to execute is better? I know it is very little, but as we start adding more and more files they do add up. We are just doing unnecessary work (reading the definition to a function), especially when we already have a reference to them in one single library (tslib).

Anyway, since many people who want to reduce their js bundle size have this library already we could piggy back from it. I would still make it a `dependency` instead of a `peer` because not everyone uses it yet. Anyway, those are my two cents (two lines of changes).

Issue #2390, opened 2017-03-23
prefer-const: should ignore anything in a declaration


andy-ms:
### Bug Report

- __TSLint version__: `master` (using `npm link`)
- __TypeScript version__: `2.3.0-dev.20170323`
- __Running TSLint via__: CLI

#### TypeScript code being linted

```ts
declare namespace N {
    let x: number;
}

N.x = 3;
```

with `tslint.json` configuration:

```json
{
    "rules": {
        "prefer-const": true
    }
}
```

#### Actual behavior

`ERROR: a.ts[2, 9]: Identifier 'x' is never reassigned; use 'const' instead of 'let'.`

#### Expected behavior

No error. `prefer-const` should skip `.d.ts` files and anything marked `declare`.

adidahiya:
Can you provide some more context / motivation around this?

Why shouldn't we lint .d.ts files? Rules generally don't have the responsibility of deciding which files they are enabled in

andy-ms:
`prefer-const` detects whether something is mutated to determine whether it should be const.
In a declaration file, the variable will of course never be mutated, since it contains only declarations. It may be mutated by external code, but tslint has no way of knowing that.

adidahiya:
ok, I buy that reasoning. makes sense

PR #2513, opened 2017-04-06
no-var-keyword: Allow global var declarations

andy-ms:
#### PR checklist

  - [X] Includes tests
- [ ] Documentation update

#### Overview of change:

The `var` keyword is useful in one case: When you want to declare a global variable.
The problem with declaring a global `const` is that this is considered a block-scoped declaration, and will warn for duplicates.

```ts
declare global {
    var x: number;
    var x: number; // OK
    const y: number;
    const y: number; // Error
}
```

This PR changes the rule to allow `var` in a position where it *declares* a global variable.
It still warns for:

* `export var x: number;` (not global)
* `var x;` (not a declaration if not in `declare global`)

Note: Also disabled fixer in declaration files, because those should almost always be `const` instead of `let`, and we can't detect `prefer-const` in declarations (#2390).

#### CHANGELOG.md entry:

[enhancement] `no-var-keyword`: Allow global var declarations

### DIFF
--- a/src/language/utils.ts
+++ b/src/language/utils.ts
@@ -16,6 +16,7 @@
  */
 
 import * as path from "path";
+import { isBlockScopedVariableDeclarationList } from "tsutils";
 import * as ts from "typescript";
 
 import {IDisabledInterval, RuleFailure} from "./rule/rule";
@@ -52,10 +53,6 @@ export function hasModifier(modifiers: ts.ModifiersArray | undefined, ...modifie
  */
 export function isBlockScopedVariable(node: ts.VariableDeclaration | ts.VariableStatement): boolean {
     return isBlockScopedVariableDeclarationList(node.kind === ts.SyntaxKind.VariableDeclaration ? node.parent! : node.declarationList);
-}
-
-export function isBlockScopedVariableDeclarationList(node: ts.VariableDeclarationList): boolean {// tslint:disable-next-line no-bitwise
-    return isNodeFlagSet(node, ts.NodeFlags.Let | ts.NodeFlags.Const);
 }
 
 export function isBlockScopedBindingElement(node: ts.BindingElement): boolean {
--- a/src/language/utils.ts
+++ b/src/language/utils.ts
@@ -51,12 +51,11 @@ export function hasModifier(modifiers: ts.ModifiersArray | undefined, ...modifie
  * which indicates this is a "let" or "const".
  */
 export function isBlockScopedVariable(node: ts.VariableDeclaration | ts.VariableStatement): boolean {
-    const parentNode = (node.kind === ts.SyntaxKind.VariableDeclaration)
-        ? (node as ts.VariableDeclaration).parent
-        : (node as ts.VariableStatement).declarationList;
+    return isBlockScopedVariableDeclarationList(node.kind === ts.SyntaxKind.VariableDeclaration ? node.parent! : node.declarationList);
+}
 
-    return isNodeFlagSet(parentNode!, ts.NodeFlags.Let)
-        || isNodeFlagSet(parentNode!, ts.NodeFlags.Const);
+export function isBlockScopedVariableDeclarationList(node: ts.VariableDeclarationList): boolean {// tslint:disable-next-line no-bitwise
+    return isNodeFlagSet(node, ts.NodeFlags.Let | ts.NodeFlags.Const);
 }
 
 export function isBlockScopedBindingElement(node: ts.BindingElement): boolean {
--- a/src/rules/noVarKeywordRule.ts
+++ b/src/rules/noVarKeywordRule.ts
@@ -15,7 +15,7 @@
  * limitations under the License.
  */
 
-import { isModuleDeclaration, isVariableDeclarationList } from "tsutils";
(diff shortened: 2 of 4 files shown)


src/language/utils.ts
```diff
@@ -51,12 +51,11 @@ export function hasModifier(modifiers: ts.ModifiersArray | undefined, ...modifie
...
 export function isBlockScopedVariable(node: ts.VariableDeclaration | ts.VariableStatement): boolean {
-    const parentNode = (node.kind === ts.SyntaxKind.VariableDeclaration)
-        ? (node as ts.VariableDeclaration).parent
-        : (node as ts.VariableStatement).declarationList;
+    return isBlockScopedVariableDeclarationList(node.kind === ts.SyntaxKind.VariableDeclaration ? node.parent! : node.declarationList);
+}
 
-    return isNodeFlagSet(parentNode!, ts.NodeFlags.Let)
-        || isNodeFlagSet(parentNode!, ts.NodeFlags.Const);
+export function isBlockScopedVariableDeclarationList(node: ts.VariableDeclarationList): boolean {// tslint:disable-next-line no-bitwise
```
ajafff: this function is already available from `tsutils`, so there is no need to duplicate it here

src/rules/noVarKeywordRule.ts
```diff
@@ -37,46 +38,50 @@ export class Rule extends Lint.Rules.AbstractRule {
...
-                && !Lint.isBlockScopedVariable(node)) {
-            this.reportFailure(node.declarationList);
-        }
-
-        super.visitVariableStatement(node);
-    }
+function walk(ctx: Lint.WalkContext<void>): void {
+    const { sourceFile } = ctx;
+    return ts.forEachChild(ctx.sourceFile, function cb(node: ts.Node): void {
+        switch (node.kind) {
```
ajafff: you could simplify the condition and replace the whole switch statement with this:
```ts
if (isVariableDeclarationList(node) && !isBlockScopedVariableDeclarationList(node) &&
    (node.parent!.kind !== ts.SyntaxKind.VariableStatement || !isGlobalVarDeclaration(node.parent!)))
  fail(node);
```

you could even consider inlining `fail`

src/rules/noVarKeywordRule.ts
```diff
@@ -37,46 +38,50 @@ export class Rule extends Lint.Rules.AbstractRule {
...
+        return ts.forEachChild(node, cb);
+    });
 
-    public visitForOfStatement(node: ts.ForOfStatement) {
-        this.handleInitializerNode(node.initializer);
-        super.visitForOfStatement(node);
+    function fail(node: ts.Node): void {
+        // Don't apply fix in a declaration file, because may have meant 'const'.
+        const fix = sourceFile.isDeclarationFile ? undefined : new Lint.Replacement(node.getStart(), "var".length, "let");
+        ctx.addFailureAtNode(Lint.childOfKind(node, ts.SyntaxKind.VarKeyword)!, Rule.FAILURE_STRING, fix);
```
ajafff: there is no need for `addFailureAtNode` with `childOfKind`, you already computed the position of the var keyword for the fix one line above.

src/rules/noVarKeywordRule.ts
```diff
@@ -37,46 +38,50 @@ export class Rule extends Lint.Rules.AbstractRule {
...
+        || parent.kind === ts.SyntaxKind.ModuleBlock && isDeclareGlobal(parent.parent!);
+}
 
-    private reportFailure(node: ts.Node) {
-        const nodeStart = node.getStart(this.getSourceFile());
-        this.addFailureAt(nodeStart, "var".length, Rule.FAILURE_STRING,
-            this.createReplacement(nodeStart, "var".length, "let"));
-    }
+function isDeclareGlobal(node: ts.Node): boolean {
+    return isModuleDeclaration(node) && node.name.kind === ts.SyntaxKind.Identifier && node.name.text === "global";
```
ajafff: `ModuleBlock.parent` is always a `ModuleDeclaration`. `typescript@next` contains the correct types.
for now it is save to just assert the type

src/rules/noVarKeywordRule.ts
```diff
@@ -37,46 +38,50 @@ export class Rule extends Lint.Rules.AbstractRule {
...
-    }
+        return ts.forEachChild(node, cb);
+    });
 
-    public visitForOfStatement(node: ts.ForOfStatement) {
-        this.handleInitializerNode(node.initializer);
-        super.visitForOfStatement(node);
+    function fail(node: ts.Node): void {
+        // Don't apply fix in a declaration file, because may have meant 'const'.
+        const fix = sourceFile.isDeclarationFile ? undefined : new Lint.Replacement(node.getStart(), "var".length, "let");
```
ajafff: consider passing `sourceFile` as argument to `node.getStart()`

src/rules/noVarKeywordRule.ts
ajafff: you could even simplify this further:
```ts
return (node.flags & ts.NodeFlags.GlobalAugmentation) !== 0;
```
In addition you avoid false positived for `namespace foo.global {}` and `namespace foo { export namespace global {} }` (those are also good candidates for a test)

You can even consider inlining this condition.

PR #2515, opened 2017-04-06
Rewrite and fix object-literal-key-quotes

ajafff:
#### PR checklist

  - [x] Includes tests
- [ ] Documentation update

#### Overview of change:

[bugfix] `object-literal-key-quotes`: does no longer require quotes for property names containing digits
[bugfix] `object-literal-key-quotes`: correctly stringify numbers when fixing

### DIFF
--- a/test/rules/object-literal-key-quotes/always/test.ts.fix
+++ b/test/rules/object-literal-key-quotes/always/test.ts.fix
@@ -6,8 +6,8 @@ const o = {
   "hyphens-need-quotes": null,
   [computed]: 456,
   "123": "hello", // failure
-  "1e4": "something", // failure
-  ".123": "float", // failure
+  "10000": "something", // failure
+  "0.123": "float", // failure
   '123': 'numbers do not need quotes',
   '010': 'but this one does.',
   '.123': 'as does this one',
--- a/src/rules/objectLiteralKeyQuotesRule.ts
+++ b/src/rules/objectLiteralKeyQuotesRule.ts
@@ -15,8 +15,19 @@
  * limitations under the License.
  */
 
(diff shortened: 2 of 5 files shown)


yarn.lock
```diff
@@ -1674,10 +1674,14 @@ tslint@latest:
     tsutils "^1.1.0"
     update-notifier "^2.0.0"
 
-tsutils@^1.1.0, tsutils@^1.4.0:
+tsutils@^1.1.0:
   version "1.4.0"
   resolved "https://registry.yarnpkg.com/tsutils/-/tsutils-1.4.0.tgz#84f8a83df9967d35bf1ff3aa48c7339593d64e19"
 
+tsutils@^1.6.0:
```
adidahiya: sigh, let's try to de-dupe this entry again? not sure if it matters though

test/rules/object-literal-key-quotes/always/test.ts.fix
```diff
@@ -6,8 +6,8 @@ const o = {
   "hyphens-need-quotes": null,
   [computed]: 456,
   "123": "hello", // failure
-  "1e4": "something", // failure
-  ".123": "float", // failure
+  "10000": "something", // failure
```
adidahiya: how does the fix replacement know to expand 1e4 to 1000?

ajafff: `1e4` is a numeric literal that is automatically expanded to 1000 by the javascript vm before stringifying. Fixing this to `"1e4"` would actually break the program.

```ts
> Object.keys({1e4: "foo"})
[ '10000' ]
> String(1e4)==="10000"
true
```

ajafff:
@adidahiya rebased and resolved merge conflicts

PR #2519, opened 2017-04-07
Add no-unnecessary-type-assertion rule

mitchellwills:
#### PR checklist

  - [X] Includes tests

#### Overview of change:
Adds rule that warns when a cast has no effect (the uncast expression has the same type as the cast expression).

The tests are broken because the lint check typecheck is not running with strict null checks ("string|undefined" type gets reported as "string" by typechecker). Adding the strictNullChecks fixed this for me in an internal repo, but doesn't seem to work here for some reason (may be ts2.2 vs ts2.3?). Any ideas what needs to change to make it work?

I compare the two types using ===. From my testing this seems to work, but there doesn't seem to be any documentation that indicates this behavior can be relied on. Is there a preferred way to check the equality of types?


palantirtech:
Thanks for your interest in [palantir/tslint](https://github.com/palantir/tslint), @mitchellwills! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

mitchellwills:
Sure i'll take a stab at the changes you suggested. Any idea on how to enable strict null types for the tests (I guess I could just switch it to a union type, but it would probably be good to test both)?

mitchellwills:
Migrated to AbstractWalker and added/fixed tests

andy-hanson:
Microsoft/TypeScript#13502 would make this rule much better. Presumably it won't warn for `myCat as Animal` now.

mitchellwills:
Hmm yah that would be pretty neat. I wonder if some people would still want to use a cast since there are some cases where removing those casts would cause errors.

```
let x = 'a' as string|number;
x = 5;
```
Although they should probably use `let x: string|number = 'a';`. Don't know if people actually do that

mitchellwills:
re CLA: I am covered under the Google corporate CLA. I just made my membership of the google organization public though and it doesn't look like there's a way for me to retrigger the CLA bot.

src/rules/noUnnecessaryCastRule.ts
```diff
@@ -0,0 +1,81 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import * as ts from "typescript";
+import * as Lint from "../index";
+
+export class Rule extends Lint.Rules.TypedRule {
+  /* tslint:disable:object-literal-sort-keys */
+  public static metadata: Lint.IRuleMetadata = {
```
adidahiya: minor: please use 4-space indentation (as per the root `.editorconfig`)

src/rules/noUnnecessaryCastRule.ts
```diff
@@ -0,0 +1,81 @@
...
+    ruleName: "no-unnecessary-cast",
+    description: `Warns if a cast does not change the type of an expression.`,
+    options: null,
+    optionsDescription: "Not configurable",
+    type: "typescript",
+    typescriptOnly: true,
+    requiresTypeInfo: true,
+  };
```
adidahiya: looks like you implemented a fixer. can you add `hasFix: true` to the metadata?

src/rules/noUnnecessaryCastRule.ts
```diff
@@ -0,0 +1,81 @@
...
+  /* tslint:enable:object-literal-sort-keys */
+
+  public static FAILURE_STRING = "This cast is unnecessary.";
```
adidahiya: Please be more descriptive in the failure string and explain that the cast doesn't change the type of the expression. Also, to be slightly pedantic here (well within the rights of a _linter_), this is more correctly referred to as a type _assertion_ in TS syntax, so let's refer to it as such (we have other rules that relate to type assertions). The name should be changed to `no-unnecessary-type-assertion`.

`FAILURE_STRING = "This assertion is unnecessary since it does not change the type of the expression."`

src/rules/noUnnecessaryCastRule.ts
```diff
@@ -0,0 +1,81 @@
...
+
+    return ts.forEachChild(sourceFile, cb);
+  }
+
+  private verifyCast(node: ts.TypeAssertion|ts.NonNullExpression|
+                     ts.AsExpression) {
+    const castType = this.typeChecker.getTypeAtLocation(node);
+    const uncastType = this.typeChecker.getTypeAtLocation(node.expression);
+
+    if (uncastType && castType && uncastType === castType) {
```
adidahiya: please avoid implicit type coercion to boolean (I'd like to turn on `strict-boolean-expressions`). do explicit null checks `uncastType != null && castType != null && ...`

src/rules/noUnnecessaryCastRule.ts
mitchellwills: Reformatted all the files

src/rules/noUnnecessaryCastRule.ts
mitchellwills: Added

src/rules/noUnnecessaryCastRule.ts
mitchellwills: Added suggested failure string and renamed the rule everywhere. The AST is a little misleading here since it refers to "<type>expr" as a type assertion and the others as something else.

src/rules/noUnnecessaryCastRule.ts
mitchellwills: Fixed implicit boolean coercion

Issue #2509, opened 2017-04-06
no-unnecessary-callback-wrapper false positive


cartant:
### Bug Report

- __TSLint version__: 5.0.0
- __TypeScript version__: 2.2.2
- __Running TSLint via__: SublimeLinter

#### TypeScript code being linted

```ts
return this.patch(this.actions
    .ofType(...Object.keys(effects))
    .map((action) => effects[action.type](action))
    .first()
);
```

with `tslint.json` configuration:

```json
{
    "defaultSeverity": "error",
    "jsRules": {},
    "rules": {
        ...
        "no-unnecessary-callback-wrapper": { "severity": "warning" },
        ...
    }
}
```

#### Actual behavior

A warning is raised for the arrow function that's passed to the `map` operator.

#### Expected behavior

I would not have expected a warning, as the `action` parameter is used to look up the function that's to be called; it's not simply forwarded to a directly-callable function.

PR #2524, opened 2017-04-08
no-unnecessary-callback-wrapper: Handle non-identifier functions

andy-hanson:
#### PR checklist

  - [X] Includes tests
- [ ] Documentation update

#### Overview of change:

#2510 is too simple; there are cases of non-identifier calls that are still an unnecessary callback, like `x => f()(x)`. We can solve #2509 by looking for uses of the parameters in the called expression, so that `x => x(x)` is still not a failure.

#### CHANGELOG.md entry:

[enhancement] `no-unnecessary-callback-wrapper`: Handle non-identifier functions

### DIFF
--- a/src/rules/noUnnecessaryCallbackWrapperRule.ts
+++ b/src/rules/noUnnecessaryCallbackWrapperRule.ts
@@ -15,7 +15,8 @@
  * limitations under the License.
  */
 
-import { hasModifier, isArrowFunction, isCallExpression, isIdentifier, isSpreadElement } from "tsutils";
+import { hasModifier, isArrowFunction, isCallExpression, isElementAccessExpression, isIdentifier,
+    isPropertyAccessExpression, isSpreadElement } from "tsutils";
 import * as ts from "typescript";
 
 import * as Lint from "../index";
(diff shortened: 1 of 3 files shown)


ajafff:
I disagree with some changes made here. `(x) => foo()(x)` is no unnecessary wrapper, because `foo()` could have side effects. It would change the semantics of the program if the callback is never called or called more than once.
Every CallExpression whose expression is not an identifier should be allowed.

+1 for allowing `(x) => x(x)`

andy-hanson:
That's technically true... but that would be a pretty cruel place to hide a side effect. Most of the time code is better off evaluating `foo()` only once, instead of once per callback.

ajafff:
> but that would be a pretty cruel place to hide a side effect. Most of the time code is better off evaluating foo() only once, instead of once per callback.

Agreed, but that's definitely not the responsibility of this rule.

What if the arrow function is not used as callback. It's totally valid to write all functions as arrow function. If that's the case, side effects are expected and it's not just an "unnecessary wrapper"

andy-hanson:
OK, I've reverted to only check for identifiers.

src/rules/noUnnecessaryCallbackWrapperRule.ts
```diff
@@ -44,40 +44,42 @@ export class Rule extends Lint.Rules.AbstractRule {
...
             arg = arg.expression;
         }
-        if (!isIdentifier(name) || !isIdentifier(arg) || name.text !== arg.text) {
-            return false;
-        }
-    }
-    return true;
+        return isIdentifier(name) && isIdentifier(arg) && name.text === arg.text
+            // If the invoked expression is one of the parameters, bail.
+            && expression.text !== name.text;
```
ajafff: this is the only line that really changed
please consider keeping the diff small and just add this line to the source as it is on master.
Btw. sorry for the rather big refactor and therefore unnecessary big diff in #2510 :confused:

andy-hanson:
Updated changelog entry to reflect what this PR ended up actually doing.

Issue #1391, opened 2016-07-12
Making all number literals have consistent format


dusanlukic:
Thanks to JavaScript’s general float number type there are no differences between `1.0` and `1` or between `0.5` and `.5`.

We could add a rule to require leading `0.`, which would basically prevent number literals like `.1` and to require trailing `.0`. The later one would prevent number literals like `12` so it could be optional.

``` typescript
"number-format": {
    "pre": true | false, // whether to always include a preceding 0.
    "post": true | false // whether to always include a trailing .0
}
```

JoshuaKGoldberg:
I'm having problems trying to implement this rule. What is the `visit` call for number literals?

JoshuaKGoldberg:
@dusanlukic looks like this can be closed?

adidahiya:
closing for now. future requests can be new issues

PR #2526, opened 2017-04-08
Add `number-literal-format` rule

andy-hanson:
#### PR checklist

  - [X] Includes tests
- [X] Documentation update

#### Overview of change:

Added the `number-literal-format` rule, which warns on `.5` and `0.50` (both: prefer `0.5`).
This could also be extended to implement #1391 if anyone wants that.

#### CHANGELOG.md entry:

[new-rule] `number-literal-format`

### DIFF
--- a/src/rules/numberLiteralFormatRule.ts
+++ b/src/rules/numberLiteralFormatRule.ts
@@ -15,7 +15,7 @@
  * limitations under the License.
  */
 
-import * as utils from "tsutils";
+import { isNumericLiteral } from "tsutils";
 import * as ts from "typescript";
 
 import * as Lint from "../index";
@@ -33,7 +33,9 @@ export class Rule extends Lint.Rules.AbstractRule {
     };
     /* tslint:enable:object-literal-sort-keys */
 
+    public static FAILURE_STRING_LEADING_0 = "Exponent should not have a leading '0'.";
     public static FAILURE_STRING_TRAILING_0 = "Number literal should not have a trailing '0'.";
+    public static FAILURE_STRING_TRAILING_DECIMAL = "Number literal should not end in '.'.";
     public static FAILURE_STRING_LEADING_DECIMAL = "Number literal should begin with '0.' and not just '.'.";
 
     public apply(sourceFile: ts.SourceFile): Lint.RuleFailure[] {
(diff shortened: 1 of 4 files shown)


src/rules/numberLiteralFormatRule.ts
```diff
@@ -0,0 +1,69 @@
...
+        if (text.length <= 1) {
+            return;
+        }
+
+        if (text.startsWith(".")) {
+            ctx.addFailureAtNode(node, Rule.FAILURE_STRING_LEADING_DECIMAL);
+        }
+
+        // Allow '10', but not '1.0'
+        if (text.endsWith("0") && text.includes(".")) {
```
ajafff: you will get false positives for `1.1e10` and will not catch `1.0e1`
consider `node.getText(sourceFile).replace(/[eE]\d+^/, "")`

src/rules/numberLiteralFormatRule.ts
```diff
@@ -0,0 +1,69 @@
...
+    });
+
+    function check(node: ts.NumericLiteral): void {
+        // Apparently the number literal '0.0' has a '.text' of '0'.
+        const text = node.getText(sourceFile);
+        if (text.length <= 1) {
+            return;
+        }
+
+        if (text.startsWith(".")) {
```
ajafff: consider adding a check for `.endsWith('.')` for literals like `1.` (or `1.e1` - see my other comment)

src/rules/numberLiteralFormatRule.ts
```diff
@@ -0,0 +1,90 @@
...
+        if (isNumericLiteral(node)) {
+            return check(node);
+        }
+        return ts.forEachChild(node, cb);
+    });
+
+    function check(node: ts.NumericLiteral): void {
+        // Apparently the number literal '0.0' has a '.text' of '0', so use '.getText()' instead.
+        const text = node.getText(sourceFile);
+        if (text.includes("e")) {
```
ajafff: `e` can also be uppercase
also consider removing this if statement in favor of some thing like this:
```ts
const [num, exp] = text.split(/e/i);
if (exp !== undefined) {
   // check exponent for leading zero
}
// check num, you don't need an extra function and a callback for adding the failure
```

src/rules/numberLiteralFormatRule.ts
```diff
@@ -0,0 +1,90 @@
...
+                ctx.addFailureAt(node.getStart(sourceFile), num.length, msg);
+            });
+        } else {
+            return checkFormat(text, (msg) => { ctx.addFailureAtNode(node, msg); });
+        }
+    }
+}
+
+function checkFormat(text: string, fail: (message: string) => void): void {
+    if (text.length <= 1) {
```
ajafff: add `|| !text.includes(".")` and remove it from the other condition below

PR #2533, opened 2017-04-09
cyclomatic-complexity: Convert to function

andy-hanson:
#### PR checklist

  - [ ] Includes tests
- [ ] Documentation update

#### Overview of change:

Just a refactor.

### DIFF
--- a/src/rules/cyclomaticComplexityRule.ts
+++ b/src/rules/cyclomaticComplexityRule.ts
@@ -15,6 +15,7 @@
  * limitations under the License.
  */
 
+import { isFunctionScopeBoundary, isIdentifier } from "tsutils";
 import * as ts from "typescript";
 import * as Lint from "../index";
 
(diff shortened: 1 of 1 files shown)


src/rules/cyclomaticComplexityRule.ts
```diff
@@ -80,139 +78,53 @@ export class Rule extends Lint.Rules.AbstractRule {
...
-            case ts.SyntaxKind.AmpersandAmpersandToken:
-                this.incrementComplexity();
-                break;
-            default:
-                break;
+function walk(ctx: Lint.WalkContext<{ threshold: number }>): void {
+    const { options: { threshold } } = ctx;
+    let complexity = 0;
+
+    return ts.forEachChild(ctx.sourceFile, function cb(node: ts.Node): void {
```
ajafff: unrelated: I guess it's inteded to not check the complexity of the SourceFile?

andy-hanson: That's the way it worked before. Could be updated now, but it would be a breaking change.

Issue #2474, opened 2017-04-02
TSLint expect trailing comma on multiline function args in React


brunolm:
### Bug Report

- __TSLint version__: 4.5.1
- __TypeScript version__: 2.2.2
- __Running TSLint via__: VSCode

#### TypeScript code being linted

```ts
<div className="members">
  { members.map(member =>
    <img key={ member.id } src={ member.img } />
  ) }
</div>
```

with `tslint.json` configuration:

```json
{
  "extends": [
    "tslint:latest"
  ]
}
```

which adds

```json
    "trailing-comma": [
      true,
      {
        "multiline": "always",
        "singleline": "never"
      }
    ]
```

#### Actual behavior
`[tslint] Missing trailing comma (trailing-comma)` after img tag

#### Expected behavior
No errors

Is there a way to add exceptions for cases like this?

andy-hanson:
A simpler example:
```ts
[1, 2, 3].map(x =>
    x + 1 // Missing trailing comma
);
```

Not sure what the rule for an exception would be. Personally I would just format that as:

```ts
<div className="members">
  {
    members.map(member =>
      <img key={ member.id } src={ member.img } />);
  }
</div>
```

brunolm:
Cool, that works (without the semicolon there). Thank you.

I was initially thinking of something like:

```
"trailing-comma": [
  true,
  {
    "multiline": "always",
    "singleline": "never",
    "jsx": "never"
  }
]
```

I don't know how complex that would be and if it would be worth implementing that.

andy-hanson:
It doesn't look like this is really about jsx though. See my first example above.

Issue #2367, opened 2017-03-17
TSLint expect trailing comma on multiline function args, which is invalid JavaScript


rcchen:
### Bug Report

- __TSLint version__: 4.5.1
- __TypeScript version__: 2.1.6
- __Running TSLint via__: (pick one) CLI

#### TypeScript code being linted

```ts
const arr = [];
arr.push(
  "foo",
  "bar"
);
```

with `tslint.json` configuration:

```json
{
  "extends": ["tslint:latest", "tslint-react"]
}
```

#### Actual behavior

TSLint expects trailing comma after the last argument to `arr.push`

#### Expected behavior

TSLint should not expect trailing comma after the last argument as that's not valid JS syntax.

ajafff:
> as that's not valid JS syntax

But it will eventually be valid JS syntax. http://2ality.com/2015/11/trailing-comma-parameters.html

It's also valid typescript syntax. And that's what this package is made for.
If you don't like the default config provided by tslint, you can still disable the rule or set it to `"never"` to ban trailing commas completely. https://palantir.github.io/tslint/rules/trailing-comma/

rcchen:
Sure, but `tslint` picked up the ability to lint JS code in `4.0.0`, and as of date, it's not valid JS syntax.

Given a mixed JS/TS codebase where the linter is being applied across both types of files, there's an issue present.

adidahiya:
@rcchen you're right; the best thing you can do for now is have a separate `tslint.json` file for linting JS that extends your main tslint config and disables the `trailing-comma` rule. I see how that's non-ideal since you miss out on object & array literal checking. I think we should eventually make rules smarter about these kinds of things in the rare cases where JS / TS syntax diverges

adidahiya:
I don't think #2367 fixes this exactly. If someone _does_ want to lint for trailing commas in multiline function args in TS but also use the same config for JS files, they won't be able to do so with the built-in configs.

adidahiya:
Actually, I like @ajafff's suggestion for the solution (https://github.com/palantir/tslint/issues/2775#issuecomment-302139330). Just use the `jsRules` block to override lint behavior in JS files: 

```json
{
    "jsRules": {
        "trailing-comma": [true, {"multiline": {"arguments": "never", ...}}]
    }
}
```

Issue #2286, opened 2017-03-02
Feature Request: trailing-comma rule customization


anvish:
Allow separate settings for all covered cases, e.g. allow function parameters to have different settings from object initializers.

adidahiya:
Why? How does it help to treat these things differently?

smdern:
> Why? How does it help to treat these things differently?

Data structures and function signatures are not the same. Adding a trailing comma inside of a data structure makes it easier for you to add/remove re-order if necessary. Where as adding an additional argument to a function has greater consideration. 

Personally I prefer to see trailing commas on multiline objects + lists and no trailing commas when it comes to functions.

smdern:
Having a level of config similar to eslint would be ideal

http://eslint.org/docs/rules/comma-dangle

PR #2538, opened 2017-04-09
allow more specific config in trailing-comma rule

vitosamson:
#### PR checklist

- [x] Addresses an existing issue: #2286, #2474, #2367
  - [x] Includes tests
- [ ] Documentation update

#### Overview of change:
Allows additional config customization in the `trailing-comma` rule, similar to ESLint's [`comma-dangle`](http://eslint.org/docs/rules/comma-dangle) rule. This resolves issues such as https://github.com/palantir/tslint/issues/2286, https://github.com/palantir/tslint/issues/2474, https://github.com/palantir/tslint/issues/2367.

The rule config can be specified in the following ways:

```json
{
    "trailing-comma": [
        true,
        { "multiline": "always" }
    ]
}
```
or
```json
{
    "trailing-comma": [
        true,
        {
            "multiline": {
                "arrays": "always",
                "objects": "always",
                "functions": "never",
                "typeLiterals": "ignore",
                "imports": "ignore",
                "exports": "never"
            }
        }
    ]
}
```

This change is backwards compatible as setting `multiline` or `singleline` to a string will automatically apply that string to all the sub-options.

If the user specifies an object but doesn't include all the keys, any missing keys will default to `ignore`. I wasn't sure how strict to make the default - ESLint defaults to `never` for any missing keys. Open to any guidance if you want to make the default stricter.

#### CHANGELOG.md entry:

[new-rule-option] `trailing-comma` adds more granular options to specify trailing commas for arrays, objects, functions, type literals, imports, and exports


palantirtech:
Thanks for your interest in [palantir/tslint](https://github.com/palantir/tslint), @vitosamson! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

src/rules/trailingCommaRule.ts
```diff
@@ -20,11 +20,98 @@ import * as ts from "typescript";
...
+    functions?: OptionValue;
+    imports?: OptionValue;
+    objects?: OptionValue;
+    typeLiterals?: OptionValue;
+}
+
 interface Options {
-    multiline?: "always" | "never";
-    singleline?: "always" | "never";
+    multiline?: OptionValue | CustomOptionValue;
```
ajafff: You don't need `OptionValue` here, as you normalize it anyway. You don't need to cast when using it below

src/rules/trailingCommaRule.ts
```diff
@@ -20,11 +20,98 @@ import * as ts from "typescript";
...
+        options.multiline = {
+            arrays: options.multiline,
+            exports: options.multiline,
+            functions: options.multiline,
+            imports: options.multiline,
+            objects: options.multiline,
+            typeLiterals: options.multiline,
+        };
+    } else {
+        options.multiline = Object.assign({}, defaultOptions, options.multiline);
```
ajafff: as `undefined` is handled the same as `"ignore"`, I think there is no need to explicitly default everything to `"ignore"`.
Just use the user provided config object in this case

src/rules/trailingCommaRule.ts
```diff
@@ -86,21 +187,21 @@ class TrailingCommaWalker extends Lint.AbstractWalker<Options> {
                 case ts.SyntaxKind.NamedImports:
                 case ts.SyntaxKind.NamedExports:
                     this.checkList((node as ts.ArrayLiteralExpression | ts.BindingPattern | ts.NamedImportsOrExports).elements,
-                                   node.end);
+                                   node.end, node);
```
ajafff: add the check if the node should be handled in this switch statement instead of passing `node` and adding another big switch statement

test/rules/trailing-comma/multiline-custom/always/tslint.json
```diff
@@ -0,0 +1,32 @@
...
+        "trailing-comma": [
+            true,
+            {
+                "multiline": {
+                    "arrays": "always",
+                    "exports": "always",
+                    "functions": "always",
+                    "imports": "always",
+                    "objects": "always",
+                    "typeLiterals": "always"
```
ajafff: there should be a test that uses a mix of all possible values

src/rules/trailingCommaRule.ts
```diff
@@ -148,26 +249,66 @@ class TrailingCommaWalker extends Lint.AbstractWalker<Options> {
         }
         const token = getChildOfKind(node, closeTokenKind, this.sourceFile);
         if (token !== undefined) {
-            return this.checkComma(list.hasTrailingComma, list, token.end);
+            return this.checkComma(list.hasTrailingComma, list, token.end, node);
         }
     }
 
-    private checkList(list: ts.NodeArray<ts.Node>, closeElementPos: number) {
+    private checkList(list: ts.NodeArray<ts.Node>, closeElementPos: number, originalNode: ts.Node) {
```
ajafff: related to comment above: change this parameter to `option: keyof CustomOptionValue`

src/rules/trailingCommaRule.ts
vitosamson: Yeah, I wasn't sure about this. Since I'm assigning the string value to the object keys in `normalizeOptions` I assumed it would make the type checker think that I was assigning e.g. `options.multiline.arrays` a `CustomOptionValue` if `options.multiline` is only declared as a `CustomOptionValue`. But this doesn't seem to cause any issues.

src/rules/trailingCommaRule.ts
vitosamson: Fair point, are you OK with having `ignore` as the default value? I mentioned in the PR that I wasn't sure how strict the default should be.

ajafff: The options should default to ignore. At least that's what I'd expect it to do.

vitosamson: Also need to handle the case where the user doesn't specify a `multiline` or `singleline` in the config, in which case we'd just default to an empty object. I'm wondering if explicitly having the `defaultOptions` adds at least a bit of clarity to the code.

src/rules/trailingCommaRule.ts
vitosamson: This doesn't work for the first few cases where array, object, imports and exports all fall through to the same case handler. I'd have to add `this.checkList(...)` for each case in `ArrayLiteral`, `ArrayBinding`, `ObjectBinding`, etc.

ajafff: sure, you need to duplicate some case blocks, but that's ok
they are only grouped that way, because the structure of their AST node is similar. separating them might even make the code easier to understand

vitosamson:
@ajafff I made some of the changes you suggested and added tests for mixed options. I left the `defaultOptions` since that handles both cases where the user either provides an incomplete object or omits the key altogether, and IMO adds some clarity to the code that missing options will receive a `ignore` value. Let me know your thoughts.

src/rules/trailingCommaRule.ts
```diff
@@ -20,11 +20,98 @@ import * as ts from "typescript";
 
 import * as Lint from "../index";
 
+type OptionValue = "always" | "never" | "ignore";
+
+interface CustomOptionValue {
+    arrays?: OptionValue;
```
ajafff: no need to make the properties optional, `Object.assign` defaults them all to "ignore"

src/rules/trailingCommaRule.ts
```diff
@@ -20,11 +20,98 @@ import * as ts from "typescript";
...
+    arrays: "ignore",
+    exports: "ignore",
+    functions: "ignore",
+    imports: "ignore",
+    objects: "ignore",
+    typeLiterals: "ignore",
+};
+
+const normalizeOptions = (options: Options): Options => {
+    if (typeof options.multiline === "string") {
```
ajafff: consider making this a separate function as both `multiline` and `singleline` are handled the same.

src/rules/trailingCommaRule.ts
```diff
@@ -113,7 +219,8 @@ class TrailingCommaWalker extends Lint.AbstractWalker<Options> {
                 case ts.SyntaxKind.ConstructorType:
                 case ts.SyntaxKind.FunctionType:
                 case ts.SyntaxKind.CallSignature:
-                    this.checkListWithEndToken(node, (node as ts.SignatureDeclaration).parameters, ts.SyntaxKind.CloseParenToken);
+                    this.checkListWithEndToken(node, (node as ts.SignatureDeclaration).parameters,
+                                                ts.SyntaxKind.CloseParenToken, "functions");
```
ajafff: IIRC I have read a request to disinguish between `"arguments"` and `"parameters"` somewhere in the issues. I don't have strong opinion on this, so I leave it up to anyone else to make a decision

ajafff: ignore this comment, I think I mixed this up with something else

PR #2544, opened 2017-04-11
Update dependencies and lint for no-unsafe-any

andy-hanson:
#### PR checklist

  - [ ] Includes tests
- [ ] Documentation update

#### Overview of change:

The `no-unsafe-any` rule mostly just required appropriate casts, but in the case of `FormatterFunction` it seems to have discovered a bug -- formatters actually seem to have to be new-able. So created a `FormatterStatic` instead. This is technically a breaking change, but code that compiled against `FormatterFunction` would have been wrong.


src/language/formatter/formatter.ts
```diff
@@ -46,6 +46,10 @@ export interface IFormatterMetadata {
 
 export type ConsumerType = "human" | "machine";
 
+export interface FormatterStatic {
```
ajafff: rename to FormatterConstructor?

src/ruleLoader.ts
```diff
@@ -20,13 +20,12 @@ import * as path from "path";
...
 import { getRelativePath } from "./configuration";
 import { showWarningOnce } from "./error";
-import { AbstractRule } from "./language/rule/abstractRule";
-import { IDisabledInterval, IOptions, IRule } from "./language/rule/rule";
+import { IDisabledInterval, IOptions, IRule, RuleStatic } from "./language/rule/rule";
 import { arrayify, camelize, dedent } from "./utils";
 
 const moduleDirectory = path.dirname(module.filename);
 const CORE_RULES_DIRECTORY = path.resolve(moduleDirectory, ".", "rules");
-const cachedRules = new Map<string, typeof AbstractRule | null>(); // null indicates that the rule was not found
```
ajafff: actually I found this to be clearer

src/language/rule/rule.ts
```diff
@@ -20,6 +20,11 @@ import * as ts from "typescript";
 import {arrayify, flatMap} from "../../utils";
 import {IWalker} from "../walker";
 
+export interface RuleStatic {
```
ajafff: same as above

src/ruleLoader.ts
```diff
@@ -153,9 +152,9 @@ function loadCachedRule(directory: string, ruleName: string, isCustomPath = fals
         }
     }
 
-    let Rule: typeof AbstractRule | null = null;
+    let Rule: RuleStatic | null = null;
     if (absolutePath != null) {
```
ajafff: this comparison can be strict

src/rules/noInferredEmptyObjectTypeRule.ts
```diff
@@ -87,7 +87,7 @@ class NoInferredEmptyObjectTypeRule extends Lint.ProgramAwareRuleWalker {
         let isAnonymous: boolean;
         if (ts.ObjectFlags == null) {
             // typescript 2.0.x specific code
```
ajafff: this branch can be removed since support for ts 2.0.x was dropped in v5.0.0

src/utils.ts
```diff
@@ -44,7 +44,7 @@ export function objectify(arg: any): any {
  * E.g. "foo-bar" -> "fooBar"
  */
 export function camelize(stringWithHyphens: string): string {
-    return stringWithHyphens.replace(/-(.)/g, (_, nextLetter) => nextLetter.toUpperCase());
+    return stringWithHyphens.replace(/-(.)/g, (_, nextLetter) => (nextLetter as string).toUpperCase());
```
ajafff: consider adding the type annotation to the parameter instead of asserting it

src/ruleLoader.ts
andy-hanson: But rules aren't actually guaranteed to be implemented by `AbstractRule` subclasses, are they?

src/utils.ts
andy-hanson: I don't like that style since it behaves the same as a typecast but doesn't look like one. Was actually considering writing a lint rule for that.

src/ruleLoader.ts
```diff
@@ -123,18 +122,18 @@ function transformName(name: string) {
...
     if (fs.existsSync(fullPath + ".js")) {
-        const ruleModule = require(fullPath);
-        if (ruleModule && ruleModule.Rule) {
+        const ruleModule = require(fullPath) as { Rule: RuleConstructor } | undefined;
+        if (ruleModule !== undefined) {
             return ruleModule.Rule;
         }
     }
-    return undefined;
+    return null;
```
adidahiya: why the switch to nulls? can't we use undefined?

src/utils.ts
adidahiya: um, I don't think that's quite right @andy-hanson. type assertions (`as string`) are generally less safe than typedefs, because the assignability checks for the latter are stricter

src/runner.ts
```diff
@@ -193,8 +193,8 @@ export class Runner {
         try {
             this.processFiles(onComplete, files, program);
         } catch (error) {
-            if (error.name === FatalError.NAME) {
-                console.error(error.message);
+            if ((error as FatalError).name === FatalError.NAME) {
```
adidahiya: can you use `error: FatalError` on the line above instead?

src/rules/noUnsafeAnyRule.ts
```diff
@@ -36,7 +36,7 @@ export class Rule extends Lint.Rules.TypedRule {
     };
     /* tslint:enable:object-literal-sort-keys */
 
-    public static FAILURE_STRING = "Unsafe use of expression of type 'any'.";
+    public static FAILURE_STRING: string = "Unsafe use of expression of type 'any'.";
```
adidahiya: surely this isn't necessary?

src/rules/noConsecutiveBlankLinesRule.ts
```diff
@@ -52,11 +52,12 @@ export class Rule extends Lint.Rules.AbstractRule {
...
      */
     public isEnabled(): boolean {
-        return super.isEnabled() && (!this.ruleArguments[0] || this.ruleArguments[0] > 0);
+        const option = this.ruleArguments[0] as number | undefined;
+        return super.isEnabled() && (!option || option > 0);
     }
 
     public apply(sourceFile: ts.SourceFile): Lint.RuleFailure[] {
-        const limit: number = this.ruleArguments[0] || Rule.DEFAULT_ALLOWED_BLANKS;
+        const limit = this.ruleArguments[0] as number | undefined || Rule.DEFAULT_ALLOWED_BLANKS;
```
adidahiya: why the switch from typedef to type assertion?

src/ruleLoader.ts
ajafff: Now that it's renamed to `RuleConstructor` I'm fine with it.

src/utils.ts
andy-hanson: A type annotation in a callback is unsafe.
```ts
declare function f(cb: (x: {}) => void): void;
f((n: number) => { console.log(n + 1); }); // Huh, no error?
```

src/ruleLoader.ts
andy-hanson: We use `cachedRules.set(fullPath, Rule);`, so if `Rule` is undefined, it will look like we didn't cache a result at all.

src/runner.ts
andy-hanson: No, TS doesn't allow a type annotation there. It wouldn't be safe anyway since TS doesn't have checked exceptions.

src/rules/noUnsafeAnyRule.ts
andy-hanson: Whoops, no idea why that got there!

src/rules/noConsecutiveBlankLinesRule.ts
andy-hanson: A typedef `const x: number = thisIsAny;` is flagged by the `no-unsafe-any` rule, since it looks safe but isn't. `const x = thisIsAny as number;` is OK since that's obviously an assertion.

src/ruleLoader.ts
```diff
@@ -54,7 +53,7 @@ export function loadRules(ruleOptionsList: IOptions[],
                 } else {
                     const ruleSpecificList = enableDisableRules || [];
                     ruleOptions.disabledIntervals = buildDisabledIntervalsFromSwitches(ruleSpecificList);
-                    rules.push(new (Rule as any)(ruleOptions));
+                    rules.push(new (Rule as any)(ruleOptions) as IRule);
```
nchen63: don't need either cast

src/configuration.ts
```diff
@@ -15,6 +15,8 @@
  * limitations under the License.
  */
 
+// tslint:disable no-unsafe-any (TODO)
```
nchen63: why is this needed?

src/formatterLoader.ts
```diff
@@ -52,24 +52,24 @@ export function findFormatter(name: string | FormatterFunction, formattersDirect
...
 
-function loadFormatter(...paths: string[]) {
+function loadFormatter(...paths: string[]): FormatterConstructor | undefined {
     const formatterPath = paths.reduce((p, c) => path.join(p, c), "");
     const fullPath = path.resolve(moduleDirectory, formatterPath);
 
     if (fs.existsSync(`${fullPath}.js`)) {
         const formatterModule = require(fullPath);
-        return formatterModule.Formatter;
+        return formatterModule.Formatter; // tslint:disable-line no-unsafe-any
```
nchen63: instead of disabling rule, cast the `require` to an interface like  `{ Formatter: FormatterConstructor }`

src/formatterLoader.ts
```diff
@@ -52,24 +52,24 @@ export function findFormatter(name: string | FormatterFunction, formattersDirect
...
-function loadFormatterModule(name: string) {
+function loadFormatterModule(name: string): FormatterConstructor | undefined {
     let src: string;
     try {
         src = require.resolve(name);
     } catch (e) {
         return undefined;
     }
-    return require(src).Formatter;
+    return require(src).Formatter; // tslint:disable-line no-unsafe-any
```
nchen63: same as above comment

src/ruleLoader.ts
nchen63: maybe we should use a string literal like "not-found" or "load-error" instead of `null`

src/ruleLoader.ts
```diff
@@ -153,9 +152,9 @@ function loadCachedRule(directory: string, ruleName: string, isCustomPath = fals
         }
     }
 
-    let Rule: typeof AbstractRule | null = null;
-    if (absolutePath != null) {
-        Rule = loadRule(absolutePath, ruleName);
+    let Rule: RuleConstructor | null = null;
+    if (absolutePath !== undefined) {
+        Rule = loadRule(absolutePath, ruleName); // tslint:disable-line no-unsafe-any
```
nchen63: I don't think you need to disable the rule here

src/rules/memberOrderingRule.ts
```diff
@@ -396,6 +396,7 @@ function getOptionsJson(allOptions: any[]): { order: MemberCategoryJson[], alpha
         return { order: convertFromOldStyleOptions(allOptions), alphabetize: false }; // presume allOptions to be string[]
     }
 
+    // tslint:disable-next-line no-unsafe-any
     return { order: categoryFromOption(firstOption[OPTION_ORDER]), alphabetize: !!firstOption[OPTION_ALPHABETIZE] };
```
nchen63: instead of disabling, annotate `firstOption` as type `Options`, then call using `firstOption.order` and `firstOption.alphabetize`

src/tslint-cli.ts
```diff
@@ -15,6 +15,8 @@
  * limitations under the License.
  */
 
+// tslint:disable no-unsafe-any (TODO)
```
nchen63: I don't think you need to disable the rule here

src/ruleLoader.ts
```diff
@@ -20,13 +20,12 @@ import * as path from "path";
...
 import { showWarningOnce } from "./error";
-import { AbstractRule } from "./language/rule/abstractRule";
-import { IDisabledInterval, IOptions, IRule } from "./language/rule/rule";
+import { IDisabledInterval, IOptions, IRule, RuleConstructor } from "./language/rule/rule";
 import { arrayify, camelize, dedent } from "./utils";
 
 const moduleDirectory = path.dirname(module.filename);
 const CORE_RULES_DIRECTORY = path.resolve(moduleDirectory, ".", "rules");
-const cachedRules = new Map<string, typeof AbstractRule | null>(); // null indicates that the rule was not found
+const cachedRules = new Map<string, RuleConstructor | "not-found">(); // null indicates that the rule was not found
```
nchen63: can remove/update this comment

PR #2545, opened 2017-04-11
max-classes-per-file: Convert to function

andy-hanson:
#### PR checklist

  - [ ] Includes tests
- [ ] Documentation update

#### Overview of change:

Just a refactor. Also added a period to failure message.

### DIFF
--- a/src/rules/maxClassesPerFileRule.ts
+++ b/src/rules/maxClassesPerFileRule.ts
@@ -15,6 +15,7 @@
  * limitations under the License.
  */
 
+import { isClassLikeDeclaration } from "tsutils";
 import * as ts from "typescript";
 import * as Lint from "../index";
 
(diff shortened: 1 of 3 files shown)


src/rules/maxClassesPerFileRule.ts
```diff
@@ -47,48 +48,32 @@ export class Rule extends Lint.Rules.AbstractRule {
...
+    public static FAILURE_STRING(maxCount: number): string {
         const maxClassWord = maxCount === 1 ? "class per file is" : "classes per file are";
-        return `A maximum of ${maxCount} ${maxClassWord} allowed`;
+        return `A maximum of ${maxCount} ${maxClassWord} allowed.`;
     }
 
     public apply(sourceFile: ts.SourceFile): Lint.RuleFailure[] {
-        return this.applyWithWalker(new MaxClassesPerFileWalker(sourceFile, this.getOptions()));
+        const argument = this.ruleArguments[0];
+        const maxClasses = isNaN(argument) || (argument as number) < 0 ? 1 : argument;
```
ajafff: There's a subtle bug here. The old implementation defaults 0 to 1. The new one doesn't.
You can also simplify this line.
```ts
const maxClasses = argument > 0 ? argument : 1;
```

Issue #1940, opened 2016-12-27
[ban] Optional explanation for banned globals


vsiao:
https://github.com/palantir/tslint/pull/1385 added the ability to give an optional explanation for banned functions. eg. instead of just:

```js
["_", "filter"]
// Calls to '_.filter' are not allowed.
```

you could provide an additional string to propose an alternative:

```js
["_", "filter", "Use the native JavaScript 'myArray.filter' instead."]
// Calls to '_.filter' are not allowed. Use the native JavaScript 'myArray.filter' instead.
```

https://github.com/palantir/tslint/pull/1408 added global bans, but left out the ability to provide an optional explanation. Obviously there's the problem of input syntax (if it's an array with 2 strings `"a"` and `"b"`, is that banning `a.b` or banning `a` with message `b`?), but I'd like to open a discussion for that.

IllusionMH:
@vsiao as I can see in docs examples and [code](https://github.com/palantir/tslint/blob/master/src/rules/banRule.ts#L66-L67) - explanation isn't available for global bans.

Probably it is possible to check for characters that are not valid identifiers (spaces, special characters etc. since `ElementAccessExpression` aren't checked), or replace arrays with descriptive object
```json
{
    "path": ["object", "property"],
    "explanation": "Use native"
}
```

P.S. Is there any reason (except linted code became less readable) why `ElementAccessExpression` aren't checked in `ban` rule?

PR #2547, opened 2017-04-11
ban: Refactor and add new option format

ajafff:
#### PR checklist

- [x] Addresses an existing issue: #1940, #2821
  - [x] Includes tests
- [x] Documentation update

#### Overview of change:
Refactor and simplify banRule. 
[enhancement] `ban` new options format: allows to specify an optional explanation message for function bans, banning nested methods and using a wildcard for object of a method ban
Fixes: #1940
Fixes: #2821

This removes some ancient rather unintentional behavior, where banning `foo.bar()` would also ban `obj.foo.bar()` and even `fn().foo.bar()`.
I looked through the git history and PRs for the rule and it seems adding this "feature" was unintended.


test/rules/ban/tslint.json
```diff
@@ -2,8 +2,8 @@
   "rules": {
     "ban": [
       true,
-      ["xit"],
-      ["fit"],
+      "xit",
+      {"name": "fit"},
```
nchen63: add test for `message` in object

src/rules/banRule.ts
```diff
@@ -54,74 +66,66 @@ export class Rule extends Lint.Rules.AbstractRule {
...
-        } else if (bannedFunction.length >= 2) {
-            this.bannedFunctions.push(bannedFunction);
+function parseOptions(args: any[]): Options {
+    const functions: FunctionBan[] = [];
+    const methods: MethodBan[] = [];
+    for (const arg of args) {
+        if (typeof arg === "string") {
+            functions.push({name: arg});
+        } else if (Array.isArray(arg)) {
+            if (arg.length === 1) {
```
nchen63: handle arg.length == 0

src/rules/banRule.ts
```diff
@@ -14,34 +14,67 @@
...
         optionExamples: [
             [
                 true,
-                ["someGlobalMethod"],
-                ["someObject", "someFunction"],
-                ["someObject", "otherFunction", "Optional explanation"],
+                "eval",
+                {name: "$", message: "please don't"},
+                ["describe", "only"],
+                {object: "it", name: "only", message: "don't fucus tests"},
```
nchen63: "fucus" -> "focus"?

nchen63:
what if someone actually wanted to ban obj.foo.bar()?

ajafff:
> what if someone actually wanted to ban obj.foo.bar()?

That's really prone to errors, because it would also ban `obj2.foo.bar()` which may not be the desired behavior. Maybe we could extend the config to take an array for the `object` property.
For the above example you could then write: 
```json
{"object": ["obj", "foo"], "name": "bar"}
```

adidahiya:
bump @ajafff, can you address @nchen63's comments?

ajafff:
@adidahiya I actually did that and messed up a rebase recently ... will fix that

ajafff:
While thinking about #2821 I realized that there shouldn't be an `object` property. Instead `name` should be an array like it was before. But instead of limiting it to 2 elements for specifying the qualified name of the banned property, it should be open ended.
That would allow what @JoshuaKGoldberg requested in #2821 
```json
{"name": ["chai", "assert", "equal"], "message": "Use strictEqual instead of equal"}
```

Any thoughts about adding a wildcard to address @nchen63's comment https://github.com/palantir/tslint/pull/2547#issuecomment-293787648 to ban a property globally?
```js
{"name": ["*", "forEach"], "message": "Use a regular for loop instead"}
// bans [].forEach(...) as well as foo.forEach(...), foo.bar.forEach(...) and foo.bar.baz.forEach(...)
```

adidahiya:
> But instead of limiting it to 2 elements for specifying the qualified name of the banned property, it should be open ended.

sgtm 👍 

> Any thoughts about adding a wildcard

also seems fine, as the user is making it explicit that they are engaging in potentially wild behavior 😂

ajafff:
PR updated to allow banning nested methods as requested in #2821 and added a wildcard for method bans

PR #2552, opened 2017-04-11
callable-types: Refactor and add fixer

ajafff:
#### PR checklist

  - [x] Includes tests
- [x] Documentation update

#### Overview of change:
[bugfix] `callable-types`: suggest correct fix for interfaces with type arguments
[new-fixer] `callable-types`

### DIFF
--- a/src/rules/callableTypesRule.ts
+++ b/src/rules/callableTypesRule.ts
@@ -90,7 +90,7 @@ function renderSuggestion(call: ts.CallSignatureDeclaration,
         } else {
             res = `type ${parent.name.text} = ${res}`;
         }
-    } else if (res.endsWith(";") && parent.parent!.kind !== ts.SyntaxKind.TypeAliasDeclaration) {
+    } else if (res.endsWith(";")) {
         res = res.slice(0, -1);
     }
     return res;
--- a/src/rules/callableTypesRule.ts
+++ b/src/rules/callableTypesRule.ts
@@ -82,16 +82,14 @@ function renderSuggestion(call: ts.CallSignatureDeclaration,
     const start = call.getStart(sourceFile);
     const colonPos = call.type!.pos - 1 - start;
     const text = sourceFile.text.substring(start, call.end);
-    let res = text.substr(0, colonPos) + " =>" + text.substr(colonPos + 1);
+    const suggestion = text.substr(0, colonPos) + " =>" + text.substr(colonPos + 1);
 
     if (parent.kind === ts.SyntaxKind.InterfaceDeclaration) {
         if (parent.typeParameters !== undefined) {
-            res = `type${sourceFile.text.substring(parent.name.pos, parent.typeParameters.end + 1)} = ${res}`;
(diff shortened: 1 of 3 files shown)


src/rules/callableTypesRule.ts
```diff
@@ -29,76 +30,68 @@ export class Rule extends Lint.Rules.AbstractRule {
...
         }
-    }
+        return ts.forEachChild(node, cb);
+    });
 }
 
 /** True if there is no supertype or if the supertype is `Function`. */
-function noSupertype(heritageClauses: ts.NodeArray<ts.HeritageClause> | undefined): boolean {
-    if (!heritageClauses) {
+function noSupertype(node: ts.InterfaceDeclaration): boolean {
```
andy-hanson: Why did you change this to take `node` even when you're only using the `.heritageClauses`?

test/rules/callable-types/test.ts.lint
```diff
@@ -1,23 +1,36 @@
...
+
+interface L<T> {
+    (): T
+    ~~~~~ [Interface has only a call signature — use `type L<T> = () => T` instead.]
 }
 
 type T = {
     (): void;
-    ~~~~~~~~~ [Type literal has only a call signature — use `() => void` instead.]
+    ~~~~~~~~~ [Type literal has only a call signature — use `() => void;` instead.]
```
andy-hanson: Don't think this should have a semicolon. A type literal may not necessarily appear at the end of a statement, could be `function f(x: { (): void }): void;`, shouldn't fix to `function f(x: () => void;): void;`.

src/rules/callableTypesRule.ts
ajafff: it makes the signature shorter and the intention of the function clearer. Of course it could be destructured directly in the parameter, but I doubt this would make a difference

test/rules/callable-types/test.ts.lint
ajafff: it actually handles that already. The only exception is on type aliases. But since a TypeAlias should always end with a semicolon, I guess I can always remove the semicolon from the call signature

src/rules/callableTypesRule.ts
```diff
@@ -29,76 +30,68 @@ export class Rule extends Lint.Rules.AbstractRule {
...
+                          parent: ts.InterfaceDeclaration | ts.TypeLiteralNode,
+                          sourceFile: ts.SourceFile): string {
+    const start = call.getStart(sourceFile);
+    const colonPos = call.type!.pos - 1 - start;
+    const text = sourceFile.text.substring(start, call.end);
+    let res = text.substr(0, colonPos) + " =>" + text.substr(colonPos + 1);
+
+    if (parent.kind === ts.SyntaxKind.InterfaceDeclaration) {
+        if (parent.typeParameters !== undefined) {
+            res = `type${sourceFile.text.substring(parent.name.pos, parent.typeParameters.end + 1)} = ${res}`;
```
andy-hanson: Just return here instead of using `res =`.

Issue #2573, opened 2017-04-14
Crash with "TypeError: Cannot read property 'resolvedReturnType' of undefined" on return-undefined


Kovensky:
### Bug Report

- __TSLint version__: 5.1.0
- __TypeScript version__: 2.3.0
- __Running TSLint via__: tslint-loader

#### TypeScript code being linted

```ts
// reduced snippet
import $ from 'jquery'

$.Deferred()
.done(foo => {
  if (!foo) {
    return
  }
})
```

with `tslint.json` configuration:

```json
{
  "extends": [
    "tslint-config-standard"
  ]
}
```

#### Actual behavior

```
TypeError: Cannot read property 'resolvedReturnType' of undefined
    at Object.getReturnTypeOfSignature (.../node_modules/typescript/lib/typescript.js:31019:27)
    at getReturnKind (.../node_modules/tslint/lib/rules/returnUndefinedRule.js:121:30)
    at check (.../node_modules/tslint/lib/rules/returnUndefinedRule.js:76:34)
    at cb (.../node_modules/tslint/lib/rules/returnUndefinedRule.js:60:9)
    at visitEachNode (.../node_modules/typescript/lib/typescript.js:14674:30)
```

Also note that I had to edit `tslint/lib/linter.js` in order to be able to get anything more useful than just the error  message.

#### Expected behavior

No crash

Issue #2566, opened 2017-04-13
"return-undefined" requires explicit "return undefined" even for functions contextually-typed "any"


Kovensky:
### Bug Report

- __TSLint version__: 5.1.0
- __TypeScript version__: 2.3.0
- __Running TSLint via__: tslint-loader/webpack

#### TypeScript code being linted

```ts
$('.play-button').on('click', e => {
  if (playing) {
    return
  }
  // rest of function
})
```

with `tslint.json` configuration:

```json
{
  "extends": [
    "tslint-config-standard"
  ]
}
```

#### Actual behavior

Error on the `return`, requiring that it be a `return undefined`

#### Expected behavior

This function is a void function that just happens to be contextually typed to return `any` by the jquery type definitions; it should not be required to explicitly return `undefined`, especially not if the contextual type is `any`.

Issue #2565, opened 2017-04-13
Allow void async functions with "return-undefined"


Kovensky:
### Bug Report

- __TSLint version__: 5.1.0
- __TypeScript version__: 2.3.0
- __Running TSLint via__: webpack/tslint-loader

#### TypeScript code being linted

```ts
async function foo (bar: boolean) {
  if (!bar) {
    return
  }
}
```

with `tslint.json` configuration:

```json
{
  "extends": [
    "tslint-config-standard"
  ]
}
```

#### Actual behavior

`return-undefined` complains about the `return` not being a `return undefined`

#### Expected behavior

`return-undefined` should allow the `return` to stand, otherwise it is not possible to naturally write a void `async` function with early returns.

PR #2576, opened 2017-04-15
Fix various `return-undefined` bugs

andy-hanson:
#### PR checklist

- [X] Addresses an existing issue: #2573, #2566, #2565
  - [X] Includes tests
- [ ] Documentation update

#### Overview of change:

#2573: Handle a contextual type with ambiguous signatures and fall back to `getTypeAtLocation`.
#2566: Allow 'any'.
#2565: Treat an async function returning `Promise<void>` like a regular function returning `void`.

#### CHANGELOG.md entry:

[bugfix] `return-undefined`: Handle contextual types with ambiguous signatures; allow `any`; and handle async functions.

### DIFF
--- a/src/rules/returnUndefinedRule.ts
+++ b/src/rules/returnUndefinedRule.ts
@@ -113,7 +113,7 @@ function getReturnKind(node: FunctionLike, checker: ts.TypeChecker): ReturnKind 
     } else if (Lint.isTypeFlagSet(returnType, ts.TypeFlags.Void)) {
         return ReturnKind.Void;
     } else if (Lint.hasModifier(node.modifiers, ts.SyntaxKind.AsyncKeyword)) {
-        // TODO: Would need access to `checker.getPromisedTypeOfPromise` to do this properly.
+        // Would need access to `checker.getPromisedTypeOfPromise` to do this properly.
         // Assume that the return type is the global Promise (since this is an async function) and get its type argument.
         const typeArguments = (returnType as ts.GenericType).typeArguments;
         if (typeArguments !== undefined && typeArguments.length === 1) {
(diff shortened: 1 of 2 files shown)


src/rules/returnUndefinedRule.ts
```diff
@@ -143,3 +151,13 @@ function isFunctionLike(node: ts.Node): node is FunctionLike {
             return false;
     }
 }
+
+function isFunctionExpressionLike(node: ts.Node): node is ts.FunctionExpression | ts.ArrowFunction {
+    switch (node.kind) {
```
adidahiya: how about just `node.kind === ts.SyntaxKind.FunctionExpression || node.kind === ts.SyntaxKind.ArrowFunction`?

src/rules/returnUndefinedRule.ts
```diff
@@ -106,27 +103,38 @@ function getReturnKind(node: FunctionLike, checker: ts.TypeChecker): ReturnKind
...
 
-    const ty = contextualType || checker.getTypeAtLocation(node);
-    if (!ty) {
-        // Type error somewhere. Bail.
+    if (returnType === undefined) {
         return undefined;
+    } else if (Lint.isTypeFlagSet(returnType, ts.TypeFlags.Void)) {
+        return ReturnKind.Void;
+    } else if (Lint.hasModifier(node.modifiers, ts.SyntaxKind.AsyncKeyword)) {
+        // TODO: Would need access to `checker.getPromisedTypeOfPromise` to do this properly.
```
adidahiya: doesn't sound like this implementation is likely to change soon, so let's remove the "TODO" part and just consider this comment an implementation note rather than an expectation of a future change

PR #2585, opened 2017-04-16
Remove use of `Object.assign` and use `...` instead.

andy-hanson:
#### PR checklist

  - [ ] Includes tests
- [ ] Documentation update

#### Overview of change:

Exactly what it says in the commit message!


src/rules/trailingCommaRule.ts
```diff
@@ -56,7 +56,7 @@ const normalizeOptions = (options: Options): Options => {
             typeLiterals: options.multiline,
         };
     } else {
-        options.multiline = Object.assign({}, defaultOptions, options.multiline);
+        options.multiline = { ...defaultOptions, ...options.multiline };
```
ajafff: `options.multiline` can be undefined. Does typescript handle that case?
`option.singleline` can be undefined as well

andy-hanson: Yep, it's handled.

andy-hanson:
It's still possible since we use `--lib es6` anyway and the declaration wasn't necessary.

Issue #2564, opened 2017-04-13
semicolons rule detects unnecessary semis when in fact they are necessary.


trusktr:
### Bug Report

- __TSLint version__: 5.1.0
- __TypeScript version__: 2.2.1
- __Running TSLint via__: tslint-loader

#### TypeScript code being linted

```ts
switch((child as any).props['data-slot']) {
    case 'header':  header  = child; break
    case 'sidebar': sidebar = child; break
    case 'track':   track   = child; break
}
```

with `tslint.json` configuration:

```json
"semicolons": [true, "never"]
```

#### Actual behavior

```
ERROR in ./Timeline.tsx
/path/to/Timeline.tsx
Unnecessary semicolon (semicolon)
  14 |         React.Children.forEach(this.props.children, child => {
  15 |             switch((child as any).props['data-slot']) {
> 16 |                 case 'header':  header  = child; break
     |                                               ^
  17 |                 case 'sidebar': sidebar = child; break
  18 |                 case 'track':   track   = child; break
  19 |             }
```

#### Expected behavior

It should not complain about that.

As a workaround, I can either put the break on a new line, or just set semicolons to false to disable checking.

trusktr:
Or, does it just not want semis ever, even between statements on the same line?

Also the auto fixer will remove the semicolons and leave the code broken.

ajafff:
> Or, does it just not want semis ever, even between statements on the same line?

The rule tries to detect where ASI needs a semicolon. The behavior you are seeing is a bug.

PR #2591, opened 2017-04-16
semicolon: fix "never" option behavior when not followed by line break

ajafff:
#### PR checklist

- [x] Addresses an existing issue: #2564 
  - [x] Includes tests
- [ ] Documentation update

#### Overview of change:
[bugfix] `semicolon`: don't mark semicolon as unnecessary when the next statement is on the same line
Fixes: #2564

### DIFF
--- a/src/rules/semicolonRule.ts
+++ b/src/rules/semicolonRule.ts
@@ -97,7 +97,7 @@ class SemicolonWalker extends Lint.AbstractWalker<Options> {
                 case ts.SyntaxKind.ImportEqualsDeclaration:
                 case ts.SyntaxKind.DoStatement:
                 case ts.SyntaxKind.ExportAssignment:
-                    this.checkSemicolonAt(node);
+                    this.checkSemicolonAt(node as ts.Statement);
                     break;
                 case ts.SyntaxKind.TypeAliasDeclaration:
                 case ts.SyntaxKind.ImportDeclaration:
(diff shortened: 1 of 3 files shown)


src/rules/semicolonRule.ts
```diff
@@ -222,8 +222,19 @@ class SemicolonWalker extends Lint.AbstractWalker<Options> {
...
             }
         }
     }
+
+    private isFollowedByStatement(node: ts.Statement): boolean {
+        const nextStatement = utils.getNextStatement(node);
+        if (nextStatement === undefined) {
+            return false;
+        }
+        return ts.getLineAndCharacterOfPosition(this.sourceFile, node.end).line
```
andy-hanson: This should be a utility somewhere. `curlyRule.ts` also has an `areOnSameLine` helper. And there was some other situation I wanted a helper for this that I can't recall right now.
I think the best way to do this would be to look for a "\n" character between the end of node1 and the start of node2.
The current method would be tricked by a multi-line statement followed by another, i.e.:
```ts
const x = f(() =>
    0); const y = 2;
```

ajafff: > This should be a utility somewhere

Sure, will do that later on.

> The current method would be tricked by a multi-line statement followed by another

The rule handles your example as expected. I added it as a test to make sure this wont regress.

PR #2408, opened 2017-03-25
Lint for strict-boolean-expressions

andy-hanson:
#### PR checklist

- [ ] Addresses an existing issue: https://github.com/palantir/tslint/pull/2346#discussion_r108027137
- [ ] New feature, bugfix, or enhancement
  - [ ] Includes tests
- [ ] Documentation update

#### Overview of change:

Applies the `strict-boolean-expressions` rule to the code.

PR #2597, opened 2017-04-18
no-duplicate-variable: handle parameters

ajafff:
#### PR checklist

  - [x] Includes tests
- [x] Documentation update

#### Overview of change:
[new-rule-option] `no-dupliate-variable` adds `check-parameters` to check if variable has the same name as a parameter

This is needed because of the behavior change of `no-shadowed-variable` in my upcoming PR.
Also the name of the rule implies this behavior. At least I expected that...

The rule does not check for duplicate parameter names, because the compiler already does that.


adidahiya:
please merge develop, this will fail strict-boolean-expressions #2408

ajafff:
@adidahiya I can't confirm there is anything failing. Anyways, I rebased on master.

test/rules/no-duplicate-variable/test.ts.lint
```diff
@@ -108,6 +109,7 @@ function testArguments1(arg: number, arg: number): void {
 // failure: local var/let declarations shadow arguments.
 function testArguments2(x: number, y: number): void {
     var x = 1;
+        ~ [err % ('x')]
     let y = 2;
```
nchen63: why isn't `y` a problem?

ajafff: this rule doesn't handle block scoped declarations. these result in a compiler error, so there is no need for a linter error

adidahiya:
hmm, I dunno about this... parameters do _feel_ a little different from `var` declarations in the function body, which is why I put this check into `no-shadowed-variable`... is it really worth the breaking change? can you make the enhancements in #2598 without making this change @ajafff?

ajafff:
@adidahiya I don't think having this check in `no-shadowed-variable` is the right thing to do. `no-shadowed-variable` allows declarations with the same name in the same scope. They either merge (e.g. interface and class) or give a compiler error (e.g. two block scoped variables or block scoped variable and class). Parameters and variables merge if types match, so adding this additional check would be an ugly hack.

How about adding a new config option `"check-parameters"` to this rule to make it non-breaking?

ajafff:
@adidahiya I added a new options `"check-parameters"` to make this change non-breaking

PR #2339, opened 2017-03-14
Rewrite 'no-shadowed-variable'

andy-hanson:
#### PR checklist

- [X] Addresses an existing issue: #2214
- [X] New feature, bugfix, or enhancement
  - [X] Includes tests
- [X] Documentation update

#### Overview of change:

Complete rewrite of `no-shadowed-variable`.
Now checks for shadowing of types too (e.g. `interface Foo { m<Foo>(): Foo }`).
Also now no ignores variables named 'this'.

Issue #2214, opened 2017-02-17
Exempting "this" from the "no-shadowed-variable" rule


Rascaile:
### Bug Report

- __TSLint version__: `4.4.2`
- __TypeScript version__: `2.1.6`
- __Running TSLint via__: (please choose one) Visual Studio Code

#### TypeScript code being linted

```ts
namespace MyNamespace {

	class MyClass {
		public constructor(private myOtherClass: MyOtherClass) {}

		public myFunction(this: MyOtherClass): void {
			function handler(this: MyClass): void {}
		}
	}

	class MyOtherClass {
		public constructor() {}
	}
}
```

with `tslint.json` configuration:

```json
"no-shadowed-variable": true,
```

#### Actual behavior

    message: 'Shadowed variable: 'this' (no-shadowed-variable)' at: '7,21'

#### Expected behavior

Since typescript 2.0 we can specify the type of `this` in functions. So technically `this` is not a variable (since it's not a real parameter) that can be shadowed. I'd like it to be exempted it from this rule.

adidahiya:
looks like a bug. accepting PRs

Rascaile:
Thanks. My first instinct is to simply add an `if` statement in this rule's code to check for the first parameter and if it's name is `"this"`. But I'm not very familiar with tslint's source code so there might be a more elegant solution, like having the parameter iterator ignore the `this` assignments - but that might conflict with other rules that do need to include it. Any thoughts on this?

PR #2598, opened 2017-04-18
no-shadowed-variable: Rewrite and lint for other shadowed names

ajafff:
#### PR checklist

- [x] Addresses an existing issue: #2214, #2921
  - [x] Includes tests
- [ ] Documentation update

#### Overview of change:
[enhancement] `no-shadowed-variable`: added checks for other shadowing declarations, e.g. interfaces, classes, type parameters, imports, etc.
Fixes: #2921
[bugfix] `no-shadowed-variable`: exempt `this` parameter
Fixes: #2214 
[rule-change] `no-shadowed-variable` no longer fails for declarations in the same scope, e.g. `var foo; var foo;`. Use the rule `no-duplicate-variable` to find such errors.
Refs: #2597

The compiler already handles declarations in the same scope pretty well. I don't think we should duplicate its logic if declarations are mergeable (or its error messages if they are not). If the user doesn't want declarations to merge, there's still `no-duplicate-variable`

This is my attempt to simplify #2339.
Closes: #2339
/cc @andy-hanson


test/rules/no-shadowed-variable/test.ts.lint
```diff
@@ -215,5 +216,95 @@ function creatorFunction(constructor: MyConstructor, filter: MyFunction): void {
...
+            ~~~~ [err % ('test')]
+    }
+    let test;
+}
+
+// Allow mergeable declarations.
+interface I {}
+interface I {}
+
+// Ignore non-mergeable declarations as the compiler will complain about them anyway
```
andy-hanson: I don't see a compiler error (excepting the unused variable and implicit any) when I use that code. It's perfectly valid to write `function f() { const x = 0; return x; }`; the function `f` is shadowed.

src/rules/noShadowedVariableRule.ts
```diff
@@ -38,117 +39,137 @@ export class Rule extends Lint.Rules.AbstractRule {
...
+                addToList(parent.variablesSeen, name, identifiers);
+            });
+        }
     }
+}
 
-    private addFailureOnIdentifier(ident: ts.Identifier) {
-        const failureString = Rule.FAILURE_STRING_FACTORY(ident.text);
-        this.addFailureAtNode(ident, failureString);
+function addToList(map: Map<string, ts.Identifier[]>, name: string, identifiers: ts.Identifier[]) {
```
andy-hanson: Would make this more generic on `T` instead of `ts.Identifier`

test/rules/no-shadowed-variable/test.ts.lint
ajafff: good catch. I added some special handling for named function expressions

src/rules/noShadowedVariableRule.ts
ajafff: adding a type parameter doesn't add anything as this function is always called with the same types

andy-hanson: This function might need to be used elsewhere ([multimap](http://www.cplusplus.com/reference/map/multimap/)s show up often), but with a different type of values.

test/rules/no-shadowed-variable/test.ts.lint
```diff
@@ -215,5 +216,105 @@ function creatorFunction(constructor: MyConstructor, filter: MyFunction): void {
...
+class C<C> {
+        ~ [err % ('C')]
+    C<C>(C: C): C {
+      ~ [err % ('C')]
+         ~ [err % ('C')]
+        type C = C;
+             ~ [err % ('C')]
+        namespace C {
+                  ~ [err % ('C')]
+            interface C<C> {}
```
adidahiya: now that we're linting for more than just shadowed _variables_, I'm tempted to change the rule name in the next major release to `no-shadowed-name`...

for now, can we at least distinguish the failure messages? users might get confused if they see "shadowed variable" for an interface name or a type parameter. let's use `Shadowed variable: x` for variables and `Shadowed name: y` for everything else

ajafff: @adidahiya I changed the error message to always show `Shadowed name: x`.

Showing a different error message for variables is not really possible:
* We're not tracking if an identifier is a variable
* Variables could merge with classes, interfaces, namespaces ... are they still considered a variable
* And even if we could distinguish the kind of identifiers, it's not clear when to use which message: if a variable shadows a type? if a type shadows a variable? only if a variable shadows another variable?

ajafff:
Added handling of imports to address #2921

test/rules/no-shadowed-variable/test.ts.lint
```diff
@@ -198,8 +211,10 @@ function constructorsWithArgsParamConst(
     derivedCtor: new (...args: any[]) => any,
     baseCtors: Array<new (...args: any[]) => any>,
     args: any[]): void {
-    const args = [];
-          ~~~~       [Shadowed variable: 'args']
+    {
```
adidahiya: why the extra block here?

test/rules/no-shadowed-variable/test.ts.lint
adidahiya: yeah good points 👍

test/rules/no-shadowed-variable/test.ts.lint
ajafff: without the block there would be a compile error because the block scoped variable `args` cannot merge with the function scoped parameter `args`.
The block is needed here because the rule only handles shadowed names only in different scopes

Martinspire:
So why is this suddenly marking all my classnames as Shadowed Name?

```C:/<path>/<file>.ts
Shadowed name: 'SomeClassName' (no-shadowed-variable)
  37 |      * Some comment.
  38 |      */
> 39 |     export class SomeClassName implements SomeClassToImplement {
     |                 ^
  40 |         /**
  41 |          * Some more comments:


C:/<path>/<file>.ts
Shadowed name: 'SomeOtherClassName' (no-shadowed-variable)
  43 |      * Some comment.
  44 |      */
> 45 |     export class SomeOtherClassName implements SomeClassToImplement {
     |                 ^
  46 |         /**
  47 |          * Some more comments:


C:/<path>/<file>.ts
Shadowed name: 'AnotherClassName' (no-shadowed-variable)
  39 |      * Some comment.
  40 |      */
> 41 |     export class AnotherClassName implements SomeClassToImplement {
     |                 ^
  42 |         /**
  43 |          * Some more comments:
```

And if you look at how the classes are setup, it is like this:

```
import Something from "./something";

namespace SomeNamespace {

    /**
     * Comment about class
     */
    export class MyClass {
             <some functions, constructor and variables...>
        }
    }
}

import MyClass = SomeNamespace.MyClass;
export default MyClass;

```

I don't see how that would be faulty or shadowed

ajafff:
@Martinspire `import MyClass = SomeNamespace.MyClass;` is shadowed by the declaration of `MyClass` inside the namespace. In your case both point to the same class. The rule does not handle this specific case.

For now you can just `export default SomeNamespace.MyClass;`.
There's an open issue #3000 that tracks adding options to ignore certain declarations

Martinspire:
@ajafff but that doesn't work if you have multiple classes and interfaces to export. There can only be one default

Martinspire:
So for example if we want to keep using namespaces for typedoc, we use something like this:

```

export import ISomeItem= MyNameSpace.ISomeItem1;
export import ISomeItem2= MyNameSpace.ISomeItem2;
export import ISomeItem3= MyNameSpace.ISomeItem3;
export import ISomeItem4= MyNameSpace.ISomeItem4;
import SomeDefaultClass= MyNameSpace.SomeDefaultClass;
export default SomeDefaultClass;

```

And currently it is complaining, but there is no way to exclude this except for disabling the rule alltogether

PR #2599, opened 2017-04-18
Use isSameLine helper

ajafff:
#### PR checklist

  - [ ] Includes tests
- [ ] Documentation update

#### Overview of change:
This PR converts some uses of `ts.getLineAndCharacterOfPosition` to `tsutils.isSameLine` as suggested by @andy-hanson in https://github.com/palantir/tslint/pull/2591#discussion_r111692431

`curly` and `one-line` are not included, because they are not refactored yet

#### CHANGELOG.md entry:

[no-log]

### DIFF
--- a/src/rules/arrowReturnShorthandRule.ts
+++ b/src/rules/arrowReturnShorthandRule.ts
@@ -60,10 +60,10 @@ interface Options {
 
 function walk(ctx: Lint.WalkContext<Options>): void {
     const { sourceFile, options: { multiline } } = ctx;
-    ts.forEachChild(sourceFile, function cb(node: ts.Node): void {
+    return ts.forEachChild(sourceFile, function cb(node: ts.Node): void {
         if (utils.isArrowFunction(node) && utils.isBlock(node.body)) {
             const expr = getSimpleReturnExpression(node.body);
-            if (expr !== undefined && (multiline || !node.body.getText(sourceFile).includes("\n"))) {
+            if (expr !== undefined && (multiline || utils.isSameLine(sourceFile, node.body.getStart(sourceFile), node.body.end))) {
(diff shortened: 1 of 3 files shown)


src/rules/arrowReturnShorthandRule.ts
```diff
@@ -60,10 +60,10 @@ interface Options {
 
 function walk(ctx: Lint.WalkContext<Options>): void {
     const { sourceFile, options: { multiline } } = ctx;
-    ts.forEachChild(sourceFile, function cb(node: ts.Node): void {
+    return ts.forEachChild(sourceFile, function cb(node: ts.Node): void {
```
andy-hanson: Do you have any evidence that adding `return` makes it faster?

ajafff: I don't have any evidence and am reluctant to write a micro benchmark for this. I doubt the change here will be measurable at all.
It's just my understanding of what's helping the optimizing compiler.

andy-hanson: Time measurement isn't necessary if you think it means the difference between tail call and not tail call. If `return` made it into a tail call, then this code would not get a stack overflow:
```js
function f(x) {
    if (x === 0)
        return console.log("done");
    else
        return f(x - 1);
}
f(30000);
```

ajafff: @andy-hanson You're confusing tail call optimization (something the compiler *can* do) and tail call elimination (a language feature that the compiler must do).

The latter can be found in the ES Spec as proper tail calls. This feature will enable (almost) infinite recursion and would make your above example run without error. Unfortunately this is very unlikely to land in V8 because of https://github.com/tc39/proposal-ptc-syntax

Tail call opitimization on the other hand can optimize tail calls whether ptc is implemented or not. Unfortunately this does not enable deeper recursion.

PR #2616, opened 2017-04-20
Rewrite and simplify comment-format

ajafff:
#### PR checklist

  - [x] Includes tests
- [ ] Documentation update

#### Overview of change:
Initially I just wanted to convert the walker to a function. But then I needed to change so much it became a complete rewrite.

#### CHANGELOG.md entry:

[enhancement] `comment-format` no longer excludes comments with triple slash from linting except `/// <reference path="..."/>`


test/rules/comment-format/upper/test.ts.lint
```diff
@@ -10,7 +10,8 @@ class Clazz { // This comment is correct
         console.log("test"); //This comment has no space
                                ~~~~~~~~~~~~~~~~~~~~~~~~~ [space]
     }
-    /// <reference or something>
+    /// <not a reference>
+      ~~~~~~~~~~~~~~~~~~~ [space]
```
ajafff: Yeah, it should handle comments starting with more than 2 slashes better. The comment text should only be checked starting from the first non-slash character. 

And maybe there could be a new option to disallow more than 2 slashes

src/rules/commentFormatRule.ts
```diff
@@ -103,136 +115,87 @@ export class Rule extends Lint.Rules.AbstractRule {
...
 
-class CommentWalker extends Lint.RuleWalker {
-    private exceptionsRegExp: ExceptionsRegExp;
-    private failureIgnorePart: string = "";
-
-    constructor(sourceFile: ts.SourceFile, options: Lint.IOptions) {
-        super(sourceFile, options);
+function parseOptions(options: Array<string | IExceptionsObject>): Options {
+    return {
+        case: options.indexOf(OPTION_LOWERCASE) !== -1
```
nchen63: nested ternary is discouraged

src/rules/commentFormatRule.ts
```diff
@@ -103,136 +115,87 @@ export class Rule extends Lint.Rules.AbstractRule {
...
-                .join("|");
-
-            // regex is "start of string"//"any amount of whitespace"("any word from ignore list") followed by non alphanumeric character
-            return new RegExp(`^//\\s*(${wordsPattern})\\b`);
+        if (start === end) {
+            return;
+        }
+        const commentText = fullText.slice(start, end);
+        // whitelist //#region and //#endregion and JetBrains IDEs' "//noinspection ..."
+        if (/^(?:#(!?end)?region|noinspection\s)/.test(commentText)) {
```
nchen63: what is the `!?end` part for? This matches `!end`

ajafff: You're right, that should be `?:end`

Issue #2618, opened 2017-04-20
Rule suggestion: prefer object spread to Object.assign


danvk:
Now that TS supports the object-spread operator, the idea would be to suggest replacing code like this:

```ts
Object.assign({}, foo, bar);
```
with

```ts
{...foo, ...bar}
```

See https://github.com/airbnb/javascript#objects--rest-spread

ajafff:
@danvk You can already do that with the `ban` rule and add an appropriate message. This doesn't have automatic fixing though.

On second thought using `ban` would also disallow cases where you really want to assign properties of one object to another without creating a new one.

danvk:
Yeah, this has slightly more complex semantics than "ban" would allow. If
it's specifically mentioned in the AirBnb guide, I think it warrants a rule.

etsuo:
For what it's worth, an Object Spread does not appear to be the exact equivalent of Object.assign:
```
value = Object.assign(new Model(), value)
```
In this case, `value instanceOf Model` is true.
```
value = {...new Model(), ...value};
```
Here, value is an instance of Object.

This auto-fix just broke a bunch of my unit-tests... no biggie... just turned it off, but I figured I'd mention it in case someone thought the two were exactly equivalent.

adidahiya:
that's a good point @etsuo... the rule should probably make an exception for when the first argument to Object.assign is a `new SomeConstructor(...)` expression because spread is not functionally equivalent

danvk:
Is the issue just with new? Or with any object with a prototype other than
Object.prototype?

PR #2624, opened 2017-04-21
New rule prefer-object-spread

ajafff:
#### PR checklist

- [x] Addresses an existing issue: #2618
  - [x] Includes tests
- [x] Documentation update

#### Overview of change:
[new-rule] `prefer-object-spread`
Fixes: #2618
/cc @danvk 

I'm not sure if `functionality` is the right type in the metadata.


danvk:
Looks great once you get the tests passing!

ajafff:
In order to get the tests passing, I needed to change the sorting of the replacements. That could be considered a breaking change, but I rather think this is a bug fix. The tests of all other rules still pass.

The sorting of the replacements should be refactored afterwards. It should take into account if a replacement adds, removes or replaces text and sort accordingly.

PR #2629, opened 2017-04-23
no-conditional-assignment: rewrite and check conditional expressions

ajafff:
#### PR checklist

  - [x] Includes tests
- [x] Documentation update

#### Overview of change:
Rewrite to walker function.

The rule now allows intentional assignments like `if (foo(a = b));` or `if ((a || (a = b)).foo);`. If this is a typo the compiler should reveal that error.

#### CHANGELOG.md entry:

[bugfix] `no-conditional-assignment`: exclude intentional assignments, e.g. inside functions
[enhancement] `no-conditional-assignment` added check for conditional (ternary) expressions

### DIFF
--- a/src/rules/noConditionalAssignmentRule.ts
+++ b/src/rules/noConditionalAssignmentRule.ts
@@ -15,6 +15,10 @@
  * limitations under the License.
  */
 
+import {
+    isAssertionExpression, isAssignmentKind, isBinaryExpression, isConditionalExpression, isDoStatement, isForStatement, isIfStatement,
+    isNonNullExpression, isParenthesizedExpression, isPrefixUnaryExpression, isWhileStatement,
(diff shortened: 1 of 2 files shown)


seokirill:
Im so sorry, but is it should works how it works now?:
```typescript
let len = value ? value.length : 0;
```
Abowe codeline contains error coz it is ternary operator 🤔

seokirill:
I was wrong. error in:
```typescript
sum += !(bit ^= 1) ? parseInt(value[len], 10) : numbers[value[len]];
```

ajafff:
@seokirill the rule only forbids assignments inside of conditions.
Your first code snippet assigns a value according to the condition.
Your second example assigns a value in the condition. That's most likely a bug.

Issue #2625, opened 2017-04-21
Discussion: should developers blindly accept --fix'es


alexeagle:
I have heard of some developers running `tslint --fix` in their editor, to make the dev workflow the same as running a formatter.

I am concerned about this because I want tslint to have the freedom to:
- make fixes which might not be safe, eg. `var` -> `let`
- make fixes which try to restore developer intent rather than preserve semantics, eg. `s.trim()` -> `s = s.trim()`
- offer more than one suggested fix (we do this for some checks in http://ErrorProne.info for example)

Another way to say this is: I think a formatter should be blindly run because it has safety guarantees, and a linter's fixes should be inspected by a developer.

I think that conflating a formatter and a linter causes this rough edge, where we'll accidentally be forced into very conservative `--fix` behavior due to developer expectations. We should decide within the tslint project what the stance on this is and set developer expectations accordingly.

ajafff:
Given that the developer doesn't see what exactly changed and does not explicitly approve each fix before being applied, I'm strongly against unsafe fixes.

I'm against `var` -> `let`, because it might cause crashes at runtime. Even the compiler does not catch every variable access in the temporal dead zone. We could of course check if the fix is safe, but that would increase complexity without much benefit.
Same with `triple-equals`. Autofixing this could cause changes at runtime. `eslint` even removed the fixer from their `eqeqeq` rule for some time because of this.

Looking at `eslint` there are mostly safe fixes. The only exceptions seem to be `var` -> `let` and `==` -> `===`

If the user had the chance to assess every fix before applying, that would be a totally different story.

rich-newman:
@ajafff has asked me to comment, as the current behavior confused the hell out of me re no-var-keyword/prefer-const.

For a var in local scope (in a function) we allow a fix for var -> let (no-var-keyword).  We also allow a fix for let -> const if applied after the first fix (prefer-const).  However we currently won't allow the prefer-const rule to directly fix var -> const because we think it's too dangerous.  

This leads to rather weird behaviour currently with the recommended rule set: if you fix once a local var goes to let.  If you fix again it can go to const.

I tend to agree with @ajafff that if you run with --fix it shouldn't be possible for your code to break.  Particularly if you're running with the recommended set of rules.

If we do want potentially unsafe fixes maybe we have a --unsafe_fix option?

At the very least we should be consistent between rules though, so we don't get double fix issues like above.  That is, prefer-const and no-var-keyword should behave the same with respect to vars.

[...]

PR #2592, opened 2017-04-16
object-literal-sort-keys: rewrite, handle spread and shorthand

ajafff:
#### PR checklist

- [x] Addresses an existing issue: #1924, #2205, #2554 
- [x] New feature, bugfix, or enhancement
  - [x] Includes tests
- [x] Documentation update

#### Overview of change:
[new-rule-option] `object-literal-sort-keys` adds `ignore-case`
Fixes: #1924
[enhancement] `object-literal-sort-keys`: check shorthand properties
Fixes: #2205
[bugfix] `object-literal-sort-keys`: handle object spread correctly
Fixes: #2554

#### Is there anything you'd like reviewers to focus on?

Numbers are still excluded because it doesn't make sense to sort them alphabetically.
Methods and accessors are also still ignored.

PR #2634, opened 2017-04-25
WIP: Add fixer for object-literal-sort-keys

kastermester:
#### PR checklist

  - [x] Includes tests
- [ ] Documentation update

#### Overview of change:
This is a work in progress to add a fixer for the rule `object-literal-sort-keys`. I find sorting by hand quite annoying, especially when dealing with objects that have values that are multiline expressions.

It's a work in progress as there's a test case in here that completely breaks the code - I would love some feedback on exactly why and how this is happening such that I can get a proper fix for that ready.

#### CHANGELOG.md entry:
 * [new-fixer] `object-literal-sort-keys`

### DIFF
--- a/src/rules/objectLiteralSortKeysRule.ts
+++ b/src/rules/objectLiteralSortKeysRule.ts
@@ -47,6 +47,7 @@ class ObjectLiteralSortKeysWalker extends Lint.RuleWalker {
     private lastSortedKeyStack: string[] = [];
     private multilineFlagStack: boolean[] = [];
     private sortedStateStack: boolean[] = [];
+    private propertyNodes: Array<ts.NodeArray<ts.ObjectLiteralElementLike>> = [];
 
     public visitObjectLiteralExpression(node: ts.ObjectLiteralExpression) {
         // char code 0; every string should be >= to this
@@ -54,7 +55,9 @@ class ObjectLiteralSortKeysWalker extends Lint.RuleWalker {
         // sorted state is always initially true
         this.sortedStateStack.push(true);
         this.multilineFlagStack.push(this.isMultilineListNode(node));
+        this.propertyNodes.push(node.properties);
         super.visitObjectLiteralExpression(node);
+        this.propertyNodes.pop();
         this.multilineFlagStack.pop();
         this.lastSortedKeyStack.pop();
         this.sortedStateStack.pop();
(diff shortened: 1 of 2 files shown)


palantirtech:
Thanks for your interest in [palantir/tslint](https://github.com/palantir/tslint), @kastermester! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

ajafff:
@kastermester 
before I review your changes, there are some things to consider:
* changing the order of the properties may change the semantics, because the initializer can have side effects. There's an open issue #2625 to clarify if fixes are allowed to break the program.
```ts
let i = 0;
{
  b: ++i, // b == 1
  a: ++i, // a == 2
}
// would become
{
  a: ++i, // a == 1
  b: ++i, // b == 2
}
```
* the rule currenly only warns for the first out of order property and that's the only one that would be fixed. after fixing you may still have other unsorted properties, which can be pretty annoying
* #2592 is a complete rewrite of the rule.
  * There will be conflicts once that one lands.
  * Your changes should be easy to port to the new implementation.
  * You can then even remove the `this.propertyNodes` stuff you added.
  * #2592 add handling for object spread, so the fix must not sort beyond object spread : `{b, ...foo, c, a}` should be fixed to `{b, ...foo, a, c}` instead of `{a, b, ...foo, c}`

kastermester:
This makes sense :)

Seeing as there's discussion around the possibility of changing semantics of the program I won't comment on that here.

I fix the order of the entire object when warned on a single property, so that should not be an issue.
I have not handled the case of object spreads, but I can do that quite easily, but not sure if that is needed before the rewrite has landed?

adidahiya:
#2592 has merged, so the conflicts need to be resolved. And as @ajafff suggested, we should include tests for object spread fixes to verify that we preserve program semantics

ajafff:
@kastermester until #2625 is resolved, I'd suggest to fix only object literals without side effects.

I just added a function that checks for side effects to `tsutils`: https://github.com/ajafff/tsutils/commit/e8689de5789cc0c60ac52fdb28b7fa9bc6706c97

kastermester:
@ajafff Thanks, I will look into this :)

kastermester:
@dawsbot I still plan to finish this - but as it usually goes in life and at work - I have been busy lately :)

HideDev:
@kastermester @ajafff any news?

listepo:
@kastermester maybe you need help?

kastermester:
Sorry I have been very busy with work and personal things.

If someone wishes to take over that would be fine :) I can't make any promises on when I'll have time to look at this again.

jimbali:
Can't believe this doesn't exist yet

PR #2396, opened 2017-03-24
Fix type parameter of applyWithFunction

ajafff:
### PR checklist

- [ ] Addresses an existing issue: #0000
- [x] New feature, bugfix, or enhancement
  - [ ] Includes tests
- [ ] Documentation update

#### Overview of change:
Type of `options` should not be used to infer type parameter `T`.
Refs: https://github.com/Microsoft/TypeScript/issues/14829

#### CHANGELOG.md entry:
[no-log]

PR #2660, opened 2017-05-01
Really fix type parameter of applyWithFunction

ajafff:
#### PR checklist

- [x] Addresses an existing issue: #2396
  - [ ] Includes tests
- [ ] Documentation update

#### Overview of change:
Followup on #2396:
Using `U extends T` allows excess properties.
Using `{} & T` does stricter type checking (no more excess properties and enables linting the parameter with `no-unsafe-any`) and also fixes the inference problem with arrow functions: https://github.com/palantir/tslint/pull/2396#issuecomment-290993472

This change may technically break dependent libraries if they explicitly specified the type parameters. In reality possibly noone ever did that. On the other hand this change may reveal bugs during type checking.
#### CHANGELOG.md entry:

[api] `AbstractRule#applyWithFunction` has better type checking for its type parameter


nchen63:
`{} & T` doesn't seem to be encouraged https://github.com/Microsoft/TypeScript/issues/14829#issuecomment-298427189

ajafff:
Seems like this is now the preferred way to exclude a parameter from type inference. It's even suggested by TypeScript devs https://github.com/Microsoft/TypeScript/issues/17491#issuecomment-318720788

ajafff:
CI failed to compile. Looks like this solution will not work for all cases. 
I'll investigate further and try to find a working solution.

ajafff:
I managed to work around the type error. Cross posting https://github.com/Microsoft/TypeScript/issues/14829#issuecomment-322267089 for reference

> `T & {}` only works when not using `strictNullChecks`.
>I encountered a problem when `T` is `undefined`, because `undefined` is not assignable to `undefined & {}`.
>
>I found that mapped types also defer inference, so i tried `type InferLast<T> = Readonly<T>;`
>That does not work for types with index signatures like arrays.
>
>My final solution is a mapped type intersected with the type itself:
>
>```ts
>export type NoInfer<T> = T & {[K in keyof T]: T[K]};
>```

PR #2666, opened 2017-05-02
Prepare v5.2.0

nchen63:


src/configs/recommended.ts
```diff
@@ -96,6 +87,7 @@ export const rules = {
     "no-invalid-this": false,
     "no-misused-new": true,
     "no-namespace": true,
+    "no-object-literal-type-assertion": true,
```
adidahiya: isn't this a breaking change? this should go in latest, not recommended

PR #2669, opened 2017-05-03
Enable 'prefer-switch' rule and raise default 'min-cases'

andy-hanson:
#### PR checklist

  - [ ] Includes tests
- [X] Documentation update

#### Overview of change:

Just enables the rule and fixes new lint failures.
**EDIT**: Also raises the default `min-cases`.

#### CHANGELOG.md entry:

[enhancement] `prefer-switch`: Change default `min-cases` to 3.

### DIFF
--- a/src/language/utils.ts
+++ b/src/language/utils.ts
@@ -339,44 +339,58 @@ export function forEachComment(node: ts.Node, cb: ForEachCommentCallback) {
 
 /** Exclude leading positions that would lead to scanning for trivia inside JsxText */
 function canHaveLeadingTrivia(tokenKind: ts.SyntaxKind, parent: ts.Node): boolean {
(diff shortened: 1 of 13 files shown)


src/language/utils.ts
```diff
@@ -339,44 +339,58 @@ export function forEachComment(node: ts.Node, cb: ForEachCommentCallback) {
...
+                    // there can only be leading trivia if we are at the end of the top level element
+                    return parent.parent!.parent!.kind !== ts.SyntaxKind.JsxElement;
+                default:
+                    return true;
+            }
+
+        default:
+            return true;
     }
-    return true;
```
ajafff: you could leave this `return` and remove the `default` clauses above

andy-hanson: I like the `default` so it's easy to tell that there's no fallthrough.

PR #2670, opened 2017-05-03
Enable 'no-unnecessary-type-assertion' rule

andy-hanson:
#### PR checklist

  - [ ] Includes tests
- [ ] Documentation update

#### Overview of change:

Just enables the rule and fixes new lint failures.
The two disables look like bugs an inconsistency between typescript's language service and compilation.


src/rules/noEmptyInterfaceRule.ts
```diff
@@ -48,7 +48,7 @@ function walk(ctx: Lint.WalkContext<void>) {
             (node.heritageClauses === undefined ||
              node.heritageClauses[0].types === undefined ||
```
ajafff: you can remove this line, because it can never be undefined

andy-hanson: `strict-type-predicates` would have caught that. ⚡️

ajafff: true, but this change of the type was only introduced recently with typescript@2.3

I'd like to enable `strict-type-predicates`, too. But there will be problems with some TypeChecker APIs that are not yet compatible with strictNullChecks

adidahiya:
still some merge conflicts

PR #2676, opened 2017-05-04
Enable 'deprecation' rule

andy-hanson:
#### PR checklist

  - [ ] Includes tests
- [ ] Documentation update

#### Overview of change:

Just enables the rule and fixes failures.

### DIFF
--- a/src/rules/deprecationRule.ts
+++ b/src/rules/deprecationRule.ts
@@ -57,11 +57,11 @@ class Walker extends Lint.ProgramAwareRuleWalker {
     }
 
     for (const d of decSym.getDeclarations()) {
+      let commentNode: ts.Node = d;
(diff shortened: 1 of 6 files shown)


src/rules/deprecationRule.ts
```diff
@@ -57,11 +57,11 @@ class Walker extends Lint.ProgramAwareRuleWalker {
     }
 
     for (const d of decSym.getDeclarations()) {
+      let commentNode: ts.Node = d;
```
ajafff: out of curiosity, did the rule make that change necessary?

andy-hanson: Yes, the comment included the string "@deprecated", so the declaration of `commentNode` was considered deprecated.

PR #2681, opened 2017-05-04
Remove 'findup' dependency

andy-hanson:
#### PR checklist

  - [ ] Includes tests
- [ ] Documentation update

#### Overview of change:

`findup-sync` seems needlessly complex for our usecase. It could be replaced by a single function.
The implementatino is made more complex due to the `nocase` option having been used. This option attempts to find the file in a case-insensitive manner even on case-sensitive file systems. I think we should at least warn if we see a file like `Tslint.JSON`.


src/configuration.ts
```diff
@@ -122,8 +121,8 @@ export function findConfigurationPath(suppliedConfigFilePath: string | null, inp
         }
     } else {
         // search for tslint.json from input file location
-        let configFilePath = findup(CONFIG_FILENAME, { cwd: inputFilePath, nocase: true });
-        if (configFilePath != null && fs.existsSync(configFilePath)) {
+        let configFilePath = findup(CONFIG_FILENAME, inputFilePath);
+        if (configFilePath !== undefined && fs.existsSync(configFilePath)) {
```
ajafff: is the `fs.existsSync` here necessary? I would expect the file to exist, because we found it in the file system

src/configuration.ts
```diff
@@ -142,6 +141,35 @@ export function findConfigurationPath(suppliedConfigFilePath: string | null, inp
...
+        }
+        directory = parent;
+    }
+
+    function findFile(cwd: string): string | undefined {
+        if (fs.existsSync(path.join(cwd, filename))) {
+            return filename;
+        }
+
+        // Try reading in the entire directory and looking for a file with different casing.
```
ajafff: maybe we should deprecate that behavior. It's already inconsistent with searching in home directory which is case sensitive.

src/configuration.ts
```diff
@@ -142,6 +141,35 @@ export function findConfigurationPath(suppliedConfigFilePath: string | null, inp
 }
 
 /**
+ * Find a file by name in a directory or any ancestory directory.
+ * This is case-insensitive, so it can find 'TsLiNt.JsOn' when searching for 'tslint.json'.
+ */
+function findup(filename: string, directory: string): string | undefined {
```
ajafff: do we want to implement some caching in this function?

andy-hanson: We shouldn't be calling `findConfigurationPath` many times anyway.

ajafff: In fact we are calling it a lot. When not passing `-c` on the command line, `Runner` calls `findConfiguration` for every time is sees a file in another directory than the previous.

ajafff:
Does this really work like `findup`, especially when passing absolute paths?

andy-hanson:
Check for yourself -- the [implementation](https://github.com/js-cli/node-findup-sync/blob/master/index.js) of findup is pretty simple if you ignore globs (which we aren't using).
Could you make a separate PR for caching once this is in?

andy-hanson:
@nchen63 We still haven't resolved whether we should warn if we see `tslint.json` with different capitalization.

nchen63:
I'd be surprised if people were the using non-lowercase filename, but deprecating and removing this   in the next major rev seems like the right thing to do (probably a little overkill)

PR #2694, opened 2017-05-06
fix build

nchen63:
[no-log]


otaran:
Using `isWhiteSpaceLike` breaks compatibility with `typescript` `<= 2.3.0`.

Issue #2186, opened 2017-02-07
Autofix file should end with a newline


ryanmcnamara:
Bug Report

TSLint version: 4.0.2
TypeScript version: 2.1.4
Running TSLint via: CLI
TypeScript code being linted

```
export const x = "x";
```
Enable the file end with newline rule, and see that it fails the linter.

TSLint could autofix this error. It currently does not.

ryanmcnamara:
I wouldn't mind trying to fix this but I'm not sure when I'd have the time

ajafff:
How do you know which line ending to choose? `\r\n` or `\n`?
Of course this could be inferred from the other line endings in the file, but what if there is only a single line?

ryanmcnamara:
Good point. 
One idea is to only autofix this when there are multiple lines in the file, that way the line ending can be inferred.
Another strategy is just to guess. Not ideal, could potentially be configured, but that seems like a lot of overhead for such a small edge case. 
Does the existing TSLint rule just check that the file ends in `\r\n` *or* `\n`?

ajafff:
> Does the existing TSLint rule just check that the file ends in \r\n or \n?
It just checks if the last character is `\n`. That covers both cases.
I like the idea of fixing only multiline files. Otherwise when guessing wrong we could introduce new failures for the `linebreak-style` rule.

ryanmcnamara:
Is it possible to check what the value of the `linebreak-style` rule is while trying to autofix the `eofline` rule?

ajafff:
Nope, currently each rule is limited to its own configuration.
There is no way to access the config of another rule, although that would be pretty useful in many cases.

nchen63:
I don't think we want rules reading options that don't belong to it. I'm fine with only fixing files with multiple lines.

mohdsalaudin:
> How do you know which line ending to choose? `\r\n` or `\n`?
> Of course this could be inferred from the other line endings in the file, but what if there is only a single line?
hi there
Just put enter and add a new line and save the file this warning will disappear

PR #2695, opened 2017-05-06
Document that eofline fix is not supported for single-line files.

mxl:
#### PR checklist

- [x] Addresses an existing issue: #2186
  - [ ] Includes tests
- [x] Documentation update

#### Overview of change:

I think we should clearly state that eofline fix is not supported for single-line files. It's not obvious without inspecting [source code](https://github.com/palantir/tslint/blob/master/src/rules/eoflineRule.ts#L47) and looks like bug.

### DIFF
--- a/src/rules/eoflineRule.ts
+++ b/src/rules/eoflineRule.ts
@@ -24,6 +24,7 @@ export class Rule extends Lint.Rules.AbstractRule {
     public static metadata: Lint.IRuleMetadata = {
         ruleName: "eofline",
         description: "Ensures the file ends with a newline.",
+        descriptionDetails: "Fix for single-line files is not supported.",
         rationale: "It is a [standard convention](http://stackoverflow.com/q/729692/3124288) to end files with a newline.",
         optionsDescription: "Not configurable.",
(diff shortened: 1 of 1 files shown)


palantirtech:
Thanks for your interest in [palantir/tslint](https://github.com/palantir/tslint), @mxl! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

ajafff:
@mxl There are several rules that don't fix all violations and none of them state that explicitly. For example `prefer-const` doesn't fix uninitialized variables, `align` doesn't fix if it would remove code, ....

IMO this shouldn't be included in `description` or `FAILURE_STRING`. If you really want to have this documented, I'd suggest adding it to `Rule.metadata.descriptionDetails`.

mxl:
@ajafff That's not good if rule says that it can fix some violation but in certain cases it does not do this and that's not documented anywhere. I will add this to `descriptionDetails`. Any suggestions on how can I improve it?

PR #2696, opened 2017-05-06
Fix semicolon rule on last line without EOF line.

mxl:
#### PR checklist

- [x] New bugfix
  - [x] Includes tests

#### Overview of change:

Currently when using `"semicolon": [true, "never"]` rule if semicolon is used in last line without next EOF line then linting fails with error:

> Warning: Cannot read property 'kind' of undefined

because [here](https://github.com/palantir/tslint/blob/3ad4573b8151b37586bbf32fcd77794c5d8b3201/src/rules/semicolonRule.ts#L217) `util.getNextToken()` return value is not tested for `undefined`.

I put tests in separate directory and not in `semicolon/never/test.ts.*` to avoid accidental appending to that files that will ruin this test.

#### CHANGELOG.md entry:

[bugfix] `semicolon`

### DIFF
--- a/src/rules/semicolonRule.ts
+++ b/src/rules/semicolonRule.ts
@@ -214,17 +214,21 @@ class SemicolonWalker extends Lint.AbstractWalker<Options> {
         if (this.options.always && !hasSemicolon) {
             this.reportMissing(node.end);
         } else if (!this.options.always && hasSemicolon) {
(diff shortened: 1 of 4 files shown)


ajafff:
@mxl Thank you for taking the time to analyse and fix this bug.

Unfortunately that has already been done as part of a bigger change in #2655. That one also includes similar regression tests.

mxl:
@ajafff Glad to hear that!

PR #2700, opened 2017-05-06
space-before-function-paren: Rewrite

andy-hanson:
#### PR checklist

  - [ ] Includes tests
- [ ] Documentation update

#### Overview of change:

Just a refactor.
The previous implementation had special handling for FunctionDeclaration with an empty name; since this is a compiler error anyway, removed the special handling and its associated tests.

#### CHANGELOG.md entry:

[enhancement] `space-before-function-paren`: Handle `get`/`set` accessor

### DIFF
--- a/src/rules/spaceBeforeFunctionParenRule.ts
+++ b/src/rules/spaceBeforeFunctionParenRule.ts
@@ -109,8 +109,10 @@ function getOption(node: ts.Node, options: Options): Option | undefined {
             return options.constructor;
 
         case ts.SyntaxKind.FunctionDeclaration:
+            return options.named;
+
         case ts.SyntaxKind.FunctionExpression:
-            return (node as ts.FunctionLikeDeclaration).name !== undefined ? options.named : options.anonymous;
+            return (node as ts.FunctionExpression).name !== undefined ? options.named : options.anonymous;
 
         case ts.SyntaxKind.MethodDeclaration:
         case ts.SyntaxKind.MethodSignature:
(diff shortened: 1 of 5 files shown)


src/rules/spaceBeforeFunctionParenRule.ts
```diff
@@ -61,130 +61,64 @@ export class Rule extends Lint.Rules.AbstractRule {
...
-        const identifier = Lint.childOfKind(node, ts.SyntaxKind.Identifier);
-        const hasIdentifier = identifier !== undefined && (identifier.getEnd() !== identifier.getStart());
-        const optionName = hasIdentifier ? "named" : "anonymous";
-        const option = this.getOption(optionName);
-        const openParen = Lint.childOfKind(node, ts.SyntaxKind.OpenParenToken);
-        this.evaluateRuleAt(openParen, option);
-    }
+        case ts.SyntaxKind.MethodDeclaration:
+        case ts.SyntaxKind.MethodSignature:
+            return options.method;
```
ajafff: could you add `SetAccessor` and `GetAccessor` while you are already editing this rule?

src/rules/spaceBeforeFunctionParenRule.ts
```diff
@@ -61,130 +61,64 @@ export class Rule extends Lint.Rules.AbstractRule {
...
-
-                default:
-                    this.cachedOptions[optionName] = options[optionName];
+function walk(ctx: Lint.WalkContext<Options>): void {
+    const { options, sourceFile } = ctx;
+    ts.forEachChild(sourceFile, function cb(node: ts.Node): void {
+        const option = getOption(node, options);
+        if (option !== undefined) {
+            const openParen = Lint.childOfKind(node, ts.SyntaxKind.OpenParenToken)!;
+            const hasSpace = sourceFile.text[openParen.pos] !== "(";
```
ajafff: this line should be changed to handle comments:
```ts
const hasSpace = ts.isWhiteSpaceLike(sourceFile.text.charCodeAt(openParen.end - 2));
```

Issue #2661, opened 2017-05-01
Rule await-promise: make Promise implementation configurable


darxriggs:
### Feature Request

- __TSLint version__: 5.1.0
- __TypeScript version__: 2.3.2
- __Running TSLint via__: CLI

#### TypeScript code being linted

```ts
import * as Bluebird from 'bluebird';

function functionReturningThenable(): Bluebird.Thenable<string> {
    return Bluebird.resolve('some data');
}

async function main(): Promise<void> {
    const data = await functionReturningThenable();
    console.log(data);
}
```

with `tslint.json` configuration:

```json
{
  "rules": {
    "await-promise": true
  }
}
```

#### Actual behavior
```shell
(await-promise) src/test.ts[8, 18]: 'await' of non-Promise.
```

#### Expected behavior
No error should be reported.

There are multiple Promise implementations and interfaces out there (`JQueryPromise`, `Q.Promise`, `Thenable`, etc.). Also see [Promises/A+](https://promisesaplus.com/), Therefore the rule should be configurable like `no-floating-promises` already is.

darxriggs:
I'd like to create PR for this but I've seen https://github.com/palantir/tslint/pull/2631 ("no-floating-promises: rewrite as walker function").

So the question is if this rule should use `applyWithWalker` or `applyWithFunction` and why?

ajafff:
@darxriggs we want to move away from `RuleWalker` and therefore `ProgramAwareRuleWalker` and eventually deprecate them. `await-promise` has not been refactored to use either `AbstractWalker` or `applyWithFunction`. #2631 would be a good start.

Which one you choose is up to you. For simple rules it's preferred to use a function. When you find yourself in the need to add several closures to the walker function, you should consider using `AbstractWalker` to avoid the repeated closure allocation.
For more information about the concepts, there is some documentation at https://palantir.github.io/tslint/develop/custom-rules/walker-design.html

PR #2705, opened 2017-05-07
Await promise configurable

darxriggs:
#### PR checklist

- [x] Addresses an existing issue: #2661
  - [x] Includes tests
- [x] Documentation update

#### Overview of change:
a) Now using a function instead of a walker.
b) Consider types derived from a Promise in union types too.
c) What's considered a "Promise" is now configurable.

I'm not sure what exactly means "Documentation update" in the checklist. I updated `metadata` inside the `Rule` itself. I guess the rest is taken care of by the maintainers (npm "docs" script).

#### CHANGELOG.md entry:
- [bugfix] [`await-promise`](https://palantir.github.io/tslint/rules/await-promise/): Consider types derived from a Promise in union types too. (#2661)
- [new-rule-option] [`await-promise`](https://palantir.github.io/tslint/rules/await-promise/): What's considered a "Promise" is now configurable. (#2661)


palantirtech:
Thanks for your interest in [palantir/tslint](https://github.com/palantir/tslint), @darxriggs! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

src/rules/awaitPromiseRule.ts
```diff
@@ -34,41 +43,43 @@ export class Rule extends Lint.Rules.TypedRule {
 
     public static FAILURE_STRING = "'await' of non-Promise.";
 
-    public applyWithProgram(srcFile: ts.SourceFile, program: ts.Program): Lint.RuleFailure[] {
-        return this.applyWithWalker(new Walker(srcFile, this.getOptions(), program));
+    public applyWithProgram(sourceFile: ts.SourceFile, program: ts.Program): Lint.RuleFailure[] {
+        const promiseTypes = new Set(["Promise", ...this.ruleArguments.map(String)]);
```
ajafff: out of curiosity, why would you need to explicitly convert the ruleArguments to String?
you can just use a type assertion instead.

src/rules/awaitPromiseRule.ts
```diff
@@ -34,41 +43,43 @@ export class Rule extends Lint.Rules.TypedRule {
...
-}
+    function isPromiseType(type: ts.Type): boolean {
+        const { target } = type as ts.TypeReference;
+        const symbol = target && target.symbol;
+        return symbol !== undefined && promiseTypes.has(symbol.name);
+    }
 
-function isUnionType(type: ts.Type): type is ts.UnionType {
-    return Lint.isTypeFlagSet(type, ts.TypeFlags.Union);
+    function isUnionType(type: ts.Type): type is ts.UnionType {
```
ajafff: this function does not need to be inside `walk`

src/rules/awaitPromiseRule.ts
darxriggs: I saw `.map(String)` in another rule and thought it to be a good idea. It really makes sure that `promiseTypes` are only `String`s. A [type assertion](https://basarat.gitbooks.io/typescript/docs/types/type-assertion.html) only makes the compiler happy but at runtime another type may occur.

src/rules/awaitPromiseRule.ts
darxriggs: That's true.

The goal here was to encapsulate everything inside `walk`. So there is only `Rule` and `walk` visible in the file.

ajafff: You don't need to encapsulate that. The encapsulation you get from the file beeing an ES6 module is enough. In fact you should try to keep closures as small as possible to avoid unnecessary work at runtime.

However, it's not that important for now. I'll add this function among others to the next version of `tsutils`. So this one will be replaced anyway.

src/rules/awaitPromiseRule.ts
ajafff: If the config contains non-string values, does it really matter to convert them to string at runtime? Without explicitly converting them, they'll simply be ignored, because they are wrong either way.

darxriggs:
I updated the pull request as suggested.

darxriggs:
@ajafff I just wanted to check if there is anything else I can do so that this PR gets merged?
I already rebased the commits two times on master to resolve merge conflicts.

Issue #2721, opened 2017-05-09
no-magic-numbers triggering for array access keys


cleavera:
### Bug Report

- __TSLint version__: 5.2.1
- __TypeScript version__: 2.2.0
- __Running TSLint via__: gulp-tslint

#### TypeScript code being linted

```ts
const match: RegExpMatchArray = require.match(/^(\^\^?)?(\?)?(\^\^?)?/);
const inheritType: string = match[3];
const isOptional: boolean = !!match[2];
```

with `tslint.json` configuration:

```json
"no-magic-numbers": true,
```

#### Actual behavior
```'magic numbers' are not allowed``` fired for both array access keys
match[__3__];
match[__2__];

#### Expected behavior
Array access keys are not magic numbers so should not be flagged as such when linting

**Comment by ajafff** [comment]
what makes a number used as array index different from a number used elsewhere? they are both magic numbers.

**Comment by cleavera** [comment]
Not really, I wouldn't call a property key on an object a magic string, so what make arrays special?

What would you propose?
```
const match: RegExpMatchArray = require.match(/^(\^\^?)?(\?)?(\^\^?)?/);
const isOptionalIndexKey: number = 2;
const isOptional: boolean = !!match[isOptionalIndexKey];
```

It doesn't provide anything extra (in readability or maintainability), and make the rule a bit of a pig to use.

**Comment by adidahiya** [comment]
How often does this show up in your codebase @cleavera? Most of the examples I've seen so far have been contrived, except for the regexp match, which seems like something you can special case with a `// tslint:disable` flag.

**Comment by cleavera** [comment]
By far and away the biggest usages are in unit tests, where the code pushes different items to an array in different circumstances and the unit tests are checking that specific items in the array equal specific values.
```ts
it('should provide a link to the landing page', function() {
  expect($ctrl.actions[0]).toEqual({ state: 'core-landingPage' });
});

it('should provide a link to the person summary', function() {
  expect($ctrl.actions[1]).toEqual({ state: 'hrm-person', params: { personId: '123' } });
});

it('should provide a link to create a new person', function() {
  expect($ctrl.actions[2]).toEqual({ state: 'hrm-person-details', params: { personId: 'new' } });
});
```

In this case the numbers literally represent the number. I could have variables for the them but they provide no value whatsoever.

I would be interested to see a case where extracting the number to a named variable provides some extra level of maintainability or readability.

If I take ajafff's (rather abstract) example from the pull request and change it to conform to the rule:
```ts
[...]

PR #2725, opened 2017-05-10
no-magic-numbers: allow magic numbers for ElementAccessExpression

cleavera:
#### PR checklist

- [x] Addresses an existing issue: #2721
  - [x] Includes tests
- [ ] Documentation update

#### Overview of change:
Allow magic numbers for ElementAccessExpression fixes(#2721)

#### CHANGELOG.md entry:

[bugfix] `no-magic-number` Allow ElementAccessExpressions to use numbers directly

### DIFF
--- a/src/rules/noMagicNumbersRule.ts
+++ b/src/rules/noMagicNumbersRule.ts
@@ -47,6 +47,7 @@ export class Rule extends Lint.Rules.AbstractRule {
     public static FAILURE_STRING = "'magic numbers' are not allowed";
 
     public static ALLOWED_NODES = new Set<ts.SyntaxKind>([
+        ts.SyntaxKind.ElementAccessExpression,
         ts.SyntaxKind.ExportAssignment,
         ts.SyntaxKind.FirstAssignment,
         ts.SyntaxKind.LastAssignment,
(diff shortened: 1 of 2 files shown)


palantirtech:
Thanks for your interest in [palantir/tslint](https://github.com/palantir/tslint), @cleavera! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

ajafff:
I'm still not conviced that array indizes should should get special handling.

Consider the following code. Do you still think those are not magic numbers?
```ts
doStuff(arr[5], arr[10], arr[2]);
```

cleavera:
In your example the number itself has no meaning. The value inside of the array is what has the meaning. I would rather the value is what gets put into the variable than the key.
For example:
```ts
let people = ['Andy', 'Dave', 'Steve'];

let steve = people[2];
talkTo(steve);
// provides more meaning than
let steve = 2;
talkTo(people[steve]);
// and is less verbose but just as understandable as
let steveIndex = 2;
let steve = people[steveIndex];
talkTo(steve);
```

PR #2743, opened 2017-05-12
cyclomatic-complexity: Don't count empty switch cases

andy-hanson:
#### PR checklist

  - [X] Includes tests
- [ ] Documentation update

#### Overview of change:

`switch { case 0: case 1: case 2: return "a"; default: return "b"; }` now has a cyclomatic complexity of 2, not 4.
Also enabled the rule. Each of the disables (not marked "fixed in 5.3") should be looked into later.

#### CHANGELOG.md entry:

[enhancement] `cyclomatic-complexity`: Don't count empty switch case


src/rules/cyclomaticComplexityRule.ts
```diff
@@ -105,6 +105,7 @@ function walk(ctx: Lint.WalkContext<{ threshold: number }>): void {
 function increasesComplexity(node: ts.Node): boolean {
     switch (node.kind) {
         case ts.SyntaxKind.CaseClause:
+            return (node as ts.CaseClause).statements.length > 0;
```
ajafff: `if (foo === 0 || foo === 1) {bar()}` increases complexity twice. The will increase complexity only once with this change:
```ts
switch (foo) {
    case 0:
    case 1:
        bar();
}
```

ajafff: related: why do `else` branches or `default` clauses not increase complexity?

ajafff: btw. if revert your change here to make them equal again, you probably need to disable the rule in almost all `ts.forEachChild` callbacks in this repo

andy-hanson: Well, these two code snippets seem equally complex to me:
```ts
if (foo) {
    return 0;
}
return 1;
```

```ts
if (foo) {
    return 0;
} else {
    return 1;
}
```

andy-hanson: `foo === 0 || foo === 1` is semantically equivalent to the switch statement, but syntactically it's treated the same as any other `||` expression, which can contain arbitrarily complex behavior.
There's some debate about this issue here: http://stackoverflow.com/questions/30240236/cyclomatic-complexity-of-switch-case-statement

andy-hanson:
I'm not a big fan of cyclomatic complexity either, so I'd be happy to switch this pull request to just the change to the rule, and move it in our `tslint.json` from "TODO" to "Don't want".

adidahiya:
I don't want to enable this rule either. Please move it from "TODO" to "Don't want" instead.

src/rules/cyclomaticComplexityRule.ts
andy-hanson: There's only one possible path through that code, so I don't think it should increase complexity.

KSuttle-cng:
Are there docs on how switch statement cases affect the cyclomatic complexity score?

PR #2748, opened 2017-05-12
deprecation: Rewrite and enhance

ajafff:
#### PR checklist

  - [x] Includes tests
- [ ] Documentation update

#### Overview of change:
Rewrite to walker function.
Use TypeScript's JsDoc parser
[enhancement] `deprecation`: error message includes deprecation text if available

### DIFF
--- a/src/rules/deprecationRule.ts
+++ b/src/rules/deprecationRule.ts
@@ -20,7 +20,6 @@ import * as ts from "typescript";
 import * as Lint from "../index";
 
 export class Rule extends Lint.Rules.TypedRule {
-
     /* tslint:disable:object-literal-sort-keys */
     public static metadata: Lint.IRuleMetadata = {
         ruleName: "deprecation",
--- a/src/rules/deprecationRule.ts
+++ b/src/rules/deprecationRule.ts
@@ -15,6 +15,7 @@
  * limitations under the License.
  */
 
+import { isIdentifier } from "tsutils";
 import * as ts from "typescript";
 import * as Lint from "../index";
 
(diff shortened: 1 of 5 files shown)


src/rules/deprecationRule.ts
```diff
@@ -37,60 +37,79 @@ export class Rule extends Lint.Rules.TypedRule {
...
-
-      const range = ts.getLeadingCommentRanges(commentNode.getFullText(), 0);
-      if (range === undefined) { continue; }
-      for (const {pos, end} of range) {
-        const jsDocText = commentNode.getFullText().substring(pos, end);
-        if (jsDocText.includes("@deprecated")) {
-            this.addFailureAtNode(node, `${node.text} is deprecated.`);
+function getDeprecation(node: ts.Identifier, tc: ts.TypeChecker): string | undefined {
+    let symbol = tc.getSymbolAtLocation(node);
+    if (symbol !== undefined && Lint.isSymbolFlagSet(symbol, ts.SymbolFlags.Alias)) {
```
andy-hanson: We could use a `skipAliasSymbol` helper in tsutils.

src/rules/deprecationRule.ts
```diff
@@ -37,60 +37,79 @@ export class Rule extends Lint.Rules.TypedRule {
...
+}
 
-    if (decSym !== undefined && Lint.isSymbolFlagSet(decSym, ts.SymbolFlags.Alias)) {
-        decSym = this.getTypeChecker().getAliasedSymbol(decSym);
-    }
-    const declarations = decSym === undefined ? undefined : decSym.getDeclarations() as ts.Node[] | undefined;
-    if (declarations === undefined) {
-        super.visitIdentifier(node);
-        return;
+function isDeclaration(identifier: ts.Identifier): boolean {
```
andy-hanson: TypeScript has `ts.isDeclarationName`, but it's internal. You could make an issue on the TypeScript issue asking it to be made non-internal since it seems pretty useful.

andy-hanson: Or I could: Microsoft/TypeScript#15811

ajafff: The implementation used here is slightly different. For example typescript's implementation includes `ShorthandPropertyAssignment` which actually uses the variable and should therefore be checked by this rule. Same with `NamespaceImport`, ...

PR #2754, opened 2017-05-13
Add 'no-redundant-jsdoc' rule

andy-hanson:
#### PR checklist

  - [X] Includes tests
- [X] Documentation update

#### Overview of change:

Adds the `no-redundant-jsdoc` rule, which forbids unnecessary redundant JSDoc, as in:
```ts
/**
 * @export
 * @function
 * @name f
 * @param {number} x
 * @returns {void}
 */
export function f(x: number): void;
```

#### CHANGELOG.md entry:

[new-rule] `no-redundant-jsdoc`


src/rules/noRedundantJsdocRule.ts
```diff
@@ -0,0 +1,143 @@
...
+        case "requires":
+        case "static":
+        case "this":
+            return true;
+        default:
+            return false;
+    }
+}
+
+declare module "typescript" {
```
ajafff: please don't do that
that should be done upstream in typescript. IIRC there's already an issue with a feature request for something like `node.getJsDoc()`

adidahiya: agreed, let's wait for this to be fixed upstream.

src/rules/noRedundantJsdocRule.ts
```diff
@@ -0,0 +1,143 @@
...
+            }
+
+            default:
+                throw new Error(`Unexpected tag kind: ${ts.SyntaxKind[tag.kind]}`);
+        }
+    }
+}
+
+function isRedundantTag(tagName: string): boolean {
+    switch (tagName) {
```
JoshuaKGoldberg: Perhaps keeping a `Set` of these would be a little more efficient & less code?

Issue #2692, opened 2017-05-05
tslint with --type-check does not recognize 'repeat' method in string


banyudu:
### Bug Report

- __TSLint version__: 5.2.0
- __TypeScript version__: 2.3.2
- __Running TSLint via__: CLI `./node_modules/.bin/tslint --type-check --project .`

#### TypeScript code being linted

```ts
const SPECIAL_VALUES = ['', ' ', '\n', '\t', 'a'.repeat(300)]
```

with `tslint.json` configuration:

```json
{
  "rules": {
    "callable-types": true,
    "class-name": true,
    "comment-format": [
      true,
      "check-space"
    ],
    "curly": true,
    "eofline": true,
    "forin": true,
    "import-blacklist": [true, "rxjs"],
    "import-spacing": true,
    "indent": [
      true,
      "spaces"
    ],
    "interface-over-type-literal": true,
    "label-position": true,
    "max-line-length": [
      true,
      140
    ],
    "member-access": false,
    "member-ordering": [
      true,
      "static-before-instance",
      "variables-before-functions"
    ],
    "no-arg": true,
    "no-bitwise": true,
    "no-console": [
      true,
      "debug",
      "info",
      "time",
      "timeEnd",
      "trace"
    ],
    "no-construct": true,
    "no-debugger": true,
    "no-duplicate-variable": true,
    "no-empty": false,
    "no-eval": true,
    "no-shadowed-variable": true,
    "no-string-literal": false,
    "no-string-throw": true,
    "no-switch-case-fall-through": true,
    "no-trailing-whitespace": true,
    "no-unused-expression": true,
    "no-use-before-declare": true,
    "no-var-keyword": true,
    "object-literal-sort-keys": false,
    "one-line": [
      true,
      "check-open-brace",
      "check-catch",
      "check-else",
      "check-whitespace"
    ],
    "prefer-const": true,
    "quotemark": [
      true,
      "single"
    ],
    "radix": true,
    "semicolon": [
      true,
      "never"
    ],
    "triple-equals": [
      true,
      "allow-null-check"
    ],
    "typedef-whitespace": [
      true,
      {
        "call-signature": "nospace",
        "index-signature": "nospace",
        "parameter": "nospace",
        "property-declaration": "nospace",
        "variable-declaration": "nospace"
      }
    ],
    "typeof-compare": true,
    "unified-signatures": true,
    "variable-name": false,
    "whitespace": [
      true,
      "check-branch",
      "check-decl",
[...]

PR #2756, opened 2017-05-13
CLI: handle directories passed with -p

ajafff:
#### PR checklist

- [x] Addresses an existing issue: #2692 
  - [x] Includes tests
- [x] Documentation update

#### Overview of change:
Search for tsconfig.json in specified directory. Exit with error if  there is none.
Ref: #2692
[enhancement] cli: `-p` option handles directories

### DIFF
--- a/src/runner.ts
+++ b/src/runner.ts
@@ -146,11 +146,14 @@ export class Runner {
         let program: ts.Program | undefined;
 
         if (this.options.project != null) {
-            if (!fs.existsSync(this.options.project)) {
+            let project: string;
+            try {
+                project = findTsconfig(this.options.project);
+            } catch (e) {
                 console.error("Invalid option for project: " + this.options.project);
                 return onComplete(1);
             }
-            program = Linter.createProgram(this.options.project);
+            program = Linter.createProgram(project);
             if (files.length === 0) {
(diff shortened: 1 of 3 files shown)


src/runner.ts
```diff
@@ -146,11 +146,14 @@ export class Runner {
         let program: ts.Program | undefined;
 
         if (this.options.project != null) {
-            if (!fs.existsSync(this.options.project)) {
+            let project: string;
+            try {
+                project = findTsconfig(this.options.project);
```
andy-hanson: I think this function should handle error handling itself (and return `string | undefined` instead of throwing an error that gets ignored.

src/runner.ts
```diff
@@ -175,9 +178,6 @@ export class Runner {
                 // if not type checking, we don't need to pass in a program object
                 program = undefined;
             }
-        } else if (this.options.typeCheck) {
```
andy-hanson: Why did you move this code?

ajafff: all static option validation is done in tslint-cli.ts
runner.ts should only be responsible to catch errors that occur at runtime

src/runner.ts
ajafff: done

PR #2765, opened 2017-05-14
Fix findup when file found in subdirectory

ajafff:
#### PR checklist

  - [x] Includes tests
- [ ] Documentation update

#### Overview of change:
The new `findup` implementation missed a `path.join(...)` and just returned the filename.
This fixes the bug and adds tests to ensure it will not regress.

#### CHANGELOG.md entry:
[no-log] because that's not yet released


andy-hanson:
Is this fixed by #2720?

ajafff:
Seems like this fixes it. You can add the tests from this PR to make sure it does

PR #2769, opened 2017-05-15
Don't call getPreEmitDiagnostics in tests

ajafff:
#### PR checklist

  - [x] Includes tests
- [x] Documentation update

#### Overview of change:
The diagnostics were ignored anyway and are not required for typechecking.
Creating a program and therefore enabling typechecked rules is now done when a tsconfig.json file is placed in the folder next to tslint.json.
That allows us to get rid of `linterOptions.typeCheck` in tslint.json. It's still supported but deprecated in case someone used it in his tests.

This reduces the time needed to run the rule tests by 50%.
I'll also submit a PR that enables typechecked rules when passing `--project` (without `--type-check`) on CLI. That will result in a 20% speedup when linting this project.

#### CHANGELOG.md entry:

[develop] testing rules with type information is enabled when a `tsconfig.json` is found next to `tslint.json`


src/test.ts
```diff
@@ -84,6 +86,11 @@ export function runTest(testDirectory: string, rulesDirectory?: string | string[
         };
         compilerOptions = ts.parseJsonConfigFileContent(config, parseConfigHost, testDirectory).options;
     }
+    // TODO remove in v6.0.0
+    if (tslintConfig !== undefined && tslintConfig.linterOptions !== undefined && tslintConfig.linterOptions.typeCheck === true) {
```
adidahiya: `linterOptions.typecheck` was never public; we can remove it immediately. https://github.com/palantir/tslint/issues/1445#issuecomment-236330061

test/rules/completed-docs/defaults/tsconfig.json
```diff
@@ -0,0 +1 @@
+{}
```
adidahiya: an empty file looks pretty weird; can we at least make it extend some common tsconfig file? perhaps located at `test/rules/test-tsconfig.json`?

ajafff: I don't want to extend because that adds unnecessary overhead. I just added `compilerOptions.target` to have some config in the file

ajafff:
Updated as suggested. Tests pass.
Added documentation.

Similar PR for `runner.ts` will be ready tomorrow morning.

Once this PR and #2762 are merged, I will refactor `test.ts` to only create the `program` once per directory.

Issue #2783, opened 2017-05-18
strict-type-predicates reports 'Expression is always true.' which is not true


Martin-Wegner:
### Bug Report

- __TSLint version__: 5.2.0
- __TypeScript version__: 2.3.2
- __Running TSLint via__: ng lint --type-check from Angular CLI

#### TypeScript code being linted

```ts
['a', 'b', 'c'].map((s: string) => {
  if (s === 'b') {
    return undefined;
  }
  return 1;
}).filter((n: number) => n !== undefined);
```

```ts
['a', 'b', 'c'].map((s: string) => {
  if (s === 'b') {
    return undefined;
  }
  return 1;
}).filter((n: number | undefined) => n !== undefined);
```

```ts
['a', 'b', 'c'].map((s: string) => {
  if (s === 'b') {
    return undefined;
  }
  return 1;
}).filter(n => n !== undefined);
```

with `tslint.json` configuration:

```json
{
  "rules": {
    "strict-type-predicates": true
  }
}

```

#### Actual behavior
> ERROR: ...: Expression is always true.

#### Expected behavior
No error.

ajafff:
Remove the type annotation `:number` which is wrong anyway. That should
make it work

Am 18.05.2017 17:19 schrieb "Martin-Wegner" <notifications@github.com>:

Martin-Wegner:
Why is `number` wrong? The mapping function maps `string` to `number`... When I use `any` the same error occurs.

ajafff:
Your function maps to `number | undefined `. Either use that as type annotation of the parameter or just omit it and let typescript infer the type. 

The bug with `any` is fixed by a recent PR

Martin-Wegner:
I omitted it and the same error occurs...

Martin-Wegner:
Using `number | undefined` also not work :(

ajafff:
That indeed looks like a bug. Do you have strictNullChecks enabled in your tsconfig.json?

Martin-Wegner:
No `strictNullChecks` is not enabled.

ajafff:
This rule does not work as expected without strictNullChecks, because the current implementation assumes that it's enabled.

I doubt it would be useful to add special handling for `strictNullChecks = false`, because the purpose of this rule is to check if you are comparing to `null` or `undefined` on non-nullable values. Everything else is already checked by the compiler.
The only check that could be performed without strictNullChecks would be somethink like this:
```ts
declare var num: number;
[...]

PR #2786, opened 2017-05-18
strict-type-predicates: warn if strictNullChecks is not enabled

ajafff:
#### PR checklist

- [x] Addresses an existing issue: #2783
  - [x] Includes tests
- [x] Documentation update

#### Overview of change:
Warn if `strict-type-predicates` is used without `strictNullChecks` and don't apply the rule. IMO that's a better behavior than just emitting wrong lint results.

This also fixes `strict-boolean-expressions` if `strict` is used instead of `strictNullChecks`.


src/language/utils.ts
```diff
@@ -425,3 +425,8 @@ export function getEqualsKind(node: ts.BinaryOperatorToken): EqualsKind | undefi
             return undefined;
     }
 }
+
+export function isStrictNullChecksEnabled(options: ts.CompilerOptions): boolean {
```
ajafff: is this the right file for this function?

src/rules/strictTypePredicatesRule.ts
```diff
@@ -52,6 +54,10 @@ export class Rule extends Lint.Rules.TypedRule {
     }
 
     public applyWithProgram(sourceFile: ts.SourceFile, program: ts.Program): Lint.RuleFailure[] {
+        if (!Lint.isStrictNullChecksEnabled(program.getCompilerOptions())) {
+            showWarningOnce("strict-type-predicates does not work without strictNullChecks");
```
ajafff: Please suggest a better warning message.

src/language/utils.ts
nchen63: I think it's okay. We'll probably have to split apart this file some time in the future

src/rules/strictTypePredicatesRule.ts
nchen63: I think it's fine. Would probably be more clear if it said `tsc option --strictNullChecks`, but if you search for `strictNullChecks`, the first hit is correct

Issue #2778, opened 2017-05-17
Make type literal delimiter semicolon by default


nchen63:
see comments in https://github.com/palantir/tslint/pull/2374

nchen63:
closed by #2787

PR #2787, opened 2017-05-19
type-literal-delimiter: Prefer ';' instead of ',', and lint for trailing delimiter

andy-hanson:
#### PR checklist

- [X] Addresses an existing issue: #2778
  - [X] Includes tests
- [X] Documentation update

#### Overview of change:

Switches the preference to `;` instead of `,`.
Checks for a trailing delimiter in multiline type literals, and for no trailing delimiter in single-line type literals.

Blocked waiting on #2789.

#### CHANGELOG.md entry:

[rule-change] `type-literal-delimiter`: Prefer `;` instead of `,`, and lint for trailing delimiter


nchen63:
Wait, semicolons seem unexpected here:

`function f(arg: { a: number; b: string })`

Yet, I get @adidahiya's comment on 
```
type t = {
  a: number,
  b: string
}
```

also being inconsistent with interface. Is it worth it to make a more complicated rule and be inconsistent? I think I'd rather see all commas instead of the 1st example.

adidahiya:
yeah I noticed that weirdness too @nchen63... I think we should add more configuration to the rule before enabling it in the built-in configs (I noticed you removed it from `tslint:latest` in the v5.3 release PR 👍)

how about we make single-line type literal checking opt-in for this rule? there could be a rule option called `"check-single-line"` that enforces the specified delimiter type?

nchen63:
I think I'm going to make a fork that doesn't use this rule so we can think about it some more and publish based on the fork

PR #2789, opened 2017-05-19
Fix failures on typescript@next

andy-hanson:
#### PR checklist

  - [ ] Includes tests
- [ ] Documentation update

#### Overview of change:

TypeScript has made some breaking changes recently (Microsoft/TypeScript#15486, Microsoft/TypeScript#15594, Microsoft/TypeScript#15887), so tslint must be updated.

Waiting on ajafff/tsutils#3.

### DIFF
--- a/package.json
+++ b/package.json
@@ -72,7 +72,7 @@
     "rimraf": "^2.5.4",
     "tslint": "latest",
     "tslint-test-config-non-relative": "file:test/external/tslint-test-config-non-relative",
-    "typescript": "^2.3.0"
+    "typescript": "next"
   },
   "license": "Apache-2.0",
   "engines": {
--- a/package.json
+++ b/package.json
@@ -44,7 +44,7 @@
     "resolve": "^1.3.2",
     "semver": "^5.3.0",
     "tslib": "^1.6.0",
-    "tsutils": "^1.9.1"
+    "tsutils": "^2.0.0"
   },
   "peerDependencies": {
     "typescript": ">=2.1.0 || >=2.1.0-dev || >=2.2.0-dev || >=2.3.0-dev || >=2.4.0-dev"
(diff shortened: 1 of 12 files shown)


src/rules/completedDocsRule.ts
```diff
@@ -389,7 +389,8 @@ class CompletedDocsWalker extends Lint.ProgramAwareRuleWalker {
     }
 
     private checkNode(node: ts.Declaration, nodeType: DocType): void {
-        if (node.name === undefined) {
+        const { name } = node as ts.NamedDeclaration;
```
ajafff: just update the type annotation of the parameter

package.json
```diff
@@ -72,7 +72,7 @@
     "rimraf": "^2.5.4",
     "tslint": "latest",
     "tslint-test-config-non-relative": "file:test/external/tslint-test-config-non-relative",
-    "typescript": "^2.3.0"
+    "typescript": "next"
```
ajafff: not sure if it makes sense to use typescript@next in development while having a yarn.lock. I guess there'll be many merge conflicts just because of the typescript version in yarn.lock

ajafff:
@andy-hanson I released `tsutils@2.0.0` with your change. That should fix the build

package.json
adidahiya: I agree, yeah I don't want this change in development

andy-hanson: It won't compile in typescript pre-`next` though. Could I just tag it to the specific daily version?

adidahiya: would that mean that tslint API consumers would need to upgrade to 2.4.0-dev as well? I'm very wary of making this change; even if the compiled tslint typings continue to work with 2.3 in _this PR_, bumping to a nightly version opens up the door to future API breaks

adidahiya: @nchen63 has another approach to fixing master that compiles with ts 2.3: https://github.com/palantir/tslint/tree/fix-master

PR #2799, opened 2017-05-20
type-literal-delimiter: Prefer comma for single-line type literal

andy-hanson:
#### PR checklist

- [X] Addresses an existing issue: https://github.com/palantir/tslint/pull/2787#issuecomment-302899411
  - [X] Includes tests
- [X] Documentation update

#### Overview of change:

If a type literal only takes up a single line, prefer `,`.

#### CHANGELOG.md entry:

[rule-change] `type-literal-delimiter`: Prefer commas for a single-line type literal


ajafff:
Why is this specific to single line type literals?
I'd prefer a second configuration option: one for type literals in type position and one for type literals used in  TypeAliasDeclarations

andy-hanson:
Combining that with `interface-over-type-literal` it seems like type literal delimiters should just always be commas.
Which is exactly what I did in #2374...

nchen63:
@andy-hanson That's a good point about `interface-over-type-literal`, but we can't assume that people will always use these rules together. 

I don't think we should differentiate between single line and multi-line since converting a long single line to multi-line would be very annoying. 

How about TypeAliasDeclarations are not checked by default, with the option for commas and semicolons. I personally prefer always commas, but I can see how some might prefer semicolons

cc: @adidahiya

adidahiya:
> I'd prefer a second configuration option: one for type literals in type position and one for type literals used in TypeAliasDeclarations

👍 I like @ajafff's and @nchen63's proposals. Sorry for the back-and-forth on this @andy-hanson...

andy-hanson:
Currently if someone wanted to set `trailing-comma` to `"never"`, then they couldn't use this rule. I think this rule should not lint for a trialing delimiter if that delimiter is a comma. And if it is a semicolon, then there should always be a trailing delimiter, like there is for interfaces.

src/rules/typeLiteralDelimiterRule.ts
```diff
@@ -20,62 +20,95 @@ import * as ts from "typescript";
...
-        optionExamples: [true],
+            Checks that type literal members are separated by commas.
+            Enforces a trailing delimiter for multi-line type literals.`,
+        optionsDescription: Lint.Utils.dedent`
+            An options object may be passed with the following keys:
+
+            * '${OPTION_TYPE_ALIAS_DELIMITER}': This may be set to one of:
+              - \`","\`: This is the default. Type aliases are treated the same as other type literals.
+              - \`";"\`: Use semicolons for type aliases instead of commas.
+              - \`"ignore\`: Do not check the delimiter used by a type alias. (Still check for a trailing delimiter.)`,
```
adidahiya: missing a quotemark after `ignore`

src/configuration.ts
```diff
@@ -44,7 +44,7 @@ export interface IConfigurationFile {
      * Other linter options, currently for testing. Not publicly supported.
      */
     linterOptions?: {
-        typeCheck?: boolean;
+        typeCheck?: boolean,
```
adidahiya: nit: would prefer to keep this as a semicolon for consistency with the rest of the interface

src/rules/typeLiteralDelimiterRule.ts
```diff
@@ -20,62 +20,95 @@ import * as ts from "typescript";
...
-        optionsDescription: "Not configurable.",
-        options: null,
-        optionExamples: [true],
+            Checks that type literal members are separated by commas.
+            Enforces a trailing delimiter for multi-line type literals.`,
+        optionsDescription: Lint.Utils.dedent`
+            An options object may be passed with the following keys:
+
+            * '${OPTION_TYPE_ALIAS_DELIMITER}': This may be set to one of:
+              - \`","\`: This is the default. Type aliases are treated the same as other type literals.
```
adidahiya: I think it would be easier to read the tslint.json configuration if these options were called `"semicolon" | "comma"`, what do you think?

src/test.ts
```diff
@@ -50,7 +50,7 @@ export interface SkippedTest {
 export interface TestResult {
     directory: string;
     results: {
-        [fileName: string]: TestOutput | SkippedTest;
+        [fileName: string]: TestOutput | SkippedTest,
```
adidahiya: nit: same here, would slightly prefer to keep this as a semicolon

src/configuration.ts
andy-hanson: So a property in an interface should be treated like a type alias?
What about a property in a class, or a return type of a method?

adidahiya:
ping @andy-hanson?

andy-hanson:
Do you want type literals in properties to always use semicolons? Should this apply to method return types too?
It looks like the rule has already been published, how should we handle changing the behavior now?

src/rules/typeLiteralDelimiterRule.ts
```diff
@@ -20,61 +20,100 @@ import * as ts from "typescript";
...
-        optionExamples: [true],
+            Checks that type literal members are separated by commas.
+            Enforces a trailing delimiter for multi-line type literals.`,
+        optionsDescription: Lint.Utils.dedent`
+            An options object may be passed with the following keys:
+
+            * '${OPTION_TYPE_ALIAS_DELIMITER}': This may be set to one of:
+              - \`"${OPTION_COMMA}"\`: This is the default. Type aliases are treated the same as other type literals.
+              - \`"${OPTION_SEMICOLON}"\`: Use semicolons for type aliases instead of commas.
+              - \`"ignore"\`: Do not check the delimiter used by a type alias. (Still check for a trailing delimiter.)`,
```
adidahiya: nit: use `${OPTION_IGNORE}`

andy-hanson:
What should be the default behavior in these situations?
* Type literal after of `type T =`
* Type literal inside a union/intersection after `type T = `
* Type literal in a property/return type
* Type literal anywhere else

adidahiya:
Sorry for the delay here... I'm going to close this since it's stale and has lots of conflicts... if anyone wants to pick it up again, my suggestion is that we should match whatever `prettier` prefers for formatting w/r/t semicolons or commas (while preserving backwards compatibility wherever possible).

PR #2800, opened 2017-05-21
Add 'early-exit' rule

andy-hanson:
#### PR checklist

  - [X] Includes tests
- [X] Documentation update

#### Overview of change:

Adds the `early-exit` lint rule, which warns on `if` statements with large bodies that could be replaced by an early exit.

Since the rule is also applied to tslint itself the diff is rather large. But that makes for a good opportunity to review the heuristics the rule uses in action.

#### CHANGELOG.md entry:

[new-rule] `early-exit`


andy-hanson:
Also, `curly` has an exception for one-line statements, but tslint itself doesn't use it. Could there be an exception added for single keywords -- `return`, `continue`, `break`, `debugger`? I think bringing these down to one line as in https://github.com/palantir/tslint/pull/2800/commits/c8c4cd8c6e423749fb0dbd0ffec8d68c5c1b87e0 is a lot nicer looking.

adidahiya:
Sorry, I don't think this needs to be a core rule. It feels pretty opinionated / niche and it could easily be implemented as a third-party rule. I'd consider adding it to core if there was more signal suggesting that users want it. I personally find early exits a little harder to parse than the alternative, so I wouldn't enable this in my projects.

andy-hanson:
There is one [style guide](https://github.com/palantir/tslint/pull/2800#event-1103154098) that suggests to always use `continue;` if possible? Maybe if the rule had separate settings for functions, loops, and switch clauses?

adidahiya:
I'm still hesitant to add it if it's going to need that much configuration. I think it's perfectly fine for it to live as a third-party rule for those who wish to enforce that style.

andy-hanson:
@ajafff Would you accept a PR at `tslint-consistent-codestyle`?

ajafff:
@andy-hanson sure, go for it. I actually wanted to write a similar rule myself but didn't have the time

PR #2803, opened 2017-05-21
Enable triple-equals and allow ==undefined

ajafff:
#### PR checklist

  - [ ] Includes tests
- [ ] Documentation update

#### Overview of change:
Enables the triple-equals rule in this repo, allow `== undefined` and fixes all failures.
This could be a first step towards enabling `no-null-keyword` and/or `strict-type-predicates`

#### CHANGELOG.md entry:
[no-log]


ajafff:
`== null` or `== undefined` check for equality to `null` and `undefined`. Before this PR `== null` was used for that.

This PR also changes some places to use strict equality. In some other places I'm not sure or both `null` and `undefined` can occur.

> Why not always check strict equality?

We can remove that option for the rule later if needed. I'd rather to enable `strict-type-predicates` and allow `== undefined` *only* where `null` and `undefined` can occur.

adidahiya:
bump on this -- needs merge

PR #2804, opened 2017-05-21
Add 'switch-final-break' rule

andy-hanson:
#### PR checklist

  - [X] Includes tests
- [X] Documentation update

#### Overview of change:

Added the 'switch-final-break' lint rule, which:
* For a `case`, enforces 'no-switch-case-fall-through' even for the last case.
* For a `default:` clause, forbids a useless `break;`.

#### CHANGELOG.md entry:

[new-rule] `switch-final-break`


src/rules/switchFinalBreakRule.ts
```diff
@@ -0,0 +1,77 @@
...
+
+export class Rule extends Lint.Rules.AbstractRule {
+    /* tslint:disable:object-literal-sort-keys */
+    public static metadata: Lint.IRuleMetadata = {
+        ruleName: "switch-final-break",
+        description: Lint.Utils.dedent`
+            Enforces consistent use of 'break;' in the final clause of a 'switch' statement.
+            A 'default' clause should never have a 'break'.
+            A 'case' clause should always have a 'break' (or other control flow exit) even if it is the last clause.`,
+        optionsDescription: "Not configurable.",
```
ajafff: consider making this rule configurable with options like "always" and "never" (and if you really want it, you can add another option to handle `default` different. although I don't understand why the default clause should be handled differently at all).

For the record, I'd prefer setting this rule to "never" in this repository to avoid unnecessary statements.

src/rules/switchFinalBreakRule.ts
```diff
@@ -0,0 +1,89 @@
...
+import * as Lint from "../index";
+
+const OPTION_ALWAYS = "always";
+
+export class Rule extends Lint.Rules.AbstractRule {
+    /* tslint:disable:object-literal-sort-keys */
+    public static metadata: Lint.IRuleMetadata = {
+        ruleName: "switch-final-break",
+        description: Lint.Utils.dedent`
+            Forbids the final clause of a switch statement to have an unnecessary \`break;\`.`,
```
ajafff: Needs to be updated to sth like: Forbids *or requires* ...

src/rules/switchFinalBreakRule.ts
```diff
@@ -0,0 +1,89 @@
...
+        const block = clause.statements[0];
+        const statements = clause.statements.length === 1 && isBlock(block) ? block.statements : clause.statements;
+        const lastStatement = last(statements);
+        if (lastStatement !== undefined && lastStatement.kind === ts.SyntaxKind.BreakStatement) {
+            ctx.addFailureAtNode(lastStatement, Rule.FAILURE_STRING_NEVER);
+        }
+    }
+}
+
+function last<T>(arr: ReadonlyArray<T>): T | undefined {
```
ajafff: I don't know why you need this one, because you still have to check for `undefined` (instead of `arr.length` without this function), but I don't have a problem with it being there

test/rules/switch-final-break/always/test.ts.lint
```diff
@@ -0,0 +1,46 @@
+switch (x) {
```
ajafff: add a test where the last clause is empty

src/rules/switchFinalBreakRule.ts
```diff
@@ -0,0 +1,89 @@
...
+    ts.forEachChild(sourceFile, function cb(node) {
+        if (isCaseBlock(node)) {
+            check(node);
+        }
+        ts.forEachChild(node, cb);
+    });
+
+    function check(node: ts.CaseBlock): void {
+        const clause = last(node.clauses);
+        if (clause === undefined || clause.statements.length === 0) { return; }
```
ajafff: if `always` is specified, the rule should also complain if the final clause is empty

src/rules/switchFinalBreakRule.ts
andy-hanson: Better to ask forgiveness than permission. I'd rather check for an undefined return value than to have some test for whether the function I'm about to call should return a defined value.

PR #2805, opened 2017-05-21
Add 'binary-expression-operand-order' rule

andy-hanson:
#### PR checklist

  - [X] Includes tests
- [X] Documentation update

#### Overview of change:

Adds the 'binary-expression-operand-order' rule, which forbids expressions like `undefined == x`.

#### CHANGELOG.md entry:

[new-rule] `binary-expression-operand-order`


src/rules/yodaRule.ts
```diff
@@ -0,0 +1,92 @@
...
+            // Allow `"foo" + x` but not `1 + x`.
+            return node.left.kind === ts.SyntaxKind.StringLiteral;
+        case ts.SyntaxKind.MinusToken:
+        case ts.SyntaxKind.SlashToken:
+        case ts.SyntaxKind.PercentToken:
+        case ts.SyntaxKind.LessThanLessThanToken:
+        case ts.SyntaxKind.GreaterThanGreaterThanToken:
+        case ts.SyntaxKind.GreaterThanGreaterThanGreaterThanToken:
+        case ts.SyntaxKind.AsteriskAsteriskToken:
+        case ts.SyntaxKind.InKeyword:
```
ajafff: CommaToken

adidahiya:
I'm not a huge fan of "yoda". How about "binary-expression-operand-order"? As per #1371 and [stylelint's conventions](https://github.com/stylelint/stylelint/blob/master/docs/user-guide/about-rules.md#about-rule-names), I lean towards naming rules with the "thing the rule applies to" preceding the "thing the rule is checking".

PR #2842, opened 2017-05-29
curly: Add "as-needed" option

andy-hanson:
#### PR checklist

  - [X] Includes tests
- [X] Documentation update

#### Overview of change:

Adds the `never` option, which forbids unnecessary curly braces.

#### CHANGELOG.md entry:

[new-rule-option] `curly`: "as-needed" option


test/rules/curly/never/test.ts.lint
```diff
@@ -0,0 +1,24 @@
+if (so) {
```
ajafff: There's one edge case that needs to be handled. You can't remove the curly braces in the following example

```ts
if (foo) {
    if (bar)
        foo(bar);
} else
    bar();
```

Issue #2856, opened 2017-06-01
xxx.d.ts files aren't being excluded


winksaville:
### Bug Report

- __TSLint version__: 5.4.1
- __TypeScript version__: 2.3.4
- __Running TSLint via__: (pick one) CLI

#### TypeScript code being linted

Linting [Alsatian](https://github.com/alsatian-test/alsatian) using `npm run review-code` which runs tslint:
```
$ tslint './**/*.ts' --exclude './**/*.d.ts' --exclude './test/**/*.ts' --exclude './node_modules/**/*.ts'

ERROR: cli/alsatian-cli-options.d.ts[3, 14]: The class property 'fileGlobs' must be marked either 'private', 'public', or 'protected'
ERROR: cli/alsatian-cli-options.d.ts[5, 14]: The class property 'timeout' must be marked either 'private', 'public', or 'protected'
ERROR: cli/alsatian-cli-options.d.ts[7, 14]: The class property 'tap' must be marked either 'private', 'public', or 'protected'
ERROR: cli/alsatian-cli-options.d.ts[9, 14]: The class property 'versionRequested' must be marked either 'private', 'public', or 'protected'
ERROR: cli/alsatian-cli-options.d.ts[11, 14]: The class property 'helpRequested' must be marked either 'priv
```

with `tslint.json` configuration:

```json
$ cat tslint.json
{
  "extends": "tslint:recommended",
  "rules": {
     "array-type": [true, "generic"],
     "arrow-parens": false,
     "max-classes-per-file": false,
     "one-line": [ true, "check-open-brace", "check-whitespace"],
     "trailing-comma": false,
     "variable-name": [true, "ban-keywords", "check-format", "allow-leading-underscore"],
     "ordered-imports": false,
     "object-literal-sort-keys": false
  },
  "jsRules": {
  },
  "rulesDirectory": [
  ]
}
```

#### Actual behavior

Errors are reported from `*.d.ts` files when using 5.4.1 but no errors are reported in 5.3.2:

#### Expected behavior
Should pass with 5.4.1

winksaville:
[Here](https://github.com/alsatian-test/alsatian/issues/401) is the error I created on Alastain.

adidahiya:
duplicate of https://github.com/palantir/tslint/issues/2855

Issue #2855, opened 2017-06-01
Only the last --exclude is honoured in 5.4.1


sandersn:
### Bug Report
#2853 doesn't fix the --exclude issue for 
- __TSLint version__: 5.4.1
- __TypeScript version__: master
- __Running TSLint via__: CLI

#### TypeScript code being linted

Typescript. Here's how to repro:

```sh
$ git clone https://github.com/Microsoft/TypeScript.git lint-failure
$ cd lint-failure
$ npm install
$ gulp lint
```

Which runs the command line: `node node_modules/tslint/bin/tslint Gulpfile.ts 'scripts/tslint/*.ts' 'src/**/*.ts' --exclude src/lib/es5.d.ts --exclude 'src/lib/*.generated.d.ts' --format stylish`

Removing the second `--exclude src/lib/*.generated.d.ts` correctly excludes src/lib/es5.d.ts, but not any of the generated d.ts files. So I think the error is that multiple excludes are ignored.

#### Actual behavior
Thousands of errors from src/lib/es5.d.ts because it is not excluded.

#### Expected behavior
No errors from src/lib/es5.d.ts

adidahiya:
looking into this now

sandersn:
Note that it seems to work on node 8. Maybe this is a change in command line parsing libraries of node?

jmquigley:
It seems that using multiple "--exclude" options on one CLI are not working (at least not for me). I have a module build that is failing. I think that it is only respecting the last one from the CLI. According to the options one should be allowed to do this: "This option can be supplied multiple times if you need multiple globs to indicate which files to exclude." (tslint --help)

The CLI for sandersn shows multiple exclude options, where the original CLI above is using only one (which is why I believe 5.4.1 fixed it for you, but not for this other use case).

adidahiya:
We did switch to commander in 5.4, that's probably also related... https://github.com/palantir/tslint/pull/2689

sandersn:
I just upgraded to node 8 and still repro the issue locally. So our build system may not even be running tslint correctly on node 8.

PR #2858, opened 2017-06-01
Fix multiple --exclude arguments

ajafff:
#### PR checklist

- [x] Addresses an existing issue: #0000
  - [ ] Includes tests
- [ ] Documentation update

#### Overview of change:
Collect all `--exclude` options.
Fixes: #2855
Fixes: #2856
Fixes: #2855

There are no test included, because I'm in a hurry right now.


sandersn:
I don't know commander -- does this still treat multiple `--exclude` arguments as an array? Or does it change the syntax to something like `--exclude "prefix/path1/*,prefix/path2/*"` ?

ajafff:
Multiple --exclude as before.
Tested it locally

Am 01.06.2017 19:35 schrieb "Nathan Shively-Sanders" <
notifications@github.com>:

adidahiya:
I added a test case, this fix seems to work

Issue #2862, opened 2017-06-01
Running --fix on .ts Files Creates Numerous Syntax Errors.


taicho:
### Bug Report

- __TSLint version__: 5.4.0 (also tried 5.4.2)
- __TypeScript version__:  2.3.2
- __Running TSLint via__: CLI

#### TypeScript code being linted

```Command
tslint -p "./tsconfig.json" --fix --type-check -c tslint.json
```

with `tslint.json` configuration:

```json
{
  "extends": [
    "tslint:recommended"
  ],
  "rules": {
    "quotemark": [true,
      "single"
    ],
    "align": [true,
      "statements",
      "arguments"
    ],
    "radix": false,
    "space-before-function-paren":false,
    "array-type": [
      true,"array"
    ]
  }
}
```

#### Actual behavior
Breaks code by inserting commas and spaces where there shouldn't be any.

File before "fix":
```
import * as winston from 'winston';

const DefaultOptions = {
    transports: [
    new winston.transports.Console({
        json: true,
        colorize: true,
        stringify:true
    })],
    rewriters:[
        (level,message,meta)=>{
            if(meta)
            {
                meta.timeStamp = new Date();
            }
            return meta;            
        }
    ]

};

export const Logger = new winston.Logger(DefaultOptions);
```
File after "fix" (notice the split-lambda, spaces between "if", commas between "export" and "const"...bizarre output; this code does not compile successfully:
```
import * as winston from 'winston';

//Modify winston here.
const DefaultOptions = {
    transports: [
    new winston.transports.Console({
        json: true,
        colorize: true,
        stringify: true,
    })],
    rewriters :[
        (level ,message ,meta )= >{
            i f(meta)
            {
                meta.timeStamp = new Date();
            }
            return meta;
        }
    ]

};

,export, const Logger = new winston.Logger(DefaultOptions);
```

#### Expected behavior
Expected to "fix" whatever fixable rules can be fixed with acceptable anomalies. Running "fix" completely rendered 59 files broken and unable to compile because of syntax errors introduced. I realize any "fix" operation is risky which is why I did a commit right before but in any case this seems like a major issue. Even "simple" files like the one above don't appear to be parsed and fixed correctly. Is this a problem on my side? Has anybody else run into this?

P.S. Also worth noting is this codebase compiles perfectly using tsc and/or gulp-typescript and/or ts-loader before the "fixes" are applied.

ajafff:
> Is this a problem on my side? Has anybody else run into this?

[...]

Issue #2641, opened 2017-04-26
Code is ruined when prefer-const and another rule are fixed on the same line


Xapphire13:
### Bug Report
When multiple rules (one of which being `prefer-const`) target the same line and you use `--fix`, the code generated isn't valid TypeScript

- __TSLint version__: 5.1.0
- __TypeScript version__: 2.2.2
- __Running TSLint via__: **CLI** and **Atom**

#### TypeScript code being linted

```ts
function test() {
    let foo= "foo";
    let foobar = foo + "bar"
}
```

with `tslint.json` configuration:

```json
{
  "defaultSeverity": "warning",
  "rules": {
    "prefer-const": true,
    "semicolon": true,
    "whitespace": [
      true,
      "check-branch",
      "check-decl",
      "check-operator",
      "check-separator",
      "check-type"
    ]
  }
}
```

#### Actual behavior
```ts
function test() {
    const f oo= "foo";
    const foobar = foo + ";bar"
}
```
#### Expected behavior
```ts
function test() {
    const foo = "foo";
    const foobar = foo + "bar";
}
```

adidahiya:
related: #2326, #2606, #2556

Issue #2326, opened 2017-03-10
prefer-const fixer emits syntactically incorrect code


hellochar:
### Bug Report

- __TSLint version__: 4.5.1
- __TypeScript version__: 2.0.6
- __Running TSLint via__: CLI: `tslint --project tsconfig.json --type-check --fix`

#### TypeScript code being linted

```ts
    let adjustedIterativeDelay = this._getAdjustedIterativeDelay(numberOfSteps);
    return this.startDelay() + adjustedIterativeDelay * (Math.max(numberOfSteps - 1, 0)) + this.stepDuration();
```

with `tslint.json` configuration:

```json
{
  "extends": "tslint:latest",
  "rules": {
    "no-namespace": false
  }
}
```

#### Actual behavior

This was the emitted code:
```
    let adjustedIterativeDelay = this._getAdjustedIterativeDelay(numberOfStepsconst    return this.startDelay() + adjustedIterativeDelay * (Math.max(numberOfSteps - 1, 0)) + this.stepDuration();
```

#### Expected behavior

Only the `let` at the start is replaced by `const`

hellochar:
Update: I turned every rule off *except* for prefer-const: true and ran the fixer and everything worked, so I think there's something weird about fixing multiple rules at once (this is part of a larger codebase that probably had dozens of fixers applied all at once)

PR #2864, opened 2017-06-02
Bugfix --project with --fix

ajafff:
#### PR checklist

- [x] Addresses an existing issue: #2862
  - [x] Includes tests
- [ ] Documentation update

#### Overview of change:
A `Program` is immutable. Hence it was never updated when applying fixes. The following fixes for the same file were applied at the wrong location.

Now we update the `Program` (create a new one) after every fix.
[bugfix] Avoid overlapping fixes when using `--project` and `--fix`
Fixes: #2862
Fixes: #2641 
Fixes: #2326
Fixes: #2877

The implementation assumes that the program has been created with the default `Host`. That may not be the case if tslint is used as a library and the program is passed from outside. We would need a new API to get that working. That's tracked by https://github.com/Microsoft/TypeScript/issues/16217


test/executable/executableTests.ts
```diff
@@ -349,15 +347,21 @@ describe("Executable", function(this: Mocha.ISuiteCallbackContext) {
                 });
         });
 
-        it("can handle multiple '--exclude' globs", (done) => {
+        it("can apply fixes from multiple rules", function(done) {
+            this.slow(4000);
+            this.timeout(6000);
+            fs.writeFileSync("test/files/project-multiple-fixes/testfile.test.ts",
+                             fs.readFileSync("test/files/project-multiple-fixes/before.test.ts", "utf-8"));
```
adidahiya: we could probably use some better infrastructure so all this juggling isn't required. but this is fine for now.

adidahiya:
@ajafff could you resolve the merge conflicts please?

ajafff:
@adidahiya done, resolved merge conflicts and fixed new lint errors

Issue #2866, opened 2017-06-02
binary-expression-operand-order wrong for all-numeric expressions


sebek64:
### Bug Report

- __TSLint version__: 5.4.2
- __TypeScript version__: 5.3.3
- __Running TSLint via__: (pick one) CLI

#### TypeScript code being linted

```ts
private static readonly x = 1000 * 3600 * 24 * 7;
```

with `tslint.json` configuration:

```json
"binary-expression-operand-order": true
```

#### Actual behavior

Produces an error:

Literal expression should be on the right-hand side of a binary expression.

#### Expected behavior

Should not produce an error.

PR #2873, opened 2017-06-03
binary-expression-operand-order: Allow both sides to be literals

andy-hanson:
#### PR checklist

- [X] Addresses an existing issue:  #2866
  - [X] Includes tests
- [ ] Documentation update

#### Overview of change:

Allow `1 + 1`.

#### CHANGELOG.md entry:

[bugfix] `binary-expression-operand-order`: Allow if both sides of the binary expression are literals.


chrismbarr:
Related and probably already covered by the merged in fix, but I ran into this issue with the following line of code
```ts
public static dayInMilliseconds = 24 * 60 * 60 * 1000;
```

Issue #2881, opened 2017-06-04
Rule to count lines of methods, functions or arrow functions


TristanFAURE:
### Bug Report

- __TSLint version__: 5.4.2
- __TypeScript version__: 2.3.0
- __Running TSLint via__: Visual Studio 

#### TypeScript code being linted

```ts
// code snippet
```

with `tslint.json` configuration:

```json

```

JoshuaKGoldberg:
This was fixed by #4224. Hooray! 🙌

PR #2882, opened 2017-06-04
Provide a new rule to check size of functions, method, constructor or arrow functions

TristanFAURE:
#### PR checklist

- [X] Addresses an existing issue: #2881
- [X] enhancement
  - [X] Includes tests
- [ ] Documentation update

#### Overview of change:
New rule, new tests, modification on all.ts

This change count the number of lines of functions, methods, constructor or arrow functions.

Maybe you could suggest me an improvement about how getting correctly the part before the block (at failure creation). First time i highlight all the function node but i got errors with my test templates


palantirtech:
Thanks for your interest in [palantir/tslint](https://github.com/palantir/tslint), @TristanFAURE! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

TristanFAURE:
I will fix the error in the build it seems my tslint had bad configuration.

Nevertheless it seems that my new rule has been verified on other files. I think I should set a default value higher than 200 (which is common on industry) what would you suggest ?

ajafff:
Another thing to consider: Would it make sense to add special handling for nested functions and classes, i.e. don't count their lines to the total?

TristanFAURE:
I will ask guys from the quality in my office thursday, do you have an example ?

ajafff:
@TristanFAURE 
Take this function for example: https://github.com/palantir/tslint/blob/master/src/rules/arrayTypeRule.ts#L59
`walk` is actually split into several small functions, but they need to be located inside the function body to close over local variables (otherwise you end up with unwieldy parameter lists). With the current implementation of this rule all lines of the nested closures `checkArrayType` and `checkTypeReference` are counted to the total of `walk`.

TristanFAURE:
Hello,
I got feedbacks from quality guys.They ask me (and i don't have the answers) if nested functions can be called by external functions ?

if no (so they are private) they consider the current behavior of the rule correct (the enclosing function's size = current function + nested)

TristanFAURE:
@ajafff I have updated tests to manage my latest feedbacks from guys of my company. If you are ok with contribution I have nothing to add

ajafff:
I'm fine with the current implementation, although I think an option to not include nested functions in the total line count would make sense. Maybe @adidahiya or @nchen63 want to share their opinions regarding this.

TristanFAURE:
Why not ! 
Another question is : I have read in software quality documents some rules about line count of executing code (without blank one and comments). 

If the rule says < 100 lines and the quality document says < 100 executing lines it is okay, the software is compliant with the rule but in other rules do you count commented and blank lines ?

TristanFAURE:
Hello it seems that there are no comments from @adidahiya or @nchen63  ?

adidahiya:
Sorry, I don't really see how this would be very useful in practice. Are there any other code quality tools that do automated checks for function line length? I'm hesitant to add this as a core rule.

TristanFAURE:
Pmd :
http://pmd.sourceforge.net/pmd-4.3.0/rules/codesize.html (excessive method length)
A standard example : 
http://www.ecss.nl/wp-content/uploads/handbooks/ecss-q-hb/ECSS-Q-HB-80-04A30March2011.doc (p64)

marines:
@adidahiya It's one of the basic clean code metrics. Long functions are complex, hard to test and less readable. Enforcing function length will make the dev to split the code into smaller pieces which will be easier to test and understand due to greater amount of function names. Also the recommended number is 10 lines per function.

TristanFAURE:
Hello do you have some updates about this PR integration ?

aervin:
My team will be using this the day it is merged. Great idea for a rule!

adidahiya:
Ok, we can add this as a core rule. I do think that an option to ignore nested functions would be very useful.

TristanFAURE:
new commits implementing options which to include nestedFunctions or not.
I hope the implementation is ok for you

src/rules/maxMethodFunctionLineCountRule.ts
```diff
@@ -0,0 +1,108 @@
...
+
+    private countLines(node: ts.Node) {
+        const includesNested = this.options.includesNested;
+        let end = node.end;
+        if (!includesNested) {
+            const firstNestedFunction = (anode: ts.Node): ts.Node | undefined => {
+                if (isFunctionWithBody(anode)) {
+                    return anode;
+                }
+                return ts.forEachChild(anode, firstNestedFunction);
```
ajafff: this approach fails for code like this:
```ts
function reallyLong() {
    let foo = function() {}; // you end here and consider this line the end of the function
    statement;
    statement;
    function foo() {}
    statement;
    // ... and so on
}
```

Instead you probably want to find the last statement that is not a function declaration:

```ts
// node: ts.FunctionLikeDeclaration
// node.body: ts.Block
for (let i = node.body.statements.length - 1; i >= 0; --i) {
    if (!isFunctionDeclaration(node.body.statements[i])) {
        break;
    }
    end = node.body.statements[i].end;
}

```

TristanFAURE:
thank you for these feedback as soon as possible i will modify my PR. If everybody is OK i will use : "exclude-nested-trailing-functions" otherwise I will use your suggestions

src/rules/maxMethodFunctionLineCountRule.ts
TristanFAURE: @ajafff I have troubles to check if my node is a ts.FunctionLikeDeclaration because I don't have isFunctionLikeDeclaration. I want to perform this test to cast my node.

Is it normal that isFunctionLikeDeclaration is missing ?

ajafff: Yeah, seems like I never implemented that typeguard in tsutils. Anyways, I suggested `isFunctionDeclaration` (without the `Like`) because that's the only declaration that's also a statement.
Every other functionlike declaration is an expression, e.g. ArrowFunction and FunctionExpression, and will therefore never be true when checking statements: `isFunctionExpression(block.statements[i])` is always false

TristanFAURE:
Hello 
I tried another way substracting to final line numbers the number of lines of subfunctions
I don't really like the way I modify **nbLinesToDrop** in my function. Do you have visitor mechanisms that can return numbers? Initially I used forEachChild result but as the doc said the visit process is stoped at first result found.

'''typescript
class MaxMethodLine extends Lint.AbstractWalker<{limit: number; includesNested: boolean}> {
            if (isFunctionWithBody(node)) {
                const limit = this.options.limit;
                const lines = this.countLines(node.body!);
                if (lines > limit) {
                    this.addFailure(node.getStart(this.sourceFile), node.body!.pos, FAILURE_STRING(lines, limit));
                }
            }
        };
        return ts.forEachChild(sourceFile, cb);
    }

    private countLines(node: ts.Node): number {
        const includesNested = this.options.includesNested;
        let nbLinesToDrop = 0;
        if (!includesNested) {
            const nbLinesToDropFunction = (anode: ts.Node): void => {
                if (isFunctionDeclaration(anode) || isFunctionWithBody(anode)) {
                    nbLinesToDrop += this.countLines(anode.body!);
                }
                ts.forEachChild(anode, nbLinesToDropFunction);
            };
            ts.forEachChild(node, nbLinesToDropFunction);
        }
        const nbLines = ts.getLineAndCharacterOfPosition(this.sourceFile, node.end).line
            - ts.getLineAndCharacterOfPosition(this.sourceFile, node.getStart(this.sourceFile)).line;
        return nbLines + 1 - nbLinesToDrop;
    }

}

ajafff:
> I'm not sure to totally understand your solution
First to have something equivalent to the "true" boolean option imagin we have a "include-nested-trailing-functions"

My proposed solution would only exclude function declarations after the last statement in the function. That's why I called it "exclude-nested-**trailing**-functions"
```ts
function foo() {
    bar();
    baz();
    return; // line count ends here
    function bar() {
    }
    function baz() {
    }
}
```
It won't cover any other case where functions are declared in between statements.

----

Your latest solution:
I think it's fine to modify `nbLinesToDrop` inside the closure. There's no better way to do that.

It will exclude small arrow functions as well:
```ts
[1, 2, 3].forEach((el) => doStuff(el)); // this line will be excluded
[1, 2, 3].map((el) => el * Math.random()).forEach((el) => doStuff(el)); // this line will be excluded TWICE
```

And it traverses the whole AST subtree multiple times which decreases performance.

We probably need to find a tradeoff between usability and complexity.

TristanFAURE:
Ok thank you for the feedback ! 
sorry to not understand first the "trailing" part. I think from a quality assurance point of view it is not possible to exclude "only trailing" ....

About the closure code, indeed it is not efficient to analyze several time the AST but I don't understand my code doesn't work with this test 👍 

 // test nested functions
    function walk1(ctx: Lint.WalkContext<Option>): void {
        const { sourceFile, options: option } = ctx;
        return ts.forEachChild(sourceFile, function cb(node): void {
            switch (node.kind) {
                case ts.SyntaxKind.ArrayType:
                    checkArrayType(node as ts.ArrayTypeNode
            }
        });

        function checkArrayType(node: ts.ArrayTypeNode): void {
        ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~   [err % ("17")]
            const { elementType, parent } = node;
            if (option === "array" || option === "array-simple" && isSimpleType(elementType)) {
                return;
            }
...

TristanFAURE:
Hello @ajafff I just updated my pull request to correctly (i hope) compute the length of method (nested including or not) after merging latest changes.

But I don't understand why I have so much changes identified ? Did I do something wrong ?

Can you help me to clean my pull request, you can also check my latest changes

TristanFAURE:
@adidahiya or @ajafff your opinion ?

VinceOPS:
Hi guys! Thank you Tristan for raising this up, I was looking for this exact rule :) (function size).
Is there any chance to see it in the next versions of TSlint? @ajafff 
Thanks :)

TristanFAURE:
Hello is somebody ok to review this PR ?

johnwiseheart:
Hi @TristanFAURE, sorry for such a delay. Were slowly getting through all the PRs and hope to review this soon!

Thanks,
TSLint team

TisnKu:
Any progress on this thread?

MonsieurMan:
So what is the status of the PR ? This would make the life of a lot of developers easier..

TristanFAURE:
@MonsieurMan I will take a look of your PR as soon as possible thanks

MonsieurMan:
@TristanFAURE TristanFAURE/tslint#2
Here it is, I did not got through the second step now that I remember 😅

jdcrecur:
Hi guys, absolutely awesome that you are working on this feature for the tslint! I cannot wait for this to be released for our team. Any idea on how long?

MonsieurMan:
@jdcrecur It is up to the tslint team now, I suggest you +1 the issue and we wait for an update.

TristanFAURE:
as you want I can cancel my PR maybe it would be interesting to have the opinion of @MonsieurMan
concerning my need, in my context we decided to finally use this : https://github.com/Microsoft/tslint-microsoft-contrib/blob/master/src/maxFuncBodyLengthRule.ts

JoshuaKGoldberg:
Ok @TristanFAURE, we'll go with that one.

Thanks so much for sending this and discussing it! Once the rule goes in, this'll be great reference as discussions around the different pros and cons of different options.

MonsieurMan:
Cool !  
Sorry I don't have much time for open source right now

PR #2886, opened 2017-06-04
Added allow-empty-catch option to no-empty

### LINKED ISSUE #1088: Request: no-empty should ignore empty catch blocks

JoshuaKGoldberg:
#### PR checklist

- [x] Addresses an existing issue: #1088
  - [x] Includes tests
- [x] Documentation update

#### Overview of change:

Allows catch blocks to be empty.
Fixes #1088.

#### CHANGELOG.md entry:

[enhancement] Added allow-empty-catch option to `no-empty`


src/rules/noEmptyRule.ts
```diff
@@ -59,6 +71,16 @@ function walk(ctx: Lint.WalkContext<void>) {
     });
 }
 
+function isExcluded(node: ts.Node, options: Options): boolean {
+    const { parent } = node;
+    if (parent === undefined) {
```
ajafff: `parent` will never be undefined here

src/rules/noEmptyRule.ts
```diff
@@ -59,6 +71,16 @@ function walk(ctx: Lint.WalkContext<void>) {
...
+        return false;
+    }
+
+    return isExcludedConstructor(parent)
+        || (options.allowEmptyCatch === true && parent.kind === ts.SyntaxKind.CatchClause);
```
ajafff: you can simply rename `isExcludedConstructor` and add this condition to its return statement

src/rules/noEmptyRule.ts
```diff
@@ -20,16 +20,26 @@ import * as ts from "typescript";
 
 import * as Lint from "../index";
 
+const ALLOW_EMPTY_CATCH = "allow-empty-catch";
+
+interface Options {
+    allowEmptyCatch?: boolean;
```
ajafff: doesn't need to be optional

src/rules/noEmptyRule.ts
JoshuaKGoldberg: Ehh, I actually prefer the explicitness. But sure.

Issue #1628, opened 2016-10-13
readonly variables


TimBo93:
What do you think about a rule, which checks whether a variable can be made read-only?
Read-Only is a TypeScript 2 Feature. The rule could maybe determined by checking whether a field is assigned only inside the constructor or the field will be initialized at the same place like the declaration.

adidahiya:
yeah, this could be pretty cool; it's similar in spirit to the `prefer-const` rule (https://github.com/palantir/tslint/issues/491)

dyst5422:
Perhaps an extension of this would be a `functional-programming-only` rule that ensures all function inputs are marked with the `Readonly<T>` annotation

dyst5422:
Just found https://www.npmjs.com/package/tslint-immutable - Super!

dyst5422:
Scratch that, definitely does not check function inputs. Will put an issue on that project though as it seems more tuned to this type of rule.

akosyakov:
It should be configurable for access modifiers. `public` and `protected` properties could be assigned outside of the current class context.

adidahiya:
Fixed by #2896

PR #2896, opened 2017-06-07
Added prefer-readonly rule with fixes

JoshuaKGoldberg:
#### PR checklist

- [x] Addresses an existing issue: #1628
  - [x] Includes tests
- [x] Documentation update

#### Overview of change:

Adds a `prefer-readonly` rule for private member variables that are never re-assigned.

Am I missing any relevant test cases?

#### CHANGELOG.md entry:

[new-rule] `prefer-readonly`


src/rules/preferReadonlyRule.ts
```diff
@@ -0,0 +1,158 @@
...
+
+            case ts.SyntaxKind.Constructor:
+                this.handleConstructor(node as ts.ConstructorDeclaration);
+                break;
+
+            case ts.SyntaxKind.PropertyDeclaration:
+                this.handlePropertyDeclaration(node as ts.PropertyDeclaration);
+                break;
+
+            case ts.SyntaxKind.PropertyAccessExpression:
```
ajafff: Checking only property access may be enough for the initial implemenation, but we should consider using the type checker later to find all references

src/rules/preferReadonlyRule.ts
```diff
@@ -0,0 +1,158 @@
...
+        const parentScope = this.scope;
+        const childScope = this.scope = new ClassScope();
+
+        ts.forEachChild(node, this.visitNode);
+
+        this.finalizeScope(childScope);
+        this.scope = parentScope;
+    }
+
+    private handleConstructor(node: ts.ConstructorDeclaration) {
```
ajafff: handle parameter properties here

src/rules/preferReadonlyRule.ts
```diff
@@ -0,0 +1,158 @@
...
+        this.scope.inConstructor = true;
+        ts.forEachChild(node, this.visitNode);
+        this.scope.inConstructor = false;
```
ajafff: that's not sufficient, functions inside the constructor cannot assign readonly properties

src/rules/preferReadonlyRule.ts
```diff
@@ -0,0 +1,158 @@
...
+    }
+
+    private markMemberVariableAsModified(name: string) {
+        if (this.scope !== undefined && !this.scope.inConstructor) {
+            this.scope.modifiedMembers.add(name);
+        }
+    }
+
+    private finalizeScope(scope: ClassScope) {
+        scope.modifiedMembers.forEach((modifiedMember) => {
```
ajafff: there needs to be some kind of stack of scopes to handle nested classes

src/rules/preferReadonlyRule.ts
```diff
@@ -0,0 +1,158 @@
...
+            && utils.isModfierFlagSet(node, ts.ModifierFlags.Private)
+            && !utils.isModfierFlagSet(node, ts.ModifierFlags.Readonly)) {
+            this.scope.declaredMembers.set(node.name.getText(), node.name as ts.Identifier);
+        }
+
+        ts.forEachChild(node, this.visitNode);
+    }
+
+    private handlePropertyAccessExpression(node: ts.PropertyAccessExpression, parent: ts.Node | undefined) {
+        if (parent !== undefined) {
```
ajafff: parent will never be undefined here

ajafff: this function should really check if `node.expression` is a subtype of the current class

src/rules/preferReadonlyRule.ts
```diff
@@ -0,0 +1,158 @@
...
+        if (sourceFile.isDeclarationFile) {
+            return;
+        }
+
+        ts.forEachChild(sourceFile, this.visitNode);
+    }
+
+    private readonly visitNode = (node: ts.Node) => {
+        switch (node.kind) {
+            case ts.SyntaxKind.ClassDeclaration:
```
ajafff: what about `ClassExpression`?

src/rules/preferReadonlyRule.ts
```diff
@@ -0,0 +1,158 @@
...
+    private handleConstructor(node: ts.ConstructorDeclaration) {
+        this.scope.inConstructor = true;
+        ts.forEachChild(node, this.visitNode);
+        this.scope.inConstructor = false;
+    }
+
+    private handlePropertyDeclaration(node: ts.PropertyDeclaration) {
+        if (this.scope !== undefined
+            && utils.isModfierFlagSet(node, ts.ModifierFlags.Private)
+            && !utils.isModfierFlagSet(node, ts.ModifierFlags.Readonly)) {
```
ajafff: needs special handling for `static` members

src/rules/preferReadonlyRule.ts
JoshuaKGoldberg: I'd really prefer not to make this a typed-only rule. Seems like an unnecessary restriction, especially given the performance concerns. Would you be ok with it not doing that?

ajafff: There's no other choice. You need to know if the property is assigned on an object of a (sub)type of the current class. Consider the following code:

```ts
class A {
    private foo;
    private bar;
    private baz;

    public copyTo(other: A) {
        other.foo = this.foo; // this one needs to be recognized as assignment
    }

    public copy(other: {[key: string]: any} {
        // both should not be handled as assignments to `A#bar` or `A#baz`
        other.bar = this.bar;
        other.baz = this.baz;
    }
}
```

JoshuaKGoldberg: That's really interesting. I didn't think of the `copyTo`, thanks.

src/rules/preferReadonlyRule.ts
```diff
@@ -0,0 +1,158 @@
...
+            this.scope.declaredMembers.set(node.name.getText(), node.name as ts.Identifier);
+        }
+
+        ts.forEachChild(node, this.visitNode);
+    }
+
+    private handlePropertyAccessExpression(node: ts.PropertyAccessExpression, parent: ts.Node | undefined) {
+        if (parent !== undefined) {
+            switch (parent.kind) {
+                case ts.SyntaxKind.BinaryExpression:
```
ajafff: also needs to handle Postfix and PrefixUnaryExpression to handle `this.foo--` and `++this.bar`

src/rules/prefer-readonly/classScope.ts
```diff
@@ -0,0 +1,132 @@
...
+        }
+    }
+
+    return false;
+}
+
+export class ClassScope {
+    private readonly privateModifiableMembers = new Map<string, ParameterOrPropertyDeclaration>();
+    private readonly privateModifiableStatics = new Map<string, ParameterOrPropertyDeclaration>();
+    private readonly variableModifications = new Map<string, VariableModifications>();
```
ajafff: you could consider making two separate lists of assigned instance members and one for static members

src/rules/prefer-readonly/classScope.ts
```diff
@@ -0,0 +1,132 @@
...
+
+export type ParameterOrPropertyDeclaration = ts.ParameterDeclaration | ts.PropertyDeclaration;
+
+function typeIsOrHasBaseType(type: ts.Type, parentType: ts.Type) {
+    if (type.symbol === undefined || parentType.symbol === undefined) {
+        return false;
+    }
+
+    for (const baseType of [type, ...type.getBaseTypes()]) {
+        if (baseType.symbol !== undefined && baseType.symbol.name === parentType.symbol.name) {
```
JoshuaKGoldberg: Agreed. I didn't find a better alternative - would welcome suggestions!

src/rules/preferReadonlyRule.ts
```diff
@@ -0,0 +1,189 @@
...
+    }
+
+    private finalizeScope(scope: ClassScope) {
+        for (const violatingNode of scope.finalizeUnmodifiedPrivateNonReadonlys()) {
+            this.complainOnNode(violatingNode);
+        }
+    }
+
+    private complainOnNode(node: ParameterOrPropertyDeclaration) {
+        const fix = Lint.Replacement.appendText(node.name.getStart(), "readonly ");
```
JoshuaKGoldberg: Sure. Is there a reason you have a preference for that?

src/rules/preferReadonlyRule.ts
```diff
@@ -0,0 +1,189 @@
...
+            return;
+        }
+
+        ts.forEachChild(sourceFile, this.visitNode);
+    }
+
+    private readonly visitNode = (node: ts.Node) => {
+        switch (node.kind) {
+            case ts.SyntaxKind.ArrowFunction:
+            case ts.SyntaxKind.FunctionExpression:
```
JoshuaKGoldberg: Sorry, how come? I'm not seeing any benefit?

src/rules/preferReadonlyRule.ts
ajafff: the new modifier is automatically placed next to the others. you don't have to worry about comments, line breaks, etc.

src/rules/preferReadonlyRule.ts
ajafff: There are a few more kinds to check here: `ClassDeclaration`, `ClassExpression`, `FunctionDeclaration`, `GetAccessor`, `SetAccessor`, `MethodDeclaration` and maybe some more:
http://www.typescriptlang.org/play/#src=class%20Foo%20%7B%0A%20%20%20%20private%20readonly%20foo%3A%20number%3B%0A%0A%20%20%20%20constructor()%20%7B%0A%20%20%20%20%20%20%20%20this.foo%20%3D%202%3B%0A%20%20%20%20%20%20%20%20let%20self%20%3D%20this%3B%0A%20%20%20%20%20%20%20%20class%20Inner%20%7B%0A%20%20%20%20%20%20%20%20%20%20%20%20prop%20%3D%20self.foo%2B%2B%3B%0A%20%20%20%20%20%20%20%20%7D%0A%20%20%20%20%20%20%20%20let%20innerObj%20%3D%20%7B%0A%20%20%20%20%20%20%20%20%20%20%20%20doStuff()%20%7B%0A%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%2B%2Bself.foo%3B%0A%20%20%20%20%20%20%20%20%20%20%20%20%7D%2C%0A%0A%20%20%20%20%20%20%20%20%20%20%20%20get%20foo()%20%7B%0A%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20return%20self.foo%2B%2B%3B%0A%20%20%20%20%20%20%20%20%20%20%20%20%7D%2C%0A%0A%20%20%20%20%20%20%20%20%20%20%20%20set%20foo(foo)%20%7B%0A%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20self.foo%20%3D%20foo%3B%0A%20%20%20%20%20%20%20%20%20%20%20%20%7D%0A%20%20%20%20%20%20%20%20%7D%0A%0A%20%20%20%20%20%20%20%20function%20fn()%20%7B%0A%20%20%20%20%20%20%20%20%20%20%20%20self.foo%2B%2B%3B%0A%20%20%20%20%20%20%20%20%7D%0A%20%20%20%20%7D%0A%7D

adidahiya:
@ajafff mind taking another look here?

src/rules/preferReadonlyRule.ts
```diff
@@ -0,0 +1,199 @@
...
+
+import * as Lint from "../index";
+import { ClassScope, ParameterOrPropertyDeclaration } from "./prefer-readonly/classScope";
+
+export class Rule extends Lint.Rules.TypedRule {
+    public static metadata: Lint.IRuleMetadata = {
+        description: "Requires that private variables are marked as `readonly` if they're never modified outside of the constructor.",
+        descriptionDetails: Lint.Utils.dedent`
+            If a private variable is only assigned to in the constructor, it should be declared as \`readonly\`.
+        `,
```
ajafff: this rule could use a `rationale` to explain why private members should be readonly. I can't come up with a good reason

src/rules/prefer-readonly/classScope.ts
```diff
@@ -0,0 +1,132 @@
...
+
+    private constructorScopeDepth = OUTSIDE_CONSTRUCTOR;
+
+    public constructor(classNode: ts.Node, typeChecker: ts.TypeChecker) {
+        this.classType = typeChecker.getTypeAtLocation(classNode);
+        this.typeChecker = typeChecker;
+    }
+
+    public addDeclaredVariable(node: ParameterOrPropertyDeclaration) {
+        if (!utils.isModfierFlagSet(node, ts.ModifierFlags.Private)
```
ajafff: Is there a reason to still use the misnamed `isModfierFlagSet`?

src/rules/prefer-readonly/classScope.ts
ajafff: That's probably ok for now.
There's the internal typescript function `isTypeInstanceOf` in `checker.ts` that does exactly what we need here. We should ask to expose this functionality.

src/rules/preferReadonlyRule.ts
```diff
@@ -0,0 +1,199 @@
...
+        const childScope = this.scope = new ClassScope(node, this.typeChecker);
+
+        ts.forEachChild(node, this.visitNode);
+
+        this.finalizeScope(childScope);
+        this.scope = parentScope;
+    }
+
+    private handleConstructor(node: ts.ConstructorDeclaration) {
+        if (this.scope !== undefined) {
```
ajafff: `this.scope` should never be undefined, because a constructor can only occur inside of a class

src/rules/preferReadonlyRule.ts
```diff
@@ -0,0 +1,199 @@
...
+
+        ts.forEachChild(node, this.visitNode);
+
+        if (this.scope !== undefined) {
+            this.scope.exitConstructor();
+        }
+    }
+
+    private handlePropertyDeclaration(node: ts.PropertyDeclaration) {
+        if (this.scope !== undefined) {
```
ajafff: same as above, `this.scope` will not be undefined

src/rules/preferReadonlyRule.ts
```diff
@@ -0,0 +1,199 @@
...
+        if (this.scope === undefined
+            || (node.operator !== ts.SyntaxKind.PlusPlusToken && node.operator !== ts.SyntaxKind.MinusMinusToken)) {
+            return;
+        }
+
+        this.scope.addVariableModification(node.operand as ts.PropertyAccessExpression);
+    }
+
+    private handleDeleteExpression(node: ts.PropertyAccessExpression) {
+        if (this.scope !== undefined) {
```
ajafff: you could move this check to `handlePropertyAccessExpression`, so you don't have to check again here and in the other methods

src/rules/preferReadonlyRule.ts
JoshuaKGoldberg: Agreed. I was torn on this - although _we_ know this, the compiler requires `!`s which I'm not a fan of. Do you still prefer `this.scope!`?

src/rules/prefer-readonly/classScope.ts
```diff
@@ -0,0 +1,127 @@
...
+
+import * as utils from "tsutils";
+import * as ts from "typescript";
+
+const OUTSIDE_CONSTRUCTOR = -1;
+const DIRECTLY_INSIDE_CONSTRUCTOR = 0;
+
+export type ParameterOrPropertyDeclaration = ts.ParameterDeclaration | ts.PropertyDeclaration;
+
+function typeIsOrHasBaseType(type: ts.Type, parentType: ts.Type) {
```
adidahiya: this probably belongs in a more generic utils file? maybe `src/language/typeUtils.ts`

src/rules/prefer-readonly/classScope.ts
```diff
@@ -0,0 +1,127 @@
...
+    for (const baseType of typeAndBaseTypes) {
+        if (baseType.symbol !== undefined && baseType.symbol.name === parentType.symbol.name) {
+            return true;
+        }
+    }
+
+    return false;
+}
+
+export class ClassScope {
```
adidahiya: I know there's one bit of `readonly`-specific logic here, but I'd slightly prefer moving this class to `src/language/classScope.ts`

I know I mentioned in another thread that it's cool to have multiple files for a rule implementation, but in this case I think we can reasonably avoid it.

JoshuaKGoldberg:
In the AppVeyor build:

> `src/linter.ts(199,18): error TS2540: Cannot assign to 'program' because it is a constant or a read-only property.`

src/language/classScope.ts
```diff
@@ -0,0 +1,108 @@
+/**
```
ajafff: I don't think this file belongs here...
I know @adidahiya suggested moving it, but this is not reusable at all.
I'd prefer to include it in the rule's source file. It will only get you to a total of about 300 lines, which is still reasonable.

src/rules/preferReadonlyRule.ts
```diff
@@ -0,0 +1,193 @@
...
+        description: "Requires that private variables are marked as `readonly` if they're never modified outside of the constructor.",
+        descriptionDetails: Lint.Utils.dedent`
+            If a private variable is only assigned to in the constructor, it should be declared as \`readonly\`.
+        `,
+        optionExamples: [true],
+        options: null,
+        optionsDescription: "Not configurable.",
+        rationale: Lint.Utils.dedent`
+            Marking never-modified variables as readonly helps enforce the code's intent of keeping them as never-modified.
+            It can also help prevent accidental changes of members not meant to be changed.`,
```
ajafff: In general I declare properties `readonly` if I don't want them to be modified, regardless of the access modifier.
This rule forces me to declare all private properties as `readonly` that are *currently* not modified. If later on I decide to modify such a property, I need to edit 2 locations in the file, which bloats the diff.
I can imagine my co-workers not making a change (or making a really awkward one instead), because they think the property is `readonly` on purpose and needs to stay that way. I think this may cause more confusion than good.

JoshuaKGoldberg: > I think this may cause more confusion than good.

Funny, I've had the opposite problem: coworkers reassigning properties that were never meant to be reassigned. 

Partially to alleviate that concern, I've added an `only-inline-lambdas` option that stops checking everything except `private member = () => {}`-style arrow functions.

JoshuaKGoldberg:
Ping @adidahiya ?

adidahiya:
lgtm -- CI status is not being reported, maybe because of the typescript@next test runner

ajafff:
@adidahiya CI was successful, but did not report. That's weird. Maybe reopening the PR and/or manually starting CI helps.

Besides this issue, there is still a TODO for you: https://github.com/palantir/tslint/pull/3490#issuecomment-344841563

Issue #2898, opened 2017-06-07
no-unused-variables autofix leaves empty lines


pablonete:
### Bug Report

- __TSLint version__: master
- __TypeScript version__: latest
- __Running TSLint via__: CLI

#### TypeScript code being linted

Issue is noticeable in [rule tests](https://github.com/palantir/tslint/blob/master/test/rules/no-unused-variable/default/import.ts.lint#L15-L20). Code to fix:

```ts
// This import statement is unused and will be deleted along with this comment.
import {a5, a6} from "a";
import {a7} from "a";
import {a8, a9, a10} from "a";
```

Which results on this [fix](https://github.com/palantir/tslint/blob/master/test/rules/no-unused-variable/default/import.ts.fix#L12-L15):

```ts
// This import statement is unused and will be deleted along with this comment.


import {a8} from "a";
```

#### Actual behavior

When autofix deletes the whole import sentence, it leaves an empty line.

#### Expected behavior

Remove the line break when the whole sentence is deleted, so expected fix should be:

```ts
// This import statement is unused and will be deleted along with this comment.
import {a8} from "a";
```

PR #2901, opened 2017-06-08
Don't leave blank lines when no-unused-variable autofix removes whole import

pablonete:
#### PR checklist

- [x] Addresses an existing issue: #2898 
- [x] New enhancement
  - [x] Includes tests

#### Overview of change:

When an unused import is deleted, the line break is removed too.
Unless there are more sentences in the same line, so the break is preserved.

- Can you think on other scenarios we need to cover?
- This is my first time contributing to TSLint and consuming TS compiler API. Let me know if there's something that can be improved.

#### CHANGELOG.md entry:

"[enhancement] `no-unused-variable`: Don't leave empty lines when import is removed"

### DIFF
--- a/src/rules/noUnusedVariableRule.ts
+++ b/src/rules/noUnusedVariableRule.ts
@@ -227,13 +227,27 @@ function addImportSpecifierFailures(ctx: Lint.WalkContext<void>, failures: Map<t
         function removeAll(errorNode: ts.Node, failure: string): void {
             const start = importNode.getStart();
             let end = importNode.getEnd();
-            if (ctx.sourceFile.getLineAndCharacterOfPosition(start).character === 0 &&
-                ctx.sourceFile.getLineEndOfPosition(end) === end) {
-                end++;
+            if (isEntireLine(start, end)) {
+                end = getNextLineStart(end);
             }
 
             const fix = Lint.Replacement.deleteFromTo(start, end);
             ctx.addFailureAtNode(errorNode, failure, fix);
+        }
+
+        function isEntireLine(start: number, end: number): boolean {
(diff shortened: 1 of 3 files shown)


palantirtech:
Thanks for your interest in [palantir/tslint](https://github.com/palantir/tslint), @pablonete! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

src/rules/noUnusedVariableRule.ts
```diff
@@ -225,7 +225,14 @@ function addImportSpecifierFailures(ctx: Lint.WalkContext<void>, failures: Map<t
         }
 
         function removeAll(errorNode: ts.Node, failure: string): void {
-            const fix = Lint.Replacement.deleteFromTo(importNode.getStart(), importNode.getEnd());
+            const start = importNode.getStart();
+            let end = importNode.getEnd();
+            if (ctx.sourceFile.getLineAndCharacterOfPosition(start).character === 0 &&
+                ctx.sourceFile.getLineEndOfPosition(end) === end) {
+                end++;
```
ajafff: that will only work if you're using `\n`. if there's `\r\n`, it will only remove `\r`

alternative to `end++`:
```ts
const nextLine = ctx.sourceFile.getLineAndCharacterOfPosition(end).line + 1;
const lineStarts = ctx.sourceFile.getLineStarts();
if (lineStarts.length !== nextLine) {
    end = lineStarts[nextLine];
}
```

pablonete: Nice suggestion, done!

Issue #2910, opened 2017-06-11
completed-docs: add option for enum members


reduckted:
The `completed-docs` rule allows documentation to be required for enums, but there is no option to require documentation for enum _members_. A new option to also check enum members would be nice.

PR #2911, opened 2017-06-11
completed-docs: Add support for enum members

reduckted:
#### PR checklist

- [x] Addresses an existing issue: #2910 
  - [x] Includes tests
- [x] Documentation update

#### Overview of change:

I've extended the `completed-docs` rule to check enum members. This is disabled by default and can be enabled by including a new argument "enum-members". This can also be configured to only require documentation for members of internal enums or exported enums.

#### CHANGELOG.md entry:

[new-rule-option] `completed-docs`: Add `enum-members` option

### DIFF
--- a/src/rules/completedDocsRule.ts
+++ b/src/rules/completedDocsRule.ts
@@ -39,6 +39,7 @@ export const ALL = "all";
 
 export const ARGUMENT_CLASSES = "classes";
 export const ARGUMENT_ENUMS = "enums";
+export const ARGUMENT_ENUM_MEMBERS = "enum-members";
 export const ARGUMENT_FUNCTIONS = "functions";
 export const ARGUMENT_INTERFACES = "interfaces";
 export const ARGUMENT_METHODS = "methods";
@@ -66,6 +67,7 @@ export type All = typeof ALL;
 export type DocType = All
     | typeof ARGUMENT_CLASSES
     | typeof ARGUMENT_ENUMS
+    | typeof ARGUMENT_ENUM_MEMBERS
     | typeof ARGUMENT_FUNCTIONS
     | typeof ARGUMENT_INTERFACES
     | typeof ARGUMENT_METHODS
@@ -165,6 +167,7 @@ export class Rule extends Lint.Rules.TypedRule {
 
             * \`"${ARGUMENT_CLASSES}"\`
             * \`"${ARGUMENT_ENUMS}"\`
+            * \`"${ARGUMENT_ENUM_MEMBERS}"\`
             * \`"${ARGUMENT_FUNCTIONS}"\`
             * \`"${ARGUMENT_INTERFACES}"\`
             * \`"${ARGUMENT_METHODS}"\`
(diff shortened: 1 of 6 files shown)


src/rules/completedDocsRule.ts
```diff
@@ -388,14 +399,20 @@ class CompletedDocsWalker extends Lint.ProgramAwareRuleWalker {
...
         const { name } = node;
         if (name === undefined) {
             return;
         }
 
+        // If we were not given a special node to use for checking
+        // the requirements, use the node that is being checked.
+        if (requirementNode === undefined) {
+            requirementNode = node;
+        }
```
JoshuaKGoldberg: You could replace these lines with `requirementNode: ts.Declaration = node`

src/rules/completedDocsRule.ts
```diff
@@ -405,19 +422,27 @@ class CompletedDocsWalker extends Lint.ProgramAwareRuleWalker {
         }
 
         const comments = symbol.getDocumentationComment();
-        this.checkComments(node, nodeType, comments);
+        this.checkComments(node, this.describeNode(nodeType), comments, requirementNode);
+    }
+
+    private describeNode(nodeType: DocType): string {
```
JoshuaKGoldberg: Seems a little hardcody to just do this to `ARGUMENT_ENUM_MEMBERS`. How about generalizing the logic by taking in both `node` and `requirementNode`? If `node` is an ancestor of `requirementNode`, append `" members"` to the description.

test/rules/completed-docs/types/test.ts.lint
```diff
@@ -138,6 +138,19 @@ enum EmptyEnum { }
  */
 enum GoodEnum { }
 
+/**
+ * ...
+ */
+enum EnumWithMember {
```
JoshuaKGoldberg: Nit: EnumWithMember**s**

src/rules/completedDocsRule.ts
```diff
@@ -39,6 +39,7 @@ export const ALL = "all";
 
 export const ARGUMENT_CLASSES = "classes";
 export const ARGUMENT_ENUMS = "enums";
+export const ARGUMENT_ENUM_MEMBERS = "enumMembers";
```
JoshuaKGoldberg: Nit: other rules have options like `"check-accessor"`, `"member-variable-declaration"`, "`array-simple`", etc. This should probably be `"enum-member"`.

src/rules/completedDocsRule.ts
reduckted: Oh cool, I didn't realise you could do that!

src/rules/completedDocsRule.ts
reduckted: Yeah, I thought about some sort of camel-case-to-words thing, but it seemed like overkill. Now that I've changed `enumMembers` to `enum-members`, I could just do:

    return nodeType.replace("-", " ");

reduckted:
What's the convention for pushing changes to a pull request? Do I make a new commit and push it, or should I be amending my previous commit and force pushing?

src/rules/completedDocsRule.ts
reduckted: `enum-member` or `enum-members`?

JoshuaKGoldberg: Ah yes, the plural :)

JoshuaKGoldberg:
I've always just pushed additional and nobody's complained yet.

reduckted:
FYI, I've just rebased these changes onto master to fix the merge conflict caused by #2924.

adidahiya:
thanks @reduckted!

Issue #2916, opened 2017-06-11
FR: Check that cases aren't duplicated in switch statement


erikvullings:
When creating a new case in a switch statement, it may happen that I duplicate an existing case. Wouldn't it be nice if this would generate an error (or at least a warning)

```ts
const a = create();
switch (a) {
  case 1:
     ...
     break;
  case 2:
     ...
     break;
  case 1:  // THIS SHOULD GENERATE AN ERROR
     ...
     break;
  default:
     ...
     break;

```

JoshuaKGoldberg:
Oh, this was fixed by #2937! 🎉

PR #2937, opened 2017-06-20
Added a no-duplicate-switch-case rule

JoshuaKGoldberg:
#### PR checklist

- [x] Addresses an existing issue: #2916
  - [x] Includes tests
- [x] Documentation update

#### Overview of change:

Added a `no-duplicate-switch-case` rule to enforce that no switch case should contain the same textual value as a previous switch statement.

#### CHANGELOG.md entry:

[new-rule] `no-duplicate-switch-case`


src/rules/noDuplicateSwitchCaseRule.ts
```diff
@@ -0,0 +1,72 @@
...
+    };
+
+    public static readonly FAILURE_STRING_FACTORY = (text: string) => `Duplicate switch case: '${text}'.`;
+
+    public apply(sourceFile: ts.SourceFile): Lint.RuleFailure[] {
+        return this.applyWithFunction(sourceFile, walk);
+    }
+}
+
+const walk = (ctx: Lint.WalkContext<void>): void => {
```
ajafff: consider converting these lambdas to regular function declarations

src/rules/noDuplicateSwitchCaseRule.ts
```diff
@@ -0,0 +1,72 @@
...
+        }
+
+        ts.forEachChild(node, cb);
+    };
+
+    const visitCaseBlock = (() => {
+        const previousCases = new Set<string>();
+
+        return (node: ts.CaseBlock): void => {
+            previousCases.clear();
```
ajafff: no need to return a closure or clear the set.
just make `previousCases` a local variable and loop over the cases.

JoshuaKGoldberg: It's a minor performance optimization to keep the same set across iterations. Do you consider the lessened clarity not worth it?

ajafff: I would guess clearing the set is as expensive as creating a new one.
I don't think it's worth optimizing here

src/rules/noDuplicateSwitchCaseRule.ts
```diff
@@ -0,0 +1,72 @@
...
+
+            for (const clause of node.clauses) {
+                if (clause.kind === ts.SyntaxKind.DefaultClause) {
+                    continue;
+                }
+
+                const text = clause.expression.getText();
```
ajafff: consider passing `ctx.sourceFile` to `.getText` to avoid unnecessary work

src/rules/noDuplicateSwitchCaseRule.ts
JoshuaKGoldberg: Is there a reason behind the preference for regular function declarations? Elsewhere I've seen folks prefer arrow lambdas for the saner scoping & lack of hoisting weirdness.

ajafff: The rest of the codebase seems to follow the pattern to use function declarations where possible and only use arrow functions where this reference matters.
I prefer regular functions because I think it's easier to notice while scanning through the code

ajafff:
@JoshuaKGoldberg The rule needs to be added to `all.ts` and `latest.ts`. That's why CI is failing.

ajafff:
I'd like to get this merged. So I updated the configuration presets and merged master.

@adidahiya do you want to take another look before merging this?

JoshuaKGoldberg:
Thanks @ajafff for the updates!

A side note: there are a few other PRs from me waiting for action (on my part). I don't know when I'll be able to get to them :(

PR #2942, opened 2017-06-22
Check type for no-string-throw

felixfbecker:
#### PR checklist

  - [ ] Includes tests
- [ ] Documentation update

#### Overview of change:

The `no-string-throw` rule currently only checks for string literals. That means it doesn't catch the following:
```ts
const msg = localize('something went wrong')
throw msg
```

Tests are failing, but I couldn't get the VS Code debugger to stop on breakpoints with the provided launch.json. Any help appreciated.

#### CHANGELOG.md entry:

Make `no-string-rule` check type of expression

### DIFF
--- a/src/rules/noStringThrowRule.ts
+++ b/src/rules/noStringThrowRule.ts
@@ -20,7 +20,7 @@ import * as ts from "typescript";
 
 import * as Lint from "../index";
 
-export class Rule extends Lint.Rules.AbstractRule {
+export class Rule extends Lint.Rules.TypedRule {
     /* tslint:disable:object-literal-sort-keys */
     public static metadata: Lint.IRuleMetadata = {
         ruleName: "no-string-throw",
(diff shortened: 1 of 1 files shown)


palantirtech:
Thanks for your interest in [palantir/tslint](https://github.com/palantir/tslint), @felixfbecker! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

ajafff:
There was already some discussion about making this rule type checked. The conclusion was that it's probably sufficient to do only static checks. (Too lazy to search for it right now, sorry)

Instead of making this rule a `TypedRule` you should consider making it an `OptionallyTypedRule`. You use the type checker if available and fall back to the old implementation otherwise.

Re failing Tests:
It seems that the test for `no-string-throw` is run in the `executableTests.ts` which spawns a new process of tslint and checks the return code or error. That's a debugging nightmare.
The easier way it to run the rule tests without the mocha tests: `npm run test:rules` and fix the errors there before running all tests. 
It may be fixed as a side effect of making the rule optionally typed as suggested above.

felixfbecker:
To share the use case: https://github.com/Microsoft/vscode/pull/29034
VS Code is localized, so error messages almost never are plain strings, but look like this
```ts
// good
throw new Error(nls.localize('TaskRunner.UNC', 'Can\'t execute a shell command on an UNC drive.')));
// bad
throw nls.localize('TaskRunner.UNC', 'Can\'t execute a shell command on an UNC drive.'));
```
without type checking, this rule can't catch the latter.

Currently many errors are actually returned with WinJS' `return Promise.wrapError()`, but the amount of `throw`n Errors will increase with a switch to `async`/`await`.

ajafff:
@felixfbecker I'm not against this change in general. I think it totally makes sense when linting a project.

I just wanted to make sure the rule still works without the type checker.

ajafff:
@felixfbecker Do you still want to get this implemented?
You'd only need to restore the old behavior if the TypeChecker is not available.

Issue #2912, opened 2017-06-11
completed-docs checks too many types of variables


reduckted:
### Bug Report

- __TSLint version__: 5.4.3
- __TypeScript version__: 2.3.4
- __Running TSLint via__: Anything

#### TypeScript code being linted

```ts
const bar = 2;
try {
} catch (ex) {
}
for (let i = 0; i < 1; i++) { }
for (let i of []) { }
for (let p in {}) { }
function foo() {
    let x = 1;
}
```

with `tslint.json` configuration:

```json
{
    "rules": {
        "completed-docs": [true, "variables"]
    }
}
```

#### Actual behavior

Errors are produced for every variable:

    ERROR: index.ts[1, 7]: Documentation must exist for variables.
    ERROR: index.ts[3, 10]: Documentation must exist for variables.
    ERROR: index.ts[5, 10]: Documentation must exist for variables.
    ERROR: index.ts[6, 10]: Documentation must exist for variables.
    ERROR: index.ts[7, 10]: Documentation must exist for variables.
    ERROR: index.ts[9, 11]: Documentation must exist for variables.

#### Expected behavior

Only an error for the first variable that is declared at the file-level is produced:

    ERROR: index.ts[1, 7]: Documentation must exist for variables.

---

Variables in catch blocks and for loops can't have documentation, so they should never be checked. 

The variable inside the function is an interesting case. If you add documentation to it, the error goes away, but does anyone document there local variables like that? It seems like a strange thing to do.

In my opinion, the intent behind wanting variables to be documented, is because they are exported or "global" to the file they are in, and those types of variables would only be declared at the file-level.

If we're happy to restrict this rule to only check variables at the file-level, then I've got a PR that's ready to go.

Issue #2335, opened 2017-03-13
[completed-docs] exported variable parsed as internal variable


bumbleblym:
### Bug Report

- __TSLint version__: 4.5.1
- __TypeScript version__: 2.2.1
- __Running TSLint via__: (pick one) CLI / ~~Node.js API / VSCode / grunt-tslint / Atom / Visual Studio / etc~~

#### TypeScript code being linted

```ts
export const Null = null
```

with `tslint.json` configuration:

```json
{
  "rules": {
    "completed-docs": [true, {
      "variables": {
        "visibilities": [
          "exported"
        ]
      }
    }]
  }
}
```

#### Actual behavior
`"all"`: lint error
`"exported"`: no lint error
`"internal"`: lint error
#### Expected behavior
`"all"`: lint error
`"exported"`: lint error
`"internal"`: no lint error

`exported` variable is being parsed as `internal`. [repro](https://github.com/bumbleblym/tslint-bug)

Also,  [documented `visibilities` type is inconsistent](https://palantir.github.io/tslint/rules/completed-docs/):
[schema](https://github.com/palantir/tslint/blob/ee721bb4fff7a3b75901cc0b29d809c157eefd5d/src/rules/completedDocsRule.ts#L104): `enum`
[example](https://github.com/palantir/tslint/blob/ee721bb4fff7a3b75901cc0b29d809c157eefd5d/src/rules/completedDocsRule.ts#L205): `array enum`
[implementation](https://github.com/palantir/tslint/blob/ee721bb4fff7a3b75901cc0b29d809c157eefd5d/src/rules/completedDocsRule.ts#L277): `array enum`

perhaps #629 should consider docs as well.

JoshuaKGoldberg:
I'll take a look after #2415 goes in.

reduckted:
This bug is caused by the fact that the `VariableDeclaration` does not have modifiers. The modifiers are on the `VariableStatement` which is a grandparent of a `VariableDeclaration`.

    VariableStatement               <-- The relevant modifiers are here.
     └─ VariableDeclarationList
         └─ VariableDeclaration     <-- This node is checked.

I've made a fix for this, but it requires some changes that I've made in PR #2911 to allow the modifiers from a different node to be used. Once that PR is merged, I'll submit another PR for this bug.

JoshuaKGoldberg:
Bump @reduckted - still interested in a followup PR?

reduckted:
[...]

PR #2950, opened 2017-06-24
completed-docs: Limit variable checking to file-level variables and use correct visibility

reduckted:
#### PR checklist

- [x] Addresses an existing issue: #2335, #2912 
  - [x] Includes tests
- [x] Documentation update

#### Overview of change:

This PR fixes two issues:

1. When checking if a variable meets the required visibility, the modifiers on the variable were being used, but the variable doesn't have modifiers - the modifiers are on the variable statement which is two levels up. These changes step up two levels to get the correct modifiers.
2. When checking variables, _all_ variables were being checked, including those inside for loops, catch statements, and local variables inside functions. This PR limits variable checking to variables that have a parent of a `SourceFile` - i.e. variables at the file-level.

As a side note, I've also fixed the `.editorconfig` file to have a rule for tslint.json files that enforces an indent size of 2 instead of the default 4, because all tslint.json files I could see used an indentation of 2.

#### CHANGELOG.md entry:

[bugfix] `completed-docs`: Uses correct visibility of variables.
[bugfix] `completed-docs`: Only checks variables at the file-level.

### DIFF
--- a/src/rules/completedDocsRule.ts
+++ b/src/rules/completedDocsRule.ts
@@ -15,6 +15,7 @@
  * limitations under the License.
  */
 
+import { isVariableDeclarationList, isVariableStatement } from "tsutils";
 import * as ts from "typescript";
 
 import * as Lint from "../index";
@@ -286,7 +287,7 @@ abstract class Requirement<TDescriptor extends RequirementDescriptor> {
 class BlockRequirement extends Requirement<IBlockRequirementDescriptor> {
     public readonly visibilities: Set<Visibility> = this.createSet(this.descriptor.visibilities);
 
-    public shouldNodeBeDocumented(node: ts.Declaration): boolean {
+    public shouldNodeBeDocumented(node: ts.Node): boolean {
         if (this.visibilities.has(ALL)) {
             return true;
         }
@@ -303,11 +304,11 @@ class ClassRequirement extends Requirement<IClassRequirementDescriptor> {
     public readonly locations: Set<Location> = this.createSet(this.descriptor.locations);
     public readonly privacies: Set<Privacy> = this.createSet(this.descriptor.privacies);
 
-    public shouldNodeBeDocumented(node: ts.Declaration) {
+    public shouldNodeBeDocumented(node: ts.Node) {
         return this.shouldLocationBeDocumented(node) && this.shouldPrivacyBeDocumented(node);
     }
 
-    private shouldLocationBeDocumented(node: ts.Declaration) {
+    private shouldLocationBeDocumented(node: ts.Node) {
         if (this.locations.has(ALL)) {
             return true;
         }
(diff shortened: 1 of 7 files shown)


src/rules/completedDocsRule.ts
```diff
@@ -395,11 +395,31 @@ class CompletedDocsWalker extends Lint.ProgramAwareRuleWalker {
...
-        this.checkNode(node, ARGUMENT_VARIABLES);
+        this.checkVariable(node);
         super.visitVariableDeclaration(node);
     }
 
-    private checkNode(node: ts.NamedDeclaration, nodeType: DocType, requirementNode: ts.Declaration = node): void {
+    private checkVariable(node: ts.VariableDeclaration) {
+        // Only check variables in variable declaration lists
+        // and not variables in catch clauses and for loops.
+        if (node.parent === undefined || node.parent.kind !== ts.SyntaxKind.VariableDeclarationList) {
```
ajafff: `node.parent` as well as `list.parent` and `statement.parent` will never be undefined. You can simply assert that using for example `node.parent!.kind`

src/rules/completedDocsRule.ts
```diff
@@ -395,11 +395,31 @@ class CompletedDocsWalker extends Lint.ProgramAwareRuleWalker {
...
+
+        const list = node.parent;
+        if (list.parent === undefined || list.parent.kind !== ts.SyntaxKind.VariableStatement) {
+            return;
+        }
+
+        // Only check variables at the file-level and not
+        // variables declared inside functions and other things.
+        const statement = list.parent;
+        if (statement.parent === undefined || statement.parent.kind === ts.SyntaxKind.SourceFile) {
```
ajafff: what about (exported) variables inside a namespace (`ts.SyntaxKind.ModuleBlock`)?

reduckted: Good point 👍

reduckted:
Just pushed up some changes suggested by @ajafff.

src/rules/completedDocsRule.ts
```diff
@@ -395,11 +396,42 @@ class CompletedDocsWalker extends Lint.ProgramAwareRuleWalker {
...
+        }
+
+        const statement = list.parent!;
+        if (!isVariableStatement(statement)) {
+            return;
+        }
+
+        // Only check variables at the namespace/module-level or file-level
+        // and not variables declared inside functions and other things.
+        let check = false;
```
ajafff: `statement.parent` can not be undefined. This check ccould be simplified to

```ts
switch (statement.parent!.kind) {
    case ts.SyntaxKind.SourceFile:
    case ts.SyntaxKind.ModuleBlock:
        break;
    default:
        return;
}
```

reduckted: Doh! Sorry, missed that on your previous comment. Am I correct in assuming it's defined as `parent?: Node` because it _can_ be undefined if the `setParentNodes` argument is `false` when you load the TypeScript source file. But since TSLint always passes `true` to that argument, the parent nodes will always be set (except on the root `SourceFile` node)?

ajafff: That's exactly how it works.

ajafff:
Thanks @reduckted

PR #2959, opened 2017-06-26
Adds new rule "space-within-parens"

aminpaks:
#### PR checklist

  - [x] Includes tests
- [ ] Documentation update

#### Overview of change:
Adds new rule to verify amount of whitespace within parentheses.

Everything.

#### CHANGELOG.md entry:

[new-rule] `space-within-parens`


palantirtech:
Thanks for your interest in [palantir/tslint](https://github.com/palantir/tslint), @aminpaks! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

src/rules/spaceWithinParensRule.ts
```diff
@@ -0,0 +1,167 @@
+/**
+ * @license
+ * Copyright 2013 Palantir Technologies, Inc.
```
ajafff: 2017

src/rules/spaceWithinParensRule.ts
```diff
@@ -0,0 +1,167 @@
...
+                if (length > this.options.size) {
+                    length -= this.options.size;
+                }
+                this.checkAndAddFailureAt(position, length, whitespaceLength);
+            }
+        }
+    }
+
+    private checkAndAddFailureAt(position: number, length: number, spaceCount: number = 0) {
+        if (this.failures.some(f => f.getStartPosition().getPosition() === position)) {
```
ajafff: do you really need to remove duplicate failures here? If yes, there's probably something wrong with the rule

src/rules/spaceWithinParensRule.ts
```diff
@@ -0,0 +1,167 @@
...
+    public walk(sourceFile: ts.SourceFile) {
+        const cb = (node: ts.Node): void => {
+            this.checkNode(node);
+            ts.forEachChild(node, cb);
+        };
+        return ts.forEachChild(sourceFile, cb);
+    }
+
+    private getOpenParenToken(node: ts.Node): ts.Node | undefined {
+        return Lint.childOfKind(node, ts.SyntaxKind.OpenParenToken);
```
ajafff: if you do this for every node, you could just use `forEachToken` from `tsutils` to visit every token instead of visiting every node and finding the paren tokens.

test/rules/space-within-parens/force-space/test.ts.lint
```diff
@@ -0,0 +1,34 @@
+// simple expression
+var x = [{
```
ajafff: add a test for multiline call expression. This should pass IMO

```ts
foo(
    bar,
    baz,
);
```

src/rules/spaceWithinParensRule.ts
```diff
@@ -0,0 +1,167 @@
...
+export class Rule extends Lint.Rules.AbstractRule {
+    /* tslint:disable:object-literal-sort-keys */
+    public static metadata: Lint.IRuleMetadata = {
+        ruleName: "space-within-parens",
+        description: "Enforces spaces within parentheses or disallow them.",
+        hasFix: true,
+        optionsDescription: Lint.Utils.dedent`
+            You may enforce the amount of whitespace within parentheses.
+        `,
+        options: {
```
ajafff: simplify to
```ts
options: {type: "number", min: 0}
```

test/rules/space-within-parens/default/tslint.json
```diff
@@ -0,0 +1,5 @@
+{
```
ajafff: please rename the folder to `nospace` or something like that since there is no default for this rule

src/rules/spaceWithinParensRule.ts
```diff
@@ -0,0 +1,167 @@
...
+            this.checkOpenParenToken(child);
+        }
+        if ((child = this.getCloseParenToken(node)) !== undefined) {
+            this.checkCloseParenToken(child);
+        }
+    }
+
+    private checkOpenParenToken(node: ts.Node) {
+        const code = this.sourceFile.text.substr(node.getFullStart());
+        const matchResult = this.openingParenRegex.exec(code);
```
ajafff: the regex solution seems a bit hacky and is not very easy to understand.
I'd prefer the following;
```ts
// pretending `node` is the OpenParenToken
let pos = node.end;
let ch = this.sourceFile.text.charCodeAt(pos);
while(ts.isWhiteSpaceSingleLine(ch)) {
    ++pos;
    ch = this.sourceFile.text.charCodeAt(pos);
}
const whitespaceCount = pos - node.end;
if (this.options.size === 0) {
    if (whitespaceCount !== 0)
        // TODO addFailure
} else if (!ts.isLineBreak(ch) /* don't require whitespace when followed by line break */ && whitespaceCount !== this.options.size) {
    // TODO fail
}
```

For the CloseParentToken you need a similar loop from `node.pos - 1` backwards.

src/rules/spaceWithinParensRule.ts
aminpaks: `forEachToken` doesn't exist in `tsutils v2.4.0`.
I tried `import { forEachToken } from 'tsutils';` no luck.

ajafff: are you sure? I just tried it (v2.4.0) and it works for me.
Maybe it's another error, e.g. unused variable.

aminpaks: Ah! it's `utils' not `tsutils'.

aminpaks: `forEachToken` has this signature `(fullText: string, kind: ts.SyntaxKind, pos: TokenPosition)` I'm using not using `fullText` and Typescript is complaining. What should I do?

src/rules/spaceWithinParensRule.ts
aminpaks: With your algorithm how should I know that if this is not the case below? Both open and close paren tokens will add a failure.
```js
console.log( );
```

src/rules/spaceWithinParensRule.ts
ajafff: Well, I really meant `import {forEachToken} from "tsutils"`. But the one in `../utils` is also ok.

Re unused parameter: either use `fullText instead of `this.sourceFile.text` or rename the parameter to `_fullText` (not the leading underscore)

src/rules/spaceWithinParensRule.ts
ajafff: in my sample implementation there was a check if `ch` is a line break to avoid failures.

When checking the CloseBrace, you could add a similar check if `ch === 40` to not fail if you reached the OpenBraceToken.

src/rules/spaceWithinParensRule.ts
aminpaks: Coming from your last comment on how to find out we are facing this case `console.log()`:
We are inside of a loop (Note: I updated the code based on `tsutils`):
```ts

    private checkNode(node: ts.Node): void {
        forEachToken(node, (token: ts.Node) => {
            if (token.kind === ts.SyntaxKind.OpenParenToken) {
                this.checkOpenParenToken(token);
            } else if (token.kind === ts.SyntaxKind.CloseParenToken) {
                this.checkCloseParenToken(token);
            }
        });
    }
```
how I'm supposed to know if I already passed the `checkOpenParenToken` to cancel the next call which is `checkCloseParentToken` of this case?

aminpaks:
I forced push and applied all your comments/notes.
Please review.

src/rules/spaceWithinParensRule.ts
aminpaks: nice!

src/rules/spaceWithinParensRule.ts
```diff
@@ -0,0 +1,164 @@
...
+
+class SpaceWhitinParensWalker extends Lint.AbstractWalker<Options> {
+    public walk(sourceFile: ts.SourceFile) {
+        return ts.forEachChild(sourceFile, (node: ts.Node): void => {
+            this.checkNode(node);
+        });
+    }
+
+    private checkNode(node: ts.Node): void {
+        forEachToken(node, (token: ts.Node) => {
```
ajafff: I actually meant to replace the `ts.forEachChild(...)` in `walk` with this call.
You can just take this method's body and put it into `walk`.

With the current implementation the rule will visit nested tokens multiple times.

src/rules/spaceWithinParensRule.ts
```diff
@@ -0,0 +1,164 @@
...
+        optionsDescription: Lint.Utils.dedent`
+            You may enforce the amount of whitespace within parentheses.
+        `,
+        options: { type: "number", min: 0 },
+        type: "style",
+        typescriptOnly: false,
+    };
+    /* tslint:enable:object-literal-sort-keys */
+
+    public static FAILURE_NO_SPACE = "No whitespace within parentheses";
```
ajafff: Can you rephrase to `Whitespace within parentheses is not allowed.` or similar?

src/rules/spaceWithinParensRule.ts
```diff
@@ -0,0 +1,164 @@
...
+    public static FAILURE_NEEDS_SPACE = "Needs whitespace within parentheses";
```
ajafff: `Needs x spaces within parentheses.`

src/rules/spaceWithinParensRule.ts
```diff
@@ -0,0 +1,164 @@
...
+    private checkOpenParenToken(tokenNode: ts.Node) {
+        let currentPos = tokenNode.end;
+        let currentChar = this.sourceFile.text.charCodeAt(currentPos);
+        const allowedSpaceCount = this.options.size;
+
+        while (ts.isWhiteSpaceSingleLine(currentChar)) {
+            ++currentPos;
+            currentChar = this.sourceFile.text.charCodeAt(currentPos);
+        }
+        if (!ts.isLineBreak(currentChar)) {
```
ajafff: if the rule is configured with 0 spaces, this condition will not show a warning for trailing whitespace right before a line break.
But that will be checked by `trailing-whitespace` anyway, so it's not strictly necessary here.

src/rules/spaceWithinParensRule.ts
```diff
@@ -0,0 +1,164 @@
...
+                } else if (whitespaceCount > 0 && whitespaceCount < allowedSpaceCount) {
+                    pos += allowedSpaceCount - whitespaceCount;
+                }
+                this.AddFailureAtWithFix(pos, length, whitespaceCount);
+            }
+        }
+    }
+
+    private checkCloseParenToken(tokenNode: ts.Node) {
+        let currentPos = tokenNode.getStart() - 1;
```
ajafff: Can be simplified to:
```ts
let currentPos = tokenNode.end - 2; // close paren token has a known width of 1, therefore node.getStart() === node.end - 1
```

Otherwise I would suggest to pass `sourceFile` as parameter to `getStart` for performance reasons.

src/rules/spaceWithinParensRule.ts
```diff
@@ -0,0 +1,164 @@
...
+        let lintMsg: string | undefined;
+        let lintFix: Lint.Replacement | undefined;
+        const allowedSpaceCount = this.options.size;
+
+        if (allowedSpaceCount === 0) {
+            lintMsg = Rule.FAILURE_NO_SPACE;
+            lintFix = Lint.Replacement.deleteText(position, length);
+        } else if (whitespaceCount < allowedSpaceCount) {
+            lintMsg = Rule.FAILURE_NEEDS_SPACE;
+            let whitespace = "";
```
ajafff:
```ts
const whitespace = " ".repeat(allowedSpaceCount - whitespaceCount);
```

test/rules/space-within-parens/force-one-space/test.ts.lint
```diff
@@ -0,0 +1,14 @@
...
+                        ~nil [0]
+                                                          ~nil [0]
+    var xyz = Conversion.rgb2xyz(r, g, b  );
+                                 ~nil [0]
+                                        ~ [1]
+    return Conversion.xyz2lab(   xyz.x, xyz.y, xyz.z);
+                               ~~ [1]
+                                                    ~nil [0]
+}
+
```
ajafff: please add a test with comments
```ts
foo(/*no param*/);
    ~nil [0]
                ~nil [0]
```

src/rules/spaceWithinParensRule.ts
```diff
@@ -0,0 +1,164 @@
...
+        if (!Number.isNaN(parsedSize) && parsedSize >= 0) {
+            size = parsedSize;
+        }
+    }
+    return {
+        size,
+    };
+}
+
+class SpaceWhitinParensWalker extends Lint.AbstractWalker<Options> {
```
ajafff: s/Whitin/Within/

src/rules/spaceWithinParensRule.ts
```diff
@@ -0,0 +1,164 @@
...
+    /* tslint:enable:object-literal-sort-keys */
+
+    public static FAILURE_NO_SPACE = "No whitespace within parentheses";
+    public static FAILURE_NEEDS_SPACE = "Needs whitespace within parentheses";
+    public static FAILURE_NO_EXTRA_SPACE(count: number): string {
+        return `No more than ${count} whitespace within parentheses allowed`;
+    }
+
+    public apply(sourceFile: ts.SourceFile): Lint.RuleFailure[] {
+        return this.applyWithWalker(new SpaceWhitinParensWalker(sourceFile, this.ruleName, parseOptions(Number(this.ruleArguments[0]))));
```
ajafff: remove the `Number(...)`, that's done by `parseOptions` anyway

src/rules/spaceWithinParensRule.ts
```diff
@@ -0,0 +1,164 @@
...
+    }
+}
+
+function parseOptions(whitespaceSize?: number): Options {
```
ajafff: change to `whitespaceSize?: any`

src/rules/spaceWithinParensRule.ts
```diff
@@ -0,0 +1,164 @@
...
+        while (ts.isWhiteSpaceSingleLine(currentChar)) {
+            --currentPos;
+            currentChar = this.sourceFile.text.charCodeAt(currentPos);
+        }
+        /**
+         * Number 40 is open parenthese char code, we skip this cause
+         * it's already been caught by `checkOpenParenToken`
+         */
+        if (!ts.isLineBreak(currentChar) && currentChar !== 40) {
+            const whitespaceCount = tokenNode.end - tokenNode.pos - 1;
```
ajafff:
```ts
const whitespaceCount = token.end - currentPos - 2;
```

src/rules/spaceWithinParensRule.ts
```diff
@@ -0,0 +1,164 @@
...
+                }
+                this.AddFailureAtWithFix(pos, length, whitespaceCount);
+            }
+        }
+    }
+
+    private checkCloseParenToken(tokenNode: ts.Node) {
+        let currentPos = tokenNode.getStart() - 1;
+        let currentChar = this.sourceFile.text.charCodeAt(currentPos);
+        const allowdSpaceCount = this.options.size;
```
ajafff: allow**e**dSpaceCount

src/rules/spaceWithinParensRule.ts
aminpaks: was getting `Unsafe use of expression of type 'any'.`

test/rules/space-within-parens/force-one-space/test.ts.lint
aminpaks: Good catch

src/rules/spaceWithinParensRule.ts
aminpaks: I though about it too.

aminpaks:
I pushed a commit. Please review :)

aminpaks:
Done

ajafff:
Merging this PR, because it's open for quite some time, code looks good and there was no objection.

Thanks @aminpaks!

Issue #2961, opened 2017-06-27
Expected curly rule schema in IDE different from what is documented


omganunicorn:
### Bug Report

- 5.4.3
- 2.3.4
- Webstorm 2017.1.4

#### TypeScript code being linted

The `tslint.json` configuration itself is where the warning happens:

```json
{
    "defaultSeverity": "error",
    "extends": [
        "tslint:recommended"
    ],
    "jsRules": {},
    "rules": {
        "array-type": [true, "generic"],
        "ban-types": [true, ["Object", "Use {} instead."]],
        "curly": [true, "ignore-same-line"],
        "interface-name": [true, "never-prefix"],
        "no-namespace": [true, "allow-declarations"],
        "no-shadowed-variable": false,
        "quotemark": [true, "single", "jsx-single"]

    },
    "rulesDirectory": []
}
```

#### Actual behavior
`"curly": [true, "ignore-same-line"]` works when linting and single-line if statements do not show an error, however, Webstorm shows the "Type is not allowed" warning for the setting when editing `tslint.json`.

#### Expected behavior
The curly rule should accept the given data type correctly like the one documented in https://palantir.github.io/tslint/rules/curly/ for its options.

ajafff:
There's nothing tslint can do to fix this.
This is an issue with Webstorm and Webstorm *probably* uses the schema from http://json.schemastore.org/tslint which is maintained externally.
If you really need it, you can contribute that change to https://github.com/schemastore/schemastore/

omganunicorn:
So I looked into the issue a bit more this morning because I also thought it might be Webstorm (hence why I edited the title), but I was able to reproduce the issue in VS Code, with only [vscode-tslint](https://github.com/Microsoft/vscode-tslint) enabled. In Webstorm I am using the tslint installation from the node_modules folder and not the global or IDE built-in one.

I was going to close this as a non-issue since the error is being given when linting a JSON and not a .ts file, however I also looked at the node_modules/tslint/rules folder, and noticed that even thought the rule works (I see no problem with `/rules/curlyRule.d.ts` or `/rules/curlyRule.js`), curly _is_ a boolean in `recommended.d.ts`:

```
export declare const jsRules: {
    "align": {
        options: string[];
    };
    "class-name": boolean;
    "curly": boolean;
...
}
```

And in recommended.js notice how align has the proper strings `option` array but curly is a boolean, again, contrary to the documentation:

```
exports.jsRules = {
    "align": {
        options: [
            "parameters",
[...]

PR #2964, opened 2017-06-27
Updated curly options in configs

omganunicorn:
#### PR checklist

- [x] Addresses an existing issue: #2961 
  - [x] Includes tests
- [ ] Documentation update

#### Overview of change:
Updated the configuration files with the correct options array for the `curly` rule to match the documentation and the actual rule's options.

### DIFF
--- a/src/configs/all.ts
+++ b/src/configs/all.ts
@@ -88,7 +88,11 @@ export const rules = {
     // Functionality
     "await-promise": true,
     // "ban": no sensible default
-    "curly": true,
+    "curly": [
+        true,
+        "as-needed",
+        "ignore-same-line",
+    ],
     "forin": true,
     // "import-blacklist": no sensible default
     "label-position": true,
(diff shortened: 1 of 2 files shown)


src/configs/all.ts
```diff
@@ -88,7 +88,11 @@ export const rules = {
     // Functionality
     "await-promise": true,
     // "ban": no sensible default
-    "curly": true,
+    "curly": [
+        true,
+        "as-needed",
+        "ignore-same-line",
+    ],
```
ajafff: This change inverts how the rule checks files for all users of those presets. This is a breaking change and is probably not what you intended.

omganunicorn:
You're right, I kept digging a bit deeper in the code, and as I added rules for linting another project, I started seeing the same issue for different rules, so I checked out configuration.ts and other files and see what you mean. This fixed my issue although the rule still worked fine, but I'll close this and look at the vscode/Schemastore.org variant specifically; my apologies and thanks for the clarification.

BehindTheMath:
Looks like this was fixed with the following PR: https://github.com/SchemaStore/schemastore/pull/318

Issue #2979, opened 2017-07-01
Rule request: no-jsdoc-types


JoshuaKGoldberg:
```typescript
/**
 * Does a thing.
 * 
 * @param {Function} action   Action to do.
 */
function do(action: () => void) {
  action();
}
```

There's no need for the `{Function}`. In code bases converted from JavaScript with filled out JSDoc annotations, it's common to have these unnecessary (and often vague or incorrect) types left over.

adidahiya:
looks like this is covered by `no-redundant-jsdoc`

PR #2980, opened 2017-07-01
Added a no-jsdoc-types rule

JoshuaKGoldberg:
#### PR checklist

- [x] Addresses an existing issue: #2979
  - [x] Includes tests
- [x] Documentation update

#### Overview of change:

Adds a `no-jsdoc-types` rule to prevent unnecessary JSDoc type annotations.

#### CHANGELOG.md entry:

[new-rule] `no-jsdoc-types`

### DIFF
--- a/src/configs/all.ts
+++ b/src/configs/all.ts
@@ -190,6 +190,7 @@ export const rules = {
     "no-angle-bracket-type-assertion": true,
     "no-boolean-literal-compare": true,
     "no-consecutive-blank-lines": true,
+    "no-jsdoc-types": true,
     "no-parameter-properties": true,
     "no-reference-import": true,
     "no-unnecessary-callback-wrapper": true,
--- a/src/rules/noJsdocTypesRule.ts
+++ b/src/rules/noJsdocTypesRule.ts
@@ -1,0 +1,67 @@
+/**
+ * @license
+ * Copyright 2013 Palantir Technologies, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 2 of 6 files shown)


src/configs/all.ts
```diff
@@ -190,6 +190,7 @@ export const rules = {
     "no-angle-bracket-type-assertion": true,
     "no-boolean-literal-compare": true,
     "no-consecutive-blank-lines": true,
+    "no-jsdoc-types": true,
```
ajafff: also add to `latest.ts`

src/rules/noJsdocTypesRule.ts
```diff
@@ -0,0 +1,66 @@
...
+    public apply(sourceFile: ts.SourceFile): Lint.RuleFailure[] {
+        return this.applyWithFunction(sourceFile, walk);
+    }
+}
+
+const violationSearcher = /\@(param|return|returns)( *)\{.*\}/g;
+const fixReplacer = /( *)\{.*\}/;
+
+function walk(ctx: Lint.WalkContext<void>) {
+    return utils.forEachComment(ctx.sourceFile, (fullText, { kind, pos }) => {
```
ajafff: instead of iterating over all comments to find and "parse" JSDoc here, you could simply use what typescript offers.
Take a look at the implementation of `forEachToken` in tsutils: https://github.com/ajafff/tsutils/blob/master/util/util.ts#L358

You can add a similar implementation here but instead of calling the callback for tokens, you call it for every node of type `ts.JSDoc`. This node already contains the parsed tags.

test/rules/no-jsdoc-types/default.ts.fix
```diff
@@ -0,0 +1,20 @@
+/**
+ * Performs an action a number of times.
+ *
+ * @param action
+ * @param action   Does a thing.
+ * @param {function} action
```
ajafff: does the fixer even fix?

JoshuaKGoldberg:
Ah, forgot about the other PR and neglected to scan through the PRs... Thanks for the review anyway.

Issue #2795, opened 2017-05-20
Trailing comma incorrectly identified inside multiline arrow function


borekb:
### Bug Report

- __TSLint version__: 5.2.0
- __TypeScript version__: 2.3
- __Running TSLint via__: CLI

#### TypeScript code being linted

```ts
commitsTableStore.setCommitRows(data.commits.map(commit =>
  new CommitRow(commit) // missing trailing comma reported here
));
```

with `tslint.json` configuration:

```json
"trailing-comma": [true, {"multiline": "always"}]
```

#### Actual behavior

Missing trailing comma reported.

#### Expected behavior

No linting error reported.

#### More info

It's caused by the "functions" checker – TSLint thinks that this is a multiline function call.

adidahiya:
huh, you're right, that is a false positive. accepting bugfix PRs

ajafff:
Cross posting my comment from https://github.com/palantir/tslint/pull/2988#issuecomment-312840722:

IMO this is not a bug or false positive. 

The above example is considered "multiline" for a reason. The docs are very clear on that:

> An array is considered “multiline” if its closing bracket is on a line after the last array element. The same general logic is followed for object literals, function typings, named import statements and function parameters.

To avoid this, you can simply rewrite it as follows:

```ts
commitsTableStore.setCommitRows(data.commits.map(commit =>
  new CommitRow(commit)));

//or
commitsTableStore.setCommitRows(data.commits.map((commit =>
  new CommitRow(commit)
))); //  note the extra parens around the arrow function
```

If we want to implement an exception for such cases, this would need a proposal what exacltly should be handled different. And it should be implemented behind an option.

adidahiya:
ah, I see your point @ajafff, and I now agree, this is not a bug or false positive. as per https://github.com/palantir/tslint/issues/2909#issuecomment-307898234, we don't have plans to to check the arity of the function being called in the trailing-comma implementation. your proposed workarounds are good 👍

borekb:
Thanks for all the context. I have to say it's quite unintuitive at first but follows some internal logic. And I didn't know you can put trailing commas in function calls just like that :)

For reference, this is how Prettier handles the situation with `--trailing-comma=es5` configuration: https://goo.gl/Juay94

```
commitsTableStore.setCommitRows(
  data.commits.map(
    commit => new CommitRow(commit) // missing trailing comma reported here
  )
);
```

PR #2988, opened 2017-07-03
Fix 'trailing-comma' wrapped expression treated as multiline

ksvitkovsky:
#### PR checklist

- [x] Addresses an existing issue: #2795 
  - [x] Includes tests
- [ ] Documentation update

#### Preface

As stated in the issue, sometimes we want multiline expressions to be treated as single-line.
That applies to long single-line expressions that we want to break for the sake of readability.

Some examples of such cases:
```ts
const foo = array.map(arrayItem =>
    new ObjectWithPrettyDamnLongConstructorName(arrayItem)
);

const foo = { bar: isStatementTrue
    ? getSomeValueForBarPropertyIfStatementIsTrue
    : getSomeValueForBarPropertyIfStatementIsFalse
};
```

Linting the code above would cause 'missing trailing comma' errors in both cases.

#### Overview of change:

Added some code to check if an expression is on the same line with a previous open token, and if so - treat it as a single-line expression.

#### CHANGELOG.md entry:

[bugfix] `trailing-comma`: treat wrapped expressions as single-line if they are on the same line

### DIFF
--- a/src/rules/trailingCommaRule.ts
+++ b/src/rules/trailingCommaRule.ts
@@ -219,7 +219,7 @@ class TrailingCommaWalker extends Lint.AbstractWalker<Options> {
 
     /* Expects `list.length !== 0` */
     private checkComma(hasTrailingComma: boolean | undefined, list: ts.NodeArray<ts.Node>, closeTokenPos: number, optionKey: OptionName) {
-        const options = isSameLine(this.sourceFile, list[list.length - 1].end, closeTokenPos)
+        const options = isSameLine(this.sourceFile, list[list.length - 1].end, closeTokenPos) || this.isWrappedSingleExpression(list)
             ? this.options.singleline
             : this.options.multiline;
         const option = options[optionKey];
@@ -229,5 +229,17 @@ class TrailingCommaWalker extends Lint.AbstractWalker<Options> {
         } else if (option === "never" && hasTrailingComma) {
             this.addFailureAt(list.end - 1, 1, Rule.FAILURE_STRING_NEVER, Lint.Replacement.deleteText(list.end - 1, 1));
         }
+    }
+
+    private isWrappedSingleExpression(list: ts.NodeArray<ts.Node>) {
+        if (list.length !== 1) {
+            return false;
+        }
+
+        const parentNodeChildren = list[0].parent!.getChildren();
+        const openTokenIndex = parentNodeChildren.findIndex((node) => node.pos === list.pos) - 1;
+        const openToken = parentNodeChildren[openTokenIndex];
+
+        return isSameLine(this.sourceFile, openToken.end, list[0].getStart());
     }
 }
--- a/test/rules/trailing-comma/multiline-always/test.ts.fix
+++ b/test/rules/trailing-comma/multiline-always/test.ts.fix
@@ -547,3 +547,11 @@ export {
 
 // don't crash
 new Foo
+
+const foo = array.map(item =>
+    mapArrayItem(item)
+);
+
+const foo = array.map(
+    item => mapArrayItem(item),
+);
--- a/test/rules/trailing-comma/multiline-always/test.ts.lint
+++ b/test/rules/trailing-comma/multiline-always/test.ts.lint
@@ -576,3 +576,12 @@ export {
 
 // don't crash
 new Foo
+
+const foo = array.map(item =>
+    mapArrayItem(item)
+);
+
+const foo = array.map(
+    item => mapArrayItem(item)
+                              ~nil [Missing trailing comma]
+);
--- a/test/rules/trailing-comma/multiline-custom/always/test.ts.lint
+++ b/test/rules/trailing-comma/multiline-custom/always/test.ts.lint
@@ -576,3 +576,12 @@ export {
 
 // don't crash
 new Foo
+
+const foo = array.map(item =>
+    mapArrayItem(item)
+);
+
+const foo = array.map(
+    item => mapArrayItem(item)
+                              ~nil [Missing trailing comma]
+);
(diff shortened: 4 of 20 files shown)


ajafff:
IMO this is not a bug or false positive. I don't know why the linked issue was labeled `Type: Bug` in the first place.

Your examples above are considered "multiline" for a reason. The docs are very clear on that: 
> An array is considered “multiline” if its closing bracket is on a line after the last array element. The same general logic is followed for object literals, function typings, named import statements and function parameters.


To avoid this, you can simply rewrite it as follows:
```ts
const foo = array.map(arrayItem =>
    new ObjectWithPrettyDamnLongConstructorName(arrayItem));

const foo = { bar: isStatementTrue
    ? getSomeValueForBarPropertyIfStatementIsTrue
    : getSomeValueForBarPropertyIfStatementIsFalse};
```


Besides that I would have some comments about the implementation, but I wanted to come to a conclusion if this is really a change we want.

ksvitkovsky:
Yeah, guess I should've asked if we really need this to be amended in the first place 😔
Anyway, could you give me those comments on the implementation?

ajafff:
> Yeah, guess I should've asked if we really need this to be amended in the first place 😔

Don't worry, it's not your fault. I'm thankful for every new contributor trying to improve tslint.

> Anyway, could you give me those comments on the implementation?

Will do later today

Until then, you can enable CircleCI builds for your fork of tslint. https://circleci.com
After that you need to set parallel builds to 4: https://circleci.com/gh/ksvitkovsky/tslint/edit#parallel-builds

src/rules/trailingCommaRule.ts
```diff
@@ -230,4 +230,16 @@ class TrailingCommaWalker extends Lint.AbstractWalker<Options> {
             this.addFailureAt(list.end - 1, 1, Rule.FAILURE_STRING_NEVER, Lint.Replacement.deleteText(list.end - 1, 1));
         }
     }
+
+    private isWrappedSingleExpression(list: ts.NodeArray<ts.Node>) {
+        if (list.length !== 1) {
+            return false;
+        }
+
+        const parentNodeChildren = list[0].parent!.getChildren();
```
ajafff: For performance reasons (although barely noticeable), you should always pass the `sourceFile` as parameter to `.getChildren()`, `getStart()` and other methods.

src/rules/trailingCommaRule.ts
```diff
@@ -230,4 +230,16 @@ class TrailingCommaWalker extends Lint.AbstractWalker<Options> {
...
+        const openTokenIndex = parentNodeChildren.findIndex((node) => node.pos === list.pos) - 1;
+        const openToken = parentNodeChildren[openTokenIndex];
```
ajafff: could be simplified to
```ts
const openToken = getPreviousToken(list[0], this.sourceFile);
// getPreviousToken needs to be imported from "tsutils"
```

src/rules/trailingCommaRule.ts
```diff
@@ -230,4 +230,16 @@ class TrailingCommaWalker extends Lint.AbstractWalker<Options> {
...
+
+        return isSameLine(this.sourceFile, openToken.end, list[0].getStart());
```
ajafff: when checking for "wrapped" expressions you also need to check if the expression ends on (or after) the next line.

Consider the following code which would pass with the current implementation:
```ts
foo(bar // trailing comma should be required here
);
```

src/rules/trailingCommaRule.ts
ajafff: you don't need `openToken` at all, because `list.pos === openToken.end`

src/rules/trailingCommaRule.ts
```diff
@@ -230,4 +230,16 @@ class TrailingCommaWalker extends Lint.AbstractWalker<Options> {
...
+    }
```
ajafff: Besides the comments above, here's how I would have written this method:
```ts
return list.length === 1 &&
    // check if the expression ends on a line following the opening token
    !isSameLine(this.sourceFile, list.pos, list[0].end) && // list[0].end is important here, because list.end could include the trailing comma
    // check if expressions starts in the same line as the opening token - you already had this
    isSameLine(this.sourceFile, list.pos, list[0].getStart(this.sourceFile);
```

src/rules/trailingCommaRule.ts
```diff
@@ -219,7 +219,7 @@ class TrailingCommaWalker extends Lint.AbstractWalker<Options> {
 
     /* Expects `list.length !== 0` */
     private checkComma(hasTrailingComma: boolean | undefined, list: ts.NodeArray<ts.Node>, closeTokenPos: number, optionKey: OptionName) {
-        const options = isSameLine(this.sourceFile, list[list.length - 1].end, closeTokenPos)
+        const options = isSameLine(this.sourceFile, list[list.length - 1].end, closeTokenPos) || this.isWrappedSingleExpression(list)
```
ajafff: If we really consider this a bug fix, that's probably ok. Otherwise this is a breaking change, because the rule now works exactly the other way around for such cases.
One possibility is to add an option to explicitly enable this new behavior.

ksvitkovsky:
@ajafff wow, thanks a lot for the feedback!

Personally, in such cases I'd like to leave close token on new line, just looks better to me. 
Though I do think that handling it explicitly is an overkill.

Drop me a comment if I need to refactor it and whether it should be enabled with an option.

ajafff:
@ksvitkovsky I'm not the one responsible for decisions in this repository. I'd like to hear what others think about this. That discussion should take place in the original issue.


There is still one thing to do. I already wrote this above. Maybe you have skipped over it:

> enable CircleCI builds for your fork of tslint. https://circleci.com
After that you need to set parallel builds to 4: https://circleci.com/gh/ksvitkovsky/tslint/edit#parallel-builds

adidahiya:
see #2795, I mistakenly labeled the issue as a bug.

PR #3006, opened 2017-07-06
Improved "complete docs" docs

304NotModified:
#### PR checklist

s
- [X] Documentation update

#### Overview of change:
- It was unclear to me if it works for JSDocs (because in another rule "JSDoc comment" is used instead of "Documentation"

- Updated the style for the inline code

### DIFF
--- a/rules/completed-docs/index.html
+++ b/rules/completed-docs/index.html
@@ -1,9 +1,9 @@
 ---
 ruleName: completed-docs
-description: Enforces documentation for important items be filled out.
+description: Enforces documentation and JSDoc comments for important items be filled out.
 optionsDescription: |-
 
-  `true` to enable for ["classes", "functions", "methods", "properties"],
+  `true` to enable for `["classes", "functions", "methods", "properties"]`,
   or an array with each item in one of two formats:
 
   * `string` to enable for that type


rules/completed-docs/index.html
```diff
@@ -1,9 +1,9 @@
 ---
 ruleName: completed-docs
-description: Enforces documentation for important items be filled out.
+description: Enforces documentation and JSDoc comments for important items be filled out.
```
ajafff: `documentation and JSDoc comments` seems redundant.
Maybe just `Enforces JSDoc for important ...`

304NotModified: but it also works for non-JS docs? (double slash)

What about:

> Enforces (JSDoc) comments for important items be filled out.

ajafff: @304NotModified Sorry for the delay

> but it also works for non-JS docs? (double slash)

After looking at the code and the tests for this rule, I'm pretty sure it only checks JSDoc.

That said, I'd prefer `Enforces JSDoc comments for important items be filled out.`
And as stated above, the changes should be done in the rule metadata in the source file instead.

304NotModified:
and created a new PR: 3823

Issue #3016, opened 2017-07-07
no-unbound-method: referencing methods that are bound in constructor should be ok


unstubbable:
### Bug Report

In https://github.com/palantir/tslint-react/issues/63 it was mentioned that a rule such as `no-unbound-method` should check for methods that are bound in the constructor (e.g. `this.handleClick = this.handleClick.bind(this);`) to avoid false positives. Although https://github.com/palantir/tslint-react/issues/63 was closed with the [introduction of `no-unbound-method`](#2089) this case does not seem to be implemented.

- __TSLint version__: 5.5.0
- __TypeScript version__: 2.4.1
- __Running TSLint via__: CLI

#### TypeScript code being linted

```ts
class C {
  public constructor() {
    this.method = this.method.bind(this);
  }
  public method(): number {
    return 1;
  }
}

const c = new C();
[0].forEach(c.method);
```

with `tslint.json` configuration:

```json
{
  "no-unbound-method": true
}
```

#### Actual behavior

⚠️ Fails with:

```
[11, 13]: Avoid referencing unbound methods which may cause unintentional scoping of 'this'.
```

#### Expected behavior

✅  Has no errors.

adidahiya:
I guess your suggestion is technically valid, but I'm not sure if the the extra complexity in this rule is worth it -- since this is TS, couldn't you just bind your methods using arrow functions?

KingHenne:
I guess arrow functions as class properties would be fine for most cases in the React world (referring to https://github.com/palantir/tslint-react/issues/63). But they do have the disadvantage that you can not call super in them, i.e. something like this is not possible without the `bind` in the constructor as far as I know:

```ts
class A {
  count: number;

  constructor() {
    this.count = 0;
    this.method = this.method.bind(this);
  }

  method() {
    this.count++;
  }
}

class B extends A {
  method() {
    super.method();
    this.count++;
    console.log(this.count);
  }
}

const b = new B();
// This would be an event handler or something:
[0].forEach(b.method); // logs 2
```

KingHenne:
closing, see https://github.com/palantir/tslint/pull/3017#issuecomment-314368510

Amaimersion:
+1 for re-open. It would be very good if someone will implement checking for binding in the constructor. There was a try (#3017), but it has failed. `no-unbound-method` rule is really a pain for React, because recommend way is to bind in the constructor, not to create arrow lambda.

JoshuaKGoldberg:
[...]

PR #3017, opened 2017-07-07
no-unbound-method: allow methods that are bound in the constructor

KingHenne:
#### PR checklist

- [x] Addresses an existing issue: #3016 
- [x] ~~New feature~~, bugfix, ~~or enhancement~~
  - [x] Includes tests
- [x] ~~Documentation update~~

#### Overview of change:

With this change no failure for `no-unbound-method` is added if the method is bound in the constructor.

I’m not that familiar with the TypeScript AST so I’m not sure whether this pretty naïve implementation could be simplified or clarified.

1. Is the approach the right way to go?
2. I tried my best to keep it as readable as possible but I still find the result rather ugly. How could this be improved?

### DIFF
--- a/test/rules/no-unbound-method/default/test.ts.lint
+++ b/test/rules/no-unbound-method/default/test.ts.lint
@@ -1,3 +1,18 @@
+class C {
+    constructor() {
+        this.foo = 1;
+        this.methodA = this.methodA.bind(this);
+    }
+    methodA() {}
+    methodB() {}
+}
+
+const c = new C();
+// OK if bound in constructor
+[0].forEach(c.methodA);
+[0].forEach(c.methodB);
+            ~~~~~~~~~ [0]
+
 interface I { m?(): void; }
 function f(i: I) {
     i.m!();
--- a/src/rules/noUnboundMethodRule.ts
+++ b/src/rules/noUnboundMethodRule.ts
@@ -115,27 +115,15 @@ function isSafeUse(node: ts.Node): boolean {
     }
 }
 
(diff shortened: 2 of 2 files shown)


palantirtech:
Thanks for your interest in [palantir/tslint](https://github.com/palantir/tslint), @KingHenne! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

test/rules/no-unbound-method/default/test.ts.lint
```diff
@@ -1,3 +1,18 @@
...
+    methodA() {}
+    methodB() {}
+}
+
+const c = new C();
+// OK if bound in constructor
+[0].forEach(c.methodA);
+[0].forEach(c.methodB);
+            ~~~~~~~~~ [0]
+
```
ajafff: There should be another test:
```ts
class D extends C {
    methodA() {}
    methodB() {}
}
const d = new D();
[0].forEach(d.methodA);
[0].forEach(d.methodB);
            ~~~~~~~~~[0]
```

I think this will not work with the current implementation.

KingHenne: Yes, you are absolutely right, this isn’t handled and also defeats [my argument](https://github.com/palantir/tslint/issues/3016#issuecomment-313842057). 😄

I will close the PR and issue then and need to either disable this rule or convert my code base to use arrow functions instead of `bind`. 😞

KingHenne:
Closing this based on feedback (see above). It was still interesting to get to know the tslint code base a bit. 😃

PR #3029, opened 2017-07-11
Update tslint and tsutils dependencies

ajafff:
#### PR checklist

  - [ ] Includes tests
- [ ] Documentation update

#### Overview of change:
Update tslint dependency

#### CHANGELOG.md entry:

[no-log]

### DIFF
--- a/package.json
+++ b/package.json
@@ -45,7 +45,7 @@
     "resolve": "^1.3.2",
     "semver": "^5.3.0",
     "tslib": "^1.7.1",
-    "tsutils": "^2.6.0"
+    "tsutils": "^2.7.1"
   },
   "peerDependencies": {
     "typescript": ">=2.1.0 || >=2.1.0-dev || >=2.2.0-dev || >=2.3.0-dev || >=2.4.0-dev || >=2.5.0-dev || >=2.6.0-dev"
--- a/package.json
+++ b/package.json
@@ -45,7 +45,7 @@
     "resolve": "^1.3.2",
     "semver": "^5.3.0",
     "tslib": "^1.7.1",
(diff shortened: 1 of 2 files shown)


yarn.lock
```diff
@@ -1486,18 +1486,20 @@ tslint@^5.4.2:
...
     semver "^5.3.0"
     tslib "^1.7.1"
-    tsutils "^2.3.0"
-
-tsutils@^2.3.0:
-  version "2.4.0"
-  resolved "https://registry.yarnpkg.com/tsutils/-/tsutils-2.4.0.tgz#ad4ce6dba0e5a3edbddf8626b7ca040782189fea"
+    tsutils "^2.5.1"
 
 tsutils@^2.5.1:
```
JKillian: we could merge this block from the lockfile with the block below for `2.6.0` satisfies the requirement for both

JKillian: (I wish Yarn would do this automatically!)

package.json
```diff
@@ -71,9 +71,9 @@
     "npm-run-all": "^4.0.2",
     "nyc": "^10.2.0",
     "rimraf": "^2.5.4",
-    "tslint": "^5.4.2",
+    "tslint": "^5.5.0",
     "tslint-test-config-non-relative": "file:test/external/tslint-test-config-non-relative",
-    "typescript": "2.4.1"
+    "typescript": "^2.4.1"
```
adidahiya: This is probably not a great idea; we should use `~2.4.1` since typescript has breaking changes across minor versions. `~` forces us to be explicit about upgrades, which is good.

ajafff: You're right. I fixed it

PR #3044, opened 2017-07-14
prefer-for-of: identify more array modifications

ajafff:
#### PR checklist

  - [x] Includes tests
- [ ] Documentation update

#### Overview of change:
[bugfix] `prefer-for-of` fixed false positives when array is modified, e.g. `arr[i]++`

### DIFF
--- a/test/rules/prefer-for-of/test.ts.lint
+++ b/test/rules/prefer-for-of/test.ts.lint
@@ -178,4 +178,20 @@ function test() {
         print();
 }
 
+for (let i = 0; i < arr.length; ++i) {
+    arr[i]++;
+}
+
+for (let i = 0; i < arr.length; ++i) {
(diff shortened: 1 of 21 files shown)


test/rules/prefer-for-of/test.ts.lint
```diff
@@ -178,4 +178,20 @@ function test() {
...
+for (let i = 0; i < arr.length; ++i) {
+    arr[i]++;
+}
+
+for (let i = 0; i < arr.length; ++i) {
+    ({...arr[i]} = foo);
+}
+
+for (let i = 0; i < arr.length; ++i) {
+    for (arr[i] of otherArr) {}
```
JKillian: Didn't even realize that this would be valid - I hope nobody ever does this in real life :P

Issue #3033, opened 2017-07-12
Rule suggestion: disallow modifying function parameters


benny-medflyt:
There is an argument that using `const` instead of `let` makes code more readable, since you can be sure at a glance that a variable will not change it's value. (See [prefer-const](https://palantir.github.io/tslint/rules/prefer-const/))

Unfortunately, in TypeScript (or JavaScript) there is no way to declare that a particular function parameter should be `const` (they are all implicitly `let`).

For example, the following function modifies the `x` parameter:

```typescript
function add(x: number, y: number): number {
    x += y;
    return x;
}
```

I imagine that a lot of projects would be happy to completely ban all such usages in their code base: all function parameters would behave as if they were `const`. I believe that this can be implemented as a tslint rule, and I suggest that such a rule be added =]

ajafff:
IMO such a rule makes sense. The implementation should not be too different from `prefer-const`. Maybe we can extract some of the logic into a reusable class or function.
Until it is implemented in tslint core, you can use a third party rule: https://github.com/vrsource/vrsource-tslint-rules#no-param-reassign

Just one question if we want to treat destructured parameters as real parameters? Would be best to make it configurable, I guess.
```ts
function foo(bar: number) {} // this is a parameter
function foo([bar, baz]: number[]) {} // technically these are local variables ... do we want to treat them as parameter?
```

PR #3045, opened 2017-07-14
new rule no-parameter-reassignment

ajafff:
#### PR checklist

- [x] Addresses an existing issue: #3033
  - [x] Includes tests
- [x] Documentation update

#### Overview of change:
[new-rule] `no-parameter-reassignment`
Fixes: #3033
/cc @benny-medflyt


benny-medflyt:
Looks awesome :1st_place_medal: 

I can't quite tell from the code... what did you end up deciding regarding destructured parameters?
I personally think that when this rule is enabled, they should also be affected and prevented from being re-assigned

ajafff:
@benny-medflyt I decided to also check destructured parameters. I don't think it's necessary to add an option to disable that behavior until there is demand for it.

ajafff:
Renamed as suggested and fixed merge conflicts

zheeeng:
I think it is acceptable reassigning params which its type is number, string, boolean. It helps to write some algorithms easy.

Issue #2985, opened 2017-07-02
Enable collaborators to manage issues and pull requests


ajafff:
Recently this repository is lacking active mainainers.
Issues pile up, Pull Requests go stale and authors lose interest, important bug fixes don't get released in a reasonable amount of time because it's simply too much for one person to handle.

This is not to blame anyone. I know you don't do this as full time job and have other stuff to do. It also seems like @adidahiya is the only maintainer recently.
But hey, this is open source. There are some pretty active and talented contributors that would be willing to help.

Get collaborators involved in triaging issues, reviewing and merging pull requests. That works great for other projects like node or the whole levelup ecosystem. IMO this will increase productivity:
* decreased number of open issues makes it easier to find bugs to fix
* ability to assign someone to issues avoids unnecessary work of others
* earlier feedback on PRs keeps authors interested
* more eyes to find bugs
* faster merging of bug fixes
* increased activity attracts more contributors
* (I can go on vacation without this project going completely inactive)

Of course there's always some risk in giving away commit access. But there are some things to consider here:
* Write a "code of conduct" for a standard way to manage this repo.
* With an increasing number of collaborators there are more eyes reviewing pull requests.
* It's git after all and everything can be reverted or even rebased out of the history.
* You retain the rights to publish a release.

Some interesting reads on the topic:
[Contributing guidelines of level projects (OPEN open source)](https://github.com/Level/community/blob/master/CONTRIBUTING.md)
[How to be an open source gardener](http://words.steveklabnik.com/how-to-be-an-open-source-gardener)

Btw. I'm not proposing this because I need the fame. I really want to get this project forward. In fact I'm frustrated by the status quo and start losing my enthusiams.

adidahiya:
I agree, I'd like to add more collaborators to this repo. I'll work on getting a code of conduct added here this week. Can you also send me an email? Address in my bio.

adidahiya:
> You retain the rights to publish a release.

I was poking around at this and I don't see a way to restrict git tagging (to publish a release) to certain users. Do you know of a way to do this?

ajafff:
> I don't see a way to restrict git tagging (to publish a release) to certain users. Do you know of a way to do this?

Initially I thought you just don't give away publish permissions for npm. But now I see that it's automatically published once the tag is added...
Unfortunately there seems to be no option to disable pushing tags to github.

An alternative might be to check CIRCLE_USERNAME in CI before actually publishing.

IllusionMH:
[...]

PR #3054, opened 2017-07-18
switch from CircleCI -> local release process

JKillian:
**Warning: When this PR is merged `NPM_TOKEN` should be deleted from the CircleCI environment variables.**

- [x] Related to https://github.com/palantir/tslint/issues/2985
- [x] Documentation update

#### Overview of change:
Switches from a GH/CircleCI based publishing flow to a local one. This allows us to add contributors to TSLint without giving them publishing/releasing permissions. 

Does the local publishing script seem reasonable? Does re-cloning the repo to a temporary directory seem reasonable? I tested the script myself, but obviously didn't publish an actual release.

Possible improvements to this PR:
* We could probably automate the creation of the git tag as well. For now, I decided to leave it as a manual step (since you have to paste the changelog into GH anyways).
* I only test the build locally with one version of TS. We could do what we do on CircleCI and make `publish:local` run the tests with multiple versions of TS.

### DIFF
--- a/README.md
+++ b/README.md
@@ -71,5 +71,6 @@ Creating a new release
 4. Commit with message `Prepare release <version>`
 5. Push your branch to GitHub and make a PR
 6. Once your PR is merged, wait for the tests to pass on CircleCI for develop
-7. Create a "Release" on GitHub with the proper tag version and notes from the changelog
-8. Run 'yarn run publish:local 4.1.0' (replace with the appropriate tag name for the release)
+7. Create a "Release" on GitHub with the proper tag version and notes from the changelog.
+   The tag should be identical to the version in `package.json`
+8. Run 'yarn run publish:local'
--- a/README.md
+++ b/README.md
@@ -68,7 +68,8 @@ Creating a new release
 
 1. Bump the version number in `package.json` and `src/linter.ts`
 2. Add release notes in `CHANGELOG.md`
-3. `yarn verify` to build the latest sources from a clean state
 4. Commit with message `Prepare release <version>`
-5. Run `npm publish`
-6. Create a git tag for the new release and push it ([see existing tags here](https://github.com/palantir/tslint/tags))
+5. Push your branch to GitHub and make a PR
+6. Once your PR is merged, wait for the tests to pass on CircleCI for develop
+7. Create a "Release" on GitHub with the proper tag version and notes from the changelog
+8. Run 'yarn run publish:local 4.1.0' (replace with the appropriate tag name for the release)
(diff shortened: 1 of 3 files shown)


IllusionMH:
In theory it is possible to use [Circle CI API](https://circleci.com/docs/api/v1-reference/#recent-builds-project) call to confirm that build for supposed release commit hash is successful.

For multiple versions tests I tried to use Docker, but on Windows there is problems on linking step during yarn install (no problems with npm). However Mac may not have this problem.

JKillian:
Both interesting ideas @IllusionMH! I'll leave those experiments for a future PR though, this one is primarily to unblock us so we can add collaborators. Publishing improvements will definitely be possible though 😁

README.md
```diff
@@ -68,7 +68,8 @@ Creating a new release
...
 1. Bump the version number in `package.json` and `src/linter.ts`
 2. Add release notes in `CHANGELOG.md`
-3. `yarn verify` to build the latest sources from a clean state
 4. Commit with message `Prepare release <version>`
-5. Run `npm publish`
-6. Create a git tag for the new release and push it ([see existing tags here](https://github.com/palantir/tslint/tags))
+5. Push your branch to GitHub and make a PR
+6. Once your PR is merged, wait for the tests to pass on CircleCI for develop
+7. Create a "Release" on GitHub with the proper tag version and notes from the changelog
+8. Run 'yarn run publish:local 4.1.0' (replace with the appropriate tag name for the release)
```
adidahiya: I thought you need `--` to forward arguments to the script?

scripts/npmPublish.sh
```diff
@@ -1,10 +1,33 @@
...
+# Publishes TSLint to NPM
+# This script must be run with yarn: "yarn run publish:local gitTagToPublish"
+# A user running this script must have Palantir NPM organization credentials
+
 set -e
 
-echo "//registry.npmjs.org/:_authToken=${NPM_TOKEN}" > .npmrc
-chmod 0600 .npmrc
+if [[ $# -ne 1 ]]; then
+    echo "usage: yarn run publish:local gitTagToPublish"
```
adidahiya: can't we grab this from `echo 'console.log("./package.json").version' | node` or something similar?

README.md
JKillian: I believe that's only needed if you're trying to pass dash options (like passing `-o` to your script or something like that)

scripts/npmPublish.sh
JKillian: hmm good point, I actually use `$npm_package_version` below

adidahiya:
I've deleted `$NPM_TOKEN` from CircleCI.

Issue #3001, opened 2017-07-06
TypeError: Cannot read property 'flags' of undefined (no-unused-variable)


danielweck:
"TypeError: Cannot read property 'flags' of undefined" with `no-unused-variable` rule.

TypeScript `2.4.1`
https://unpkg.com/typescript@2.4.1/

TSlint `5.5.0`
https://unpkg.com/tslint@5.5.0/

Running TSlint via CLI (`package.json` script with `tslint --type-check --project "./tsconfig.json" -c "./tslint.json" "./src/**/*.ts"`)

```
TypeError: Cannot read property 'flags' of undefined
    at Object.getCheckFlags (/PROJECT/node_modules/typescript/lib/typescript.js:10137:22)
    at getTypeOfSymbol (/PROJECT/node_modules/typescript/lib/typescript.js:30957:20)
    at getTypeOfNode (/PROJECT/node_modules/typescript/lib/typescript.js:47329:24)
    at Object.getTypeAtLocation (/PROJECT/node_modules/typescript/lib/typescript.js:26823:31)
    at getImplicitType (/PROJECT/node_modules/tslint/lib/rules/noUnusedVariableRule.js:269:24)
    at cb (/PROJECT/node_modules/tslint/lib/rules/noUnusedVariableRule.js:259:20)
    at visitEachNode (/PROJECT/node_modules/typescript/lib/typescript.js:15956:30)
    at Object.forEachChild (/PROJECT/node_modules/typescript/lib/typescript.js:16143:24)
    at cb (/PROJECT/node_modules/tslint/lib/rules/noUnusedVariableRule.js:264:19)
    at visitNode (/PROJECT/node_modules/typescript/lib/typescript.js:15944:20)
```

`tslint.json` configuration:
```json
{
  "extends": "tslint:recommended",
  "rules": {
    "no-floating-promises": true,
    "promise-function-async": true,
    "await-promise": true,
    "no-unused-variable": true,
    "no-unused-expression": true,
    "variable-name": [
      true,
      "ban-keywords",
      "check-format",
      "allow-leading-underscore",
      "allow-trailing-underscore",
      "allow-pascal-case"
    ],
    "ordered-imports": [
      true,
      {
        "import-sources-order": "case-insensitive",
        "named-imports-order": "lowercase-last"
      }
    ],
    "no-console": [
      false
    ],
    "interface-name": [
      false
    ]
  },
  "jsRules": {
    "ordered-imports": [
      true,
      {
[...]

Issue #2876, opened 2017-06-04
no-unused-variable error with object destructuring


cartant:
### Bug Report

- __TSLint version__: 5.4.2
- __TypeScript version__: 2.3.4
- __Running TSLint via__: CLI

#### TypeScript code being linted

To reproduce the problem, two files are required:

```ts
// main.ts
import { SomeClass } from "./some";
const person = { name: "alice" };
const { name } = person;
```

```ts
// some.ts
export class SomeClass {};
```

with `tslint.json` configuration:

```json
{
    "defaultSeverity": "error",
    "rules": { "no-unused-variable": { "severity": "error" } }
}
```

#### Actual behavior

An internal error within the rule occurs:

```
Warning: Cannot read property 'flags' of undefined
```

#### Expected behavior

TSLint should execute without internal errors and should report that `SomeClass` and `name` are both unused.

#### Additional information

The import is required to effect the internal error. If the import is omitted, no internal error occurs.

With the import, the rule follows a code path into the [`getImplicitType`](https://github.com/palantir/tslint/blob/5.4.2/src/rules/noUnusedVariableRule.ts#L277) function and the implementations of `utils.isPropertyDeclaration` and `utils.isVariableDeclaration` allow for the `{ name } = person` node to be passed to `checker.getTypeAtLocation` - from within which the error is thrown.

victornoel:
@cartant do you think this is the same as #3001?

cartant:
@victornoel Maybe. That issue doesn't seem to include enough information for the reported problem to be reproduced, so who knows?

victornoel:
@adidahiya any plan to fix this issue? it is really problematic…

adidahiya:
Is this still an issue with TSLint 5.5.0?

cartant:
@adidahiya I've revisted the harness I created and the error is still effected with:

- __TSLint version__: 5.5.0
- __TypeScript version__: 2.4.1
- __Running TSLint via__: CLI

---

    TypeError: Cannot read property 'flags' of undefined
        at Object.getCheckFlags (C:\Users\Nicholas\git\temp\tslint-no-unused-variable\node_modules\typescript\lib\typescript.js:10137:22)
        at getTypeOfSymbol (C:\Users\Nicholas\git\temp\tslint-no-unused-variable\node_modules\typescript\lib\typescript.js:30957:20)
[...]

PR #3058, opened 2017-07-19
no-unused-variable: don't crash at destructuring

ajafff:
#### PR checklist

- [x] Addresses an existing issue: #2876, #3001
  - [x] Includes tests
- [ ] Documentation update

#### Overview of change:
Avoid typescript crash when trying to get type of destructured variable declaration.
Also avoid walking the AST multiple times when `--declaration` is not enabled. That should result in better performance and fewer false negatives.

[bugfix] `no-unused-variable` fixed crash when using destructuring
Fixes: #2876
Fixes: #3001


pelotom:
Hi, we would love to get this fix. Any idea when we can expect it to make its way into a release?

adidahiya:
@pelotom we'll try to release v5.6.0 this week. in the meantime you can grab the latest commit from master

PR #3064, opened 2017-07-21
Add new rule: grouped-imports

trenneman:
#### PR checklist

  - [x] Includes tests
- [ ] Documentation update

#### Overview of change:

Adds new rule to verify imports are grouped by libraries, parent directories & current directory.

The fixes for too many/few blank lines fix a single import statement. The fix for changing the order fixes all imports statements completely. I am not sure this is OK.

#### CHANGELOG.md entry:

[new-rule] `grouped-imports`

### DIFF
--- a/src/rules/groupedImportsRule.ts
+++ b/src/rules/groupedImportsRule.ts
@@ -1,0 +1,152 @@
+/**
+ * @license
+ * Copyright 2017 Palantir Technologies, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import * as Lint from "tslint";
+import { isImportDeclaration } from "tsutils";
+import * as ts from "typescript";
+
+export class Rule extends Lint.Rules.AbstractRule {
+    /* tslint:disable:object-literal-sort-keys */
+    public static metadata: Lint.IRuleMetadata = {
+        ruleName: "grouped-imports",
+        description: "Separate import groups by blank lines.",
+        rationale: "Keeps a clear overview on dependencies.",
+        optionsDescription: "Not configurable.",
+        hasFix: true,
+        options: {},
+        optionExamples: [true],
+        type: "style",
+        typescriptOnly: false,
+    };
+    /* tslint:enable:object-literal-sort-keys */
+
+    public static IMPORT_SOURCES_SEPARATED = "Import sources within a group must not be separated by blank lines";
+    public static IMPORT_SOURCES_NOT_SEPARATED =
+        "Import sources of different groups must be separated by a single blank line";
+    public static IMPORT_SOURCES_ORDER =
+        "Import sources of different groups must be sorted by: libraries, parent directories, current directory";
+
+    public apply(sourceFile: ts.SourceFile): Lint.RuleFailure[] {
+        return this.applyWithWalker(new Walker(sourceFile, this.ruleName, this.getOptions()));
+    }
+}
+
+enum ImportStatementType {
+    LIBRARY_IMPORT = 1,
+    PARENT_DIRECTORY_IMPORT = 2,
+    CURRENT_DIRECTORY_IMPORT = 3,
+}
+
+interface ImportStatement {
+    statement: ts.Statement;
+    type: ImportStatementType;
+    lineStart: number;
+    lineEnd: number;
+}
+
+class Walker extends Lint.AbstractWalker<Lint.IOptions> {
+    private lastImportStatement: ImportStatement;
+
+    private static getImportStatementType(statement: ts.Statement): ImportStatementType {
+        const path = Walker.getImportPath(statement);
+        if (path.charAt(0) === ".") {
+            if (path.charAt(1) === ".") {
+                return ImportStatementType.PARENT_DIRECTORY_IMPORT;
+            } else {
+                return ImportStatementType.CURRENT_DIRECTORY_IMPORT;
+            }
+        } else {
+            return ImportStatementType.LIBRARY_IMPORT;
+        }
+    }
+
+    private static getImportPath(statement: ts.Statement): string {
+        const str = statement.getText();
+        let index;
+        let lastIndex;
+        index = str.indexOf("'");
+        if (index > 0) {
+            lastIndex = str.lastIndexOf("'");
+        } else {
+            index = str.indexOf("\"");
+            lastIndex = str.lastIndexOf("\"");
+        }
+        if (index < 0 || lastIndex < 0) {
+            throw new Error(`Unable to extract path from import statement \`${statement.getText()}\``);
+        }
+        return str.substring(index + 1, lastIndex);
+    }
+
+    public walk(sourceFile: ts.SourceFile): void {
+        sourceFile.statements
+            .filter(isImportDeclaration)
+            .forEach((st) => this.checkStatement(st));
+    }
+
+    private toImportStatement(statement: ts.Statement): ImportStatement {
+        return {
+            lineEnd: this.sourceFile.getLineAndCharacterOfPosition(statement.getEnd()).line,
+            lineStart: this.sourceFile.getLineAndCharacterOfPosition(statement.getStart()).line,
+            statement,
+            type: Walker.getImportStatementType(statement),
+        };
+    }
+
+    private checkStatement(statement: ts.Statement): void {
+        const importStatement = this.toImportStatement(statement);
+        if (this.lastImportStatement) {
+            this.checkImportStatement(importStatement);
+        }
+        this.lastImportStatement = importStatement;
+    }
+
+    private checkImportStatement(importStatement: ImportStatement) {
+        if (importStatement.type === this.lastImportStatement.type) {
+            if (importStatement.lineStart !== this.lastImportStatement.lineEnd + 1) {
+                const replacement = Lint.Replacement.deleteFromTo(
+                    this.lastImportStatement.statement.getEnd() + 1, importStatement.statement.getStart());
+                this.addFailureAtNode(importStatement.statement, Rule.IMPORT_SOURCES_SEPARATED, replacement);
+            }
+        } else if (importStatement.type.valueOf() < this.lastImportStatement.type.valueOf()) {
+            this.addFailureAtNode(importStatement.statement, Rule.IMPORT_SOURCES_ORDER, this.getAllImportsFix());
+        } else {
+            if (importStatement.lineStart !== this.lastImportStatement.lineEnd + 2) {
+                const replacement = Lint.Replacement.appendText(importStatement.statement.getStart(), ts.sys.newLine);
+                this.addFailureAtNode(importStatement.statement, Rule.IMPORT_SOURCES_NOT_SEPARATED, replacement);
+            }
+        }
+    }
+
+    private getAllImportsFix(): Lint.Fix {
+        const importStatements = this.sourceFile.statements.filter(isImportDeclaration);
+        const libs = importStatements.filter((st) => Walker.getImportStatementType(st) === ImportStatementType.LIBRARY_IMPORT);
+        const parent = importStatements.filter((st) => Walker.getImportStatementType(st) === ImportStatementType.PARENT_DIRECTORY_IMPORT);
+        const current = importStatements.filter((st) => Walker.getImportStatementType(st) === ImportStatementType.CURRENT_DIRECTORY_IMPORT);
+        let imports: string[] = [];
+        [libs, parent, current].forEach((statements) => {
+            if (statements.length) {
+                imports = imports.concat(statements.map((st) => st.getText()));
+                imports.push("");
+            }
+        });
+        return Lint.Replacement.replaceFromTo(
+            importStatements[0].getStart(),
+            importStatements[importStatements.length - 1].getEnd(),
+            imports.join(ts.sys.newLine),
+        );
+    }
+}
--- a/src/rules/groupedImportsRule.ts
+++ b/src/rules/groupedImportsRule.ts
@@ -16,7 +16,7 @@
  */
 
 import * as Lint from "tslint";
-import { isImportDeclaration } from "tsutils";
+import { isImportDeclaration, isTextualLiteral } from "tsutils";
 import * as ts from "typescript";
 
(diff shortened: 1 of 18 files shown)


palantirtech:
Thanks for your interest in [palantir/tslint](https://github.com/palantir/tslint), @trenneman! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

test/rules/grouped-imports/bad-order.ts.lint
```diff
@@ -0,0 +1,6 @@
+import {bar} from '../bar';
+
+import {foo} from 'foo';
+~~~~~~~~~~~~~~~~~~~~~~~~ [Import sources of different groups must be sorted by: libraries, parent directories, current directory]
```
ajafff: please add a test that has both kinds of error: unnecessary blank line and wrong order

src/rules/groupedImportsRule.ts
```diff
@@ -0,0 +1,152 @@
...
+        }
+        if (index < 0 || lastIndex < 0) {
+            throw new Error(`Unable to extract path from import statement \`${statement.getText()}\``);
+        }
+        return str.substring(index + 1, lastIndex);
+    }
+
+    public walk(sourceFile: ts.SourceFile): void {
+        sourceFile.statements
+            .filter(isImportDeclaration)
```
ajafff: what about `ImportEqualsDeclaration` where `isExternalModuleReference(statement.moduleReference)`? For example:

```ts
import fs = require('fs');
```

src/rules/groupedImportsRule.ts
```diff
@@ -0,0 +1,152 @@
...
+                return ImportStatementType.PARENT_DIRECTORY_IMPORT;
+            } else {
+                return ImportStatementType.CURRENT_DIRECTORY_IMPORT;
+            }
+        } else {
+            return ImportStatementType.LIBRARY_IMPORT;
+        }
+    }
+
+    private static getImportPath(statement: ts.Statement): string {
```
ajafff:
```ts
getImportPath(statement: ts.ImportDeclaration): string {
    if (isTextualLiteral(statement.moduleSpecifier)) {
        return statement.moduleSpecifier.text;
    }
    // TODO handle invalid grammar like `import foo from 'a'+foo+'b';`
}
```

src/rules/groupedImportsRule.ts
```diff
@@ -0,0 +1,152 @@
...
+
+interface ImportStatement {
+    statement: ts.Statement;
+    type: ImportStatementType;
+    lineStart: number;
+    lineEnd: number;
+}
+
+class Walker extends Lint.AbstractWalker<Lint.IOptions> {
+    private lastImportStatement: ImportStatement;
```
ajafff: `: ImportStatement | undefined`

src/rules/groupedImportsRule.ts
```diff
@@ -0,0 +1,152 @@
...
+            this.checkImportStatement(importStatement);
+        }
+        this.lastImportStatement = importStatement;
+    }
+
+    private checkImportStatement(importStatement: ImportStatement) {
+        if (importStatement.type === this.lastImportStatement.type) {
+            if (importStatement.lineStart !== this.lastImportStatement.lineEnd + 1) {
+                const replacement = Lint.Replacement.deleteFromTo(
+                    this.lastImportStatement.statement.getEnd() + 1, importStatement.statement.getStart());
```
ajafff: I guess this won't work for 
```ts
import foo from 'foo';import bar from 'bar'; // not the missing space after the first semicolon
```

src/rules/groupedImportsRule.ts
```diff
@@ -0,0 +1,152 @@
...
+        const importStatement = this.toImportStatement(statement);
+        if (this.lastImportStatement) {
+            this.checkImportStatement(importStatement);
+        }
+        this.lastImportStatement = importStatement;
+    }
+
+    private checkImportStatement(importStatement: ImportStatement) {
+        if (importStatement.type === this.lastImportStatement.type) {
+            if (importStatement.lineStart !== this.lastImportStatement.lineEnd + 1) {
```
ajafff: should probably be `>` instead of `!==`, see comment below

src/rules/groupedImportsRule.ts
```diff
@@ -0,0 +1,152 @@
...
+                const replacement = Lint.Replacement.deleteFromTo(
+                    this.lastImportStatement.statement.getEnd() + 1, importStatement.statement.getStart());
+                this.addFailureAtNode(importStatement.statement, Rule.IMPORT_SOURCES_SEPARATED, replacement);
+            }
+        } else if (importStatement.type.valueOf() < this.lastImportStatement.type.valueOf()) {
+            this.addFailureAtNode(importStatement.statement, Rule.IMPORT_SOURCES_ORDER, this.getAllImportsFix());
+        } else {
+            if (importStatement.lineStart !== this.lastImportStatement.lineEnd + 2) {
```
ajafff: s/!==/</

src/rules/groupedImportsRule.ts
```diff
@@ -0,0 +1,152 @@
...
+    }
+
+    private checkImportStatement(importStatement: ImportStatement) {
+        if (importStatement.type === this.lastImportStatement.type) {
+            if (importStatement.lineStart !== this.lastImportStatement.lineEnd + 1) {
+                const replacement = Lint.Replacement.deleteFromTo(
+                    this.lastImportStatement.statement.getEnd() + 1, importStatement.statement.getStart());
+                this.addFailureAtNode(importStatement.statement, Rule.IMPORT_SOURCES_SEPARATED, replacement);
+            }
+        } else if (importStatement.type.valueOf() < this.lastImportStatement.type.valueOf()) {
```
ajafff: why the `valueOf()`?

src/rules/groupedImportsRule.ts
```diff
@@ -0,0 +1,152 @@
...
+            this.addFailureAtNode(importStatement.statement, Rule.IMPORT_SOURCES_ORDER, this.getAllImportsFix());
+        } else {
+            if (importStatement.lineStart !== this.lastImportStatement.lineEnd + 2) {
+                const replacement = Lint.Replacement.appendText(importStatement.statement.getStart(), ts.sys.newLine);
```
ajafff: `ts.sys.newLine` does not always match the line endings in the current file.
Have a look at `eoflineRule.ts` to see how the line ending is inferred there.

src/rules/groupedImportsRule.ts
```diff
@@ -0,0 +1,152 @@
...
+        let imports: string[] = [];
+        [libs, parent, current].forEach((statements) => {
+            if (statements.length) {
+                imports = imports.concat(statements.map((st) => st.getText()));
+                imports.push("");
+            }
+        });
+        return Lint.Replacement.replaceFromTo(
+            importStatements[0].getStart(),
+            importStatements[importStatements.length - 1].getEnd(),
```
ajafff: I guess that will break code where there's another statement between imports:

```ts
import bar from './bar';
let baz = bar.baz;
import foo from 'foo';
```

test/rules/grouped-imports/bad-order.ts.lint
trenneman: For now, I only add a single failure & fix in case of wrong orders. I've added a test with both cases though. More on this in a later comment.

src/rules/groupedImportsRule.ts
trenneman: Is this really necessary? It adds a lot of complexity and additional testing throughout the rest of the code.

src/rules/groupedImportsRule.ts
trenneman: Added. In what case the `isTextualLiteral` returns false? Your example seems to be processed as `import foo from 'a'`.

src/rules/groupedImportsRule.ts
trenneman: True, updated the fixes.

src/rules/groupedImportsRule.ts
trenneman: Removed.

src/rules/groupedImportsRule.ts
trenneman: Thanks. I've updated this.

trenneman:
Thanks for your first remarks @ajafff.

I've fixed some of the issues. But three issues need some further thinking first:
* Preserve comments.
* How to deal with `import fs = require('fs');`-like statements?
* How to deal with statements between imports?

src/rules/groupedImportsRule.ts
ajafff: That's just my preference for documentation purposes. It helps someone else (or even you) to understand how the rule works when he needs to change something.
There's no need for any additional checks. In the places where `lastImportStatement` can be undefined, you already have a check. In all other places you can simply assert that it's never undefined, e.g. `this.lastImportStatement!.type`

ajafff:
@trenneman 

> How to deal with `import fs = require('fs');`-like statements?

It's probably fine to ignore them like `ordered-imports` does. This kind of import is only there for backwards compatibility with pre-ES6 modules anyway.

> I've thinking about the comments/additional statements. If you ask me they should all be put AFTER the imports (without deleting any of them).

That would be the best way to go. Maybe someone will complain about the wrong comment position after fixing. But until then we should keep the rule simple.
If you have trouble implementing the special handling for statements between imports, you can also choose to not implement it and just don't fix in such cases. IMO that code path is not likely to be hit often.

src/rules/groupedImportsRule.ts
trenneman: Ah great! Didn't know that was possible.

trenneman:
@ajafff Please have a look at the updated version. The walker now quits after the first incorrect import and creates a fix for all imports. The imports are removed and re-inserted in the correct order. This ensures now existing code (comments or statements) are lost.

src/rules/groupedImportsRule.ts
```diff
@@ -0,0 +1,171 @@
...
+    lineEnd: number;
+}
+
+class Walker extends Lint.AbstractWalker<Lint.IOptions> {
+    private previousImportStatement: ImportStatement | undefined;
+
+    public walk(sourceFile: ts.SourceFile): void {
+        const importsStatements = sourceFile.statements
+            .filter(isImportDeclaration)
+            .map(this.toImportStatement);
```
ajafff: use `.map(this.toImportStatement, this)`. Same for `.find(this.isBadlyPositioned, this)` below.
That way you can make both methods "real" methods (no need for an arrow function)

src/rules/groupedImportsRule.ts
```diff
@@ -0,0 +1,171 @@
...
+        if (isTextualLiteral(statement.moduleSpecifier)) {
+            return statement.moduleSpecifier.text;
+        }
+        return "";
+    }
+
+    private createFix(importStatements: ImportStatement[]): Lint.Fix {
+        const newLine = this.getEofChar(this.sourceFile);
+        const imports = this.getOrderedImports(importStatements);
+        const addition = Lint.Replacement.appendText(0, imports.join(newLine));
```
ajafff: Instead of `0` consider using the start of the first import found as insertion position.

If you use 0, you'll break source files that contain a shebang (e.g. `#! env node`) in the first line.

src/rules/groupedImportsRule.ts
```diff
@@ -0,0 +1,171 @@
...
+            end++;
+        }
+        return [start, end];
+    }
+
+    private isWhiteSpaceChar(char: string | undefined): boolean {
+        return char === undefined ? false : Lint.isWhiteSpace(char.charCodeAt(0));
+    }
+
+    private getEofChar(sourceFile: ts.SourceFile): string {
```
ajafff: s/Eof/Eol/

src/rules/groupedImportsRule.ts
```diff
@@ -0,0 +1,171 @@
...
+
+    private getOrderedImports(importStatements: ImportStatement[]): string[] {
+        const libs = importStatements.filter((imp) => imp.type === ImportStatementType.LIBRARY_IMPORT);
+        const parent = importStatements.filter((imp) => imp.type === ImportStatementType.PARENT_DIRECTORY_IMPORT);
+        const current = importStatements.filter((imp) => imp.type === ImportStatementType.CURRENT_DIRECTORY_IMPORT);
+        return [libs, parent, current].reduce((arr, imps) => {
+            if (imps.length == 0) {
+                return arr;
+            }
+            return arr.concat(imps.map((imp) => imp.statement.getText()), "");
```
ajafff: nit: consider passing `sourceFile` as parameter to `.getText()` and `.getStart()` for performance reasons.
Though that's not as relevant here, because it's only used in case of a lint error (which should not be the hot path).

trenneman:
Thanks @ajafff. I've processed your comments. `getRangeIncludingWhitespace` now only removes from the beginning of the first to the end of the last import statement.

I've also rebased on master. Let me know when I should squash as well.

src/rules/groupedImportsRule.ts
```diff
@@ -0,0 +1,179 @@
...
+
+import * as Lint from "tslint";
+import { isImportDeclaration, isTextualLiteral } from "tsutils";
+import * as ts from "typescript";
+
+export class Rule extends Lint.Rules.AbstractRule {
+    /* tslint:disable:object-literal-sort-keys */
+    public static metadata: Lint.IRuleMetadata = {
+        ruleName: "grouped-imports",
+        description: "Separate import groups by blank lines.",
```
adidahiya: needs better docs. it's unclear what import groups are from this simple description. I think the phrase "import statement" needs to appear. and the exact algorithm (libraries, parent directories, current directory) needs to be explained in `descriptionDetails` of `IRuleMetadata`.

src/rules/groupedImportsRule.ts
```diff
@@ -0,0 +1,179 @@
...
+        rationale: "Keeps a clear overview on dependencies.",
+        optionsDescription: "Not configurable.",
+        hasFix: true,
+        options: {},
+        optionExamples: [true],
+        type: "style",
```
adidahiya: I would actually argue that this falls into "maintainability"

src/rules/groupedImportsRule.ts
```diff
@@ -0,0 +1,179 @@
...
+                } else if (importStatement.lineStart !== this.previousImportStatement.lineEnd + 2) {
+                    return true;
+                }
+            }
+        }
+        this.previousImportStatement = importStatement;
+        return false;
+    }
+
+    private toImportStatement(statement: ts.ImportDeclaration): ImportStatement {
```
adidahiya: nit: slightly better name would be `importDeclarationToStatement`

trenneman:
Thanks @adidahiya 

I actually think it makes sense to merge it with the `ordered-imports` rule. Especially, because I already use it on top of that one. It will take me some time to understand that rule as it it rather large, but I think I can manage that. Shall I update this PR with a proposal?

adidahiya:
@trenneman I think opening a new PR would be best (link to it from this one).

trenneman:
Continued in #3138 .

Issue #2869, opened 2017-06-02
New rule: no-duplicate-imports


adidahiya:
Just like http://eslint.org/docs/rules/no-duplicate-imports

bad:

```ts
import { merge } from 'module';
import something from 'another-module';
import { find } from 'module';
```

good:

```ts
import { merge, find } from 'module';
import something from 'another-module';
```

chances:
> In some cases you cannot simply merge imports.

Why _should_ this cause a linting error?

If they're impossible to merge then that's not an error.

IllusionMH:
@chances can you provide examples when they cannot be merged?
If it is common case - might be added as rule option or may not be included into `recommended` config (not it is only in `all` and `latest`).

In most cases it may be causes by copy/paste or bugs in auto-imports.
Rule helps to avoid such error cases. exceptional cases can be handled with inline disable comments or rule can be disabled in your configuration.

ajafff:
That's a quote from my PR. Here the example:

```ts
import foo from 'foo';
import foo2 from 'foo'; // cannot be combined with preceding import

import * as bar from 'bar';
import {baz} from 'bar'; // cannot be combined with preceding import
```

If you find something like that in your code, it should be refactored anyway. 
"cannot simply be merged" means that the autofixer would need to rename all references to the merged import in the file.

buu700:
@ajafff, how is it recommended to refactor those cases? Right now I have a case where I need both `bar` and `baz` (and `bar.baz` would work but be annoyingly verbose), and the only workaround I can see is to disable this rule. It seems like it'd make more sense for non-fixable instances to be treated by the rule as valid.

IllusionMH:
Cases mentioned above looks like lazy copy/paste. Both cases can be normalized with hands and I don't see reason to leave them as separate imports.

@buu700 if `bar.baz` is too verbose, that what reason for `* as bar`in first place? Only to export it?
Otherwise it is better to use several named imports (`{anything, baz}`) from this module or stick to `bar.anything` for consistency.

buu700:
As mentioned, I need to use both `baz` and `bar` (as a class), which is because it's a JS library that does things that would be silly/unexpected in idiomatic TypeScript.

`bar.baz` is basically okay, and the solution I ended up going with, but there are two reasons I don't really like it:

1. Other classes reference `baz` directly, so the fact that this class is using `bar.baz` all over the place is a little inconsistent.

[...]

PR #3075, opened 2017-07-25
Implement new rule no-duplicate-imports

ajafff:
#### PR checklist

- [x] Addresses an existing issue: #2869
  - [x] Includes tests
- [x] Documentation update

#### Overview of change:
[new-rule] `no-duplicate-imports`
Fixes: #2869

Some thoughts about the current implemenation:
* There no option `includeExports` like in the ESLint rule. Do we want this?
* ImportEqualsDeclarations are ignored
* There's no fixer. That can be added later.
* In some cases you cannot simply merge imports. The rule will still complain about them (maybe the error message should be changed?):

```ts
import foo from 'foo';
import foo2 from 'foo'; // cannot be combined with preceding import

import * as bar from 'bar';
import {baz} from 'bar'; // cannot be combined with preceding import
```

### DIFF
--- a/src/configs/all.ts
+++ b/src/configs/all.ts
@@ -145,6 +145,7 @@ export const rules = {
     "max-file-line-count": [true, 1000],
     "max-line-length": [true, 120],
     "no-default-export": true,
+    "no-duplicate-imports": true,
     "no-irregular-whitespace": true,
     "no-mergeable-namespace": true,
     "no-require-imports": true,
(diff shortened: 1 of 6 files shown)


adidahiya:
> In some cases you cannot simply merge imports. The rule will still complain about them (maybe the error message should be changed?)

👍  for still complaining about them. same error message is fine for now

PR #3079, opened 2017-07-28
Only import ./test when needed

### LINKED ISSUE #3072: tslint npm package missing './test/parse'
### Bug Report

- __TSLint version__: 5.5.0
- __TypeScript version__:
- __Running TSLint via__: CLI

```ts
// code snippet
```

```json
{
  "extends": "tslint:latest",
  "rules": {
    /**
     * Security Rules. The following rules should be turned on because they find security issues
     * or are recommended in the Microsoft Secure Development Lifecycle (SDL)
     */
    "insecure-random": true,
    "no-banned-terms": true,
    "no-delete-expression": true,
    "no-disable-auto-sanitization": true,
    "no-eval": true,
    "no-function-constructor-with-string-args": true,
    "no-http-string": [true, "http://www.example.com/?.*", "http://www.examples.com/?.*"],
    "no-octal-literal": true,
    "no-reserved-keywords": true,
    "no-string-based-set-immediate": true,
    "no-string-based-set-interval": true,
    "no-string-based-set-timeout": true,
    "non-literal-require": true,
    "possible-timing-attack": true,

    /**
     * Common Bugs and Correctness. The following rules should be turned on because they find
     * common bug patterns in the code or enforce type safety.
     */
    "forin": true,
    "no-any": true,
    "no-arg": true,
    "no-bitwise": true,
    "no-conditional-assignment": true,
    "no-console": [true, "debug", "info", "log", "time", "timeEnd", "trace"],
    "no-constant-condition": true,
    "no-control-regex": true,
    "no-duplicate-case": true,
    "no-duplicate-variable": true,
    "no-empty": true,
    "no-for-in-array": false,
    "no-increment-decrement": true,
    "no-invalid-regexp": true,
    "no-invalid-this": true,
    "no-regex-spaces": true,
    "no-sparse-arrays": true,
    "no-stateless-class": true,
    "no-string-literal": true,
    "no-string-throw": true,
    "no-unnecessary-bind": true,
    "no-unnecessary-override": true,
    "no-unsafe-finally": true,
    "no-unused-expression": true,
    "no-unused-new": true,
    "no-use-before-declare": true,
    "no-with-statement": true,
    "promise-must-complete": true,
    "radix": true,
    "switch-default": true,
    "triple-equals": [true, "allow-null-check"],
    "use-isnan": true,
    "use-named-parameter": true,
    "valid-typeof": true,

    /**
     * Code Clarity. The following rules should be turned on because they make the code
     * generally more clear to the reader.
     */
    "adjacent-overload-signatures": true,
    "array-type": [true, "array"],
    "arrow-parens": false, // for simple functions the parens on arrow functions are not needed
    "callable-types": true,
    "class-name": true,
    "comment-format": [true, "check-space", "check-uppercase"],
    "export-name": false,
    "function-name": [true, {
      "method-regex": "^[a-z][\\w\\d]+$",
      "private-method-regex": "^[a-z_][\\w\\d]+$",
      "protected-method-regex": "^[a-z][\\w\\d]+$",
      "static-method-regex": "^[A-Z][\\w\\d]+$",
      "function-regex": "^[a-z][\\w\\d]+$"
    }],
    "import-name": false,
    "interface-name": [true, "never-prefix"],
    "jsdoc-format": true,
    "max-classes-per-file": [true, 3], // we generally recommend making one public class per file
    "max-file-line-count": [true, 500],
    "max-func-body-length": [true, 100, {
      "ignore-parameters-to-function-regex": "describe"
    }],
    "max-line-length": [true, 140],
    "member-access": true,
    "member-ordering": [
      true,
      {
        "order": [{
            "name": "static non-private",
            "kinds": [
              "public-static-field",
              "protected-static-field",
              "public-static-method",
              "protected-static-method"
            ]
          },
          "instance-field",
          "constructor"
        ]
      }
    ],
    "missing-jsdoc": false,
    "new-parens": true,
    "no-construct": true,
    "no-default-export": false,
    "no-empty-interfaces": true,
    "no-for-in": false,
    "no-function-expression": true,
    "no-multiline-string": true, // multiline-strings often introduce unnecessary whitespace into the string literals
    "no-null-keyword": false, // turn no-null-keyword off and use undefined to mean not initialized and null to mean without a value
    "no-parameter-properties": false,
    "no-relative-imports": false,
    "no-require-imports": true,
    "no-shadowed-variable": true,
    "no-suspicious-comment": true,
    "no-typeof-undefined": true,
    "no-unnecessary-field-initialization": true,
    "no-unnecessary-local-variable": false,
    "no-var-keyword": true,
    "no-var-requires": true,
    "no-var-self": true,
    "object-literal-sort-keys": false, // turn object-literal-sort-keys off and sort keys in a meaningful manner
    "one-variable-per-declaration": [true, "ignore-for-loop"],
    "ordered-imports": [false, "any"],
    "prefer-array-literal": true,
    "prefer-const": false,
    "prefer-for-of": true,
    "typedef": [true, "call-signature", "parameter", "property-declaration", "member-variable-declaration"],
    "underscore-consistent-invocation": true,
    "unified-signatures": true,
    "variable-name": [true, "ban-keywords", "check-format", "allow-leading-underscore"],

    /**
     * Whitespace related rules. The only recommended whitespace strategy is to pick a single format and
     * be consistent.
     */
    "align": [true, "parameters", "statements"],
    "curly": true,
    "eofline": false,
    "import-spacing": true,
    "indent": [true, "spaces"],
    "no-empty-line-after-opening-brace": false,
    "no-single-line-block-comment": true,
    "no-unnecessary-semicolons": true,
    "object-literal-key-quotes": [true, "as-needed"],
    "one-line": [true, "check-open-brace", "check-catch", "check-else", "check-whitespace"],
    "quotemark": [true, "single"],
    "semicolon": [true, "never"],
    "trailing-comma": [true, {
      "singleline": "never",
      "multiline": "always"
    }], // forcing trailing commas for multi-line
    "typedef-whitespace": [
      true,
      {
        "call-signature": "nospace",
        "index-signature": "nospace",
        "parameter": "nospace",
        "property-declaration": "nospace",
        "variable-declaration": "nospace"
      },
      {
        "call-signature": "onespace",
        "index-signature": "onespace",
        "parameter": "onespace",
        "property-declaration": "onespace",
        "variable-declaration": "onespace"
      }
    ],
    "whitespace": [true, "check-branch", "check-decl", "check-operator", "check-separator", "check-type"],

    /**
     * Controversial/Configurable rules.
     */
    "ban": false, // only enable this if you have some code pattern that you want to ban
    "import-blacklist": false, // enable and configure this as you desire
    "interface-over-type-literal": false, // there are plenty of reasons to prefer interfaces
    "no-angle-bracket-type-assertion": false, // pick either type-cast format and use it consistently
    "no-empty-interface": false, // it is currently better to use the MS version of this rule
    "no-inferred-empty-object-type": false, // if the compiler is satisfied then this is probably not an issue
    "no-internal-module": false, // only enable this if you are not using internal modules
    "no-magic-numbers": false, // by default it will find too many false positives
    "no-mergeable-namespace": false, // your project may require mergeable namespaces
    "no-namespace": false, // only enable this if you are not using modules/namespaces
    "no-reference": true, // in general you should use a module system and not /// reference imports
    "no-unexternalized-strings": false, // the VS Code team has a specific localization process that this rule enforces
    "object-literal-shorthand": false, // object-literal-shorthand offers an abbreviation not an abstraction
    "prefer-type-cast": true, // pick either type-cast format and use it consistently
    "space-before-function-paren": false // turn this on if this is really your coding standard
  },
  "rulesDirectory": [
    "node_modules/tslint-microsoft-contrib"
  ]
}
```

#### Actual behavior
running the cli below produce an error
tslint -c tslint.json -t stylish '{src,__tests__,scripts}/**/*.{ts,tsx}'
module.js:487
    throw err;
    ^

Error: Cannot find module './test/parse'
    at Function.Module._resolveFilename (module.js:485:15)
    at Function.Module._load (module.js:437:25)
    at Module.require (module.js:513:17)
    at require (internal/module.js:11:18)
    at Object.<anonymous> (D:\temp\helloworld\node_modules\tslint\lib\test.js:28
:13)
    at Module._compile (module.js:569:30)
    at Object.Module._extensions..js (module.js:580:10)
    at Module.load (module.js:503:32)
    at tryModuleLoad (module.js:466:12)
    at Function.Module._load (module.js:458:3)
#### Expected behavior
I think the './test/parse' files/modules are not publish

ajafff:
#### PR checklist

- [x] Addresses an existing issue: #3072
  - [ ] Includes tests
- [ ] Documentation update

#### Overview of change:
Only import the sources needed for rule tests if the `--test` option is used. ~That fixes #3072 for our standard users. Developers using the `--test` feature might still have trouble with the missing `./test` directory caused by `yarn clean`.~
That avoid loading unnecessary modules like `colors`, `diff`, `semver` and everything in `src/test/`

### DIFF
--- a/src/runner.ts
+++ b/src/runner.ts
@@ -32,7 +32,6 @@ import {
 import { FatalError } from "./error";
 import { LintResult } from "./index";
 import * as Linter from "./linter";
-import { consoleTestResultsHandler, runTests } from "./test";
 import { arrayify, flatMap } from "./utils";
 
 export interface Options {
@@ -141,8 +140,9 @@ async function runWorker(options: Options, logger: Logger): Promise<Status> {
     }
 
     if (options.test) {
-        const results = runTests((options.files || []).map(trimSingleQuotes), options.rulesDirectory);
-        return consoleTestResultsHandler(results) ? Status.Ok : Status.FatalError;
+        const test = await import("./test");
+        const results = test.runTests((options.files || []).map(trimSingleQuotes), options.rulesDirectory);
+        return test.consoleTestResultsHandler(results) ? Status.Ok : Status.FatalError;
     }
 
     if (options.config && !fs.existsSync(options.config)) {


aleung:
@ajafff I feel it's not necessary to make that complicate to use dynamic import. 

Only need to rename folder `test` to a new name and modify test.ts [L28](https://github.com/palantir/tslint/blob/master/src/test.ts#L28) and L29 to import from the new folder. Some test cases import from `test` and they need to be updated as well.

aleung:
I make a PR #3081 in the way I described.

ajafff:
@aleung You're right that your PR is the better way to fix #3072 

This one has another purpose: Always importing `test.ts` loads `colors`, `diff` and `semver` which will not be used in over 99,9% of all uses. Although it might not be noticeable, the dynamic import can avoid this unnecessary work.

Issue #2308, opened 2017-03-07
*space-before-function-paren* with async arrow function with generic type parameter


ibratoev:
### Bug Report

- __TSLint version__: 4.5.1
- __TypeScript version__: 2.1.4
- __Running TSLint via__: CLI

#### TypeScript code being linted

```ts
const a = async <T>(x: string) => x;
```

with `tslint.json` configuration:

```json
{
  "extends": "tslint:latest",
  "rulesDirectory": [
    "node_modules/tslint-eslint-rules/dist/rules"
  ],
  "rules": {
    "quotemark": [
      true,
      "single",
      "avoid-escape",
      "jsx-double"
    ],
    "whitespace": [
      true,
      "check-branch",
      "check-decl",
      "check-module",
      "check-operator",
      "check-separator",
      "check-type"
    ],
    "no-inferrable-types": true,
    "no-reference": false,
    "no-var-requires": false,
    "object-literal-sort-keys": false,
    "interface-name": [false],
    "no-console": [false],
    "member-access": [false],
    "no-extra-semi": true,
    "no-namespace": [true, "allow-declarations"],
    "interface-over-type-literal": false,
    "max-classes-per-file": [false]
  }
}
```

#### Actual behavior
Error: space-before-function-paren
#### Expected behavior
No error because of the generic type parameter.

ajafff:
Fixed by #3085

PR #3085, opened 2017-07-31
Fix 'space-before-function-paren' for anonymous/arrow generic functions

ksvitkovsky:
#### PR checklist

- [x] Addresses an existing issue: #2308
  - [x] Includes tests
- [ ] Documentation update

Fix considers the following cases as exceptions and skips whitespace check:
- arrow function with type parameters
- anonymous functions with type parameters


Jameskmonger:
Will this fix the following case?

    const foo = <T>(a: T) => {};

ajafff:
@Jameskmonger The added tests tell that your code snippet would be valid after this PR lands

ajafff:
Thanks @ksvitkovsky

Jameskmonger:
@hisuwh This fixes the issue you were talking about.

Issue #2511, opened 2017-04-06
improve docs for no-reference import


chrismbarr:
https://palantir.github.io/tslint/rules/no-reference-import/

The **entire** documentation for this rule is as follows:

>"Don’t if you import “foo” anyway."

I have no idea what that means or what this rule does based on this description.  "Don't" what?

adidahiya:
<reference> probably needs to be escaped in the HTML.

adidahiya:
.... and I did it again here. `<reference>` path imports are disallowed

nchen63:
should escape in code generation, not in metadata

hyndsite:
@adidahiya maybe I am missing something, but is that any different than the [no-reference ](https://palantir.github.io/tslint/rules/no-reference/) rule?

PR #3087, opened 2017-08-01
Escape code in no-reference-import documentation metadata

chrismbarr:
#### PR checklist

- [x] Addresses an existing issue: #2511
  - [ ] Includes tests
- [x] Documentation update

#### Overview of change:
This should fix the documentation for the for the `no-reference-import` rule to actually be readable and understood.  Right now it makes no sense.
currently: http://palantir.github.io/tslint/rules/no-reference-import


chrismbarr:
There's 'yer problem!
image

Issue #2122, opened 2017-01-26
Lint whitespace on object spread


adidahiya:
### Config

New option on `whitespace` rule called "check-spread"

### Failing code

```ts
const thing = {
    ... otherThing,
};
```

### Passing code

```ts
const thing = {
    ...otherThing,
};
```

PR #3089, opened 2017-08-01
New rule option: 'check-rest-spread' for 'whitespace' rule

ksvitkovsky:
#### PR checklist

- [x] Addresses an existing issue: #2122
  - [x] Includes tests
- [ ] Documentation update

#### CHANGELOG.md entry:

[new-rule-option] `whitespace`: Add `check-rest-spread` option (#2122)

### DIFF
--- a/src/rules/whitespaceRule.ts
+++ b/src/rules/whitespaceRule.ts
@@ -64,8 +64,8 @@ export class Rule extends Lint.Rules.AbstractRule {
         typescriptOnly: false,
     };
 
-    public static FAILURE_STRING = "missing whitespace";
-    public static FAILURE_STRING_2 = "invalid whitespace";
+    public static FAILURE_STRING_MISSING = "missing whitespace";
+    public static FAILURE_STRING_INVALID = "invalid whitespace";
 
     public apply(sourceFile: ts.SourceFile): Lint.RuleFailure[] {
         return this.applyWithFunction(sourceFile, walk, parseOptions(this.ruleArguments));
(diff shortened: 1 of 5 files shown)


src/rules/whitespaceRule.ts
```diff
@@ -27,6 +27,7 @@ const OPTION_DECL = "check-decl";
 const OPTION_OPERATOR = "check-operator";
 const OPTION_MODULE = "check-module";
 const OPTION_SEPARATOR = "check-separator";
+const OPTION_SPREAD = "check-spread";
```
ajafff: I'd prefer `check-rest-spread` which checks object and array spread as well as object and array rest and rest parameters

src/rules/whitespaceRule.ts
```diff
@@ -192,6 +196,13 @@ function walk(ctx: Lint.WalkContext<Options>) {
                 if (options.decl && initializer !== undefined) {
                     checkForTrailingWhitespace((type !== undefined ? type :  name).getEnd());
                 }
+                break;
+
+            case ts.SyntaxKind.SpreadAssignment:
```
ajafff: This also checks `foo(...bar)`, which is correct if this option checks rest *and* spread. 
It would also need to check `BindingElement` and `Parameter` where `dotDotDotToken` is not undefined and `SpreadElement`.

ksvitkovsky:
@ajafff thank you for the feedback! will get back to it soon

adidahiya:
thanks @ksvitkovsky

ksvitkovsky:
my pleasure :blush:

yak80:
Last build cause error, looks like prob in this option:
```
ERROR: src/components/app/App.ts[42, 11]: missing whitespace
```
```ts
    /**
     * Vue component 'preFetch' hook, that is used in SSR to do required
     * things with given Vuex store before page rendering.
     *
42:  * @param store    Vuex store, to perform actions on.
     *
     * @return    Resolved promise with array of all users, that exist
     *            in the system.
     */
```

Looks like it has conflict with option `check-operator`.

Default `whitespace` configuration: 
```
"whitespace": [true, "check-branch", "check-operator", "check-typecast"])
```

If remove `check-operator` from configuration then all fine and error fixed:
```
"whitespace": [true, "check-branch", "check-typecast"]
```

Issue #3051, opened 2017-07-18
New rule: no-submodule-imports


adidahiya:
I'd like a new rule to disallow reaching into submodules in import paths. Similar to [eslint-plugin-import/no-internal-modules](https://github.com/benmosher/eslint-plugin-import/blob/ccd93942a0b9f59f1b698672ba167ece1aa9f5f1/docs/rules/no-internal-modules.md) (but we can't call it that since TypeScript has its own notion of "internal module").

```ts
// bad
import { Keys } from "@blueprintjs/core/dist/common";

// good
import { Keys } from "@blueprintjs/core";
```

with rule config

```json
"no-submodule-imports": {
  "options": [
    "@blueprinjs/core"
  ]
}
```

aervin:
Will this be very similar to the "import-blacklist" rule? If so, I think I could give it a try.

ajafff:
Yes, this rule will be very similar to import-blacklist. In fact there will
only be a different condition to show an error. The rest of the logic can
be exactly the same.

Am 02.08.2017 18:07 schrieb "aervin_" <notifications@github.com>:

Will this be very similar to the "import-blacklist" rule? If so, I think I
could give it a try.

—
You are receiving this because you are subscribed to this thread.
Reply to this email directly, view it on GitHub
<https://github.com/palantir/tslint/issues/3051#issuecomment-319719610>,
or mute
the thread

PR #3091, opened 2017-08-02
Add no-submodule-imports rule

aervin:
Attempt at the functionality requested here: [#3051](https://github.com/palantir/tslint/issues/3051)

This essentially extends the import-blacklist rule with a custom implementation of checkForBannedImport(). 

optionsDescription: A list of modules whose submodules are blocked from being imported directly.


palantirtech:
Thanks for your interest in [palantir/tslint](https://github.com/palantir/tslint), @aervin! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

aervin:
Fixing CI issue now

src/rules/importBlacklistSubmodulesRule.ts
```diff
@@ -0,0 +1,102 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import {
+    isCallExpression, isExternalModuleReference, isIdentifier, isImportDeclaration, isImportEqualsDeclaration, isTextualLiteral,
```
adidahiya: nit: please split these imports into multiple lines

src/rules/importBlacklistSubmodulesRule.ts
```diff
@@ -0,0 +1,102 @@
...
+import * as Lint from "../index";
+
+export class Rule extends Lint.Rules.AbstractRule {
+    /* tslint:disable:object-literal-sort-keys */
+    public static metadata: Lint.IRuleMetadata = {
+        ruleName: "import-blacklist-submodules",
+        description: Lint.Utils.dedent`
+            Disallows importing any submodule of the listed modules.`,
+        rationale: Lint.Utils.dedent`
+            Not sure on this exactly...`,
```
adidahiya: Try this:

Submodules of some packages are treated as private APIs and the import paths may change without deprecation periods. It's best to stick with top-level package exports.

src/rules/importBlacklistSubmodulesRule.ts
```diff
@@ -0,0 +1,102 @@
...
+        optionsDescription: "A list of modules whose submodules are blacklisted.",
```
adidahiya: "A list of __packages__ whose submodules ..."

src/rules/importBlacklistSubmodulesRule.ts
```diff
@@ -0,0 +1,102 @@
...
+                type: "string",
+            },
+            minLength: 1,
+        },
+        optionExamples: [true, [true, "rxjs", "lodash", "@blueprintjs/core"]],
+        type: "functionality",
+        typescriptOnly: false,
+    };
+
+    public static FAILURE_STRING = "This submodule is blacklisted, try importing from a parent module";
```
adidahiya: How about:

"Submodule import paths from this package are disallowed; import from the root instead"

src/rules/importBlacklistSubmodulesRule.ts
```diff
@@ -0,0 +1,102 @@
...
+
+    public isEnabled(): boolean {
```
adidahiya: I don't think you need to override this method, we don't do it elsewhere

src/rules/importBlacklistSubmodulesRule.ts
```diff
@@ -0,0 +1,102 @@
...
+                ts.forEachChild(statement, findRequire);
+            }
+        }
+    }
+
+    private checkForBannedImport(expression: ts.Expression) {
+        if (isTextualLiteral(expression)) {
+            let blacklistOption = "";
+
+            this.options.forEach((option: string) => {
```
adidahiya: please use a for-of loop, not `forEach`

test/rules/import-blacklist-submodules/test.ts.lint
```diff
@@ -0,0 +1,25 @@
+import { submodule } from '@blueprintjs/core';
```
adidahiya: nit: please use double quotes to be consistent with the default config presets (`quotemark: [true, "double"]`)

test/rules/import-blacklist-submodules/test.ts.lint
```diff
@@ -0,0 +1,25 @@
...
+import { submodule } from '@blueprintjs/core/a/b/c/sub-module';
+                                            ~~~~~~~~~~~~~~~~~ [0]
+
+import sub-module = require('@blueprintjs/core');
+
+import sub-module = require('@blueprintjs/core/a/b/c/sub-module');
+                                              ~~~~~~~~~~~~~~~~~ [0]
+
+// non-static imports cannot be checked
+import {Observable} from rxjs;
```
adidahiya: is this valid TS?

test/rules/import-blacklist-submodules/test.ts.lint
```diff
@@ -0,0 +1,25 @@
...
+import forOwn = require(lodash);
+
```
adidahiya: can you test dynamic import syntax too? You'll probably need to make a new test file that applies to TS 2.4 only (so CI knows not to run the test with TS parser <v2.4). see an example here https://github.com/palantir/tslint/blob/30f83444fe1a0f2f999c95e7297f54361e4ad9eb/test/rules/no-reference/ts240.ts.lint#L1

adidahiya:
also, can you please remove the package-lock.json file? We use yarnpkg lockfiles instead.

aervin:
@adidahiya thanks for all the feedback! I'll work on this later today. May ask for further guidance if that's alright.

test/rules/import-blacklist-submodules/test.ts.lint
aervin: I was confused by this as well. I relied heavily on the [import-blacklist test code](https://github.com/palantir/tslint/blob/master/test/rules/import-blacklist/test.ts.lint) to get my test case up and running. Could use some clarification on this point.

ajafff: I introduced this test case back then. 
It's not valid typescript, but the parser consumes it. The AST can contain any expression as module specifier and tslint should not crash because it only expects string literals.

src/rules/importBlacklistSubmodulesRule.ts
```diff
@@ -0,0 +1,102 @@
...
+    public apply(sourceFile: ts.SourceFile): Lint.RuleFailure[] {
+        return this.applyWithWalker(new ImportBlacklistSubmodulesWalker(sourceFile, this.ruleName, this.ruleArguments));
+    }
+}
+
+class ImportBlacklistSubmodulesWalker extends Lint.AbstractWalker<string[]> {
+    public walk(sourceFile: ts.SourceFile) {
+        const findRequire = (node: ts.Node): void => {
+            if (isCallExpression(node) && node.arguments.length === 1 &&
+                isIdentifier(node.expression) && node.expression.text === "require") {
```
ajafff: Add `|| node.expression.kind === ts.SyntaxKind.ImportKeyword` here to make it work with dynamic imports

aervin: TS is yelling at me unless I use `node.expression.originalKeywordKind`--I was grouping w/ `node.expression.text === "require"` like so:
`
isIdentifier(node.expression) && (node.expression.text === "require" || node.expression.originalKeywordKind === ts.SyntaxKind.ImportKeyword)
`

Am I way off here?

ajafff: `(isIdentifier(node.expression) && node.expression.text === "require" || node.expression.kind  === ts.SyntaxKind.ImportKeyword)`

test/rules/import-blacklist-submodules/test.ts.lint
aervin: You mean something like:
`const forOwn = require(lodash).forOwn`?

src/rules/noSubmoduleImportsRule.ts
```diff
@@ -0,0 +1,106 @@
...
+    /* tslint:disable:object-literal-sort-keys */
+    public static metadata: Lint.IRuleMetadata = {
+        ruleName: "no-submodule-imports",
+        description: Lint.Utils.dedent`
+            Disallows importing any submodule of the listed modules.`,
+        rationale: Lint.Utils.dedent`
+            Submodules of some packages are treated as private APIs and the import
+            paths may change without deprecation periods. It's best to stick with
+            top-level package exports.`,
+        optionsDescription: "A list of packages whose submodules are blacklisted.",
```
ajafff: Consider making this a whitelist and ban all submodule imports by default. 
That means we're only interested in imports that are no relative or absolute paths: `!/^(..?(\/|$)|\/)/.test(path)`
It should handle scoped and unscoped package names.

aervin: Okay great, will tackle it sometime this weekend. Thanks!

aervin: And just to clarify: you mean that `import { forOwn } from "lodash";` and `import { Something } from "@angular/core";` would both be allowed by default, right?

ajafff: Correct, those are always allowed. And `import * as forOwn from "lodash/forown";` as well as `import { NgModel } from "@angular/forms/src/directives"` would be banned by default unless `lodash` or `@angular/forms` are whitelisted respectively.

src/rules/noSubmoduleImportsRule.ts
```diff
@@ -0,0 +1,133 @@
...
+
+            /**
+             * A submodule is being imported.
+             * Check if its path contains any
+             * of the whitelist packages.
+             */
+            let whitelisted = false;
+
+            for (const option of this.options) {
+                if (expression.text.indexOf(option) !== -1) {
```
ajafff: `.startsWith(option +"/")

src/rules/noSubmoduleImportsRule.ts
```diff
@@ -0,0 +1,133 @@
...
+                    whitelisted = true;
```
ajafff: Just return here

src/rules/noSubmoduleImportsRule.ts
```diff
@@ -0,0 +1,133 @@
...
+                    break;
+                }
+            }
+
+            if (whitelisted) { return; }
+
+            this.addFailure(
```
ajafff: Simply use `addFailureAtNode(expression, Rule.FAILURE_STRING)`

src/configs/all.ts
```diff
@@ -117,6 +117,7 @@ export const rules = {
     "no-string-literal": true,
     "no-string-throw": true,
     "no-sparse-arrays": true,
+    // "no-submodule-imports": no sensible default
```
ajafff: Can be enabled by default

aervin:
I'll tackle these latest change requests this morning. @ajafff do I need to add some additional test coverage for dynamic imports?

aervin:
Help @ajafff! CI build error is beyond my control I think. Needs retriggered?

src/rules/noSubmoduleImportsRule.ts
```diff
@@ -0,0 +1,129 @@
...
+            } else {
+                ts.forEachChild(statement, findDynamicImport);
+            }
+        }
+    }
+
+    private checkForBannedImport(expression: ts.Expression) {
+        if (isTextualLiteral(expression)) {
+            if (isAbsoluteOrRelativePath(expression.text)) { return; }
+            if (!isSubmodulePath(expression.text)) { return; }
```
ajafff: Both if statements can be combined with the outer if.

src/rules/noSubmoduleImportsRule.ts
```diff
@@ -0,0 +1,129 @@
...
+             * Check if its path contains any
+             * of the whitelist packages.
+             */
+            for (const option of this.options) {
+                if (expression.text.startsWith(`${option}/`)) {
+                    return;
+                }
+            }
+
+            this.addFailureAtNode(
```
ajafff: Could be written in one line

src/rules/noSubmoduleImportsRule.ts
```diff
@@ -0,0 +1,129 @@
...
+function isAbsoluteOrRelativePath(path: string): boolean {
+    return /^(..?(\/|$)|\/)/.test(path);
+}
+
+function isScopedPath(path: string): boolean {
+    return path[0] === "@";
+}
+
+function isSubmodulePath(path: string): boolean {
+    const pathDepth: number = path.split("/").length;
```
ajafff: Could be simplified to
```ts
return path.split("/").length > (isScopedPath(path) ? 2 : 1);
```

ajafff:
CI is very flaky lately. I restarted it. 

There's still the problem with the changed file permissions. Have a look at the `Changes` tab of this PR to see what I mean. I can't see your tests because there are too many changed files to display.

aervin:
I think I've addressed all requested changes thus far. Happy to make further edits, however.

ajafff:
Sorry for the delay. I'm currently on vacation and don't have that much time to offer and mobile internet connection is pretty bad. 
However, the code looks good to me. Thanks for your great work. 

We may also want to enable this rule in the `tslint:latest` preset.

aervin:
Don't let me bug you @ajafff-- have a great vacation and thanks for all your help!

src/rules/noSubmoduleImportsRule.ts
```diff
@@ -0,0 +1,115 @@
...
+    public static FAILURE_STRING = "Submodule import paths from this package are disallowed; import from the root instead";
+
+    public apply(sourceFile: ts.SourceFile): Lint.RuleFailure[] {
+        return this.applyWithWalker(new NoSubmoduleImportsWalker(sourceFile, this.ruleName, this.ruleArguments));
+    }
+}
+
+class NoSubmoduleImportsWalker extends Lint.AbstractWalker<string[]> {
+    public walk(sourceFile: ts.SourceFile) {
+        const findDynamicImport = (node: ts.Node): void => {
```
adidahiya: this doesn't need to be a `const`, please make it a plain helper `function` at the end of the file

src/rules/noSubmoduleImportsRule.ts
```diff
@@ -0,0 +1,115 @@
...
+                if (isExternalModuleReference(statement.moduleReference) && statement.moduleReference.expression !== undefined) {
+                    this.checkForBannedImport(statement.moduleReference.expression);
+                }
+            } else {
+                ts.forEachChild(statement, findDynamicImport);
+            }
+        }
+    }
+
+    private checkForBannedImport(expression: ts.Expression) {
```
adidahiya: this is a static function (doesn't use `this`), you can make it `static` or move to a plain helper function

adidahiya:
I'm going to merge this and look into the minor style comments separately.

aervin:
Can open a fresh PR for those updates later tonight. Do you imagine many more import-related rules coming along in future? If so, might be nice to keep a template/list of all the different import syntax possibilities that TsLint supports/tests. Becomes a sort of copy and paste job for anyone who needs to test their import-related rule.

src/rules/noSubmoduleImportsRule.ts
ajafff: It can't. It uses the lexical this binding

src/rules/noSubmoduleImportsRule.ts
ajafff: Only if `this addFailureAtNode` is moved out of the function. This one should stay as is.

ajafff:
@aervin I'm planning to add a function to `tsutils` that finds all imports in a file. That will reduce code duplication in tslint rules

aervin:
@ajafff that would help a very great deal. I was referring to what would go in an import rule's test.ts.lint in the above comment. I just found it tedious to make sure I was covering all bases in the rule's tests.

PR #3104, opened 2017-08-07
Prepare release 5.6.0

adidahiya:


adidahiya:
eh, actually going to merge a few PRs before doing this

Issue #2818, opened 2017-05-24
Extend whitespace rule for pipe check


CSchulz:
At the moment it is possible to have inconsistent use of whitespaces around pipe:

```
public statusMessage: string| undefined;
```

I would like to have an extension for the whitespace rule, which allows or disallows space around the pipe.

adidahiya:
accepting PRs for a new rule option for the whitespace rule

ajafff:
Fixed by #3083

PR #2884, opened 2017-06-04
Added "check-type-list" as option to whitespace rule

JoshuaKGoldberg:
It'll now check whitespace around the operators in syntax lists (`|` and `&`).
Fixes #2818.

#### PR checklist

- [x] Addresses an existing issue: #2818
- [x] New feature, bugfix, or enhancement
  - [x] Includes tests
- [x] Documentation update

#### Overview of change:

Also changed references to the const options to use the actual option variable and alphabetized the listings. Figured it'd be a little cleaner this way.
The `maxLength` for options was listed at 7 despite there originally being 8 options (any particular reason?), so I bumped it to 9 for this 9th option.

#### CHANGELOG.md entry:

[enhancement] Added type list checking to `whitespace`

PR #3106, opened 2017-08-08
Whitespace rule enhancement: pipe character support

aervin:
#### PR checklist

- [x] Addresses an existing issue: #[2818](https://github.com/palantir/tslint/issues/2818)
  - [x ] Includes tests
- [ ] Documentation update

#### Overview of change:
Adds support for the bar (or "pipe") character in the whitespace rule.  Users can add "check-pipe" to option list to enable. 

Are additional tests required?


adidahiya:
Sorry, there are already two existing PRs that do this, I would prefer to merge one of those: #2884 or #3106.

"pipe" is ambiguous when we already have a `check-operator` option. I think a better option name is `check-type-operator`.

aervin:
No worries sounds good

PR #3110, opened 2017-08-09
Testing enhancement: run tests for single rule

aervin:
#### PR checklist

  - [ ] Includes tests
- [ ] Documentation update

#### Overview of change:
Allows for running the tests of a specific rule directory. 
Example command: `npm run test:rules whitespace`

Example output:
tslint-test-indiv-rule

Just thought I'd offer this up in case others are interested.

### DIFF
--- a/test/ruleTestRunner.ts
+++ b/test/ruleTestRunner.ts
@@ -23,9 +23,34 @@ import { consoleTestResultHandler, runTest } from "../src/test";
 /* tslint:disable:no-console */
 console.log();
 console.log(colors.underline("Testing Lint Rules:"));
-/* tslint:enable:no-console */
 
 const testDirectories = glob.sync("test/rules/**/tslint.json").map(path.dirname);
+const rule: string = process.argv[2];
+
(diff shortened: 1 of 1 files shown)


adidahiya:
the TSLint CLI actually supports this already. See Tips & Tricks here: https://palantir.github.io/tslint/develop/testing-rules/

does that work for you / is it ergonomic enough?

aervin:
That's way better than what I've come up with lol. Just overlooked this part of the docs is all. Thanks!

aervin:
@adidahiya there was one nice perk of this setup I just remembered: the ability to pass an argument via launch.json for realtime debugging of specific rules in vscode. Maybe it's just on my machine, but rules like "whitespace" trigger additional tests for unrelated rules. This feature might already be possible w existing setup, but I'm unsure. Live debugging in vscode is very helpful to me personally, but don't know if warrants a change to the ruleTestRunner...

aervin:
can pass the relative path just as easily! sorry and thanks again

Issue #3063, opened 2017-07-21
quotemark config option ordering matters?


alexwalter88:
### Bug Report

- __TSLint version__: 5.5.0
- __TypeScript version__: 2.4.2
- __Running TSLint via__: tslint-language-service in VSCode

#### TypeScript code being linted

```ts
let k = "blah";
```

with `tslint.json` configuration:

```json
{
  "rules": {
    "quotemark": [
      true,
      "avoid-template",
      "single"
    ]
  }
}
```

#### Actual behavior
no error
#### Expected behavior
`" should be ' (quotemark)` error

If I put `"single"` after `true` in the config array, then it works. I tested rearranging the other possible config options like `"jsx-double"`, and those all seem to work regardless of order, but as soon as `"single"` is not the first config option, the entire rule stops working.

Is this by design or a bug? I couldn't see any documentation on the ordering of config options mattering. Could this be affecting other rules too? I ordered all of my config options alphabetically because I assumed order wouldn't matter.

alexwalter88:
Welp found the place in the code is explicitly checking the first value of the config options: https://github.com/palantir/tslint/blob/master/src/rules/quotemarkRule.ts#L78-L85

I would have seen the error too if I was not using tslint-language-service, but Typescript doesn't yet support logging from plugins for end users. Although, this should probably be mentioned in the rule documentation.

ajafff:
You're right, this rule expects `"single"` or `"double"` to be the first option.
That should at least be documented.

#889 also tracks validating the config to show an error in such cases.

aervin:
Would it be better if we searched through argument list to find OPTION_SINGLE or OPTION_DOUBLE and throw error if no such argument provided?

PR #3114, opened 2017-08-10
quotemark-rule: single/double options can be at any index in arg list

aervin:
- [x] Addresses an existing issue: [#3063](https://github.com/palantir/tslint/issues/3063)
  - [ ] Includes tests
- [ ] Documentation update

Added a couple util functions to quotemark rule. Checks argument list for OPTION_SINGLE or OPTION_DOUBLE in such a way that argument order doesn't matter. Currently, isEnabled() will return false if no single/double option is specified. Maybe defaulting to double quotes makes more sense?

### DIFF
--- a/src/rules/quotemarkRule.ts
+++ b/src/rules/quotemarkRule.ts
@@ -1,6 +1,6 @@
 /**
  * @license
- * Copyright 2017 Palantir Technologies, Inc.
+ * Copyright 2013 Palantir Technologies, Inc.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -72,10 +72,6 @@ export class Rule extends Lint.Rules.AbstractRule {
         return `${actual} should be ${expected}`;
     }
 
-    public isEnabled(): boolean {
-        return super.isEnabled() && hasSingleDoublePreference(this.ruleArguments);
-    }
-
     public apply(sourceFile: ts.SourceFile): Lint.RuleFailure[] {
         const args = this.ruleArguments;
         const quoteMark = getQuotemarkPreference(args) === OPTION_SINGLE ? "'" : '"';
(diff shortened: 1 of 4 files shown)


src/rules/quotemarkRule.ts
```diff
@@ -141,3 +133,21 @@ function walk(ctx: Lint.WalkContext<Options>) {
         ts.forEachChild(node, cb);
     });
 }
+
+function hasSingleDoublePreference(args: any[]): boolean {
+    for (const arg of args) {
```
ajafff: you can simply `return getQuotemarkPreference(args) !== undefined;`

src/rules/quotemarkRule.ts
```diff
@@ -75,18 +73,12 @@ export class Rule extends Lint.Rules.AbstractRule {
     }
 
     public isEnabled(): boolean {
-        return super.isEnabled() && (this.ruleArguments[0] === OPTION_SINGLE || this.ruleArguments[0] === OPTION_DOUBLE);
+        return super.isEnabled() && hasSingleDoublePreference(this.ruleArguments);
```
ajafff: If we remove this method, the rule will default to double quotes? I think that's more reasonable than just doing nothing.

src/rules/quotemarkRule.ts
```diff
@@ -75,18 +73,12 @@ export class Rule extends Lint.Rules.AbstractRule {
...
     }
 
     public apply(sourceFile: ts.SourceFile): Lint.RuleFailure[] {
         const args = this.ruleArguments;
-        if (args.length > 0) {
-            if (args[0] !== OPTION_SINGLE && args[0] !== OPTION_DOUBLE) {
-                showWarningOnce(`Warning: First argument to 'quotemark' rule should be "${OPTION_SINGLE}" or "${OPTION_DOUBLE}"`);
```
ajafff: It's funny that this code would show the correct warning, but was never executed because the rule was disabled.

src/rules/quotemarkRule.ts
```diff
@@ -141,3 +133,21 @@ function walk(ctx: Lint.WalkContext<Options>) {
...
+            return true;
+        }
+    }
+    return false;
+}
+
+function getQuotemarkPreference(args: any[]): string | undefined {
+    for (const arg of args) {
+        if (arg === OPTION_SINGLE || arg === OPTION_DOUBLE) {
+            return (arg === OPTION_SINGLE) ? OPTION_SINGLE : OPTION_DOUBLE;
```
ajafff: just `return arg;`

src/rules/quotemarkRule.ts
```diff
@@ -1,6 +1,6 @@
 /**
  * @license
- * Copyright 2013 Palantir Technologies, Inc.
+ * Copyright 2017 Palantir Technologies, Inc.
```
ajafff: I think this one should stay unchanged

src/rules/quotemarkRule.ts
aervin: I totally agree and yes removing that call will set default to double quotes.

Issue #3111, opened 2017-08-10
Lint against static-only classes


lfairy:
Users who come from a Java-style OO language may wrap their utility functions in an extra class, instead of putting them at the top level. In this case, they probably should instead use inner namespaces or remove the nesting altogether.

I propose a `no-static-only-class` rule that warns on

* a class
* with only static members
* which does not:
    * extend another class or implement an interface, or
    * have a non-empty constructor

ajafff:
Fixed by #3119

PR #3119, opened 2017-08-11
new-rule: no-unnecessary-class

aervin:
#### PR checklist

- [x] Addresses an existing rule suggestion: #[3111](https://github.com/palantir/tslint/issues/3111)
  - [x] Includes tests
- [ ] Documentation update

#### Overview of change:
Adds the "no-static-only-classes" rule as outlined in the suggestion post. 

This rule ignores a class which:
- extends an existing class
- implements an interface
- has anything other than whitespace in its constructor

- Does this rule need to take constructor parameters into account? 
- Still learning the different APIs--is there a better way to detect empty block scopes (see isEmptyConstructor function)? 
- Help w/ description/rationale


src/verify/lines.ts
```diff
@@ -15,7 +15,9 @@
  */
 
 // Use classes here instead of interfaces because we want runtime type data
+/* tslint:disable:no-static-only-classes */
 export class Line { }
+/* tslint:enable:no-static-only-classes */
```
lfairy: I would prefer that an empty class is *not* counted as a static-only class, since IMO that doesn't fit with the spirit of the lint.

test/rules/no-static-only-classes/test.ts.lint
```diff
@@ -0,0 +1,60 @@
...
+    constructor() {}
+}
+
+export class PascalClass {
+    public static helper(): void {}
+    private Helper(): boolean {
+        return true;
+    }
+    constructor() { this.that = what; }
+}
```
lfairy: I'd like a test for shorthand properties as well, e.g.:

```typescript
export class PascalClass {
    public static helper(): void {}
    constructor(private that: string) {}
}
```

This should not raise a warning because the `private that` declares a non-static property.

src/verify/lines.ts
aervin: You're right I'll add a check for that.

src/rules/noStaticOnlyClassesRule.ts
```diff
@@ -0,0 +1,106 @@
...
+    } else if (classMembers.length === 1 && classMembers[0].kind === ts.SyntaxKind.Constructor) {
+        return isEmptyConstructor(classMembers[0]);
+    } else {
+        return false;
+    }
+}
+
+function isEmptyConstructor(member: ts.ClassElement): boolean {
+    if (member.kind === ts.SyntaxKind.Constructor
+        && isFunctionWithBody(member)
```
ajafff: this line is not needed because you already check if it has a body on the next line

src/rules/noStaticOnlyClassesRule.ts
```diff
@@ -0,0 +1,106 @@
...
+        && member.body !== undefined) {
```
ajafff: Add `&& !member.parameters.some(isParameterProperty)` to exit if the constructor has parameter properties

src/rules/noStaticOnlyClassesRule.ts
```diff
@@ -0,0 +1,106 @@
...
+        return true;
+    } else if (classMembers.length === 1 && classMembers[0].kind === ts.SyntaxKind.Constructor) {
+        return isEmptyConstructor(classMembers[0]);
+    } else {
+        return false;
+    }
+}
+
+function isEmptyConstructor(member: ts.ClassElement): boolean {
+    if (member.kind === ts.SyntaxKind.Constructor
```
ajafff: use `isConstructorDeclaration(member)` instead

src/rules/noStaticOnlyClassesRule.ts
```diff
@@ -0,0 +1,106 @@
...
+        && isFunctionWithBody(member)
+        && member.body !== undefined) {
+        return member.body.getFullText().trim().replace(/\s+/g, "") === "{}";
```
ajafff:
```ts
return member.body.statements.length === 0
```

aervin:
Thanks for all the feedback! I'll get started later today.

src/rules/noStaticOnlyClassesRule.ts
```diff
@@ -0,0 +1,106 @@
...
+            if (isClassDeclaration(statement)
+                && !hasExtendsClause(statement)
+                && !hasImplementsClause(statement)
+                && !isEmptyClass(statement)) {
+                for (const member of statement.members) {
+                    if (!hasStaticModifier(member.modifiers) && !isEmptyConstructor(member)) {
+                        return;
+                    }
+                }
+                if (statement.name !== undefined) {
```
aervin: I didn't know this is a possibility. What should the failure message look like in this scenario? I've been underlining the class name to indicate a problem with the class itself.

src/rules/noStaticOnlyClassesRule.ts
```diff
@@ -0,0 +1,94 @@
...
+
+function hasExtendsClause(statement: ts.ClassDeclaration): boolean {
+    return (statement.heritageClauses !== undefined) && (statement.heritageClauses[0].token === ts.SyntaxKind.ExtendsKeyword);
+}
+
+function isEmptyClass(statement: ts.ClassDeclaration): boolean {
+    const classMembers = statement.members;
+    if (classMembers.length === 0) {
+        return true;
+    } else if (classMembers.length === 1 && classMembers[0].kind === ts.SyntaxKind.Constructor) {
```
ajafff: why do you need to check if the class has an empty constructor?
you can simplify this whole function body to `return statement.members === 0;` without changing the semantics of the rule.

src/rules/noStaticOnlyClassesRule.ts
```diff
@@ -0,0 +1,94 @@
...
+                    this.addFailure(node.name.pos + 1, node.name.end, Rule.FAILURE_STRING);
+                }
+            }
+            return ts.forEachChild(node, checkIfStaticOnlyClass);
+        };
+        ts.forEachChild(sourceFile, checkIfStaticOnlyClass);
+    }
+}
+
+function hasExtendsClause(statement: ts.ClassDeclaration): boolean {
```
ajafff: Can you please rename the parameter `statement` to something like `declaration`. Same for the function below.

src/rules/noStaticOnlyClassesRule.ts
```diff
@@ -0,0 +1,94 @@
...
+        const checkIfStaticOnlyClass = (node: ts.Node): void => {
+            if (isClassDeclaration(node)
+                && !hasExtendsClause(node)
+                && !isEmptyClass(node)) {
+                for (const member of node.members) {
+                    if (!hasModifier(member.modifiers, ts.SyntaxKind.StaticKeyword) && !isEmptyConstructor(member)) {
+                        return;
+                    }
+                }
+                if (node.name !== undefined) {
```
ajafff: If you show the error on the `class` keyword instead of the class' name, you don't need this condition.

```ts
this.addFailureAtNode(getChildOfKind(node, ts.SyntaxKind.ClassKeyword, this.sourceFile)!, Rule.FAILURE_STRING);
```

src/rules/noStaticOnlyClassesRule.ts
```diff
@@ -0,0 +1,83 @@
...
+
+function hasExtendsClause(declaration: ts.ClassDeclaration): boolean {
+    return (declaration.heritageClauses !== undefined) && (declaration.heritageClauses[0].token === ts.SyntaxKind.ExtendsKeyword);
+}
+
+/**
+ * An "empty class" for our purposes.
+ */
+function isEmptyClass(declaration: ts.ClassDeclaration): boolean {
+    return declaration.members.length === 0 || declaration.members.every(isConstructorDeclaration);
```
aervin: `declaration.members.every(isConstructorDeclaration)` because possibility of overloaded constructor? Would like to know if this check is necessary...

ajafff: That check makes sense to correctly handle constructor overloads.
On a side note, you could remove the length check here because `Array.prototype.every` returns `true` for empty arrays.

aervin: Your reviews always make me go "oh right... duh!" lol Taking care of this now. However, I only see one suggestion rather than two?

ajafff: You're, I forgot to save the last comment before submitting :unamused:

src/rules/noStaticOnlyClassesRule.ts
```diff
@@ -0,0 +1,83 @@
...
+}
+
+class NoStaticOnlyClassesWalker extends Lint.AbstractWalker<string[]> {
+    public walk(sourceFile: ts.SourceFile) {
+        const checkIfStaticOnlyClass = (node: ts.Node): void => {
+            if (isClassDeclaration(node) && !isEmptyClass(node) && !hasExtendsClause(node)) {
+                for (const member of node.members) {
+                    if (isConstructorWithShorthandProps(member) ||
+                       (!isConstructorDeclaration(member) && !hasModifier(member.modifiers, ts.SyntaxKind.StaticKeyword))) {
+                        return;
```
ajafff: This `return` will cause nested classes to not be checked. Although this might not occur often in real world code, I think it makes sense to also find this error:

```ts
class Foo {
    public prop = 1;
    constructor() {
        class Bar {
        ~~~~~ [lorem ipsum ... static class]
            static PROP = 2;
        }
    }
}
```

aervin: Working on a fix but could use your input:

Currently this code is passing in the test file:
`
class Foo {
  constructor() {
    class Bar {
      private a: SomeType;
      static PROP = 2;
    }
  }
}
`

Personally, I would not want `class Foo` in my codebase, but should it throw an error?

aervin: bad example-- what if `private a` were `public a`?

ajafff: I'd say as soon as there is an instance property or method, the rule should not complain. Therefore `class Bar` in your example is allowed.
`class Foo` on the other hand is pretty useless without any property or method. But it's out of scope for this rule (at least with the current name).

But you could just rename the rule to `no-unnecessary-class` (or similar) and check for
* only static members (as it currently does) -> could be converted to a namespace or plain object
* constructor only classes -> could be converted to a function
* a mix of the above
* completely empty classes? - maybe add an option to ignore them

aervin: I'd say empty classes are the last priority of this rule. It should never complain about an empty class. I think we both agree that is a separate rule (if anything). 

I really like the idea of `no-unnecessary-class`, but user-config would need some discussion? We could start with a sensible default setting for the rule. I don't see much point in a class with static members and constructors only... Is there a real-life coding scenario where this is useful? I'd be interested to learn.

lfairy: > I don't see much point in a class with static members and constructors only... Is there a real-life coding scenario where this is useful?

It can be useful for writing compilers, where classes simulate discriminated unions (much like "case classes" in Scala). For example:

```typescript
class While extends Statement {
    constructor(public condition: Expression, public body: Statement[]) {}
}
class Break extends Statement {}
class Continue extends Statement {}
```

Then you'd check if a statement is a `break` using `stmt instanceof Break`.

aervin: Well that makes a lot of sense @lfairy. Have you looked over the test cases for this rule? What do you make of this rule's scope?

src/rules/noUnnecessaryClassRule.ts
```diff
@@ -0,0 +1,126 @@
...
+                    return;
+                }
+            }
+            return ts.forEachChild(node, checkIfStaticOnlyClass);
+        };
+        ts.forEachChild(sourceFile, checkIfStaticOnlyClass);
+    }
+
+    private hasOption(option: string): boolean {
+        return this.options.indexOf(option) !== -1;
```
ajafff: Consider parsing the options upfront and passing an object to the constructor of your Walker.

src/rules/noUnnecessaryClassRule.ts
```diff
@@ -0,0 +1,126 @@
...
+const OPTION__ALLOW_STATIC_ONLY = "allow-static-only";
+
+class NoUnnecessaryClassWalker extends Lint.AbstractWalker<string[]> {
+    public walk(sourceFile: ts.SourceFile) {
+        const checkIfStaticOnlyClass = (node: ts.Node): void => {
+            if (isClassDeclaration(node) && !hasExtendsClause(node)) {
+                if (node.members.length === 0 && !this.hasOption(OPTION__ALLOW_EMPTY_CLASS)) {
+                    this.addFailureAtNode(getChildOfKind(node, ts.SyntaxKind.ClassKeyword)!, Rule.FAILURE_EMPTY_CLASS);
+                    return;
+                } else if (node.members.length == 0) {
```
ajafff: use a nested if statement above instead of duplicating parts of the condition in else-if

src/rules/noUnnecessaryClassRule.ts
```diff
@@ -0,0 +1,126 @@
...
+
+const OPTION__ALLOW_CONSTRUCTOR_ONLY = "allow-constructor-only";
+const OPTION__ALLOW_EMPTY_CLASS = "allow-empty-class";
+const OPTION__ALLOW_STATIC_ONLY = "allow-static-only";
+
+class NoUnnecessaryClassWalker extends Lint.AbstractWalker<string[]> {
+    public walk(sourceFile: ts.SourceFile) {
+        const checkIfStaticOnlyClass = (node: ts.Node): void => {
+            if (isClassDeclaration(node) && !hasExtendsClause(node)) {
+                if (node.members.length === 0 && !this.hasOption(OPTION__ALLOW_EMPTY_CLASS)) {
```
ajafff: Consider moving the remainder of this function to a separate method for readability.

src/rules/noUnnecessaryClassRule.ts
```diff
@@ -0,0 +1,126 @@
...
+                    this.addFailureAtNode(getChildOfKind(node, ts.SyntaxKind.ClassKeyword)!, Rule.FAILURE_EMPTY_CLASS);
+                    return;
+                } else if (node.members.length == 0) {
+                    return;
+                }
+
+                if (node.members.some(isConstructorWithClassDeclaration)) {
```
ajafff: I'd prefer not to have this special case ignored. If there is only a constructor that contains a class declaration, the class is still constructor only.

src/rules/noUnnecessaryClassRule.ts
```diff
@@ -0,0 +1,126 @@
...
+                    return ts.forEachChild(node, checkIfStaticOnlyClass);
+                }
+
+                const allMembersAreConstructors = node.members.every(isConstructorDeclaration);
+
+                if (allMembersAreConstructors && !this.hasOption(OPTION__ALLOW_CONSTRUCTOR_ONLY)) {
```
ajafff: You need to check if the constructor has parameter properties, either by default or add an option for that.

aervin:
Thanks @ajafff! I'll try to get to these changes later today.

aervin:
Hey @ajafff, I think I've addressed all change requests besides the additional logic-wrapping function for readability. I'm happy to fix that, but wanted to make sure the rule was implementing the functionality you imagined first.

src/rules/noUnnecessaryClassRule.ts
```diff
@@ -0,0 +1,149 @@
...
+
+const OPTION__ALLOW_CONSTRUCTOR_ONLY = "allow-constructor-only";
+const OPTION__ALLOW_EMPTY_CLASS = "allow-empty-class";
+const OPTION__ALLOW_STATIC_ONLY = "allow-static-only";
+
+class NoUnnecessaryClassWalker extends Lint.AbstractWalker<string[]> {
+    public walk(sourceFile: ts.SourceFile) {
+        const checkIfUnnecessaryClass = (node: ts.Node): void => {
+            if (isClassDeclaration(node) && !hasExtendsClause(node)) {
+                return this.checkMembers(node, checkIfUnnecessaryClass);
```
ajafff: don't return here

src/rules/noUnnecessaryClassRule.ts
```diff
@@ -0,0 +1,149 @@
...
+            !node.members.some(isNonStaticMember)
+        ) {
+            this.addFailureAtNode(
+                getChildOfKind(node, ts.SyntaxKind.ClassKeyword, this.sourceFile)!,
+                Rule.FAILURE_STATIC_ONLY,
+            );
+        }
+
+        /* Check for classes nested in constructors */
+        for (const member of node.members) {
```
ajafff: you can remove this block if you remove the return at the call site (see my other comment above)

src/rules/noUnnecessaryClassRule.ts
```diff
@@ -0,0 +1,149 @@
...
+        typescriptOnly: false,
+    };
+    /* tslint:enable:object-literal-sort-keys */
+
+    public static FAILURE_CONSTRUCTOR_ONLY = "Every member of this class is a constructor. Use functions instead.";
+    public static FAILURE_STATIC_ONLY = "Every member of this class is static. Use namespaces or plain objects instead.";
+    public static FAILURE_EMPTY_CLASS = "This class has no members.";
+
+    public apply(sourceFile: ts.SourceFile): Lint.RuleFailure[] {
+        return this.applyWithWalker(new NoUnnecessaryClassWalker(sourceFile, this.ruleName, this.ruleArguments));
```
ajafff: we typically parse the options before passing them to the Walker.
I mean something like this:
```ts
new NoUnnecessaryClassWalker(sourceFile, this.ruleName, {
    allowConstructorOnly: this.ruleArguments.indexOf(OPTION__ALLOW_CONSTRUCTOR_ONLY) !== -1,
    allowEmptyClass: this.ruleArguments.indexOf(OPTION__ALLOW_EMPTY_CLASS) !== -1,
    allowStaticOnly: this.ruleArguments.indexOf(OPTION__ALLOW_STATIC_ONLY) !== -1,
})
```

test/rules/no-unnecessary-class/allow-constructor-only/test.ts.lint
```diff
@@ -0,0 +1,156 @@
...
+            }
+        }
+    }
+}
+
+class Foo {
+    constructor() {
+        class Bar {
+            public static helper(): void {}
+            private private Helper(): boolean {
```
ajafff: double `private`

aervin:
@ajafff I think the description/rationale still needs to be updated. Do you have ideas?

ajafff:
The rationale is still correct.
The description needs to be amended. I'm not the best advisor for such texts. Maybe: "Disallows classes that are not strictly necessary."

adidahiya:
thanks @aervin!

PR #3120, opened 2017-08-11
Refine no-unnecessary-type-assertion rule to not flag necessary assertions

calebegg:
Type assertions prevent a type from being widened in a couple of
situations, such as in an object literal or the inferred return type of
a function.

### DIFF
--- a/src/rules/noUnnecessaryTypeAssertionRule.ts
+++ b/src/rules/noUnnecessaryTypeAssertionRule.ts
@@ -15,6 +15,7 @@
  * limitations under the License.
  */
 
+import { isObjectFlagSet, isTypeFlagSet } from "tsutils";
 import * as ts from "typescript";
 import * as Lint from "../index";
 
@@ -66,9 +67,9 @@ class Walker extends Lint.AbstractWalker<void> {
         }
 
         if (node.kind !== ts.SyntaxKind.NonNullExpression &&
-            ((castType.flags & ts.TypeFlags.Literal) ||
-                (castType.flags & ts.TypeFlags.Object &&
-                (castType as ts.ObjectType).objectFlags & ts.ObjectFlags.Tuple))) {
+            (isTypeFlagSet(castType, ts.TypeFlags.Literal) ||
+                isTypeFlagSet(castType, ts.TypeFlags.Object) &&
+                isObjectFlagSet(castType as ts.ObjectType, ts.ObjectFlags.Tuple))) {
             // It's not always safe to remove a cast to a literal type or tuple
             // type, as those types are sometimes widened without the cast.
             return;
--- a/src/rules/noUnnecessaryTypeAssertionRule.ts
+++ b/src/rules/noUnnecessaryTypeAssertionRule.ts
@@ -65,6 +65,15 @@ class Walker extends Lint.AbstractWalker<void> {
             return;
         }
 
+        if (node.kind !== ts.SyntaxKind.NonNullExpression &&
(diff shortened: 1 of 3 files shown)


calebegg:
I think the test failure is unrelated; looks like a timeout.

test/rules/no-unnecessary-type-assertion/test.ts.fix
```diff
@@ -69,3 +69,13 @@ function func(aOrB: TypeA|TypeB) {
         let z = aOrB;
     }
 }
+
+// Expecting no warning for these assertions as they are not unnecessary.
+
+type Bar = 'bar';
+const data = {
+    x: 'foo' as 'foo',
+    y: 'bar' as Bar,
```
ajafff: I don't get why you need this behavior. A type annotation would also work

```ts
const data = {x: 'foo', y: Bar} = {x: 'foo', y: bar}
```

Same goes for the arrow function below: just add the return type:
```ts
[1, 2, 3, 4, 5].map((x): [number, string] => [x, 'A' + x]);
```


I guess it's useful if the assertion is only a little part of a bigger object where you don't want to write the whole type?

calebegg: It's a relatively common pattern in our very large codebase where we're trying to turn on this rule. I don't think there's anything so wrong with the pattern that it would make sense to ban it. It keeps you from having to repeat the keys, at least.

Skimming through my list of breakages, I also see:

```typescript
const FOO = {
  bar: 'a' as 'a',
  baz: 'stuff',
  // ... many more properties without 'as'
```

i.e., not all properties are cast like this. For those I would hate to have to make it so much more verbose.

I also see instances of:

```typescript
for (const foo of ['bar' as 'bar', 'baz' as 'baz']) { /* ... */ }
```

Also, a search turned up a match in tslint itself:
https://github.com/palantir/tslint/blob/master/src/linter.ts#L110

src/rules/noUnnecessaryTypeAssertionRule.ts
```diff
@@ -65,6 +66,15 @@ class Walker extends Lint.AbstractWalker<void> {
             return;
         }
 
+        if (node.kind !== ts.SyntaxKind.NonNullExpression &&
+            (isTypeFlagSet(castType, ts.TypeFlags.Literal) ||
+                isTypeFlagSet(castType, ts.TypeFlags.Object) &&
```
ajafff: you can use `isObjectType(castType)` to avoid the assertion in the next line

test/rules/no-unnecessary-type-assertion/test.ts.fix
ajafff: That makes sense, thank you for the detailed explanation

src/rules/noUnnecessaryTypeAssertionRule.ts
calebegg: Done, thanks!

ajafff:
Related TypeScript PR that *would* make this change unnecessary: https://github.com/Microsoft/TypeScript/pull/17785

Let's see if it lands.

calebegg:
That won't solve all of the cases, plus it appears the goal is to put it behind a flag. I'd prefer to land this and adjust if necessary after that lands + is in a current release of TS + that flag is included in `--strict`. How does that sound?

ajafff:
> I'd prefer to land this and adjust if necessary after that lands + is in a current release of TS

That's what I was planning to do. I just wanted to leave this open for a few days, so others can have a look, too.

PR #3129, opened 2017-08-14
no-submodule-imports: consider exact matches as whitelisted

ajafff:
#### PR checklist

  - [x] Includes tests
- [x] Documentation update

#### Overview of change:
Today I wanted to use the new `no-submodule-imports` rule. I need to whitelist `@angular/core/testing` for example, but doing that will check if the import starts with `@angular/core/testing/` (note the trailing slash).
This PR also tests for exact matches, allowing me to whitelist `@angular/core/testing` instead of `@angular/core`

#### CHANGELOG.md entry:

[enhancement] `no-submodule-imports` allows whitelisting of submodules like `@angular/core/testing`

### DIFF
--- a/src/rules/noSubmoduleImportsRule.ts
+++ b/src/rules/noSubmoduleImportsRule.ts
@@ -36,15 +36,14 @@ export class Rule extends Lint.Rules.AbstractRule {
             Submodules of some packages are treated as private APIs and the import
             paths may change without deprecation periods. It's best to stick with
             top-level package exports.`,
-        optionsDescription: "A list of packages whose submodules are whitelisted.",
+        optionsDescription: "A list of whitelisted package or submodule names.",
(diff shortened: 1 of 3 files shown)


src/rules/noSubmoduleImportsRule.ts
```diff
@@ -102,10 +99,6 @@ class NoSubmoduleImportsWalker extends Lint.AbstractWalker<string[]> {
     }
 }
 
-function isAbsoluteOrRelativePath(path: string): boolean {
-    return /^(..?(\/|$)|\/)/.test(path);
```
ajafff: This one was actually wrong. it needed `\.` instead of just `.`
Anyways, we now use a typescript function which also check for `C:\foo\bar` as of typescript@2.5.0

PR #3130, opened 2017-08-14
Fix lint error in no-invalid-template-strings

ajafff:
[no-log] removed trailing whitespace


ajafff:
Fix was included in #3132

PR #3135, opened 2017-08-15
no-unnecessary-qualifier: better fix for ts@2.5

ajafff:
#### PR checklist

  - [ ] Includes tests
- [ ] Documentation update

#### Overview of change:
#3052 made `no-unnecessary-qualifier` work with upcoming changes in typescript@2.5
In the meantime typescript added a new API to do this in a more correct way: https://github.com/Microsoft/TypeScript/pull/17457
/cc @andy-ms 

#### CHANGELOG.md entry:

[no-log]


ajafff:
@andy-ms That new API never made it into the 2.5 release. I had to add back the workaround of comparing the declarations array to make it work with 2.5.1

And now we probably have to keep it around until we drop support for typescript@2.5.x

PR #3139, opened 2017-08-18
refactor newlineBeforeReturn to support more variants

CSchulz:
#### PR checklist

  - [x] Includes tests
- [x] Documentation update

#### Overview of change:

Rename newlineBeforeReturn to newline and add some more options for adding newlines.


ajafff:
I agree that it makes sense to rename and enhance the rule instead of adding yet another whitespace related rule. Nevertheless we need to keep `newline-before-return` around until the next major version to avoid a breaking change.
But that's pretty easy. Just export the walker of your new rule and use it from `newline-before-return` with the configuration to only require a blank line before return statements.
Also add a deprecation message to the metadata of `newline-before-return`

src/rules/newlineRule.ts
```diff
@@ -0,0 +1,150 @@
...
+import * as Lint from "../index";
+ 
+enum OptionState {
+    always = "always",
+    ignore = "ignore",
+    never = "never",
+}
+ 
+const ALWAYS_IGNORE_OR_NEVER = {
+    enum: OptionState,
```
ajafff: this is only used in the metadata? I don't think this works with string enums. It actually expects a string array here.

src/rules/newlineRule.ts
```diff
@@ -0,0 +1,150 @@
...
+type Option = "always" | "ignore" | "never";
+type Options = Partial<Record<OptionName, Option>>;
+ 
+function parseOptions(json: Option | Options | undefined): Options {
+    const options: Options = { };
+    for (const optionName of optionNames) {
+        options[optionName] = typeof json === "object" ? json[optionName] : json === undefined ? "always" : json;
+    }
+    return options;
+}
```
ajafff: I'd suggest the following (note that this defaults to ignore, which might be more intuitive)
```ts
type Options = Record<OptionName, Option>;
function parseOptions(json: Partial<Options> | undefined) {
    const default: Options = {return: "ignore", class: "ignore", functionBlock: "ignore", block: "ignore"};
    return {...default, json};
}
```

src/rules/newlineRule.ts
```diff
@@ -0,0 +1,150 @@
...
+export class Rule extends Lint.Rules.AbstractRule {
+    /* tslint:disable:object-literal-sort-keys */
+    public static metadata: Lint.IRuleMetadata = {
+        ruleName: "newline",
+        description: "Requires or disallows blank lines before return, after class declaration, function block declaration or block.",
+        rationale: "Helps maintain a readable style in your codebase.",
+        optionsDescription: Lint.Utils.dedent`
+            Following arguments may be optionally provided:
+ 
+            * \`"return"\` checks for empty line before return when not the only line in the block.
```
ajafff: s/line/statement/

src/rules/newlineRule.ts
```diff
@@ -0,0 +1,150 @@
...
+            * \`"class"\` checks for empty line after class declaration.
```
ajafff: checks for empty line at the start of a class body

src/rules/newlineRule.ts
```diff
@@ -0,0 +1,150 @@
...
+            * \`"functionBlock"\` checks for empty line after function block declaration.
```
ajafff: at the start of a function block

src/rules/newlineRule.ts
```diff
@@ -0,0 +1,150 @@
...
+            },
+            type: "object",
+        },
+        optionExamples: [[true, {return: "always", class: "never"}]],
+        type: "style",
+        typescriptOnly: false,
+    };
+    /* tslint:enable:object-literal-sort-keys */
+ 
+    public static FAILURE_STRING_FACTORY(kind: boolean, nodeType: string) {
```
ajafff: Use a string enum for `kind` to make it easier to understand

src/rules/newlineRule.ts
```diff
@@ -0,0 +1,150 @@
...
+                if (prev === undefined) {
+                    // return is not within a block (e.g. the only child of an IfStatement) or the first statement of the block
+                    // no need to check for preceding newline
+                    return;
+                }
+ 
+                this.checkForEmptyLine(prev, node, this.options.return == OptionState.always, 'return');
+            }
+ 
+            if (isClassDeclaration(node) && this.options.class != OptionState.ignore) {
```
ajafff: can be an `else if`

src/rules/newlineRule.ts
```diff
@@ -0,0 +1,150 @@
...
+                this.checkForEmptyLine(
+                    getChildOfKind(node, ts.SyntaxKind.OpenBraceToken, this.sourceFile)!, node.members[0],
```
ajafff: you need to check if the class has any members

src/rules/newlineRule.ts
```diff
@@ -0,0 +1,150 @@
...
+        const cb = (node: ts.Node): void => {
+            if (node.kind === ts.SyntaxKind.ReturnStatement && this.options.return != OptionState.ignore) {
+                const prev = getPreviousStatement(node as ts.Statement);
+                if (prev === undefined) {
+                    // return is not within a block (e.g. the only child of an IfStatement) or the first statement of the block
+                    // no need to check for preceding newline
+                    return;
+                }
+ 
+                this.checkForEmptyLine(prev, node, this.options.return == OptionState.always, 'return');
```
ajafff: prefer `===` here and in all other places

src/rules/newlineRule.ts
```diff
@@ -0,0 +1,150 @@
...
+                    getChildOfKind(node, ts.SyntaxKind.OpenBraceToken, this.sourceFile)!, node.members[0],
+                    this.options.class == OptionState.always, 'first member',
+                );
+            }
+ 
+            if (isBlock(node) &&
+                    (isFunctionWithBody(node.parent!) && this.options.functionBlock != OptionState.ignore
+                        || this.options.block != OptionState.ignore)) {
+                this.checkForEmptyLine(
+                    node.getChildAt(0), node.statements[0], this.options.functionBlock == OptionState.always, 'first statement',
```
ajafff: you need two different branchen for `block` and `functionBlock` because you need to check different config options

src/rules/newlineRule.ts
```diff
@@ -0,0 +1,150 @@
...
+                );
+            }
+            return ts.forEachChild(node, cb);
+        };
+        return ts.forEachChild(sourceFile, cb);
+    }
+ 
+    private checkForEmptyLine(node: ts.Node, firstNode: ts.Node, lineRequired: boolean, nodeType: string) {
+        let start = node.end;
```
ajafff: consider replacing the `node: ts.Node` parameter with `start: number` and pass the position directly. That avoids accessing tokens with `getChildAt`

src/rules/newlineRule.ts
ajafff: instead of `getChildOfKind(...)` you could simply pass `node.members.pos` as first parameter. see my other comment in `checkForEmptyLine` below

src/rules/newlineRule.ts
ajafff: instead of `node.getChildAt(...)` you could simply pass `node.statements.pos` as first parameter. see my other comment in `checkForEmptyLine` below

ajafff: you need to check if the block contains any statement

src/rules/newlineRule.ts
```diff
@@ -0,0 +1,150 @@
...
+        let line = ts.getLineAndCharacterOfPosition(this.sourceFile, start).line;
+        let firstNodeStart = firstNode.getStart();
```
ajafff: please provide the `sourceFile` parameter to `getStart()` for performance reasons

test/rules/newline/none/test.tsx.lint
```diff
@@ -1,18 +1,19 @@
...
-
+ 
 <div>{ [].map((child: any) =>
     <span />;
 ) }</div>
-
-
-[0]: Missing blank line before return
+ 
+
```
ajafff: do these trailing whitespaces serve a purpose?

src/rules/newlineRule.ts
```diff
@@ -0,0 +1,144 @@
...
+export class Rule extends Lint.Rules.AbstractRule {
+    /* tslint:disable:object-literal-sort-keys */
+    public static metadata: Lint.IRuleMetadata = {
+        ruleName: "newline",
+        description: "Requires or disallows blank lines before return, after class declaration, function block declaration or block.",
+        rationale: "Helps maintain a readable style in your codebase.",
+        optionsDescription: Lint.Utils.dedent`
+            Following arguments may be optionally provided:
+ 
+            * \`"return"\` checks for empty listane before return when not the only statement in the block.
```
ajafff: s/listane/line/

src/rules/newlineRule.ts
```diff
@@ -0,0 +1,144 @@
...
+            * \`"class"\` checks for empty line at the start of a class body.
+            * \`"functionBlock"\` checks for empty line at the start of a function block.
+            * \`"block"\` checks for empty line after block declaration.`,
```
ajafff: s/after/in the first line of a/

src/rules/newlineRule.ts
```diff
@@ -0,0 +1,144 @@
...
+* limitations under the License.
+*/
+ 
+import { getPreviousStatement, isBlock, isClassDeclaration, isFunctionWithBody } from "tsutils";
+import * as ts from "typescript";
+import * as Lint from "../index";
+
+export enum ViolationType {
+    needed = "needed",
+    needless = "needless",
```
ajafff: prefer "unnecessary"

src/rules/newlineRule.ts
ajafff: as discussed offline: I don't like the option names.
I'd prefer "before-return" here.
I'm also in favor of "start-of-class" to allow adding "before-class" and "after-class" later on.

And to keep naming consistent, you should prefer "function-block" over "functionBlock"

src/rules/newlineRule.ts
```diff
@@ -0,0 +1,144 @@
...
+            type: "object",
+        },
+        optionExamples: [[true, {return: "always", class: "never"}]],
+        type: "style",
+        typescriptOnly: false,
+    };
+    /* tslint:enable:object-literal-sort-keys */
+ 
+    public static FAILURE_STRING_FACTORY(kind: ViolationType, nodeType: string) {
+        const kindMsg = kind === ViolationType.needed ? 'Missing' : 'Unneeded';
```
ajafff: Why don't you use the string value of the enum?

PR #3145, opened 2017-08-20
Normalize the filepath so VS's error list has the correct interactions.

### LINKED ISSUE #2977: msbuild formatter should change  / to \ in paths
### Bug Report

- __TSLint version__: 5.4.3
- __TypeScript version__:2.4.1
- __Running TSLint via__:  Visual Studio

I'm using tslint withing VS Enterprise 2017 with the `msbuild` formatter. Although the errors get reported correctly in the Error List window if you double click on one of them it opens the file but does not go to the line where the error happens. This is because the path are using unix style slashes so they look something like:
`T:/Projects/Test/test.ts`
Which seems to mess up the auto navigate to the error. if the path is changed to:
`T:\Projects\Test\test.ts`
The correct line is opened.

WGroenestein:
Resolves #2977

### DIFF
--- a/src/formatters/msbuildFormatter.ts
+++ b/src/formatters/msbuildFormatter.ts
@@ -15,11 +15,11 @@
  * limitations under the License.
  */
 
+import * as path from "path";
+
 import { AbstractFormatter } from "../language/formatter/abstractFormatter";
 import { IFormatterMetadata } from "../language/formatter/formatter";
 import { RuleFailure } from "../language/rule/rule";
-import * as path from "path";
-
 import { camelize, dedent } from "../utils";
 
 export class Formatter extends AbstractFormatter {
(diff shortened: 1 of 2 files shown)


palantirtech:
Thanks for your interest in [palantir/tslint](https://github.com/palantir/tslint), @WGroenestein! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

WGroenestein:
Only tested it in VS on windows

src/formatters/msbuildFormatter.ts
```diff
@@ -18,6 +18,7 @@
 import { AbstractFormatter } from "../language/formatter/abstractFormatter";
 import { IFormatterMetadata } from "../language/formatter/formatter";
 import { RuleFailure } from "../language/rule/rule";
+import * as path from "path";
```
adidahiya: the imports are not sorted so you have a lint error here.

Issue #3127, opened 2017-08-14
Do something against the CircleCI timeouts


ajafff:
Our mocha tests time out quite often on CircleCI. Especially node 0 which tests with node v4.x is prone to fail because it's slower than all other. It's annoying to always restart builds until one eventually succedes.

Some time ago I already increased the timeout for some test cases from 4000 to 8000 ms.
Do we want to increase the timeout further or try [some other CI](https://github.com/ligurio/Continuous-Integration-services/blob/master/continuous-integration-services-list.md)?

adidahiya:
Which tests are slow? We should skip any tests which rely on an extended timeout and try to make them faster. I'd prefer to stay on CircleCI.

ajafff:
I looked through the failed build of the last weeks. We have a winner: It's the test for `tslint --project --fix`.
There are some other IO heavy tests that fail occasionally, but >90% of the time it's the first one in the list:
* Executable --project flag can apply fixes from multiple rules (8000ms)
* Configuration loadConfigurationFromPath with config not relative to tslint extends with package installed relative to tslint (2000ms)
* Executable --project flag works with '--exclude' (4000ms)
* Linter apply fixes to correct files (2000ms)

Possible fix for the first list item: use `runner.ts` directly instead of spawning a new process. That avoid loading all sources and (de)optimizing them again. Related: #2587

PR #3153, opened 2017-08-23
Speed up test for --project --fix

ajafff:
#### PR checklist

- [x] Addresses an existing issue: #3127
  - [ ] Includes tests
- [ ] Documentation update

#### Overview of change:
We don't need to execute and fix that many rules to ensure everything works as expected. By removing `whitespace` from the test's `tslint.json` the execution time is reduced from 4.7s to 2.7s on my computer.
That's enough to not run into a timeout even if CI is really slow.
Fixes: #3127

Rewriting the test to not spawn a new process only reduces the execution time by 0.2s - 0.5s

#### CHANGELOG.md entry:
[no-log]

### DIFF
--- a/test/files/project-multiple-fixes/before.test.ts
+++ b/test/files/project-multiple-fixes/before.test.ts
@@ -5,15 +5,15 @@ const DefaultOptions = {
     new winston.transports.Console({
         json: true,
         colorize: true,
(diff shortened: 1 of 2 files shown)


adidahiya:
Nice, good call.

Issue #3147, opened 2017-08-22
trailingComma is required after rest parameter


thgreasi:
### Bug Report

- __TSLint version__: 5.6.0
- __TypeScript version__: 2.4.0
- __Running TSLint via__: CLI

#### TypeScript code being linted

```ts
export const foo = (
	a: any,
	...rest: any[]
) => {};
```

with `tslint.json` configuration:

```json
{
  "extends": "tslint:recommended",
  "rules": {
    "indent": [ true, "tabs" ],
    "jsdoc-format": true,
    "whitespace": [
      false,
      "check-type"
    ],
    "object-literal-sort-keys": false,
    "only-arrow-functions": [false],
    "quotemark": [true, "single", "avoid-escape"],
    "max-classes-per-file": [false],
    "max-line-length": [180],
    "member-access": false,
    "member-ordering": [false],
    "no-consecutive-blank-lines": false,
    "no-shadowed-variable": false,
    "no-var-requires": true,
    "arrow-parens": false,
    "no-angle-bracket-type-assertion": false,
    "no-console": [false],
    "no-empty-interface": false,
    "no-string-literal": false,
    "object-literal-key-quotes": [true, "as-needed"],
    "interface-name": [false],
    "interface-over-type-literal": false,
    "variable-name": [
      true,
      "ban-keywords",
      "check-format",
      "allow-leading-underscore",
      "allow-pascal-case"
    ],
    "semicolon": [true, "always", "ignore-bound-class-methods"],
    "trailing-comma": [
      true,
      {
        "multiline": {
          "objects": "always",
          "arrays": "always",
          "functions": "always",
          "typeLiterals": "ignore"
        }
      }
    ]
  }
}
```

#### Actual behavior
`Error: x:y trailing-comma  Missing trailing comma` on the line of `...rest: any[]`

#### Expected behavior
It shouldn't error since trailing commas are not allowed after spreaded parameters.
See: https://github.com/prettier/prettier/issues/1079#issuecomment-288693961

ajafff:
> trailing commas are not allowed after spreaded parameters.

Is there an official spec that states this?
[...]

PR #3176, opened 2017-08-28
trailing-comma: add option to forbid trailing comma after rest

ajafff:
#### PR checklist

- [x] Addresses an existing issue: #3147
  - [x] Includes tests
- [x] Documentation update

#### Overview of change:
Add an option `"specCompliant"` to the rule. Enabling that option forbids trailing commas after object and array rest as well as rest parameters. That makes the linted code compatible with the ES spec and the implementation in JS VMs.
Fixes: #3147

Do you know a better option name than `specCompliant`?

#### CHANGELOG.md entry:

[new-rule-option] `trailing-comma` adds option `"specCompliant"` to make it compatible with the ES spec regarding trailing commas after object/array rest and rest parameters.


IllusionMH:
Shouldn't be the problem, but tests can be added for combination of arguments destructuring and regular params later and vice versa
```ts
function combined([a, ...arr], {b, ...obj}, regular) {}
                                                   ~ [Missing trailing comma]

function combined([a, ...arr,], {b, ...obj,}, regular,) {}
                            ~ [Forbidden trailing comma]
                                          ~ [Forbidden trailing comma]
```

ajafff:
@IllusionMH I just added tests as suggested

> Is it worth to consider adding `specCompliant: true` to `tslint:recommended` config at least for next major update?

We should definitely enable it for .js files. And to keep it consistent we should also enable it for .ts files.
Users who want to use trailing commas everywhere in their typescript files, can disable it using `specCompliant: false`.

I'm thinking about renaming the option to something like `"mode": "spec"` and `"mode": "consistent"`. Thoughts?

adidahiya:
What does `"mode": "consistent"` convey? "spec compliant" sounds a little more intuitive to me, but I would suggest making it even more verbose as `esSpecCompliant`

ajafff:
> What does `"mode": "consistent"` convey?

It would enable you to consistently use traling commas everywhere in your TypeScript code. Not a great idea on second thought, since it's inconsistent with how JavaScript works.

Renamed to `"esSpecCompliant"` as suggested.

adidahiya:
master is broken (semantic merge conflict?) but this PR looks gtg

ajafff:
That's not a merge conflict. The typescript@next changed the error message (and detection algorithm) for unused variables.
I'm working on a fix right now.

thgreasi:
Woohoo!

Issue #3190, opened 2017-08-31
object-literal-sort-keys: allow grouping with new lines


Merott:
### Feature Request

The `object-literal-sort-keys` can sometimes get in the way of logically grouping properties of an object. By allowing additional new lines inside the object literal, it makes it possible to logically group properties while keeping the elements within each group in alphabetical order, like this:

```ts
const person = {
  id: 123

  height: 180
  weight: 85

  preferences: {
    ...
  }
}
```

instead of:

```ts
const person = {
  height: 180
  id: 123
  preferences: {
    ...
  }
  weight: 85
}
```

ajafff:
Fixed by #3191

PR #3191, opened 2017-08-31
object-literal-sort-keys: Allow grouping of object properties via additional blank lines

Merott:
#### PR checklist

- [x] Addresses an existing issue: #3190  
  - [x] Includes tests
- [x] Documentation update

#### Overview of change:

Changes the behaviour of `object-literal-sort-keys` when using alphabetical ordering, and allows object keys to be grouped together separated with blank lines.

#### CHANGELOG.md entry:

[enhancement] [`object-literal-sort-keys`](https://palantir.github.io/tslint/rules/object-literal-sort-keys/): allow grouping of object properties via additional blank lines when using alphabetical ordering.

### DIFF
--- a/src/rules/objectLiteralSortKeysRule.ts
+++ b/src/rules/objectLiteralSortKeysRule.ts
@@ -32,14 +32,19 @@ export class Rule extends Lint.Rules.OptionallyTypedRule {
     /* tslint:disable:object-literal-sort-keys */
     public static metadata: Lint.IRuleMetadata = {
         ruleName: "object-literal-sort-keys",
-        description: "Checks ordering of keys in object literals.",
+        description: Lint.Utils.dedent`
+            Checks ordering of keys in object literals.
+
+            When using the default alphabetical ordering, additional blank lines may be used to group
+            object properties together while keeping the elements within each group in alphabetical order.
+        `,
         rationale: "Useful in preventing merge conflicts",
         optionsDescription: Lint.Utils.dedent`
             By default, this rule checks that keys are in alphabetical order.
             The following may optionally be passed:
 
             * "${OPTION_IGNORE_CASE}" will to compare keys in a case insensitive way.
-            * "${OPTION_MATCH_DECLARATION_ORDER} will prefer to use the key ordering of the contextual type of the object literal, as in:
+            * "${OPTION_MATCH_DECLARATION_ORDER}" will prefer to use the key ordering of the contextual type of the object literal, as in:
 
                 interface I { foo: number; bar: number; }
                 const obj: I = { foo: 1, bar: 2 };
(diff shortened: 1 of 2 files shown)


palantirtech:
Thanks for your interest in [palantir/tslint](https://github.com/palantir/tslint), @Merott! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

src/rules/objectLiteralSortKeysRule.ts
```diff
@@ -184,6 +193,10 @@ function walk(ctx: Lint.WalkContext<Options>, checker?: ts.TypeChecker): void {
     }
 }
 
+function hasNewLineBefore(sourceFile: ts.SourceFile, element: ts.ObjectLiteralElement) {
+    return /\r?\n\r?\n/.test(sourceFile.text.slice(element.getFullStart(), element.getStart(sourceFile)));
```
ajafff: This will also match for blank lines inside comments, for example:

```ts
let foo = {
    foo: 1,
    /* I need some space to express myself

       and here the comment continues */
    bar: 2,
};
```
You can have a look at `newline-before-return` to see how I would check for a blank line before, between or after comments (but not inside).
https://github.com/palantir/tslint/blob/master/src/rules/newlineBeforeReturnRule.ts#L62-L81

test/rules/object-literal-sort-keys/allow-newline-grouping/test.ts.lint
```diff
@@ -0,0 +1,41 @@
+var passA = {
+    a: 1,
+    b: 2,
+    d: 4,
+
+    c: 3,
+    e: 5,
+};
```
ajafff: please add one of these test cases to an existing test to check if the new behavior is not enabled without the option

Merott:
Thank you @ajafff - I've pushed up an update based on your feedback.

adidahiya:
This rule change makes sense to me. This would work similar to `ordered-imports` which allows import groups separated by newlines. In fact we might just want to make this the default behavior and not keep it behind a rule option, to be consistent with import statement linting behavior. Thoughts @ajafff @Merott?

Merott:
It would make me personally happy to have this behaviour by default, similar to`ordered-imports` as you mentioned. However, given that this behaviour could be considered a breaking change if it were to be applied by default, I  thought about introducing it as an option instead. 

I should also mention that `ordered-imports` doesn't support grouping with new lines within each import statement. Thet'd be closer in terms  of syntax to what my proposed rule option here does, and is something that I'd also like to have.

ajafff:
Making this the default behavior is not a breaking change. Code that passes now will also pass after this change.
I think this makes sense to be the default.

Merott:
@ajafff OK - I will take out the option and update the PR accordingly.

Merott:
Ready for a re-review 👍🏻.

src/rules/objectLiteralSortKeysRule.ts
```diff
@@ -137,6 +142,10 @@ function walk(ctx: Lint.WalkContext<Options>, checker?: ts.TypeChecker): void {
                     break;
                 case ts.SyntaxKind.ShorthandPropertyAssignment:
                 case ts.SyntaxKind.PropertyAssignment:
+                    if (hasBlankLineBefore(ctx.sourceFile, property)) {
```
ajafff: This can be pretty expensive when checked for each property. I'd prefer to only check this right before adding a failure

Merott: Done.

ajafff:
Thanks @Merott

PR #3195, opened 2017-09-02
Mark inputFilePath as optional in findConfiguration()

lumaxis:
#### PR checklist

  - [ ] Includes tests
- [ ] Documentation update

#### Overview of change:

When `suppliedConfigFilePath` is passed to `findConfigurationPath()`, `inputFilePath` is completely ignored and should therefore be optional.

#### CHANGELOG.md entry:

[bugfix] Correctly mark `inputFilePath` as an optional parameter in `Configuration.findConfiguration()`

### DIFF
--- a/src/configuration.ts
+++ b/src/configuration.ts
@@ -90,8 +90,10 @@ const BUILT_IN_CONFIG = /^tslint:(.*)$/;
  * of the search for a configuration.
  * @returns Load status for a TSLint configuration object
  */
-export function findConfiguration(configFile: string | null, inputFilePath: string): IConfigurationLoadResult {
-    const configPath = findConfigurationPath(configFile, inputFilePath);
+export function findConfiguration(configFile: string | null, inputFilePath: string): IConfigurationLoadResult;
+export function findConfiguration(configFile: string, inputFilePath?: string): IConfigurationLoadResult;
+export function findConfiguration(configFile: string | null, inputFilePath?: string): IConfigurationLoadResult {
+    const configPath = findConfigurationPath(configFile, inputFilePath!);
     const loadResult: IConfigurationLoadResult = { path: configPath };
 
     try {
@@ -112,7 +114,9 @@ export function findConfiguration(configFile: string | null, inputFilePath: stri
  * @returns An absolute path to a tslint.json file
  * or undefined if neither can be found.
  */
-export function findConfigurationPath(suppliedConfigFilePath: string | null, inputFilePath: string) {
+export function findConfigurationPath(suppliedConfigFilePath: string | null, inputFilePath: string): string | undefined;
+export function findConfigurationPath(suppliedConfigFilePath: string, inputFilePath?: string): string | undefined;
+export function findConfigurationPath(suppliedConfigFilePath: string | null, inputFilePath?: string): string | undefined {
     if (suppliedConfigFilePath != null) {
         if (!fs.existsSync(suppliedConfigFilePath)) {
             throw new FatalError(`Could not find config file at: ${path.resolve(suppliedConfigFilePath)}`);
@@ -123,7 +127,7 @@ export function findConfigurationPath(suppliedConfigFilePath: string | null, inp
         // convert to dir if it's a file or doesn't exist
         let useDirName = false;
         try {
-            const stats = fs.statSync(inputFilePath);
+            const stats = fs.statSync(inputFilePath!);
             if (stats.isFile()) {
                 useDirName = true;
             }
@@ -132,11 +136,11 @@ export function findConfigurationPath(suppliedConfigFilePath: string | null, inp
             useDirName = true;
         }
         if (useDirName) {
-            inputFilePath = path.dirname(inputFilePath);
+            inputFilePath = path.dirname(inputFilePath!);
         }
 
         // search for tslint.json from input file location
-        let configFilePath = findup(CONFIG_FILENAME, inputFilePath);
+        let configFilePath = findup(CONFIG_FILENAME, inputFilePath!);
         if (configFilePath !== undefined) {
             return path.resolve(configFilePath);
         }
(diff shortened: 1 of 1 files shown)


palantirtech:
Thanks for your interest in [palantir/tslint](https://github.com/palantir/tslint), @lumaxis! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

src/configuration.ts
```diff
@@ -90,7 +90,7 @@ const BUILT_IN_CONFIG = /^tslint:(.*)$/;
  * of the search for a configuration.
  * @returns Load status for a TSLint configuration object
  */
-export function findConfiguration(configFile: string | null, inputFilePath: string): IConfigurationLoadResult {
+export function findConfiguration(configFile: string | null, inputFilePath?: string): IConfigurationLoadResult {
```
ajafff: to prevent call like `findConfiguration(null, undefined);` this function needs overloads. just add these lines above:

```ts
export function findConfiguration(configFile: string | null, inputFilePath: string): IConfigurationLoadResult;
export function findConfiguration(configFile: string, inputFilePath?: string): IConfigurationLoadResult
```

src/configuration.ts
```diff
@@ -112,14 +112,14 @@ export function findConfiguration(configFile: string | null, inputFilePath: stri
  * @returns An absolute path to a tslint.json file
  * or undefined if neither can be found.
  */
-export function findConfigurationPath(suppliedConfigFilePath: string | null, inputFilePath: string) {
+export function findConfigurationPath(suppliedConfigFilePath: string | null, inputFilePath?: string) {
```
ajafff: same as above

src/configuration.ts
```diff
@@ -112,14 +112,14 @@ export function findConfiguration(configFile: string | null, inputFilePath: stri
...
     if (suppliedConfigFilePath != null) {
         if (!fs.existsSync(suppliedConfigFilePath)) {
             throw new FatalError(`Could not find config file at: ${path.resolve(suppliedConfigFilePath)}`);
         } else {
             return path.resolve(suppliedConfigFilePath);
         }
-    } else {
+    } else if (inputFilePath) {
```
ajafff: `inputFilePath !== undefined` to satisfy the linter

src/configuration.ts
lumaxis: Good idea! For some reason I didn't even think of overloads here 🙂

src/configuration.ts
lumaxis: @ajafff I tried adding that here too but then I run into a strange compiler error that I can't really make sense of:

```typescript
src/configuration.ts(95,58): error TS2345: Argument of type 'string | undefined' is not assignable to parameter of type 'string'.
  Type 'undefined' is not assignable to type 'string'.
error Command failed with exit code 2.
```

From what I understand, this _should_ work but the TS compiler is ignoring the other overload? Not sure. Any ideas? Thanks!

src/configuration.ts
lumaxis: Absolutely. Sorry I missed this, thanks!

src/configuration.ts
```diff
@@ -90,7 +90,8 @@ const BUILT_IN_CONFIG = /^tslint:(.*)$/;
  * of the search for a configuration.
  * @returns Load status for a TSLint configuration object
  */
-export function findConfiguration(configFile: string | null, inputFilePath: string): IConfigurationLoadResult {
+export function findConfiguration(configFile: string | null, inputFilePath: string): IConfigurationLoadResult;
+export function findConfiguration(configFile: string, inputFilePath?: string): IConfigurationLoadResult {
```
ajafff: You actually need 2 overloads and one implementation. Right now there is only 1 overload.
```ts
export function findConfiguration(configFile: string | null, inputFilePath: string): IConfigurationLoadResult;
export function findConfiguration(configFile: string, inputFilePath?: string): IConfigurationLoadResult;
export function findConfiguration(configFile: string | null, inputFilePath?: string): IConfigurationLoadResult {
   ...
}
```

src/configuration.ts
```diff
@@ -90,7 +90,8 @@ const BUILT_IN_CONFIG = /^tslint:(.*)$/;
...
     const configPath = findConfigurationPath(configFile, inputFilePath);
```
ajafff: you will get a compile error on this line, because the parameter types do not match any overload.
You could work around it with this dirty hack:
```ts
findConfigurationPath(configFile, inputFilePath!); // note the ! after inputFilePath
```
Otherwise you need an if..else and still need to use a nonNullAssertion

src/configuration.ts
```diff
@@ -112,14 +113,15 @@ export function findConfiguration(configFile: string | null, inputFilePath: stri
...
+export function findConfigurationPath(suppliedConfigFilePath: string | null, inputFilePath: string): string | undefined;
+export function findConfigurationPath(suppliedConfigFilePath: string, inputFilePath?: string): string | undefined {
     if (suppliedConfigFilePath != null) {
         if (!fs.existsSync(suppliedConfigFilePath)) {
             throw new FatalError(`Could not find config file at: ${path.resolve(suppliedConfigFilePath)}`);
         } else {
             return path.resolve(suppliedConfigFilePath);
         }
-    } else {
+    } else if (inputFilePath !== undefined) {
```
ajafff: On second thought, you don't need this condition. If `suppliedConfigPath === null`, then `inputFilePath` cannot be undefined. You can simply assert that using the nonNullAssertion operator `inputFilePath!` in the else branch

ajafff:
Thanks @lumaxis 

Note that github does not send notifications when you force push changes to your branch. I only noticed by chance.

PR #3196, opened 2017-09-02
Refactor and fix no-unsafe-any

ajafff:
#### PR checklist

  - [x] Includes tests
- [ ] Documentation update

#### Overview of change:
* No longer ignore certain expression like property name in destructuring or expressions on both sides of a comparison.
* Check destructuring assignments and destructuring initializer.
* Disallow destructuring of `any`, e.g. `function foo({a, b}: any){}`
* Stricter checks for class property initializer when the property is declared in the superclass with a non-any type.
* Check `yield` expressions.
* Exclude type nodes completely.
* Ignore `export = x` and `export default x`
* Better handling of JSX.
* Allow `any` in certain locations when there is no contextual type, e.g. `declare let x: any; let obj = {x};` or `() => x`.
* Don't add redundant failures: if we know that `x` of `x.y` is of type `any` and we also reported that error, we don't add another for `x.y`, `x.y.z` or `x.y()`, etc.
* Fixed all TODOs

#### CHANGELOG.md entry:

[enhancement] `no-unsafe-any` checks more expressions, for example destructuring, `yield`, property initializer


src/rules/noUnsafeAnyRule.ts
```diff
@@ -40,248 +40,319 @@ export class Rule extends Lint.Rules.TypedRule {
...
-    /** @param anyOk If true, this node will be allowed to be of type *any*. (But its children might not.) */
-    function cb(node: ts.Node, anyOk?: boolean): void {
+    public walk(sourceFile: ts.SourceFile) {
+        if (sourceFile.isDeclarationFile) {
+            return; // Not possible in a declaration file.
+        }
+        sourceFile.statements.forEach(this.noCheck);
+    }
+
+    private noCheck = (node: ts.Node) => void this.visitNode(node);
```
adidahiya: this method name is a little unintuitive; can you make it more verbose and add a comment explaining its usage?

src/rules/noUnsafeAnyRule.ts
```diff
@@ -40,248 +40,319 @@ export class Rule extends Lint.Rules.TypedRule {
...
+            return false;
         }
+        this.addFailureAtNode(node, Rule.FAILURE_STRING);
+        return true;
     }
 
-    /** OK for this value to be 'any' if that's its contextual type. */
-    function checkContextual(arg: ts.Expression): void {
-        return cb(arg, /*anyOk*/ isAny(checker.getContextualType(arg)));
+    private checkContextual(node: ts.Expression, allowIfNoContextualType?: boolean) {
```
adidahiya: minor: would prefer the verbosity of naming this `checkContextualType`

Issue #3143, opened 2017-08-19
Error message is hidden when used with node API (for example, with tslint-loader)


MadaraUchiha:
### Bug Report

- __TSLint version__: 5.6.0
- __TypeScript version__: 2.4.1
- __Running TSLint via__: NodeJS API / `tslint-loader`

with `tslint.json` configuration:

```json
{
  "rulesDirectory": "./nonexistent"
}
```

When run with the `tslint` CLI, displays correct error.

#### Actual behavior

When run using `tslint-loader` as part of running `webpack`, displays the following error:
```
ERROR in ./src/index.tsx
Module build failed: Error
    at Object.FatalError (/home/madara/projects/Foo/node_modules/tslint/lib/error.js:31:23)
    at Function.findConfiguration (/home/madara/projects/Foo/node_modules/tslint/lib/configuration.js:55:27)
    at parseConfigFile (/home/madara/projects/Foo/node_modules/tslint-loader/index.js:45:26)
    at resolveOptions (/home/madara/projects/Foo/node_modules/tslint-loader/index.js:35:27)
    at Object.module.exports (/home/madara/projects/Foo/node_modules/tslint-loader/index.js:144:17)
 @ multi react-hot-loader/patch ./src/index
```

#### Expected behavior
Expected the error to contain 

```
Failed to load /home/madara/projects/Foo/tslint.json: Could not find custom rule directory: ./nonexistent
```

Like it does when you run `tslint -p .` in the project's directory.

---

The reason this happens is because of `error.ts`'s `class FatalError extends Error`, which is compiled with an ES5 target. Errors cannot be extended in this way (`ParentConstructor.call(this, args)`) before ES2015.

The potential solutions I can see are:

1. Bump the TypeScript compilation targets to ES2015 (which is a potentially breaking change, for example: custom rules will probably have to be compiled to ES2015 as well, ES5 will likely fail with "constructor AbstractSomething must be called with `new`"
2. Don't extend Error in this fashion, find some other way to hide the stack trace for CLI.

WDYT?

PR #3211, opened 2017-09-10
Fix prototype chain of FatalError

ajafff:
#### PR checklist

- [x] Addresses an existing issue: #3143
  - [ ] Includes tests
- [ ] Documentation update

#### Overview of change:
https://github.com/Microsoft/TypeScript-wiki/blob/master/Breaking-Changes.md#extending-built-ins-like-error-array-and-map-may-no-longer-work
Fixes: #3143

#### CHANGELOG.md entry:
[no-log]

### DIFF
--- a/src/runner.ts
+++ b/src/runner.ts
@@ -121,8 +121,8 @@ export async function run(options: Options, logger: Logger): Promise<Status> {
     try {
         return await runWorker(options, logger);
     } catch (error) {
-        if ((error as FatalError).name === FatalError.NAME) {
-            logger.error((error as FatalError).message);
+        if (error instanceof FatalError) {
+            logger.error(error.message);
             return Status.FatalError;
         }
         throw error;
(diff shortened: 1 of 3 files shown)


src/runner.ts
```diff
@@ -121,8 +121,8 @@ export async function run(options: Options, logger: Logger): Promise<Status> {
     try {
         return await runWorker(options, logger);
     } catch (error) {
-        if ((error as FatalError).name === FatalError.NAME) {
-            logger.error((error as FatalError).message);
+        if (error instanceof FatalError) {
```
adidahiya: I would generally avoid `instanceof`, but in this case it's fine since `FatalError` is declared within the tslint package. If it was an import from another package, `instanceof` is prone to fail depending on which version of that package is used at runtime (there could be multiple versions installed in `node_modules`).

PR #3224, opened 2017-09-14
Add option to disallow modifying properties of parameters

trenneman:
#### PR checklist

  - [x] Includes tests
- [x] Documentation update

#### Overview of change:
Added an option to `no-parameter-reassignment` to disallow adding/modifying/deleting the properties of argument objects as well.

#### CHANGELOG.md entry:

[new-rule-option] [`no-parameter-reassignment`](https://palantir.github.io/tslint/rules/no-parameter-reassignment/): Add `props` option


ajafff:
When implementing this rule I decided not to add this code.
The purpose of this rule is to prevent reassigning parameters, because you cannot declare them as `const`.
Modifying objects passed as parameter can be prevented by choosing the correct type. You can declare it as follows:

```ts
type DeepReadonly<T> = {
     readonly [K in keyof T]: DeepReadonly<T[K]>
};
function foo(bar: Readonly<A>, baz: DeepReadonly<B>) {
    bar.prop++; // compile error
    bar.nested.prop++; // no compile error, because Readonly is shallow

    baz.prop++; // compile error
    baz.nested.prop++; // compile error
}
```

After all you should prefer compile errors over lint errors. That's why I'm 👎  on this

---

Btw: "parameter properties" !== "parameter's properties"

trenneman:
That's true. But I can't enforce using `DeepReadonly` over a whole project. Adding a linter rule makes this behaviour consistent over all functions.

ajafff:
Ok, if I leave my opinion aside, there are still some cases that are currently not detected by your implementation:
```ts
foo.bar.baz++;
foo[0]++;
(foo).bar++;
// and combinations of the above

// this is a bit tricky and maybe not worth the effort
const {bar} = foo:
bar.baz++; // DeepReadonly would catch this one
```

Also note that your parameters are not immutable with this option. Collections can still be modified through their methods, e.g. Array.prototype.push.

---

> I can't enforce using DeepReadonly over a whole project

You can. A rule to enforce that would be way easier than this rule.

adidahiya:
I have the same concerns as @ajafff; I don't think this belongs in the core rule set since it's tricky to get right and is much more natural as a compiler check. Feel free to implement it as a custom rule though.

trenneman:
OK, I agree. I kinda underestimated the impact/complexity of this. Thanks for reviewing so quickly, @ajafff @adidahiya.

PR #3226, opened 2017-09-14
Enable if-else in test markup, fix build

ajafff:
#### PR checklist

  - [x] Includes tests
- [ ] Documentation update

#### Overview of change:
Fixes build failures with typescript@next

Instead of duplicating all tests that have different output between typescript versions, I chose to add support for #if #else and #endif directives to the test markup. The syntax is inspired by C/C++.
The directives are processed before the file is actually parsed. Therefore it can be used to conditionally set the text of a message substitution.
Currently you can only test the typescript version.

I'll update the docs afterwards.

#### CHANGELOG.md entry:

[no-log] will add a changelog entry to the docs PR

### DIFF
--- a/src/verify/parse.ts
+++ b/src/verify/parse.ts
@@ -14,6 +14,7 @@
  * limitations under the License.
  */
 
+import * as semver from "semver";
 import * as ts from "typescript";
 import { format } from "util";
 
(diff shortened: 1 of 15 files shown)


src/verify/parse.ts
```diff
@@ -41,6 +42,57 @@ export function getTypescriptVersionRequirement(text: string): string | undefine
...
+    const result = [];
+    let collecting = true;
+    let state = State.Initial;
+    for (const line of lines) {
+        if (line.startsWith("#if typescript")) {
+            if (state !== State.Initial) {
+                throw lintSyntaxError("#if directives cannot be nested");
+            }
+            state = State.If;
+            collecting = semver.satisfies(tsVersion, line.slice(14).trim());
```
adidahiya: instead of `.slice(14)`, can you include the string literal you're splitting with? `"if typescript ".length`

PR #3233, opened 2017-09-17
New rule no-return-await

ajafff:
#### PR checklist

  - [x] Includes tests
- [x] Documentation update

#### CHANGELOG.md entry:

[new-rule] `no-return-await`


develar:
👎 to include this rule by default to latest. I don't want to spend hours to debug why my try-catch doesn't work some time later (or another colleague in my team). Just because I forgot to insert missed `await`. i.e. when you modify your code and this return statement will be not latest in the function body.

adidahiya:
> Just because I forgot to insert missed await

@develar that sounds like the job of another lint rule -- check out no-floating-promises

PR #3238, opened 2017-09-19
Add new ban-parens rule

calebegg:
Since the definition of “unnecessary” is by its nature very opinionated, I made this rule very open-ended, with a nice set of default “unnecessary” parens that I hope should be relatively uncontroversial.

Let me know if there are defaults you think I should either add or remove.


src/rules/noUnnecessaryParensRule.ts
```diff
@@ -0,0 +1,190 @@
...
+        test(node: ts.ParenthesizedExpression | ts.ParenthesizedTypeNode) {
+            return isNodeOfKind(isParenthesizedExpression(node) ? node.expression : node.type, name);
+        },
+    })).concat(
+        asChildOf.map((name: string) => {
+            const [parentKind, whichChild] = name.split(".");
+            return {
+                message: `the ${whichChild} child of an expression of type ${parentKind}`,
+                test(node: ts.ParenthesizedExpression | ts.ParenthesizedTypeNode) {
+                    if (node.parent == undefined) {
```
ajafff: That condition will never be true

src/rules/noUnnecessaryParensRule.ts
```diff
@@ -0,0 +1,190 @@
...
+            [{
+                withChild: [
+                    "Identifier",
+                    "LiteralExpression",
+                ],
+                asChildOf: [
+                    "VariableDeclaration.initializer",
+                    "ParenthesizedExpression.expression",
+                    "CallExpression.arguments",
+                    "ExpressionStatement.expression",
```
ajafff: All of these options require knowledge of typescript's AST. I don't think that's practicle for most TSLint users.

Maybe the rule should just find *all* unnecessary parens by default and provide options to disable some of them. Something like `"allow-typeof"` to allow `(typeof foo)==="string"` for example. These can be added on demand.

Also with these options, there's currently no way to ban parens on `"*.type"`, e.g. `let foo: (string)` and `foo(param: (1)) {}`

ajafff: There's currently no way to ban `(1 + 2) + 3` without banning `(1 + 2) * 3`, too.

src/rules/noUnnecessaryParensRule.ts
calebegg: True, but I thought the check would be preferable to the `!`. Switched to the `!`.

src/rules/noUnnecessaryParensRule.ts
calebegg: > Maybe the rule should just find all unnecessary parens

So, I didn't want to get into litigating what's "necessary". I don't want to tell people to not do `(true && false) || true` or something, when I think that's probably preferable to leaving out the parentheses since it's hard to remember boolean prescience.

Instead, I want to ban them where they're *clearly* unnecessary, and leave the edge cases for human reviewers (or the judgement of the author).

I feel like having a list of allowances would get out of hand.

If you think this approach is fundamentally unsuitable, that's fine, just let me know. I'll just use it on my project.

> Also with these options, there's currently no way to ban parens on "*.type", e.g. let foo: (string) and foo(param: (1)) {}
 
Why not? I added some examples and tests.

calebegg: Maybe the documentation could point to http://astexplorer.net. That's how I find these AST names.

src/rules/noUnnecessaryParensRule.ts
```diff
@@ -0,0 +1,197 @@
...
+    }
+
+    public apply(sourceFile: ts.SourceFile): Lint.RuleFailure[] {
+        return this.applyWithFunction(sourceFile, walk, this.ruleArguments[0] as Options);
+    }
+}
+
+function isNodeOfKind(node: ts.Node, kindName: string) {
+    switch (kindName) {
+        case "LiteralExpression":
```
ajafff: consider adding `case "*": return true;` to allow for `*.type`

src/rules/noUnnecessaryParensRule.ts
```diff
@@ -0,0 +1,197 @@
...
+        // ex: ((1 + 1)) + 2
+        "ParenthesizedExpression.expression",
+        // ex: foo((a), b)
+        "CallExpression.arguments",
+        // ex: Foo<(string|number), string>
+        "TypeReference.typeArguments",
+        // ex: (foo.bar());
+        "ExpressionStatement.expression",
+        // ex: function foo((a: string), b: number) {}
+        "SignatureDeclaration.parameters",
```
ajafff: `SignatureDeclaration` is not a `SyntaxKind`. It's the base type for `FunctionDeclaration`, `FunctionExpression`, `ConstructorDeclaration`, `IndexSignature`, ...
This option will not work as intended.
Also note that a parameter cannot be wrapped in parens

src/rules/noUnnecessaryParensRule.ts
```diff
@@ -0,0 +1,197 @@
...
+        // ex: let x: (string|number) = 3;
+        "VariableDeclaration.type",
+        // ex: function(foo: (number|string)) {}
+        "Parameter.type",
```
ajafff:
```ts
// foo[(bar + "baz")]
"ElementAccessExpression.argumentExpression",
// `${(foo)}`
"TemplateSpan.expression"
```

src/rules/noUnnecessaryParensRule.ts
```diff
@@ -0,0 +1,197 @@
...
+    ] : ctx.options.withChild;
+    const asChildOf = ctx.options.default ? [
+        // ex: let x = (1 + foo());
+        "VariableDeclaration.initializer",
+        // ex: type x = (string|number);
+        "TypeAliasDeclaration.type",
+        // ex: ((1 + 1)) + 2
+        "ParenthesizedExpression.expression",
+        // ex: foo((a), b)
+        "CallExpression.arguments",
```
ajafff: also `"NewExpression.arguments"`

src/rules/noUnnecessaryParensRule.ts
```diff
@@ -0,0 +1,197 @@
...
+        // ex: Foo<(string|number), string>
+        "TypeReference.typeArguments",
```
ajafff: `"*.typeArguments"`?

src/rules/noUnnecessaryParensRule.ts
ajafff: Adding a link to astexplorer.net sounds like a good idea.

ajafff:
@calebegg please merge master to fix the build

src/rules/noUnnecessaryParensRule.ts
calebegg: Ah, good call on both counts. I was being overenthusiastic. :-)

test/rules/no-unnecessary-parens/test.ts.lint
```diff
@@ -0,0 +1,33 @@
+let x = (3);
```
ajafff: There's one edge case that would result in invalid code:
```ts
let y, z;
let x = (y = 1, z = 2);
```

I hope nobody writes such code.
Nontheless there should probably be an exception for `ts.SyntaxKind.BinaryExpression` where `operatorToken.kind === ts.SyntaxKind.CommaToken`

src/rules/noUnnecessaryParensRule.ts
```diff
@@ -0,0 +1,205 @@
...
+        "CallExpression.arguments",
+        "NewExpression.arguments",
+        // ex: Foo<(a|b), c>; foo<(a)>();
+        "*.typeArguments",
+        // ex: (foo.bar());
+        "ExpressionStatement.expression",
+        // ex: let x: (string|number) = 3;
+        "VariableDeclaration.type",
+        // ex: function(foo: (number|string)) {}
+        "Parameter.type",
```
ajafff: I'm still in favor of using `"*.type"`. I can't come up with a case where parens are desired or even required.

calebegg: Done.

test/rules/no-unnecessary-parens/test.ts.lint
calebegg: Ugh, the comma operator is the worst. On my shortlist of new tslint rules is one to ban it entirely.

I made it so that it doesn't suggest a fix if the child uses a comma operator, but it still flags it as an issue. How does that sound?

calebegg:
Hey, that's my own commit! :-) Fixed.

src/rules/noUnnecessaryParensRule.ts
```diff
@@ -0,0 +1,216 @@
...
+        // ex (x.a).b
+        "PropertyAccessExpression",
+        // ex (f());
+        "CallExpression",
+    ] : ctx.options.withChild;
+    const asChildOf = ctx.options.default ? [
+        // ex: let x = (1 + foo());
+        "VariableDeclaration.initializer",
+        // ex: type x = (string|number);
+        "TypeAliasDeclaration.type",
```
ajafff: this is redundant now, but I'm fine with leaving it in for documentation

src/rules/noUnnecessaryParensRule.ts
```diff
@@ -0,0 +1,216 @@
...
+        "ParenthesizedExpression.expression",
+        // ex: foo((a), b); new Foo((a));
+        "CallExpression.arguments",
+        "NewExpression.arguments",
+        // ex: Foo<(a|b), c>; foo<(a)>();
+        "*.typeArguments",
+        // ex: (foo.bar());
+        "ExpressionStatement.expression",
+        // ex: let x: (string|number) = 3;
+        "VariableDeclaration.type",
```
ajafff: also redundant

src/rules/noUnnecessaryParensRule.ts
calebegg: Yeah, that's true. I guess it does give a better error message, kinda, to have the more specific ones.

I'd be fine either way. Up to you.

calebegg:
Makes sense. Renamed.

calebegg:
FYI, updated to fix a minor issue that came up when running over our codebase -- `typeof(x)` and `throw(y)` become `typeofx` and `throwy`.

calebegg:
Fixed more issues that cropped up in our codebase -- `(0).toString()` became `0.toString()`, which doesn't compile.

```
return (
   foo);
```

becomes

```
return 
  foo;
```

which also doesn't compile.

src/rules/banParensRule.ts
```diff
@@ -0,0 +1,241 @@
...
+                    isBinaryExpression(node.expression) &&
+                    node.expression.operatorToken.kind === ts.SyntaxKind.CommaToken) {
+                    replacement = [];
+                }
+
+                if (!(
+                    // Don't flag `(0).foo()`, because `0.foo()` doesn't work.
+                    (isParenthesizedExpression(node) &&
+                        isNumericLiteral(node.expression) &&
+                        node.parent != undefined &&
```
ajafff: `node.parent` cannot be `undefined`

src/rules/banParensRule.ts
```diff
@@ -0,0 +1,241 @@
...
+                        isPropertyAccessExpression(node.parent)) ||
+                    // Don't flag `return (\nfoo)`, since the parens are necessary.
+                    (isParenthesizedExpression(node) &&
+                        node.parent != undefined &&
```
ajafff: same here

src/rules/banParensRule.ts
```diff
@@ -0,0 +1,241 @@
...
+                        isReturnStatement(node.parent) &&
+                        ctx.sourceFile.text[node.getStart() + 1] === "\n")
```
ajafff: `isSameLine(ctx.sourceFile, node.expression.pos, node.expression.getStart(ctx.sourceFile))`

src/rules/banParensRule.ts
```diff
@@ -0,0 +1,241 @@
...
+                // Don't suggest a fix for a (hopefully rare) pattern where
+                // removing the parentheses would almost always be bad, e.g.
+                // let x = (y = 1, z = 2);
+                if (isParenthesizedExpression(node) &&
+                    isBinaryExpression(node.expression) &&
+                    node.expression.operatorToken.kind === ts.SyntaxKind.CommaToken) {
+                    replacement = [];
+                }
+
+                if (!(
```
ajafff: Consider extracting this condition to a function and invoke it in the condition in line 199
```ts
if (isParenthesizedExpression(node) && !parensAreNecessary(node, ctx.sourceFile) || isParenthesizedType(node))
```

src/rules/banParensRule.ts
calebegg: Done (had to tweak it a bit; used `node.expression.getStart()` and `node.getStart()`)

src/rules/banParensRule.ts
calebegg: Done

src/rules/banParensRule.ts
calebegg: Changed to !

For another case, this broke the type narrowing, so I had to add a cast.

src/rules/banParensRule.ts
calebegg: Done

src/rules/banParensRule.ts
```diff
@@ -0,0 +1,252 @@
...
+        case "Keyword":
+            return node.kind > ts.SyntaxKind.FirstKeyword && node.kind < ts.SyntaxKind.LastKeyword;
+        case "*":
+            return true;
+        default:
+            return node.kind === syntaxKindMapping[kindName];
+    }
+}
+
+function isParenthesizedType(node: ts.Node): node is ts.ParenthesizedTypeNode {
```
ajafff: note to myself: remove this after the next release of `tsutils`

src/rules/banParensRule.ts
```diff
@@ -0,0 +1,252 @@
...
+ * Checks some exceptional cases where the parentheses likely are still required.
+ */
+function parensAreNecessary(node: ts.ParenthesizedExpression, sourceFile: ts.SourceFile) {
+    return (
+        // Don't flag `(0).foo()`, because `0.foo()` doesn't work.
+        (isNumericLiteral(node.expression) &&
+            isPropertyAccessExpression(node.parent!)) ||
+        // Don't flag `return (\nfoo)`, since the parens are necessary.
+        (isReturnStatement(node.parent!) &&
+            !isSameLine(sourceFile, node.getStart(), node.expression.getStart())) ||
```
ajafff: please pass the `sourceFile` parameter to `.getStart()` to avoid unnecesary lookups.
I'm also very confident that you can replace `node.getStart()` with `node.expression.pos`. The former gives you the start of the open paren, the latter gives the end of the open paren. But that difference doesn't matter for `isSameLine`

src/rules/banParensRule.ts
```diff
@@ -0,0 +1,252 @@
...
+        // Don't flag parens around destructuring assignment
+        (isBinaryExpression(node.expression) &&
+            node.expression.operatorToken.kind === ts.SyntaxKind.EqualsToken &&
+            isObjectLiteralExpression(node.expression.left) &&
+            isExpressionStatement(node.parent!)) ||
+        // Don't flag parentheses in an arrow function's body
+        (isArrowFunction(node.parent!) &&
+            (node.parent as ts.ArrowFunction).body === node));
```
ajafff: you can remove the second part of the condition. A `ParenthesizedExpression` can only occur in the body of an `ArrowFunction`

src/rules/banParensRule.ts
```diff
@@ -0,0 +1,252 @@
...
+}
```
ajafff: We should stop adding more exceptions or it will get out of hand. Users should not ban parens around certain Expressions, for example BinaryExpression and ConditionalExpression. (maybe add this to the docs?)
Otherwise you would need to allow `BinaryExpression` and `ConditionalExpression` when it is `(New|Call|PropertyAccess|ElementAccess|Void|Await|Delete)Expression.expression`, e.g. `(foo.bar || foo.baz)()`, `(foo ? bar : baz)()`, ...

... and probably many more

calebegg: Added documentation.

I agree it's getting complex. If you think there are any default syntax kinds that aren't worth the exceptions, let me know. It's running cleanly over our codebase now with defaults, so I am at least somewhat confident that we've found most of the strange cases.

src/rules/banParensRule.ts
calebegg: Done.

src/rules/banParensRule.ts
calebegg: Done. Sorry, I wrestled with this a bit until I got something working and then didn't pay more attention. I didn't really know the difference between `.getStart` and `.pos` (though I think I get it now). Is there documentation for this stuff somewhere?

What's the benefit of passing `sourceFile` in?

ajafff: `.pos` is the position where the parser started to parse the current node (or token). `node.pos` is the same value as the preceding token's `.end`.
`node.getStart()` is the position where the first token of a node starts. The difference between `.pos` and `.getStart()` is caused by trivia (whitespace and comments).
Also note that `node.pos === node.getFullStart()` and `node.end === node.getEnd()`

Re: passing `sourceFile` parameter:
I wrote this down in https://palantir.github.io/tslint/develop/custom-rules/performance-tips.html
> There are serveral methods that have an optional parameter sourceFile. Don’t omit this parameter if you care for performance. If ommitted, typescript needs to walk up the node’s parent chain until it reaches the SourceFile. This can be quite costly when done frequently on deeply nested nodes.

You also find there why I prefer  not to call `.getStart()`:
> node.getStart() scans the source to skip all the leading trivia. Although barely noticeable, this operation is not for free. If you need the start position of a node more than once per function, consider caching it.

calebegg: Got it, makes sense. Thanks for the tips!

calebegg:
Anything else you need from me for this?

ajafff:
@calebegg nope, I just want @adidahiya to sign this off before landing, but it seems like he's busy lately.

test/rules/ban-parens/test.ts.lint
```diff
@@ -0,0 +1,46 @@
...
+(0).toLocaleString('en-US', {style: 'percent'});
+return (
+    0);
+({a, b} = obj);
+let foo = () => ({a: 1}.a);
+
+// No issues with
+let z = (x) => x + 1;
+let x = (1 + 1) * 2;
+(f()) + 1;
```
adidahiya: can you also add some tests for common JSX syntax?

```ts
const foo = (
    <div>text</div>
);

<button ref={(el => this.buttonEl = el)} />
             ~                        ~

<button ref={el => (this.buttonEl = el)} />
                   ~                  ~
```

calebegg: Added two tests. I don't really know what I'm doing with JSX :-) so let me know if they could be improved. (The 2nd and 3rd you gave aren't errors with my current test settings)

adidahiya: Ok, the 3rd example I gave should not be an error, but I think the 2nd one _should_ so that we can align more closely with `prettier`. I would  Would you be able to update your rule implementation so that my 2nd snipped does cause a lint failure?

test/rules/ban-parens/test.tsx.lint
```diff
@@ -0,0 +1,9 @@
+const foo = (
+            ~
+    <div>text</div>
+~~~~~~~~~~~~~~~~~~~
+);
```
adidahiya: these should not be banned by default. parens around multiline jsx statements are _enforced_ in the default tslint-react config. 

this is why I wanted you to add these tests -- the default config should not conflict with tslint-react.

calebegg: I see -- I wish you'd just said that.

I don't know if it'll be easy to do this, if jsx expressions can occur anywhere. That eliminates most of the rules. I can just exclude jsx files entirely from this check, maybe?

adidahiya: Sorry if I wasn't clear upfront. Excluding `.tsx` files completely in this rule is not a good option because it will feel pretty inconsistent for TSX users -- there's lots of regular TS code in `.tsx` files. I haven't really dug into the internals of this rule implementation here (deferring to @ajafff) but I am concerned about the overall UX of this rule for tslint-react users, so I'd like you to update it before merging.

ajafff:
@calebegg I would avoid any hard coded special handling for JsxElements or .jsx files.
Instead you could add a config option for exceptions that override the blacklisted locations.

Consider this code snippet from the tests:
```tsx
const foo = (
    <div>text</div>
);
```
The blacklist disallows parens in `VariableDeclaration.initializer` while the whitelist requires parens around `JsxElement`. Therefore no error.

@adidahiya I'd rather not distinguish between single-line and multi-line JsxElements.
Do we also need to whitelist `JsxSelfClosingElement`? It will mostly be on a single line but can also span multiple lines when there are many attributes.

In the near future TypeScript will add support for `JsxFragment`. We should whitelist it now to be prepared for that feature.

test/rules/ban-parens/test.ts.lint
calebegg: Done.

calebegg:
Makes sense. Done.

src/rules/banParensRule.ts
```diff
@@ -0,0 +1,282 @@
...
+                        childOrChildren.indexOf(node) !== -1 :
+                        childOrChildren === node;
+                },
+            };
+        }));
+
+    return ts.forEachChild(ctx.sourceFile, function cb(node: ts.Node): void {
+        if ((isParenthesizedExpression(node) && !parensAreNecessary(node, ctx.sourceFile)) || isParenthesizedType(node)) {
+            const restriction = restrictions.find((r) => r.test(node));
+            if (restriction != undefined) {
```
ajafff: consider moving the check for `exceptions` to this condition. that avoid checking exceptions for every blacklist entry.

also: prefer `!==` over `!=`

src/rules/banParensRule.ts
```diff
@@ -0,0 +1,282 @@
...
+        // ex: function(foo: (number|string)) {}
+        "*.type",
+        // foo[(bar + "baz")]
+        "ElementAccessExpression.argumentExpression",
+        // `${(foo)}`
+        "TemplateSpan.expression",
+        ...ctx.options.asChildOf,
+    ] : ctx.options.asChildOf;
+    const exceptions = ctx.options.default ? [
+        "JsxElement",
```
ajafff: I think it's best to also whitelist `JsxSelfClosingElement`

ajafff:
You need to merge latest master to fix the CI failures

src/rules/banParensRule.ts
calebegg: Oh, that makes much more sense!

I actually moved it to above restrictions.find so that it can just skip the testing entirely for exception nodes.

src/rules/banParensRule.ts
calebegg: Done

ajafff:
@adidahiya PTAL

@calebegg merging master caused some new lint failures:

```
/home/ubuntu/tslint/src/rules/completed-docs/tagExclusion.ts
ERROR: 78:43   ban-parens  Don't include parentheses around an expression of type LiteralExpression

/home/ubuntu/tslint/src/rules/maxLineLengthRule.ts
ERROR: 99:28   ban-parens  Don't include parentheses around an expression of type Identifier

/home/ubuntu/tslint/src/rules/noUnsafeAnyRule.ts
ERROR: 102:45  ban-parens  Don't include parentheses around the initializer child of an expression of type VariableDeclaration
```

adidahiya:
Ok, I have some more feedback but I apologize if this is poorly timed. I haven't been able to keep up with all the external PRs submitted to this repo recently.

I don't think this rule in its current form belongs in the core rule set. The configuration options are esoteric and unapproachable for most users. This is mainly because they expose internals of the AST semantics to the end user -- I'm drawing a line there when it comes to developer experience. I think we need to limit the scope of possible configuration because otherwise users will get confused by the error messages it produces and will have a hard time figuring out how to configure the rule's options to make their linter pass. Alternatively, it would be fine to keep this in a separate ruleset package for power users of TSLint.

calebegg:
Closing. It would have been great to get this feedback earlier on in the process.

PR #3239, opened 2017-09-19
Fix an issue where necessary type assertions were being flagged by no-unnecessary-type-assertions

calebegg:


src/rules/noUnnecessaryTypeAssertionRule.ts
```diff
@@ -69,7 +69,11 @@ class Walker extends Lint.AbstractWalker<void> {
         if (node.kind !== ts.SyntaxKind.NonNullExpression &&
             (isTypeFlagSet(castType, ts.TypeFlags.Literal) ||
                 isObjectType(castType) &&
-                isObjectFlagSet(castType, ts.ObjectFlags.Tuple))) {
+                isObjectFlagSet(castType, ts.ObjectFlags.Tuple)) ||
+            // Sometimes tuple types don't have ObjectFlags.Tuple set, like when
+            // they're being matched against an inferred type
+            (castType.getProperties().some((symbol) => !isNaN(Number(symbol.name))))) {
```
ajafff: This basically checks if there is at least one property which name is a number?
That took me a while to figure out. Maybe a comment would help here.

calebegg: Done.

PR #3249, opened 2017-09-26
New rule dead-type-branch

rjamet:
#### PR checklist

  - [X] Includes tests
- [ ] Documentation update

#### Overview of change:

That rule looks at if-blocks and tries to flag dead branches. More specifically, this searches conditions of if for some type-based tests (typeof, instanceof, type guards in functions), and remembers which variables are looked at, as we consider it a strong signal for type-based tests. We then look for its usage under that condition and the following else / else if statements, and if in there, an occurence of the expression is typed as `never` (per the typechecker), then flag the branch as it's likely to be dead.

To be clear, the rule itself does none of the type arithmetic. Most of the complexity is there to ensure we only flag things typed as never when that is likely to be a result of type tests. 

First of all, that's my first PR into TSLint, so I probably missed a lot of conventions and so on: if so, let me know. This is still a bit rough around the edges too (especially documentation), but I wanted to start the conversation early, to know if you'd be interested in this rule.

Areas that should probably be improved:
- User-facing documentation (my first idea was to use this internally at Google, I haven't looked at your conventions there yet)
- Performance, especially around the way I pass `taggedExpr` around, recursively copying it. It might or might not matter, I'm not sure.
- The error message and the rule name aren't that great, I welcome any suggestion there.
- The code is not trivial, so there might be more bugs in there.
- I uncovered very surprising behaviors around type inference and structural typing (though it's not yet clear whether that's a bug or not or what causes that), see https://github.com/Microsoft/TypeScript/issues/18560. That could result in false-positives.
- It conflicts in principle with some patterns of exhaustiveness checks for enums. Let's say you use string literal types, then you could write a series of `if`/`else if` for each value, then a final `else throw("unexpected enum value" + v);` in case the enum evolves: that's going to be flagged as dead code.

cc @bowenni @mprobst 

#### CHANGELOG.md entry:

[new-rule] `dead-type-branch`

### DIFF
--- a/src/rules/deadTypeBranchRule.ts
+++ b/src/rules/deadTypeBranchRule.ts
@@ -1,0 +1,232 @@
+import * as Lint from 'tslint';
+import * as tsutils from 'tsutils';
+import * as ts from 'typescript';
+
+type PerProgramConfig = {
+  tc: ts.TypeChecker,
+  ctx: Lint.WalkContext<void>,
+  // Format: filename|startPosition|endPosition
+  // We use this Map to flag only the highest-level block / expressions.
+  flaggedRanges: Set<string>,
+};
+
+/**
+ * This linter rule tries to find dead branches in programs by looking for
+ * unsatisfiable typing constraints.
+ */
+export class Rule extends Lint.Rules.TypedRule {
+  public static metadata: Lint.IRuleMetadata = {
+    ruleName: 'dead-type-branch',
+    description: 'This linter rule tries to find dead branches in programs ' +
+        'by looking for unsatisfiable typing constraints.',
+    rationale: 'Unsatisfiable type constraits indicate unreachable code.',
+    options: null,
+    optionsDescription: '',
+    type: 'typescript',
+    typescriptOnly: true,
+  };
+
+  applyWithProgram(sourceFile: ts.SourceFile, program: ts.Program):
+      Lint.RuleFailure[] {
+    return this.applyWithFunction(sourceFile, (ctx) => this.walk(ctx, program));
+  }
+
+  /**
+   * Looks for if statements to analyze for dead branches.
+   */
+  private static searchForIfStatements(
+      node: ts.Node, config: PerProgramConfig,
+      previouslyTaggedExprs?: Set<string>): void {
+    const taggedExpr = Rule.copyOrInit(previouslyTaggedExprs);
+
+    if (!tsutils.isIfStatement(node)) {
+      return node.forEachChild(
+          n => Rule.searchForIfStatements(n, config, taggedExpr));
+    }
+
+    const ifNode: ts.IfStatement = node;
+    Rule.searchForTypeGuardsInCondition(ifNode.expression, config)
+        .forEach(g => taggedExpr.add(g));
+
+    if (taggedExpr.size !== 0) {
+      Rule.searchForExpressionsTypedAsNever(
+          ifNode.thenStatement, config, taggedExpr);
+      if (ifNode.elseStatement &&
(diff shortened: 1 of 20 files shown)


palantirtech:
Thanks for your interest in [palantir/tslint](https://github.com/palantir/tslint), @rjamet! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

src/rules/deadTypeBranchRule.ts
```diff
@@ -0,0 +1,232 @@
...
+
+  applyWithProgram(sourceFile: ts.SourceFile, program: ts.Program):
+      Lint.RuleFailure[] {
+    return this.applyWithFunction(sourceFile, (ctx) => this.walk(ctx, program));
+  }
+
+  /**
+   * Looks for if statements to analyze for dead branches.
+   */
+  private static searchForIfStatements(
```
ajafff: We generally try to avoid static methods on the rule. Consider using function declarations.

src/rules/deadTypeBranchRule.ts
```diff
@@ -0,0 +1,232 @@
...
+      node: ts.Node, config: PerProgramConfig,
+      previouslyTaggedExprs?: Set<string>): void {
+    const taggedExpr = Rule.copyOrInit(previouslyTaggedExprs);
+
+    if (!tsutils.isIfStatement(node)) {
+      return node.forEachChild(
+          n => Rule.searchForIfStatements(n, config, taggedExpr));
+    }
+
+    const ifNode: ts.IfStatement = node;
```
ajafff: you don't need this type annotation

ajafff:
> It conflicts in principle with some patterns of exhaustiveness checks for enums. Let's say you use string literal types, then you could write a series of if/else if for each value, then a final else throw("unexpected enum value" + v); in case the enum evolves: that's going to be flagged as dead code.

That's actually a big deal. This pattern is used a lot AFAICT.

test/rules/dead-type-branch/function-type-guards/test.ts.lint
```diff
@@ -0,0 +1,40 @@
...
+else if (isA(ab)) { ab; }
+
+
+if (isA(ab)) { ab; }
+else if (isB(ab)) { ab; }
+else if (isC(ab)) { ab; }
+                  ~~~~~~~ [This block is unreachable: TypeScript infers the type of 'ab' as 'never' at this location, which means the previous type guard is always false.]
+else { ab; }
+     ~~~~~~~ [This block is unreachable: TypeScript infers the type of 'ab' as 'never' at this location, which means the previous type guard is always false.]
+
```
ajafff: consider the following:
```ts
function foo(v: 1 | 2) {
    switch (v) {
        case 1:
        case 2:
            return v;
        default:
            doSomethingWithoutV();
    }
}
```
If I'm not mistaken, the rule will not show a warning on the default clause because it doesn't use `v`.
IIRC there's no way to retrieve the type of a symbol at a location where that symbol is not used. That means the TypeChecker cannot tell you that `v` is `never` because `v` is not used in the default clause.

That said, it seems like this rule only prevents unintentional usage of a variable which type is narrowed to `never`.
I guess that could be achieved without control flow analysis.

rjamet: You're right: if you're not using the value in a block or in its parents, nothing will warn. But it seems risky to rely on that value not being used in that pattern, no ?

(btw, the rule currently doesn't touch case statements at all, but the idea is the same)

Issue #3256, opened 2017-09-28
no-empty-interface should not fail on an interface that resolves types


bronson:
### Bug Report

Related to https://github.com/palantir/tslint/issues/2065

- __TSLint version__: 5.7.0
- __TypeScript version__: 2.5.2
- __Running TSLint via__: Atom

#### TypeScript code being linted

```ts
interface CustomPromise extends Promise<string|false> {
}
```
This throws the error, "An interface declaring no members is equivalent to its supertype".

That's not really true, is it?  Here, CustomPromise is not equivalent to Promise.

andy-hanson:
Why not `type CustomPromise = Promise<string | false>`?

ajafff:
I guess that's because you cannot extend a type alias.

PR #3260, opened 2017-09-28
no-empty-interface: allow providing type arguments for extended type

ajafff:
#### PR checklist

- [x] Addresses an existing issue: #3256
  - [x] Includes tests
- [ ] Documentation update

#### Overview of change:
[bugfix] `no-empty-interface` allows providing type arguments for extended type
Fixes: #3256

The rule now ignores every interface whose parent has type arguments. That leads to some false-negatives:
```ts
interface Base<T> {
    prop: T;
}

interface Good extends Base<string> {}
interface Bad<T> extends Base<T> {}
interface ProbablyGood<T extends string> extends Base<T> {}
interface ProbablyGood2<T = string> extends Base<T> {}

interface Base2<T, U = void> {
    prop: T,
    prop2: U;
}

interface Bad2<T, U> extends Base2<T, U> {}
interface Good2<U, T> extends Base2<T, U> {}
interface Good3<T> extends Base2<T> {} // looks like `Bad`, but isn't ... would require the type checker
```

There will be no error on any of these declarations. One might expect errors on `Bad` and `Bad2` (and maybe even on `ProbablyGood`, `ProbablyGood2` if the base type has the same default and/or constraint) but that requires the type checker to find out about all type parameters of the base type.

I don't know if these rare cases are worth the effort. Anyways, that should be done in a separate PR.

### DIFF
--- a/src/rules/noEmptyInterfaceRule.ts
+++ b/src/rules/noEmptyInterfaceRule.ts
@@ -45,13 +45,22 @@ function walk(ctx: Lint.WalkContext<void>) {
     return ts.forEachChild(ctx.sourceFile, function cb(node: ts.Node): void {
         if (isInterfaceDeclaration(node) &&
             node.members.length === 0 &&
-            (node.heritageClauses === undefined ||
-             // allow interfaces that extend 2 or more interfaces
-             node.heritageClauses[0].types.length < 2)) {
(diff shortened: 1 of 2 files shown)


andy-ms:
Could this be put behind an option? I'd like to still get errors for `interface T extends U<string> {}` on DefinitelyTyped as this can just be `type T = U<string>;`.

Issue #3022, opened 2017-07-10
no-invalid-this should not complain on functions with contextual this


NN---:
### Bug Report

- __TSLint version__: 5.5.0
- __TypeScript version__: 2.4.1
- __Running TSLint via__: gulp

#### TypeScript code being linted

```ts
export function f<T>(this: Observable<T>): Observable<T> {
 return this.map(i=>i); // this here is defined by parameter
}
```

with `tslint.json` configuration:

```json
"no-invalid-this": true
```

#### Actual behavior
Error

#### Expected behavior
No error

ajafff:
Fixed by #3267

PR #3267, opened 2017-10-01
[bug-fix] no-invalid-this / ignore functions w "this" param

aervin:
#### PR checklist

- [x] Addresses an existing issue: [#3022](https://github.com/palantir/tslint/issues/3022)
  - [x] Includes tests
- [ ] Documentation update

#### Overview of change:
Functions with contextual "this" will be ignored.

Are additional test cases needed?

### DIFF
--- a/src/rules/noInvalidThisRule.ts
+++ b/src/rules/noInvalidThisRule.ts
@@ -19,6 +19,8 @@ import * as ts from "typescript";
 
 import * as Lint from "../index";
 
+import { isThisParameter } from "tsutils";
+
 const OPTION_FUNCTION_IN_METHOD = "check-function-in-method";
 const DEPRECATED_OPTION_FUNCTION_IN_METHOD = "no-this-in-function-in-method";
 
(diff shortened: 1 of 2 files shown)


test/rules/no-invalid-this/enabled/test.ts.lint
```diff
@@ -1,3 +1,21 @@
+export function f<T>(this: Observable<T>): Observable<T> {
+    const nestedFunction = (this) => {
```
ajafff: arrow functions cannot have a `this` parameter
We don't need to worry about it, because TypeScript shows an error. But we shouldn't use it in any test

src/rules/noInvalidThisRule.ts
```diff
@@ -96,3 +101,7 @@ function walk(ctx: Lint.WalkContext<boolean>): void {
         ts.forEachChild(node, cb);
     });
 }
+
+function hasContextualThis(node: ts.FunctionDeclaration): boolean {
```
ajafff: please rename to `hasThisParameter` (or completely inline it)
This has nothing to do with **contextual** `this`. It's about explicitly declared `this` parameters.

Contextual `this` looks as follows:

```ts
declare function callWithThis(fn: (this: Array<any>) => void): void;

callWithThis(function() {
    console.log(this.length); // this is allowed, because `this` is contextually typed to be an array
});
```
To detect that you would need the TypeChecker. I don't think that's worth it, because TypeScript provides the option `--noImplicitThis` which makes this rule obsolete.

ajafff:
Thanks @aervin

Issue #2953, opened 2017-06-24
make object-literal-shorthand configurable


soheilpro:
Is it possible to make ```object-literal-shorthand``` configurable?

I believe using object literal shorthands is a bad idea and I would like to disallow them in my code. But currently, ```object-literal-shorthand``` rule is not configurable and always assumes they must be enforced.

adidahiya:
ok, we'd accept a PR that adds rule option for `"never"` and `"always"` (defaulting to `"always"`)

ajafff:
Fixed by #3268

vyshkant:
@adidahiya @ajafff WDYT about providing other options like in https://eslint.org/docs/rules/object-shorthand?

PR #3268, opened 2017-10-01
[enhancement] object-literal-shorthand / config to disallow shorthand

aervin:
#### PR checklist

- [x] Addresses an existing issue: [#2953](https://github.com/palantir/tslint/issues/2953)
  - [x] Includes tests
- [x] Documentation update

#### Overview of change:
Can pass argument "never" to object-literal-shorthand rule to disallow shorthand syntax. Currently, default is "always".


src/rules/objectLiteralShorthandRule.ts
```diff
@@ -15,46 +15,76 @@
...
     public apply(sourceFile: ts.SourceFile): Lint.RuleFailure[] {
-        return this.applyWithFunction(sourceFile, walk);
+        return this.applyWithFunction(sourceFile, walk, {
+            allowShorthandAssignments: this.ruleArguments.indexOf("never") === -1,
+        });
     }
 }
 
-function walk(ctx: Lint.WalkContext<void>) {
+function walk(ctx: Lint.WalkContext<Options>) {
```
ajafff: The branches for both options don't share any code. Consider splitting into two separate functions. That will make it easier to understand and avoids unnecessary CPU cycles.

src/rules/objectLiteralShorthandRule.ts
```diff
@@ -15,46 +15,76 @@
...
+    const { options: { allowShorthandAssignments } } = ctx;
     return ts.forEachChild(ctx.sourceFile, function cb(node): void {
-        if (isPropertyAssignment(node)) {
-            if (node.name.kind === ts.SyntaxKind.Identifier &&
+        if (isShorthandAssignment(node) && !allowShorthandAssignments) {
+            ctx.addFailureAtNode(node, Rule.SHORTHAND_ASSIGNMENT);
```
ajafff: fixing the shorthand to longhand syntax shouldn't be too hard I guess

src/rules/objectLiteralShorthandRule.ts
```diff
@@ -63,9 +93,16 @@ function walk(ctx: Lint.WalkContext<void>) {
     });
 }
 
-function handleLonghandMethod(name: ts.PropertyName, initializer: ts.FunctionExpression, sourceFile: ts.SourceFile): [string, Lint.Fix] {
+function handleLonghandMethod(
+    name: ts.PropertyName,
+    initializer: ts.FunctionExpression,
+    sourceFile: ts.SourceFile,
+): [string, Lint.Fix] {
```
ajafff: Please avoid all these purely stylistic changes. They don't add any value to the PR and make the diff harder to reason about.

aervin: I think I have a TS formatter formatting on save. I'll take care of it.

src/rules/objectLiteralShorthandRule.ts
```diff
@@ -63,6 +103,19 @@ function walk(ctx: Lint.WalkContext<void>) {
     });
 }
 
+function fixShorthandToLonghand(node: ts.ShorthandPropertyAssignment | ts.MethodDeclaration): Lint.Fix {
+    let replacementText = isMethodDeclaration(node) ? ": function" : `: ${node.name.getText()}`;
```
ajafff: you can substitute `node.name.getText()` with `node.name.text`

test/rules/object-literal-shorthand/never/test.ts.lint
```diff
@@ -0,0 +1,52 @@
...
+  b: "hello",
+  c: 'c',
+  ["a" + "nested"]: {
+    x: x
+  }
+};
+
+const asyncFn = {
+  foo: async function() {},
+  bar: async function*() {}
```
ajafff: needs a test for async shorthand

src/rules/objectLiteralShorthandRule.ts
```diff
@@ -63,6 +103,19 @@ function walk(ctx: Lint.WalkContext<void>) {
...
+
+    replacementText = isGeneratorFunction(node) ? `${replacementText}*` : replacementText;
+
+    const fixes: Lint.Fix = [Lint.Replacement.appendText(node.name.end, replacementText)];
+    if (isGeneratorFunction(node)) {
+        const asteriskPosition = (node as ts.MethodDeclaration).asteriskToken!.getStart();
+        fixes.push(Lint.Replacement.replaceFromTo(asteriskPosition, asteriskPosition + 1, ""));
+    }
```
ajafff: needs special handling for `async` methods as well.

before you do that, just a suggestion to simplify the fix: replace the name with the `function` keyword and insert the name plus the colon at the start of the declaration.
That way you don't have to handle the async keyword or the asterisk at all

src/configs/all.ts
```diff
@@ -204,7 +204,7 @@ export const rules = {
     "no-unnecessary-type-assertion": true,
     "number-literal-format": true,
     "object-literal-key-quotes": [true, "consistent-as-needed"],
-    "object-literal-shorthand": true,
+    "object-literal-shorthand": [true, "always"],
```
ajafff: With the current implementation `"always"` is not a valid option.

aervin:
I'm struggling to simplify the fixer function b/c of this case:
```
var obj4 = {
  async* f() {
   yield 1;
   yield 2;
   yield 3;
  }
};
```

src/rules/objectLiteralShorthandRule.ts
```diff
@@ -63,6 +98,32 @@ function walk(ctx: Lint.WalkContext<void>) {
     });
 }
 
+function fixShorthandToLonghand(node: ts.ShorthandPropertyAssignment | ts.MethodDeclaration): Lint.Fix {
+    const isGenerator = (node as ts.MethodDeclaration).asteriskToken !== undefined;
+    const isAsync = hasModifier(node.modifiers, ts.SyntaxKind.AsyncKeyword);
+    let
+    replacementStart = (isAsync && node.modifiers !== undefined) ? getModifier(node, ts.SyntaxKind.AsyncKeyword)!.getStart() : -1;
+    replacementStart = (isGenerator && !isAsync) ? (node as ts.MethodDeclaration).asteriskToken!.getStart() : -1;
+    replacementStart = replacementStart === -1 ? node.name.getStart() : replacementStart;
```
ajafff: the `async` keyword is the only modifier allowed here, so `replacementStart` can always be `node.getStart()`

src/rules/objectLiteralShorthandRule.ts
```diff
@@ -63,6 +98,32 @@ function walk(ctx: Lint.WalkContext<void>) {
...
+
+    const fixes: Lint.Fix = [
+        isMethodDeclaration(node)
+            ? Lint.Replacement.replaceFromTo(
+                    replacementStart,
+                    node.name.end,
+                    `${node.name.getText()}:${isAsync ? " async" : ""} function${isGenerator ? "*" : ""}`,
+                )
+            /* tslint:disable:trailing-comma */
+            : Lint.Replacement.appendText(node.name.getStart(), `${node.name.text}: `)
```
ajafff: consider moving the fix for ShorthandPropertyAssignment out of the function. It will make many conditions and type assertions unnecessary

src/rules/objectLiteralShorthandRule.ts
```diff
@@ -78,3 +139,10 @@ function handleLonghandMethod(name: ts.PropertyName, initializer: ts.FunctionExp
     }
     return [prefix + sourceFile.text.substring(nameStart, name.end), fix];
 }
+
+function isShorthandAssignment(node: ts.Node): boolean {
+    return (
+        isShorthandPropertyAssignment(node) ||
+        (isMethodDeclaration(node) && node.parent !== undefined ? node.parent.kind === ts.SyntaxKind.ObjectLiteralExpression : false)
```
ajafff: `node.parent` cannot be undefined
also no need for the conditional expression, just use `&&`

test/rules/object-literal-shorthand/never/test.ts.lint
```diff
@@ -0,0 +1,62 @@
...
+  x,
+  ~ [OBJECT_LITERAL_DISALLOWED]
+  a: 123,
+  b: "hello",
+  c: 'c',
+  ["a" + "nested"]: {
+    x: x
+  }
+};
+[OBJECT_LITERAL_DISALLOWED]: Shorthand property assignments have been disallowed.
```
ajafff: please add a test case with class methods. these should be ignored by the rule

aervin: You mean class members which instantiate object literals with shorthand props?

ajafff: I mean regular method declarations inside a class. They are ignored by the rule. I just want a test to ensure this won't regress.

```ts
class C {
    doStuff() {} // shouldn't be an error
}
```

aervin: Ahh of course. Thanks

src/rules/objectLiteralShorthandRule.ts
```diff
@@ -15,33 +15,70 @@
...
+        );
     }
 }
 
-function walk(ctx: Lint.WalkContext<void>) {
+function disallowShorthandWalker(ctx: Lint.WalkContext<void>) {
+    return ts.forEachChild(ctx.sourceFile, function cb(node): void {
+        if (isShorthandAssignment(node)) {
+            ctx.addFailureAtNode(
+                node,
```
ajafff: looking at the tests, I think it's better to just show the error at `node.name`

aervin: Went ahead and updated the error messages for shorthand-only too

ajafff:
> Went ahead and updated the error messages for shorthand-only too

I was just about to comment on that. I think showing the error only at the name is a bit too narrow for those errors.
If you revert that, I think this PR is good to go. Sorry for the back and forth.

aervin:
Haha maybe a bit presumptuous on my part. Anyway, we made it! Thanks as always for your help/patience.

ajafff:
Thanks @aervin

Issue #3136, opened 2017-08-15
prefer-conditional-expression: check returns in addition to assignments


ajafff:
`prefer-conditional-expression` complains about this code:

```ts
let v: number;
if (condition) {
    v = 1;
} else {
    v = 2;
}
// prefer
const v = condition ? 1 : 2;
```

I'd like the rule to also check `return`. That means it should complain about this code as well:

```ts
if (condition) {
    return 1;
} else {
    return 2;
}
// prefer
return condition ? 1 : 2;
```

I don't see a difference to the existing behavior. That's why I think we don't need to add an option for this new check.

JoshuaKGoldberg:
☠️ *TSLint's time has come!* ☠️ 

TSLint is no longer accepting most feature requests per #4534. See **[typescript-eslint.io](https://typescript-eslint.io)** for the new, shiny way to lint your TypeScript code with **ESLint**. ✨ 

It was a pleasure open sourcing with you all!

JoshuaKGoldberg:
🤖 Beep boop! 👉  **TSLint is deprecated** 👈 _(#4534)_  and you should **switch to [typescript-eslint](https://typescript-eslint.io)**! 🤖 

🔒 This issue is being locked to prevent further unnecessary discussions. Thank you! 👋

PR #3271, opened 2017-10-03
check for return statements in conditional expressions

ivanryuu:
#### PR checklist

- [x] Addresses an existing issue: #3136 
  - [x] Includes tests
- [ ] Documentation update

#### Overview of change:
[enhancement] the prefer-conditional-expression now checks for return statements

### DIFF
--- a/src/language/walker/ruleWalker.ts
+++ b/src/language/walker/ruleWalker.ts
@@ -57,11 +57,7 @@ export class RuleWalker extends SyntaxWalker implements IWalker {
     }
 
     public hasOption(option: string): boolean {
-        if (this.options !== undefined) {
-            return this.options.indexOf(option) !== -1;
-        } else {
-            return false;
-        }
+        return this.options !== undefined ? this.options.indexOf(option) !== -1 : false;
     }
 
     /** @deprecated Prefer `addFailureAt` and its variants. */
--- a/src/language/walker/ruleWalker.ts
+++ b/src/language/walker/ruleWalker.ts
@@ -57,7 +57,7 @@ export class RuleWalker extends SyntaxWalker implements IWalker {
     }
 
     public hasOption(option: string): boolean {
-        return this.options !== undefined ? this.options.indexOf(option) !== -1 : false;
+        return this.options !== undefined && this.options.indexOf(option) !== -1;
     }
 
     /** @deprecated Prefer `addFailureAt` and its variants. */
--- a/src/rules/preferConditionalExpressionRule.ts
+++ b/src/rules/preferConditionalExpressionRule.ts
@@ -15,7 +15,7 @@
  * limitations under the License.
  */
 
-import { isBinaryExpression, isBlock, isExpressionStatement, isIfStatement, isSameLine } from "tsutils";
+import { isBinaryExpression, isBlock, isExpressionStatement, isIfStatement, isReturnStatement, isSameLine  } from "tsutils";
 import * as ts from "typescript";
 
 import * as Lint from "../index";
@@ -45,6 +45,8 @@ export class Rule extends Lint.Rules.AbstractRule {
     };
     /* tslint:enable:object-literal-sort-keys */
 
+    public static FAILURE_STRING_RETURN = "Use a conditionial expression instead of returning in multiple places.";
+
     public static FAILURE_STRING(assigned: string): string {
         return `Use a conditional expression instead of assigning to '${assigned}' in multiple places.`;
     }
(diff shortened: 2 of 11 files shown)


palantirtech:
Thanks for your interest in [palantir/tslint](https://github.com/palantir/tslint), @ivanryuu! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

src/rules/preferConditionalExpressionRule.ts
```diff
@@ -107,3 +117,20 @@ function getAssigned(node: ts.Statement, sourceFile: ts.SourceFile): ts.Expressi
 function nodeEquals(a: ts.Node, b: ts.Node, sourceFile: ts.SourceFile): boolean {
     return a.getText(sourceFile) === b.getText(sourceFile);
 }
+
+function detectReturn(ifStatement: ts.IfStatement): boolean {
```
ajafff: instead of duplicating the `detect` function, I'd suggest modifying `getAssigned`:
* add a condition for ReturnStatement
* return an object to indicate what kind of statement it is (return or assignment) including the expression if it is an assignment
* `nodeEquals` checks if either both are returns OR both are assignments and the expressions are equal

Otherwise `detectReturn` and `hasReturn` would also need to handle the `checkElseIf` option

test/rules/prefer-conditional-expression/check-else-if/test.ts.lint
```diff
@@ -51,4 +51,33 @@ if (true) {
...
+~~ [err2]
+    return 1;
+} else {
+    return 2;
+}
+
+if (true) {
+    return 1;
+} else if (true) {
+       ~~ [err2]
```
ajafff: IMO the error should be located only at the first `if`

test/rules/prefer-conditional-expression/default/test.ts.lint
```diff
@@ -50,4 +50,33 @@ if (true) {
...
+~~ [err2]
+    return 1;
+} else {
+    return 2;
+}
+
+if (true) {
+    return 1;
+} else if (true) {
+       ~~ [err2]
```
ajafff: without `"check-else-if"` there should be no error on `if-elseif-else` statements

test/rules/prefer-conditional-expression/check-else-if/test.ts.lint
ivanryuu: I added it there because only the last set of such statements, IMO, should be changed to a conditional expression. When you have more than two return statements in an if-elseif-block, it would quickly get messy, and I thought that it would be clearer to show it there.

```javascript
if (x) {
    return 1;
} else if (y) {
   return 2;
} else {
   return 3;
}

// would become
if (x) {
   return 1;
}
return y ? 2 : 3;

// instead of
return x ? 1 : (y ? 2 : 3);
```

ajafff: That's why we have the option `"check-else-if"`. If enabled, you need to write these nasty nested conditional expressions. If not enabled, the rule completely ignores such cases

ajafff:
Please run `yarn verify`. The new check results in some lint errors in this project that need to be fixed.

src/rules/preferConditionalExpressionRule.ts
```diff
@@ -56,54 +58,86 @@ export class Rule extends Lint.Rules.AbstractRule {
...
+    ts.forEachChild(thenStatement, cb);
+}
+
+function detect({ thenStatement, elseStatement }: ts.IfStatement, sourceFile: ts.SourceFile, elseIf: boolean, count: number = 1)
+    : ConditionalExpressionWrapper | undefined {
     if (elseStatement === undefined || !elseIf && elseStatement.kind === ts.SyntaxKind.IfStatement) {
         return undefined;
     }
-    const elze = isIfStatement(elseStatement) ? detect(elseStatement, sourceFile, elseIf) : getAssigned(elseStatement, sourceFile);
+    const elze = isIfStatement(elseStatement) ? detect(elseStatement, sourceFile, elseIf, ++count) : getWrapper(elseStatement, sourceFile);
```
ajafff: why do you need the `count` parameter? it seems to be unused

src/language/walker/ruleWalker.ts
```diff
@@ -57,11 +57,7 @@ export class RuleWalker extends SyntaxWalker implements IWalker {
     }
 
     public hasOption(option: string): boolean {
-        if (this.options !== undefined) {
-            return this.options.indexOf(option) !== -1;
-        } else {
-            return false;
-        }
+        return this.options !== undefined ? this.options.indexOf(option) !== -1 : false;
```
ajafff: theres no need for the conditional expression, just use `return this.options !== undefined && this.options.indexOf(option) !== -1`

src/utils.ts
```diff
@@ -100,11 +96,7 @@ export function stripComments(content: string): string {
         } else if (m4 !== undefined) {
             // A line comment. If it ends in \r?\n then keep it.
             const length = m4.length;
-            if (length > 2 && m4[length - 1] === "\n") {
-                return m4[length - 2] === "\r" ? "\r\n" : "\n";
-            } else {
-                return "";
-            }
+            return length > 2 && m4[length - 1] === "\n" ? (m4[length - 2] === "\r" ? "\r\n" : "\n") : "";
```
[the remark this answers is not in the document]
ivanryuu: I noticed that bug in #3350 while fixing the lint errors and added some code to address it. I also added some tests that showed it was working.

adidahiya:
Enhancement vs. breaking change: generally I see it as an enhancement but I'm sure it will cause lots of upgrade pain, so I think we should treat it as a break and only enable this new check with a new rule option. We can later choose to make the option enabled by default in the next major version.

johnwiseheart:

Issue #3253, opened 2017-09-27
New rule: newline-per-chained-call


jaufgang:
Please add a rule based on this one from tslint:

https://eslint.org/docs/rules/newline-per-chained-call

This rule requires a newline after each call in a method chain or deep member access. Computed property accesses such as instance[something] are excluded.

Chained method calls on a single line without line breaks are harder to read, so some developers place a newline character after each method call in the chain to make it more readable and easy to maintain.

**Fails:**
```js
d3.select("body").selectAll("p").data([4, 8, 15, 16, 23, 42 ]).enter().append("p").text(function(d) { return "I'm number " + d + "!"; });
```

**Passes:**
```js
d3
    .select("body")
    .selectAll("p")
    .data([
        4,
        8,
        15,
        16,
        23,
        42
    ])
    .enter()
    .append("p")
    .text(function (d) {
        return "I'm number " + d + "!";
    });
```

ajafff:
Some thoughts:

Add an option to allow the first property access to be on the same line?
```ts
foo.doStuff()
    .doMoreStuff()
    .finish();
```
What about this:
```ts
someArray
    .map(
        (el) => el,
    ).filter(
        (el) => !!el,
    ).forEach(
        (el) => console.log(el),
    );
```

Does the threshold depend on the number of calls or the number of property accesses?
```ts
some.deeply.nested.property.pop();
// vs.
some
    .deeply
    .nested
    .property
    .pop();
```

aervin:
I like this rule idea. Personally, I prefer that the first prop be on its own line. I also think the rule should depend on the number of property accesses, not method calls alone. 

Does styling inline callback args deserve its own rule?

JoshuaKGoldberg:
This was just released - issue can be closed? :)

mmkal:
@ajafff @adidahiya @aervin can I make a request that this rule be made configurable? It's pretty much unusable as-is in projects that use jest, because this sort of thing is a violation:
```js
expect(foo).toEqual(bar)
```
i.e. if we were to turn it on, every single test assertion would have to be 'fixed'. And after fixing, they'd look worse.

Some suggestions for configuration options:

#### `all-or-nothing`
```json
{ "rules": { "newline-per-chained-call": "all-or-nothing" } }
```
either all chained methods get a new line, or none of them do. e.g. these are ok:
```js
someObj.foo().bar()
someObj
    .foo()
    .bar()
```
but this isn't:
```js
someObj.foo()
    .bar()
[...]

PR #3278, opened 2017-10-04
[new-rule] newline-per-chained-call

aervin:
#### PR checklist

- [x] Addresses an existing issue: [#3253](https://github.com/palantir/tslint/issues/3253)
  - [x] Includes tests
- [x] Documentation update

#### Overview of change:
Verifies that chains containing more than n expressions be broken apart onto new lines. 

Configuration is still up in the air. It would be very easy to allow users to configure a "threshold" for the number of chained calls on a single line:
```
{
  "rules": {
    "newline-per-chained-call": [true, 2 /* the max chain length permitted on a single line */]
  }
}
```
Additionally, there's the question of the `this` keyword. Should it count as a "link" in the chain? Currently, `this` is ignored. Maybe an `ignore-this-keyword` config option would be useful?

The description/rationale may not be sufficient or what the original poster had in mind. Happy to change those.

### Changelog

[new-rule] `newline-per-chained-call`


src/configs/all.ts
```diff
@@ -193,6 +193,7 @@ export const rules = {
     "match-default-export-name": true,
     "new-parens": true,
     "newline-before-return": true,
+    // "newline-per-chained-call": Errors
```
ajafff: I don't know if we should exclude it. We should rather find a reasonable default value.

From the "Errors" I can deduce that there are many lint errors within this project if you enable the rule. In this case you can just disable it in `tslint.json`

src/rules/newlinePerChainedCallRule.ts
```diff
@@ -0,0 +1,111 @@
...
+interface Options {
+    maxChainLength: number;
+}
+
+export class Rule extends Lint.Rules.AbstractRule {
+    /* tslint:disable:object-literal-sort-keys */
+    public static metadata: Lint.IRuleMetadata = {
+        ruleName: "newline-per-chained-call",
+        description: Lint.Utils.dedent`
+            Requires that chained property accessors be broken apart onto separate lines.`,
```
ajafff: `chained property accessors` or `chained method calls`?

src/rules/newlinePerChainedCallRule.ts
```diff
@@ -0,0 +1,111 @@
...
+        );
+    }
+}
+
+function getChainLength(node: ts.PropertyAccessExpression): number {
+    let chainLength = 1;
+    const nextAccessorOrCallExpression = (nextNode: ts.Expression): void => {
+        if (
+            isIdentifier(nextNode) ||
+            (isPropertyAccessExpression(nextNode) && !isThisKeyword(nextNode))
```
ajafff: what's that condition supposed to do? If `nextNode` is a `PropertyAccessExpression` it will never be `ThisKeyword`.

src/rules/newlinePerChainedCallRule.ts
```diff
@@ -0,0 +1,111 @@
...
+        return (
+            chainLength > this.options.maxChainLength &&
+            node.getText().split("\n").length < chainLength
+        );
+    }
+}
+
+function getChainLength(node: ts.PropertyAccessExpression): number {
+    let chainLength = 1;
+    const nextAccessorOrCallExpression = (nextNode: ts.Expression): void => {
```
ajafff: this could just be a `while` loop

src/rules/newlinePerChainedCallRule.ts
ajafff: also note that the corresponding eslint rule only counts method calls and not property accesses

src/rules/newlinePerChainedCallRule.ts
```diff
@@ -0,0 +1,111 @@
...
+        }
+        return { maxChainLength };
+    }
+}
+
+class NewlinePerChainedCallWalker extends Lint.AbstractWalker<Options> {
+    public walk(sourceFile: ts.SourceFile) {
+        const checkForUnbrokenChain = (node: ts.Node): void => {
+            if (this.hasUnbrokenChain(node)) {
+                return this.addFailureAtNode(node, Rule.FAILURE_STRING);
```
ajafff: I don't think the `return` here is a good idea. I know it's there to avoid multiple errors on the same node, but it also ignores everything else that's nested:

```ts
// tslint:disable-line (I intentionally dont'want errors on this line)
just().do().stuff(
    // Unfortunately I don't get an error on the nested expression :(
    some().really().long().method().chain()
);
```

src/rules/newlinePerChainedCallRule.ts
aervin: So method calls on their own line if the chain length--including prop accessors--is greater than n? Or should chain length include method calls only?

ajafff: I guess method calls only is the best to start with. Otherwise you'll get a lot more errors with little benefit.
For example `chai` assertions:
```ts
expect({a: 1}).to.not.have.property('b');
```

aervin: Good point. Your single-line `chai` example looks fine to me, but if it had a third call I'd want each call to be on its own line:
```
expect({ a: 1 })
    .to.not.have
    .property('b').more.accessors // Should this be okay?
    .methodCall(1);
```

aervin: And this makes me realize that my scheme of counting `\n` chars is not very flexible...

ajafff: I guess it's best to check if there is a newline after the method call. No counting at all.
```ts
// allowing
expect({ a: 1 })
    .to.not.have.property('b');
// disallowing
expect({ a: 1 }).to.not.have
    .property('b')
```
The only exception is the first method call:
```ts
// not allowed
foo.bar()
    .baz();
// same with
this.bar().
    .baz();

// allowed
foo
    .bar()
    .baz();
```

aervin: Okay nice, that's a much better solution. I have only one thing I'd change in your examples, and it's purely stylistic. I think property accessors between method calls look better like this:
```
expect({ a: 1 })
    .to.not.have
    .property('b');
```
... the condition being that prop accessors between method calls cannot be on the same line as either method call. I think? It may be too convoluted for a config option. I think this is the behavior I would want in the default setting, though.

src/rules/newlinePerChainedCallRule.ts
```diff
@@ -0,0 +1,66 @@
...
+        return this.applyWithWalker(
+            new NewlinePerChainedCallWalker(sourceFile, this.ruleName, undefined),
+        );
+    }
+}
+
+class NewlinePerChainedCallWalker extends Lint.AbstractWalker<void> {
+    public walk(sourceFile: ts.SourceFile) {
+        const checkForUnbrokenChain = (node: ts.Node): void => {
+            if (isCallExpression(node) && needsNewline(node)) {
```
ajafff: `&& isPropertyAccessExpression(node.expression)`

src/rules/newlinePerChainedCallRule.ts
```diff
@@ -0,0 +1,66 @@
...
+            return ts.forEachChild(node, checkForUnbrokenChain);
+        };
+        return ts.forEachChild(sourceFile, checkForUnbrokenChain);
+    }
+}
+
+function needsNewline(node: ts.CallExpression): boolean {
+    const rawExpressionText = node
+        .getFullText()
+        .substr((node.expression as ts.CallExpression).expression.getFullText().length);
```
ajafff: does the same with less work:
```ts
const rawExpressionText = sourceFile.text.substring((node.expression as ts.PropertyAccessExpression).expression.end, node.end);
```

test/rules/newline-per-chained-call/default/test.ts.lint
```diff
@@ -0,0 +1,78 @@
...
+                        .another(function(result) { return result.hello.Yay! }).wrong();
+                                                                                ~~~~~ [ERROR]
+
+SomeClass.propA.helloYay((a: number) => {
+                ~~~~~~~~ [ERROR]
+    return a + 1;
+});
+
+this.some.nested();
+          ~~~~~~ [ERROR]
```
ajafff: I don't consider one method call a chain. Either always use a minimum of 2 or make the threshold configurable

src/rules/newlinePerChainedCallRule.ts
```diff
@@ -0,0 +1,66 @@
...
+                    Rule.FAILURE_STRING,
+                );
+            }
+            return ts.forEachChild(node, checkForUnbrokenChain);
+        };
+        return ts.forEachChild(sourceFile, checkForUnbrokenChain);
+    }
+}
+
+function needsNewline(node: ts.CallExpression): boolean {
```
ajafff: Greatly simplified version, should work too:
```ts
function needsNewline(node: ts.PropertyAccessExpression, sourceFile: ts.SourceFile): boolean {
    return !isSameLine(sourceFile, node.expression.end, node.name.pos);
}
```

src/rules/newlinePerChainedCallRule.ts
```diff
@@ -0,0 +1,66 @@
...
+        );
+    }
+}
+
+class NewlinePerChainedCallWalker extends Lint.AbstractWalker<void> {
+    public walk(sourceFile: ts.SourceFile) {
+        const checkForUnbrokenChain = (node: ts.Node): void => {
+            if (isCallExpression(node) && needsNewline(node)) {
+                this.addFailureAtNode(
+                    (node.expression as ts.PropertyAccessExpression).name,
```
ajafff: Does it make sense to include the dot in the error's range?
That makes more sense if the dot is on the previous line:
```ts
foo.
    bar().
    baz();
```

Would be as simple as
```ts
this.addFailure(node.expression.name.pos - 1, node.expression.name.end, Rule.FAILURE_STRING);
```

src/rules/newlinePerChainedCallRule.ts
```diff
@@ -0,0 +1,85 @@
...
+                callExists = true;
+                return;
+            }
+            return checkForCall(child.expression as
+                | ts.CallExpression
+                | ts.PropertyAccessExpression);
+        }
+    };
+    ts.forEachChild(node, checkForCall);
+    return callExists;
```
ajafff: `ts.forEachChild` is not the right choice here. The implementation would complain about this:
```ts
this.foo(bar());
```

Instead you can just recurse into the AST by accessing the `expression` property:
```ts
function hasChildCall(node: ts.PropertyAccessExpression): boolean {
    let {expression} = node;
    while (isPropertyAccessExpression(expression) || isElementAccessExpression(expression)) {
        ({expression} = expression);
    }
    return expression.kind === ts.SyntaxKind.CallExpression;
}
```

also note that you need to handle `ElementAccessExpression` for cases like this:
```ts
this.getFoo()[0].toString();
```

src/rules/newlinePerChainedCallRule.ts
```diff
@@ -0,0 +1,85 @@
...
+    }
+}
+
+function hasChildCall(node: ts.CallExpression): boolean {
+    let callExists = false;
+    const checkForCall = (child: ts.CallExpression | ts.PropertyAccessExpression): void => {
+        if (isCallExpression(child) || isPropertyAccessExpression(child)) {
+            if (isCallExpression(child)) {
+                callExists = true;
+                return;
```
ajafff: if you `return true` here, `forEachChild` will return early without visiting the remaining children.
you could then use `return ts.forEachChild(child, checkForCall)` and `return ts.forEachChild(node, checkForCall) === true` below

src/rules/newlinePerChainedCallRule.ts
```diff
@@ -0,0 +1,85 @@
...
+            }
+            return checkForCall(child.expression as
+                | ts.CallExpression
```
ajafff: this line will not be reached for `CallExpression`
but the assertion is unnecessary anyway

test/rules/newline-per-chained-call/default/test.ts.lint
```diff
@@ -0,0 +1,79 @@
+const y: string[] = _observable
```
ajafff: please add the two test cases from my comment above

and in addition:

```ts
this.foo()["bar"](); // no error here
foo().bar(); // error here? the current implementation adds one
```

src/rules/newlinePerChainedCallRule.ts
ajafff: nevermind, misread your code

src/rules/newlinePerChainedCallRule.ts
ajafff: I see you already kinda did it this way. I was confused by the closure and the call to `forEachChild`.
As noted above, `forEachChild` is wrong here. And once you get rid of it, you don't need the callback. So you will end up with a loop like my example code.

ajafff:
How did we decide about the first method call? Is it allowed on the first line or not?
```ts
this.foo() // is valid with the current implementation
    .bar();
```

aervin:
Thanks for the feedback! Concerning the first method call, I was thinking of the `this` keyword in particular and how it's exactly 4 characters (a common indentation length) and how often the keyword is used... Basically, I'm not sure about enforcing this
```typescript
this
    .foo()
    .bar()
```
when this
```typescript
this.foo()
    .bar()
```
saves a line and looks just as orderly.

ajafff:
I looked at the examples of the ESLint rule again. They seem to allow the first call on the same line. So I guess we can leave the current behavior unchanged.

```ts
_.chain({})
  .map(foo)
  .filter(bar);
```

In addition they also allow property access before the method call. Our current implementation doesn't allow this. I still don't have a strong opinion on this one. Maybe we can just add an option later if users demand it.
```ts
obj
  .prop.method()
  .method2()
  .method3().prop;
```

Adding a configurable threshold can also be done later.

ChrisHuntCal:
Is there anything I can do to help move this forward?  We would love to be able to implement this into our style guide!

aervin:
I would too @ChrisHuntCal, but lately school and work have left little room for open source. I have no problem with someone continuing this (or taking a fresh crack at it). I do remember there being a PR for an "indentation" helper for tslint. If that was merged in, could this rule have a fixer?

ChrisHuntCal:
Cool!  I'll try to get acquainted with the code and see where I get.

aervin:
Sounds good @ChrisHuntCal. I think the most pressing concern is addressing this case: `this.someMethod()["elementAccessExpression"]()`. The rule should throw at the element access expression but currently ignores them.

ajafff:
@aervin `this.someMethod()["elementAccessExpression"]()` should **not** be an error.
Inserting a line break before the element access can cause a lot of confusion, because it looks like an array.

I just meant you shouldn't stop recursing on an ElementAccessExpression when searching for method calls.

aervin:
@ajafff this may be okay then? I added the following test cases:
```ts
this.getFoo()[0].toString(); // errors here

foo().bar(); // errors here

this.foo()["bar"](); // no problems with this
```

package.json
```diff
@@ -81,4 +81,4 @@
   "engines": {
     "node": ">=4.8.0"
   }
-}
+}
```
ajafff: please revert this change

aervin:
@ajafff `package.json` fixed. CI error is a timeout issue.

JarvisPrestidge:
@aervin @ajafff Any movement on this would be greatly appreciated :slightly_smiling_face:

et:
Hey hopefully I can ask a quick question about this.
Is it possible to enforce the indentation on the newline?

e.g. both of these pass

```ts
  private isValidStageName(name: string): boolean {
    return Object.values(PipelineStages)
    .includes(name);
  }
```

```ts
  private isValidStageName(name: string): boolean {
    return Object.values(PipelineStages)
      .includes(name);
  }
```

but I would only like the second to pass.

PR #3282, opened 2017-10-05
typeCheck ignored when project is unset

smasala:
`typeCheck: true` is ignored if `project` is undefined. This fix sets a reasonable default `project` value.

Made with: @markuskeunecke

#### PR checklist

  - [ ] Includes tests
- [ ] Documentation update

### DIFF
--- a/src/runner.ts
+++ b/src/runner.ts
@@ -173,15 +173,15 @@ async function runLinter(options: Options, logger: Logger): Promise<LintResult> 
     return doLinting(options, files, program, logger);
 }
 
-function resolveFilesAndProgram({ files, project, exclude, outputAbsolutePaths }: Options): { files: string[]; program?: ts.Program } {
+function resolveFilesAndProgram({ files, typeCheck, project, exclude, outputAbsolutePaths }: Options): { files: string[]; program?: ts.Program } {
     // remove single quotes which break matching on Windows when glob is passed in single quotes
     const ignore = arrayify(exclude).map(trimSingleQuotes);
 
-    if (project === undefined) {
+    if (!typeCheck && project === undefined) {
         return { files: resolveGlobs(files, ignore, outputAbsolutePaths) };
     }
 
-    const projectPath = findTsconfig(project);
+    const projectPath = findTsconfig(project || ".");
     if (projectPath === undefined) {
         throw new FatalError(`Invalid option for project: ${project}`);
     }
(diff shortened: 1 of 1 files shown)


palantirtech:
Thanks for your interest in [palantir/tslint](https://github.com/palantir/tslint), @smasala! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

ajafff:
Thanks for putting this PR together. In this case however, it would've been better to open an issue to discuss this before doing the actual work.

Why do you need that change? The documentation clearly states that you can use `--type-check` only with `--project`. When using the CLI you even get an error when using `--type-check` without `--project`.

Since `runner.ts` is not meant to be a public API, I don't see the need for changing the behavior.

Also note that `--type-check` is no longer needed for type checked rules to work and is likely to be deprecated soon.

smasala:
@ajafff `type-check` is required so that `ts loaders` for webpack evaluate checks requiring type information, such as `completed-docs`. Since `type-check:false` (default) is able to find and use the `tsconfig.json` it is not intuitive that the 'project' config is required.

ajafff:
> Since type-check:false (default) is able to find and use the tsconfig.json it is not intuitive that the 'project' config is required.

You are able to find *some* tsconfig.json. You can't be sure to use the right one.

> type-check is required so that ts loaders for webpack evaluate checks requiring type information, such as completed-docs.

I don't get the "ts loaders for webpack" part.
Anyways, you're right that `completed-docs` requires type information. But since [tslint@5.4.0](https://github.com/palantir/tslint/blob/master/CHANGELOG.md#star-non-breaking-api-changes) you **only** need `--project` to make it work. `--type-check` does a lot more than that and is probably not what you want it to do.

smasala:
Then are both the documentation and the `warning` messages are confusing.

If `typeCheck` and `project` are not defined but `completed-docs` in `tslint.json` is, then the `warning`:
```
Warning: The 'completed-docs' rule requires type information.
```
Is confusing, don't you think?

However, rules defined in `tslint.json` that **do not** require `typeChecking` are evaluated regardless and the `tsconfig.json` in this instance is somehow found.

When `typeCheck: true` is set then the output is the same as before. 

Although `typeCheck` is no longer required (as you mentioned above), it is even more confusing that `project` must suddenly be set to evaluate `type requiring` rules.

PR no longer valid. Opened Issue

Issue #3287, opened 2017-10-05
[site update suggestion] Add mention of astexplorer.net


aervin:
### Suggestion
I think most anyone charged with managing a Typescript codebase can see the value in `tslint`, but some may not see a clear path to implementing custom rules. The AST in particular can be mystifying upon first introduction. This [AST explorer tool](http://astexplorer.net/) (and there may be others like it) is very useful for bridging the gap in understanding between the Typescript one writes and the object representation `tslint` uses for linting. I would have benefited personally had such a tool been linked in the docs, so I thought I'd share. I'm proposing something between a mention of this tool in the "Custom rules" portion of the docs, or perhaps even a "Resources" section on the "Develop" page.

ajafff:
Agreed. I think it belongs to "Develop". Having a separate page for "Resources" only makes sense if there's more than one link.
Also make sure to add a note that astexplorer.com is not always up to date. For example right now typescript@latest is 2.5.3 and astexplorer is still on 2.4.1

ajafff:
Fixed by #3295

PR #3295, opened 2017-10-07
[Documentation] adds mention of astexplorer.net to 'Develop' page

aervin:
#### PR checklist

- [x] Addresses an existing issue: [#3287](https://github.com/palantir/tslint/issues/3287)
  - [ ] Includes tests
- [x] Documentation update

#### Overview of change:
Adds mention of [astexplorer.net](http://astexplorer.net/) to the 'Develop' page.

### DIFF
--- a/docs/develop/custom-rules/index.md
+++ b/docs/develop/custom-rules/index.md
@@ -41,7 +41,7 @@ class NoImportsWalker extends Lint.RuleWalker {
 }
 ```
 
-Given a walker, TypeScript's parser visits the AST using the visitor pattern. So the rule walkers only need to override the appropriate visitor methods to enforce its checks. For reference, the base walker can be found in [syntaxWalker.ts](https://github.com/palantir/tslint/blob/master/src/language/walker/syntaxWalker.ts).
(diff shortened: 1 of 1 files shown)


docs/develop/custom-rules/index.md
```diff
@@ -41,7 +41,7 @@ class NoImportsWalker extends Lint.RuleWalker {
 }
 ```
 
-Given a walker, TypeScript's parser visits the AST using the visitor pattern. So the rule walkers only need to override the appropriate visitor methods to enforce its checks. For reference, the base walker can be found in [syntaxWalker.ts](https://github.com/palantir/tslint/blob/master/src/language/walker/syntaxWalker.ts).
+Given a walker, TypeScript's parser visits the AST using the visitor pattern. So the rule walkers only need to override the appropriate visitor methods to enforce its checks. For reference, the base walker can be found in [syntaxWalker.ts](https://github.com/palantir/tslint/blob/master/src/language/walker/syntaxWalker.ts). To see what your Typescript file or snippet looks like as an AST, visit [astexplorer.net](http://astexplorer.net/) (__note__: later versions of Typescript may not be supported).
```
ajafff: Consider rephrasing to:
"(note: current version of TypeScript may not be supported, yet)"

Issue #3232, opened 2017-09-17
await-promise: misses check 'for await (... of ...)'


ajafff:
### Bug Report

- __TSLint version__: 5.7.0
- __TypeScript version__: 2.5.2
- __Running TSLint via__: CLI

#### TypeScript code being linted

```ts
async function main(param: Array<Promise<number>>) {
    for await (const foo of param) {
        foo; // foo is of type Promise<number>, because 'for await' does not await each entry
        // 'for await' is only useful when iterating AsyncIterableIterator<T>
    }
}
```

with `tslint.json` configuration:

```json
"await-promise": true
```

#### Actual behavior
No warning
#### Expected behavior
Warning when using `for await` on anything that is not `AsyncIterableIterator<T>`

PR #3297, opened 2017-10-07
await-promise: check for-await-of

ajafff:
#### PR checklist

- [x] Addresses an existing issue: #3232
  - [x] Includes tests
- [ ] Documentation update

#### Overview of change:
[enhancement] `await-promise` enforces that `for-await-of` is only used with `AsyncIterableIterator`
Fixes: #3232

Depends on: #3296

### DIFF
--- a/src/rules/awaitPromiseRule.ts
+++ b/src/rules/awaitPromiseRule.ts
@@ -41,8 +41,8 @@ export class Rule extends Lint.Rules.TypedRule {
     };
     /* tslint:enable:object-literal-sort-keys */
 
-    public static FAILURE_STRING = "'await' of non-Promise.";
-    public static FAILURE_FOR_AWAIT_OF = "'for-await-of' of non-AsyncIterable.";
+    public static FAILURE_STRING = "Invalid 'await' of a non-Promise value.";
+    public static FAILURE_FOR_AWAIT_OF = "Invalid 'for-await-of' of a non-AsyncIterable value.";
 
     public applyWithProgram(sourceFile: ts.SourceFile, program: ts.Program): Lint.RuleFailure[] {
         const promiseTypes = new Set(["Promise", ...this.ruleArguments as string[]]);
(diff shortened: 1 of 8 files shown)


src/rules/awaitPromiseRule.ts
```diff
@@ -54,27 +55,33 @@ function walk(ctx: Lint.WalkContext<Set<string>>, tc: ts.TypeChecker) {
...
-        return bases !== undefined && bases.some(couldBePromise);
+    function isPromiseType(name: string) {
+        return promiseTypes.has(name);
     }
+}
 
-    function isPromiseType(type: ts.Type): boolean {
-        const { target } = type as ts.TypeReference;
-        return target !== undefined && target.symbol !== undefined && promiseTypes.has(target.symbol.name);
+function containsType(type: ts.Type, predicate: (name: string) => boolean): boolean {
```
adidahiya: seems like a utility we might want in `language/utils.ts`, what do you think?

src/rules/awaitPromiseRule.ts
```diff
@@ -42,6 +42,7 @@ export class Rule extends Lint.Rules.TypedRule {
     /* tslint:enable:object-literal-sort-keys */
 
     public static FAILURE_STRING = "'await' of non-Promise.";
+    public static FAILURE_FOR_AWAIT_OF = "'for-await-of' of non-AsyncIterable.";
```
adidahiya: I realized both of these error messages are a bit too terse -- how about 

- "Invalid 'await' of a non-Promise value"
- "Invalid 'for-await-of' of a non-AsyncIterable value"

src/rules/awaitPromiseRule.ts
ajafff: Makes sense. We also use a similar function in `return-undefined` and probably in other rules, too.
We need to find a better abstraction than the one used here to make it work for other cases. Will do that in a follow-up PR.

Issue #3234, opened 2017-09-18
return-undefined false positive with async function


nicolashenry:
### Bug Report

- __TSLint version__: 5.7.0
- __TypeScript version__: 2.5.2
- __Running TSLint via__: CLI / VSCode

#### TypeScript code being linted

```ts
declare function test(cb: () => Promise<void> | void): void;

test(async () => {
	return;
});
```

with `tslint.json` configuration:

```json
{
  "defaultSeverity": "warning",
  "rules": {
    "return-undefined": true
  }
}
```

#### Actual behavior

```
[tslint] Value-returning function should use `return undefined;`, not just `return;`. (return-undefined)
```

#### Expected behavior

No error

Issue #2812, opened 2017-05-23
return-undefined false positive


cartant:
### Bug Report

- __TSLint version__: 5.3.0
- __TypeScript version__: 2.3.3
- __Running TSLint via__: CLI

#### TypeScript code being linted

```ts
export function t(p: (a: any) => any): void {}
t((a) => { return undefined; });
```

with `tslint.json` configuration:

```json
{
  "rules": {
    "return-undefined": { "severity": "error" }
  }
}

```

#### Actual behavior

```
ERROR: ...: `void` function should use `return;`, not `return undefined;`.
```

#### Expected behavior

No error, as the function parameter is typed as returning `any` - not `void`.

Prior to updating to the latest version of TSLint, the configuration did not see an error effected. No error is effected if 5.2.0 is used.

adidahiya:
looks like a direct result of #2731. notice the test baseline that got changed: https://github.com/palantir/tslint/pull/2731/files#diff-bd90f0718fe7a8725e9db572ef333339R36

can you provide a bit more context around the code your'e trying to lint here? should it really be typed `any`?

... either way, this behavior is kind of surprising, so we should probably revert that part of the rule change @andy-hanson

cartant:
@adidahiya Sure. The typings aren't mine and `any` is reasonably appropriate for the usage.

It's a [`transactionUpdate` parameter](https://github.com/firebase/firebase-js-sdk/blob/v4.0.0/typings/firebase.d.ts#L291) in the Firebase JavaScript client:

```ts
transaction (
  transactionUpdate : (a : any ) => any ,
  onComplete ? : (a : Error | null , b : boolean ,
  c : firebase.database.DataSnapshot | null
) => any , applyLocally ? : boolean ) : firebase.Promise < any > ;
```

I need to implement that interface in a mock.

cartant:
@adidahiya And the code the fails the linting is [the test for the mock](https://github.com/cartant/firebase-nightlight/blob/v2.0.2/source/mock-ref-spec.ts#L1673-L1676):

```ts
return mockRef
    .transaction((value) => {

        return undefined;
    })
    .then(({ committed, snapshot }) => {
```

Firebase transactions return `undefined` if nothing is updated.

cartant:
Seeing as though this is a breaking change, is it going to be considered a bug? Or should I implement the trivial workaround that'll be required if this is indeed how you want the rule to work?

adidahiya:
I consider it a bug. Will label it as such.

PR #3298, opened 2017-10-07
return-undefined: various bug fixes

ajafff:
#### PR checklist

- [x] Addresses an existing issue: #3234, #2812
  - [x] Includes tests
- [ ] Documentation update

#### Overview of change:
[bugfix] `return-undefined` fixed regressions: once again allows anything if return type is `any`
Fixes: #2812
[bugfix] `return-undefined` declared return type takes precedence over contextual type
[bugfix] `return-undefined` handles union return types in async functions
Fixes: #3234

### DIFF
--- a/src/language/utils.ts
+++ b/src/language/utils.ts
@@ -93,11 +93,13 @@ export function someAncestor(node: ts.Node, predicate: (n: ts.Node) => boolean):
     return predicate(node) || (node.parent !== undefined && someAncestor(node.parent, predicate));
 }
 
-export function ancestorWhere<T extends ts.Node>(node: ts.Node, predicate: (n: ts.Node) => boolean): ts.Node | undefined {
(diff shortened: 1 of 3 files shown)


src/language/utils.ts
```diff
@@ -93,11 +93,13 @@ export function someAncestor(node: ts.Node, predicate: (n: ts.Node) => boolean):
     return predicate(node) || (node.parent !== undefined && someAncestor(node.parent, predicate));
 }
 
-export function ancestorWhere<T extends ts.Node>(node: ts.Node, predicate: (n: ts.Node) => boolean): ts.Node | undefined {
+export function ancestorWhere<T extends ts.Node>(node: ts.Node, predicate: (n: ts.Node) => n is T): T | undefined;
```
adidahiya: nice, I like the type guard. do we still need the other non-type-guard function signature?

Issue #2693, opened 2017-05-05
Bypass checking URLs in comments for 'max-line-length' rule


roman-petrov:
The case is we have some long URLs in comments in our code base . These URLs sometimes are quite long, so we have to use `tslint:disable-next-line:max-line-length` before long URLs. My feature request/suggestion is to modify `max-line-length` rule to allow long URLs in comments, or add an option to allow such URLs.

ajafff:
There was a similar request to allow long lines for import statements. IIRC the resolution was that tslint won't fix this and you should just disable the rule for that line. 

Personally I'd like to add some options to the rule, too. 

In the mean you can use `max-len` from `tslint-eslint-rules`

roman-petrov:
Just a suggestion: maybe it makes sense to just add an option like `ignoreRegexp` which will contain array of regular expressions?..

bennyn:
@roman-petrov, we have the same issue with our code. To give an example:

```typescript
/**
 * Definition of OAuth 2.0 Scopes for Google Plus.
 * @see https://developers.google.com/resources/api-libraries/documentation/plus/v1/java/latest/com/google/api/services/plus/PlusScopes.html
 */
export default class PlusScopes {
  // Know the list of people in your circles, your age range, and language
  public static PLUS_LOGIN = 'https://www.googleapis.com/auth/plus.login';
  // Know who you are on Google
  public static PLUS_ME = 'https://www.googleapis.com/auth/plus.me';
  // View your email address
  public static USERINFO_EMAIL = 'https://www.googleapis.com/auth/userinfo.email';
  // View your basic profile info
  public static USERINFO_PROFILE = 'https://www.googleapis.com/auth/userinfo.profile';
}
```

The URL next to `@see` is too long when using `"extends": "tslint:recommended"` in `tslint.json`.

In ESLint there is a [ignoreComments](http://eslint.org/docs/rules/max-len#ignorecomments) option which lets you bypass long comments. Would be nice if TSLint had something similar.

adidahiya:
We'd take a PR for a rule option that exempts URLs from the rule.

xjose97x:
@adidahiya  please also long regular expressions.

adidahiya:
@xjose97x I'm reluctant to add regular expressions for now because of the performance impact. Can you describe your use cases for regex?

xjose97x:
@adidahiya I have a regex for checking email addresses:
[...]

PR #3099, opened 2017-08-04
Feat: ignore pattern for max-line-length

DanielKucal:
#### PR checklist

- [X] Addresses an existing issue: #1123, #1872
- [X] Enhancement
  - [X] Includes tests
- [X] Documentation update - if it is generated from the comments then yes...  However I don't know how to extend parameters taken by `max-line-length` to avoid this warning:
image

#### Overview of change:
We can pass a regexp (as a string) to `max-line-length` rule to ignore some lines, e.g. long imports. Example usage:
```
"max-line-length": [true, 40, "^import |^export | implements "],
```
will ignore the following lines:
```typescript
import { SomeLongInterfaceName } from '../../../nested/directory/structure/target';
class MyClass implements SomeLongInterfaceName, OnInit, OnDestroy, OnViewChange {}
export { MyClass, SomeLongInterfaceName };
```

#### Is there anything you'd like reviewers to focus on?
Please help me get rid of the `Additional items are not allowed` warning.

#### CHANGELOG.md entry:
[new-rule-option], [enhancement] allow to define ignore pattern for `max-line-length` rule

PR #3303, opened 2017-10-08
init implementation; adds config option to ignore urls

aervin:
#### PR checklist

- [x] Addresses an existing issue: [#2693](https://github.com/palantir/tslint/issues/2693)
  - [x] Includes tests
- [ ] Documentation update

#### Overview of change:
Adds config option `"ignore-urls"`. When the option is passed, lines containing URLs are ignored by the rule.

This does not address requests for ignoring regexps. Given the rule's current implementation, is there a way to check for regexps efficiently?


src/rules/maxLineLengthRule.ts
```diff
@@ -49,15 +68,28 @@ export class Rule extends Lint.Rules.AbstractRule {
...
+                }
+            }
+            ctx.addFailureAt(line.pos, line.contentLength, Rule.FAILURE_STRING_FACTORY(ctx.options.limit));
         }
     }
 }
+
+function hasUrl(line: LineRange, sourceFile: ts.SourceFile): boolean {
+    const lineText = sourceFile.text.substr(line.pos, line.contentLength);
+    const forUrl = /[-a-zA-Z0-9@:%._\+~#=]{2,256}\.[a-z]{2,6}\b([-a-zA-Z0-9@:%_\+.~#?&//=]*)/g;
```
ajafff: You don't want to maintain this thing. just use a package like https://www.npmjs.com/package/url-regex

src/rules/maxLineLengthRule.ts
```diff
@@ -49,15 +68,28 @@ export class Rule extends Lint.Rules.AbstractRule {
...
 
-function walk(ctx: Lint.WalkContext<number>) {
-    const limit = ctx.options;
+function walk(ctx: Lint.WalkContext<Options>) {
     for (const line of getLineRanges(ctx.sourceFile)) {
-        if (line.contentLength > limit) {
-            ctx.addFailureAt(line.pos, line.contentLength, Rule.FAILURE_STRING_FACTORY(limit));
+        if (line.contentLength > ctx.options.limit) {
+            if (ctx.options.ignoreUrls) {               //
+                if (hasUrl(line, ctx.sourceFile)) {     // To minimize regex searches
```
ajafff: can be combined with the enclosing condition

src/rules/maxLineLengthRule.ts
```diff
@@ -29,12 +36,24 @@ export class Rule extends Lint.Rules.AbstractRule {
...
+                    minimum: "1",
+                },
+                {
+                    type: "string",
+                    enum: [OPTION_IGNORE_URL],
+                },
+            ],
         },
-        optionExamples: [[true, 120]],
+        optionExamples: [[true, 120], [true, 140, "ignore-urls"]],
```
ajafff: Should be either a number or an object with a `limit` property, see https://github.com/palantir/tslint/pull/3099#discussion_r131516378

src/rules/maxLineLengthRule.ts
```diff
@@ -29,12 +36,24 @@ export class Rule extends Lint.Rules.AbstractRule {
             Limiting the length of a line of code improves code readability.
             It also makes comparing code side-by-side easier and improves compatibility with
             various editors, IDEs, and diff viewers.`,
-        optionsDescription: "An integer indicating the max length of lines.",
+        optionsDescription: Lint.Utils.dedent`
+            An integer indicating the max length of lines. Additionally, the \`\"ignore-urls\"\` option
+            can be passed to ignore lines with long URLs.
```
ajafff: do we really want to ignore the line or just subtract the url's length from the line length?
FWIW ESLint ignores the whole line

ajafff:
> This does not address requests for ignoring regexps. Given the rule's current implementation, is there a way to check for regexps efficiently?

There's the possibility to use `getTokenAtPosition(line.start)` and `getNextToken(...)` until you reach the end of the line. But that's better done in a follow-up PR.

src/rules/maxLineLengthRule.ts
aervin: Won't this search for URLs even if user passes `ignore-urls` option?

src/rules/maxLineLengthRule.ts
aervin: I'm good with either approach. Subtracting line length adds a bit of overhead obviously (there could be multiple URLs on a single line, for instance), but I think it may improve the quality of the rule in general.

ajafff: Adding an option like `"ignore-trailing-comments"` later on will eventually need this logic anyway.
That's why I'm asking for opinions.

aervin: SGTM--the behavior you're suggesting is probably a more intuitive interpretation of `ignore-urls` anyway, now that I think about it.

src/rules/maxLineLengthRule.ts
aervin: I ran into a problem when trying to add the `url-regex` package. I'll look over this again soon. Sorry for the delay.

ajafff: I guess you have problems because there are no types included in the package and also nothing available as `@types/url-regex`?
You can simply define types for the module in this repository. I also had to do this: https://github.com/palantir/tslint/blob/master/src/custom-types.d.ts

src/rules/maxLineLengthRule.ts
```diff
@@ -44,20 +69,52 @@ export class Rule extends Lint.Rules.AbstractRule {
...
+
+function isOptions(arg: number | Options): arg is Options {
+    return (arg as Options).limit !== undefined;
+}
+
+function lineLengthMinusUrl(line: LineRange, sourceFile: ts.SourceFile): number {
+    return sourceFile
+            .text
+            .substr(line.pos, line.contentLength)
+            .replace(urlRegex(), "")
```
ajafff: consider calling `urlRegex` only once and store the result in a constant

src/rules/maxLineLengthRule.ts
```diff
@@ -44,20 +69,52 @@ export class Rule extends Lint.Rules.AbstractRule {
...
+function walk(ctx: Lint.WalkContext<Options>) {
     for (const line of getLineRanges(ctx.sourceFile)) {
-        if (line.contentLength > limit) {
-            ctx.addFailureAt(line.pos, line.contentLength, Rule.FAILURE_STRING_FACTORY(limit));
+        if (
+            line.contentLength > ctx.options.limit &&
+            !(
+                ctx.options.ignoreUrls &&
+                hasUrl(line, ctx.sourceFile) &&
+                lineLengthMinusUrl(line, ctx.sourceFile) < ctx.options.limit
```
ajafff: `<=`

src/rules/maxLineLengthRule.ts
```diff
@@ -44,20 +69,52 @@ export class Rule extends Lint.Rules.AbstractRule {
...
+    return sourceFile
+            .text
+            .substr(line.pos, line.contentLength)
+            .replace(urlRegex(), "")
+            .length;
+}
+
+function parseOptions(args: number[] | Options[]): Options {
+    return isOptions(args[0])
+        ? args[0] as Options
```
ajafff: you should apply default values here as well, because the user can also provide only `Partial<Options>`

src/rules/maxLineLengthRule.ts
```diff
@@ -44,20 +69,52 @@ export class Rule extends Lint.Rules.AbstractRule {
...
+function isOptions(arg: number | Options): arg is Options {
+    return (arg as Options).limit !== undefined;
+}
+
+function lineLengthMinusUrl(line: LineRange, sourceFile: ts.SourceFile): number {
+    return sourceFile
+            .text
+            .substr(line.pos, line.contentLength)
+            .replace(urlRegex(), "")
+            .length;
```
ajafff: I'm wondering what users would expect us to do here.
Consider a limit of 140 and a line with length 180. The line includes 2 urls with 20 characters each.
IMO the rule should still complain. There should be an exception for **long** urls that cannot be broken apart. But short urls can be moved to a separate line without a problem.
I'm thinking about a threshold of 1/2 or 1/3 of the limit.
What do you think?

src/rules/maxLineLengthRule.ts
```diff
@@ -29,12 +36,30 @@ export class Rule extends Lint.Rules.AbstractRule {
             Limiting the length of a line of code improves code readability.
             It also makes comparing code side-by-side easier and improves compatibility with
             various editors, IDEs, and diff viewers.`,
-        optionsDescription: "An integer indicating the max length of lines.",
+        optionsDescription: Lint.Utils.dedent`
+            An integer indicating the max length of lines, or a configuration object with two
+            properties: \`\"ignoreUrls\"\` and \`\"limit\"\`.
```
ajafff: the default of 140 should be documented

src/rules/maxLineLengthRule.ts
aervin: I agree the rule should still complain, however it may be confusing for users: they would explicitly config the rule to `{ ignoreUrls: true }`, but the rule may still complain about URLs sometimes. This could be fixed with something like renaming the option to `ignoreLongUrls`?

ajafff: You're right. Your suggestion makes sense.
Does it make sense to allow users to configure what is considered "long"? For example `{ignore(Long)Urls: 100}`
On the other hand we would repeat that configuration when adding an option to ignore long regular expression literals.

aervin: I'm on the fence. Part of me thinks this is within the scope of a "sensible" default setting. That is, we pick a setting--the `100` from your last example--and enforce it quietly?

ajafff: Agreed. Let's just got with a sensible default and see how it pans out.
We shouldn't use a fixed limit though. I know there are people using a limit of 80 for max line length. Let's just use 1/2 of `limit` to also allow some indentation on the same line.

aervin:
TODO: Tests for lines w multiple, smaller URLs

src/rules/maxLineLengthRule.ts
```diff
@@ -44,20 +71,71 @@ export class Rule extends Lint.Rules.AbstractRule {
...
+    const lineText = sourceFile.text.substr(line.pos, line.contentLength);
+    for (const url of lineText.match(URL_REGEX)!) {
+        if (url.length >= minUrlLength) {
+            longUrlLength += url.length;
         }
     }
+    return lineText.length - longUrlLength;
+}
+
+function parseOptions(args: number[] | Options[]): Options {
```
ajafff: instead of the array, you could pass only the first element. that would make your typeguard work.
the no-unsafe-any failure can be avoided by using `parseOptions(this.ruleArguments[0] as any)`

src/rules/maxLineLengthRule.ts
```diff
@@ -44,20 +71,71 @@ export class Rule extends Lint.Rules.AbstractRule {
...
+                Rule.FAILURE_STRING_FACTORY(ctx.options.limit),
+            );
+        }
+    }
+}
+
+function buildOptions(options: Partial<Options>): Options {
+    options.ignoreLongUrls =
+        options.ignoreLongUrls === undefined ? false : options.ignoreLongUrls;
+    options.limit = options.limit === undefined ? DEFAULT_LIMIT : options.limit;
```
ajafff: consider creating a new object instead of modifying the original. maybe the caller wants to reuse the object later (unlikely but possible)

src/rules/maxLineLengthRule.ts
```diff
@@ -29,12 +37,31 @@ export class Rule extends Lint.Rules.AbstractRule {
             Limiting the length of a line of code improves code readability.
             It also makes comparing code side-by-side easier and improves compatibility with
             various editors, IDEs, and diff viewers.`,
-        optionsDescription: "An integer indicating the max length of lines.",
+        optionsDescription: Lint.Utils.dedent`
+            An integer indicating the max length of lines, or a configuration object with two
+            properties: \`\"ignoreLongUrls\"\` and \`\"limit\"\`. If no configuration is provided, a
```
ajafff: Please add a note that urls are considered long if they are at least half of `limit` and that their length is subtracted from the total line length.

aervin:
Regarding the rule's documentation, I think I'll attempt a more thorough explanation of the configuration object.

aervin:
@ajafff Should we still go through w this one?

ajafff:
@aervin It makes sense to continue here. AFAICT there were no pending changes requested. This only needs a merge and should be good to go.

I'm a little distracted by my own project lately, so ping me here when this is ready for a final review.

JKillian:
@aervin think we should update this, or just let it go? I'm sort of inclined to thing that https://github.com/palantir/tslint/pull/3099 should suffice - it should be easy enough to write a sim ple regex to match most cases of a URL. That said, I'm not opposed to having this option to make things easier for users if you want to take the time to update this

aervin:
@JKillian Happy to let this one go. The only worthwhile PR I have open is for max-line-length, IMO. Close this if you like!

Issue #746, opened 2015-10-19
Whitespace before curly braces


richardm:
When check-branch is true, whitespace is required after an if / then, but not before a corresponding opening curly brace or after a closing curly brace.

As a result, we have an awkward scenario where check-branch only requires code to look like this:

    function(str){
      if (str && str.length > 20){
        return str.substr(0, 20) + '...';
      }else {
        return str;
      }
    }

In my experience, when teams want to enforce spaces after the branch, they also want to enforce spaces before the opening curly on line 2 and after the closing curly on line 4, like this:

    function(str){
      if (str && str.length > 20) { // require space before opening curly
        return str.substr(0, 20) + '...';
      } else { // require space after closing curly
        return str;
      }
    }

In addition, there is no way to require `function(str){` to be formatted with whitespace as `function (str) {`.

Would pull requests adding this functionality be accepted? Any suggestions for naming those rule options?

glen-84:
+1 This is really needed. Somewhat related is #628.

riggerthegeek:
+1. I like to enforce whitespace after
```
function foo () {
}
```

danvk:
@adidahiya @JKillian I'd be interested in implementing this. What would your preference on syntax be? `"whitespace": [true, "check-curly"]`?

glen-84:
`check-brace` maybe?

ollwenjones:
+1 also import statements. if trying to enforce:
```
import { MyThing } from './the.module';
```
over
```
import {MyThing} from './the.module';
```
or else you inevitably end up with
```
import { MyThing} from './the.module';
```

fmontes:
Hi guys,

Is this still no possible?

Regards.

ospfranco:
Any word on this? it would be really good for my team since we are migrating from a pure JS ESLinted with airbnb style guide, and the new TS code does not automatically do this for us.

glen-84:
Of the 3 issues mentioned in the initial comment:

1. Space before opening brace: Handled by `one-line` rule (`check-whitespace`) *and* `whitespace` rule (`check-preblock`)
2. Space after closing brace: No available rules (although I've only looked up until TSLint 4.4.2)
3. Space before function parenthesis: Handled by `space-before-function-paren` rule.

byrondover:
Whitespace rules around destructuring assignment and import / export statements, as @ollwenjones mentioned, is what I'm really after.

[...]

PR #3308, opened 2017-10-11
whitespace: add option check-postblock-prekeyword

yubaoquan:
check-postblock-prekeyword

#### Example:

before lint:
```
try {
}catch {
}finally {
}

if () {
}else {
}
```
after lint:
```
try {
} catch {
} finally {
}


if () {
} else {
}
```

#### PR checklist

- [x] Addresses an existing issue: #746
  - [x] Includes tests
- [x] Documentation update

#### CHANGELOG.md entry:

[new-rule] `check-postblock-prekeyword`
[new-rule]

### DIFF
--- a/src/rules/whitespaceRule.ts
+++ b/src/rules/whitespaceRule.ts
@@ -123,12 +123,7 @@ function walk(ctx: Lint.WalkContext<Options>) {
                 }
                 break;
             }
-            // check for catch block.
-            case ts.SyntaxKind.CatchClause:
-                if (options.postblockPrekeyword && matchPostblockPreKeyword(node)) {
-                    checkForTrailingWhitespace(node.end);
-                }
-                break;
+
             case ts.SyntaxKind.Block:
                 if (options.preblock) {
                     checkForTrailingWhitespace(node.getFullStart());
@@ -320,16 +315,6 @@ function walk(ctx: Lint.WalkContext<Options>) {
                 }
         }
     });
-    function matchPostblockPreKeyword(node: ts.Node): boolean {
-        const parent = node.parent!;
-        const isDoWhile = utils.isDoStatement(parent);
-        const isIfElse = utils.isIfStatement(parent)
-            && parent.elseStatement !== undefined
-            && parent.elseStatement !== node;
-        const isTryCatchFinally = utils.isTryStatement(parent)
-            && (parent.tryBlock === node || parent.catchClause === node);
-        return isDoWhile || isIfElse || isTryCatchFinally;
-    }
     function checkEqualsGreaterThanTokenInNode(node: ts.Node): void {
         if (!options.operator) {
             return;
@@ -370,4 +355,18 @@ function walk(ctx: Lint.WalkContext<Options>) {
         const fix = Lint.Replacement.deleteText(position, 1);
         ctx.addFailureAt(position, 1, Rule.FAILURE_STRING_INVALID, fix);
     }
+}
+
+function matchPostblockPreKeyword(node: ts.Node): boolean {
+    const parent: ts.Node = node.parent!;
+    const isDoWhile = utils.isDoStatement(parent);
+    const isIfElse = utils.isIfStatement(parent)
+        && parent.elseStatement !== undefined
+        && parent.elseStatement !== node;
+    const hasTry = utils.isTryStatement(parent)
+        && parent.tryBlock === node;
+    const hasFinally = utils.isCatchClause(parent)
+        && parent.parent!.finallyBlock !== undefined;
+    const isTryCatchFinally = hasTry || hasFinally;
+    return isDoWhile || isIfElse || isTryCatchFinally;
 }
--- a/src/rules/whitespaceRule.ts
+++ b/src/rules/whitespaceRule.ts
@@ -123,25 +123,18 @@ function walk(ctx: Lint.WalkContext<Options>) {
                 }
                 break;
             }
-
+            // check for catch block.
+            case ts.SyntaxKind.CatchClause:
+                if (options.postblockPrekeyword && matchPostblockPreKeyword(node)) {
+                    checkForTrailingWhitespace(node.end);
+                }
+                break;
             case ts.SyntaxKind.Block:
                 if (options.preblock) {
                     checkForTrailingWhitespace(node.getFullStart());
                 }
-                if (options.postblockPrekeyword) {
-                    const nextToken = utils.getNextToken(node as ts.Block)!;
-                    if (!utils.isSameLine(sourceFile, node.end, nextToken.getStart())) {
-                        break;
-                    }
-                    switch (nextToken.kind) {
-                        case ts.SyntaxKind.WhileKeyword:
-                        case ts.SyntaxKind.ElseKeyword:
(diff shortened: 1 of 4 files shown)


palantirtech:
Thanks for your interest in [palantir/tslint](https://github.com/palantir/tslint), @yubaoquan! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

yubaoquan:
signed the individual CLA on October 12, 2017 at 12:03 AM. @palantirtech

src/rules/whitespaceRule.ts
```diff
@@ -50,14 +51,17 @@ export class Rule extends Lint.Rules.AbstractRule {
             * \`"check-type"\` checks for whitespace before a variable type specification.
             * \`"check-typecast"\` checks for whitespace between a typecast and its target.
             * \`"check-type-operator"\` checks for whitespace between type operators \`|\` and \`&\`.
-            * \`"check-preblock"\` checks for whitespace before the opening brace of a block`,
+            * \`"check-preblock"\` checks for whitespace before the opening brace of a block.
+            * \`"check-postblock-prekeyword"\` checks for whitespace between the closing brace of a block
+            and the following keyword (\`} else\` / \`} catch\` / \`} while\`).`,
```
ajafff: what's the difference to `one-line` with `"check-whitespace"`?

src/rules/whitespaceRule.ts
```diff
@@ -122,6 +128,21 @@ function walk(ctx: Lint.WalkContext<Options>) {
...
+                    const nextToken = utils.getNextToken(node as ts.Block)!;
+                    if (!utils.isSameLine(sourceFile, node.end, nextToken.getStart())) {
+                        break;
+                    }
+                    switch (nextToken.kind) {
+                        case ts.SyntaxKind.WhileKeyword:
+                        case ts.SyntaxKind.ElseKeyword:
+                        case ts.SyntaxKind.CatchKeyword:
+                        case ts.SyntaxKind.FinallyKeyword:
+                            if (node.end === nextToken.getStart()) {
```
ajafff: you can simply use `checkForTrailingWhitespace(node.end)`

src/rules/whitespaceRule.ts
```diff
@@ -122,6 +128,21 @@ function walk(ctx: Lint.WalkContext<Options>) {
                 if (options.preblock) {
                     checkForTrailingWhitespace(node.getFullStart());
                 }
+                if (options.postblockPrekeyword) {
+                    const nextToken = utils.getNextToken(node as ts.Block)!;
+                    if (!utils.isSameLine(sourceFile, node.end, nextToken.getStart())) {
+                        break;
+                    }
+                    switch (nextToken.kind) {
+                        case ts.SyntaxKind.WhileKeyword:
```
ajafff: This may do a bit too much. That `WhileKeyword` could be from the `while (...) {}` statement following the block.
That may not be a big deal, because it will only complain about such code:
```ts
if (condition) { /* code */ }while(condition) {};
                             ~ [missing whitespace]
```

Anyway, there's an easier way to do this check without `getNextToken`:

```ts
const parent = node.parent!;
if (isDoStatement(parent) || // Block is followed by WhiteKeyword
    isIfStatement(parent) && parent.elseStatement !== undefined && parent.thenStatement === node || // Block is followed by ElseKeyword
    isTryStatement(parent) && (
        parent.tryBlock === node || // Block is followed by CatchKeyword or FinallyKeyword
        parent.catchClause !== undefined && parent.catchClause.block === node // Block is followed by FinallyKeyword
    )
) {
    // do stuff
}
```
Well, while writing it down I realized it's not really easier. But it's clearer what kind of statement is checked.

src/rules/whitespaceRule.ts
yubaoquan: I reviewed the `one-line` rule code and its test case. Didn't find that rule solve the situation of this PR. And I tried to add `one-line` with all its options to my test file, the linter didn't report error of missing whitespace between closing bracket and `else`

src/rules/whitespaceRule.ts
yubaoquan: OK. I wrote a function to check conditions as you mentions.

src/rules/whitespaceRule.ts
yubaoquan: Changed as you said.

src/rules/whitespaceRule.ts
ajafff: #3200 includes the check for whitespace before `else`.
The only thing missing is whitespace before `while` in `do {...} while (...);` loops. But that could easily be added.

yubaoquan: So is my pr going to be merged?

src/rules/whitespaceRule.ts
```diff
@@ -306,7 +320,16 @@ function walk(ctx: Lint.WalkContext<Options>) {
...
     });
-
+    function matchPostblockPreKeyword(node: ts.Node): boolean {
+        const parent = node.parent!;
+        const isDoWhile = utils.isDoStatement(parent);
+        const isIfElse = utils.isIfStatement(parent)
+            && parent.elseStatement !== undefined
+            && parent.elseStatement !== node;
+        const isTryCatchFinally = utils.isTryStatement(parent)
+            && (parent.tryBlock === node || parent.catchClause === node);
```
ajafff: You can use the following to avoid the additional `case` for `CatchClause` above (sorry my initial suggestion got that part wrong):
```ts
const isTryCatchFinally = utils.isTryStatement(parent) && parent.tryBlock === node ||
    utils.isCatchClause(parent) && parent.parent!.finallyBlock !== undefined;
```

src/rules/whitespaceRule.ts
```diff
@@ -306,7 +320,16 @@ function walk(ctx: Lint.WalkContext<Options>) {
                 }
         }
     });
-
+    function matchPostblockPreKeyword(node: ts.Node): boolean {
```
ajafff: Consider moving this function out of the `walk` closure. It doesn't use any variables from that scope, so it doesn't need to be here.

src/rules/whitespaceRule.ts
ajafff: I have no preference into which rule the functionality should go. However, we generally try to avoid duplication.
I'll leave that decision to @adidahiya

src/rules/whitespaceRule.ts
yubaoquan: Done

src/rules/whitespaceRule.ts
yubaoquan: Done

src/rules/whitespaceRule.ts
adidahiya: @yubaoquan do you generally use `one-line` with `"check-whitespace"` in your lint config? if so, I think we should just add the additional check @ajafff suggested to `one-line` instead of a new option here.

JKillian:
it's a little confusing, but functionality should go in the `one-line` rule likely, see https://github.com/palantir/tslint/pull/3308#discussion_r153646487

PR #3322, opened 2017-10-15
deprecate --type-check

ajafff:
#### PR checklist

  - [ ] Includes tests
- [x] Documentation update

#### Overview of change:
Deprecated `--type-check`. It's no longer necessary. Most users don't read the changelog and may not have noticed it.
It breaks consumers that expect a certain output format. And it confuses users, because `--exclude` does not work as expected.

We should remove it in v6.0

#### CHANGELOG.md entry:

[deprecation] CLI argument `--type-check` is no longer necessary and will be removed in the next major version

### DIFF
--- a/docs/usage/cli/index.md
+++ b/docs/usage/cli/index.md
@@ -46,7 +46,7 @@ Options:
 -s, --formatters-dir  formatters directory
 -t, --format          output format (prose, json, stylish, verbose, pmd, msbuild, checkstyle, vso, fileslist, codeFrame)  [default: "prose"]
 --test                test that tslint produces the correct output for the specified directory
---type-check          enable type checking when linting a project
+--type-check          (deprecated) enable type checking when linting a project
 -v, --version         current version
 ```
 
@@ -128,8 +128,8 @@ tslint accepts the following command-line options:
     files will be linted.
 
 --type-check
-    Enables the type checker when running linting rules. --project must be
-    specified in order to enable type checking.
+    (deprecated) Enables the type checker when running linting rules.
+    --project must be specified in order to enable type checking.
 
 -v, --version:
     The current version of tslint.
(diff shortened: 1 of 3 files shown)


chrismbarr:
Does this now mean that type checking is always enabled when running tslint?  If so, I like it!

ajafff:
@chrismbarr type checking and enabling typed rules is not the same thing:

Type checking does a full check of your entire project to find compile errors before the actual linting is done. It does not affect the lint result. This is deprecated now and will be removed in a future release. Type checking should be done by the compiler, not TSLint.

Typed rules are not enabled by default. Technically that's not possible. But you only need `--project` now to enable typed rules. Before v5.4.0 you needed `--project --type-check`. The latter is no longer necessary and therefore deprecated.

chrismbarr:
Yes, that's what I meant - thanks!  To be fair though a "typed rule" was enabled with a CLI flag named `--type-check` - so you can see why the terminology is a bit confusing.

But this is a good change, I like it. 1 flag to enable this feature instead of 2 is much better.

mcalthrop:
Just following up on this... I am running `tslint` version `5.7.0`, and using VS Code v1.17.2 (currently the latest release).

Per advice in this PR, I have removed the `--type-check` parameter, and now have this in my `"scripts"` section of my `package.json`:

``` json
"tslint": "tslint --project ./tsconfig.json",
```

The issue I now have is that in VS Code no longer highlights type errors.

I realise that `tslint` is no longer providing this functionality, and any typing error will be flagged at compilation time – but of course I still want to see these errors when developing.

I have done some searching, but not yet found a solution... any ideas?

Thanks in advance.

Matt

mcalthrop:
Oops... looks like it was just some kind of refresh issue. I restarted VS Code a couple of times, and now it's disappeared.

As you were! :-D

Issue #3323, opened 2017-10-16
member-access rule should trigger with readonly parameter property


cyrilgandon:
### Bug Report

- __TSLint version__: 5.7.0
- __TypeScript version__: 2.5.3
- __Running TSLint via__:  VSCode

#### TypeScript code being linted

```ts
class Point {
    constructor(
        readonly x: number,
        readonly y: number
    ) { }
}
```

with `tslint.json` configuration:

```json
{
  "rules": {
    "member-access": [true, "check-accessor"]
  }
}
```

#### Actual behavior
No Error
#### Expected behavior
It should trigger the [`member-access` rule](https://palantir.github.io/tslint/rules/member-access/):
>The class property 'x' must be marked either 'private', 'public', or 'protected' (member-access)

The same code, written differently, triggers the error:
```ts
class Point {
    readonly x: number; // <-- Error
    readonly y: number; // <-- Error
    constructor(x: number, y: number) {
        this.x = x;
        this.y = y;
    }
}
```

Maybe there is a need for a new configuration of this rule, like `check-parameter-property`, because I'm not sure that it make sense to have it under `check-accessor`.

ajafff:
Fixed by #3325

PR #3325, opened 2017-10-16
[enhancement] member-access rule on parameter properties

cyrilgandon:
#### PR checklist

- [x] Addresses an existing issue: #3323
  - [x] Includes tests
- [ ] Documentation update

#### Overview of change:
Check parameter property for rule member-access.
```
class myClass {
    constructor (
        readonly x: number, // <--- trigger member-access rule
        public readonly y: number, // <--- trigger member-access rule if no-public
    ) { }
}
```

Do we need to treat this as an optional configuration, or should it be by default.  
There is already  `no-public`, `check-accessor` and `check-constructor`.   
Does a `check-parameter-property` make sense, or just let the rule be by default like the other properties.

#### CHANGELOG.md entry:

[new-rule-option] `check-parameter-property` for [member access](https://palantir.github.io/tslint/rules/member-access/) rule

### DIFF
--- a/src/rules/memberAccessRule.ts
+++ b/src/rules/memberAccessRule.ts
@@ -15,7 +15,7 @@
  * limitations under the License.
  */
 
-import { getChildOfKind, getModifier, getNextToken, getTokenAtPosition, isClassLikeDeclaration, isConstructorDeclaration } from "tsutils";
+import * as utils from "tsutils";
 import * as ts from "typescript";
 
 import { showWarningOnce } from "../error";
@@ -24,11 +24,13 @@ import * as Lint from "../index";
 const OPTION_NO_PUBLIC = "no-public";
 const OPTION_CHECK_ACCESSOR = "check-accessor";
 const OPTION_CHECK_CONSTRUCTOR = "check-constructor";
+const OPTION_CHECK_PARAMETER_PROPERTY = "check-parameter-property";
 
 interface Options {
     noPublic: boolean;
     checkAccessor: boolean;
     checkConstructor: boolean;
+    checkParameterProperty: boolean;
 }
 
 export class Rule extends Lint.Rules.AbstractRule {
@@ -42,15 +44,16 @@ export class Rule extends Lint.Rules.AbstractRule {
 
             * \`"no-public"\` forbids public accessibility to be specified, because this is the default.
             * \`"check-accessor"\` enforces explicit visibility on get/set accessors
-            * \`"check-constructor"\`  enforces explicit visibility on constructors`,
+            * \`"check-constructor"\`  enforces explicit visibility on constructors
+            * \`"check-parameter-property"\`  enforces explicit visibility on parameter properties`,
         options: {
             type: "array",
             items: {
                 type: "string",
-                enum: [OPTION_NO_PUBLIC, OPTION_CHECK_ACCESSOR, OPTION_CHECK_CONSTRUCTOR],
+                enum: [OPTION_NO_PUBLIC, OPTION_CHECK_ACCESSOR, OPTION_CHECK_CONSTRUCTOR, OPTION_CHECK_PARAMETER_PROPERTY],
             },
             minLength: 0,
-            maxLength: 3,
+            maxLength: 4,
         },
         optionExamples: [true, [true, OPTION_NO_PUBLIC], [true, OPTION_CHECK_ACCESSOR]],
         type: "typescript",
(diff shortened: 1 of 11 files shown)


src/rules/memberAccessRule.ts
```diff
@@ -96,6 +96,13 @@ function walk(ctx: Lint.WalkContext<Options>) {
                 }
             }
         }
+        if (isConstructorDeclaration(node)) {
```
ajafff: you can move this check into the `for...of` loop above and use `child` instead of `node`. A `ConstructorDeclaration` can only occur in `ClassLikeDeclaration.members`.

You can also add `&& child.body !== undefined` to avoid checking constructor overload signatures.

src/rules/memberAccessRule.ts
```diff
@@ -96,6 +96,13 @@ function walk(ctx: Lint.WalkContext<Options>) {
...
+            for (const param of node.parameters) {
+                if (ts.isParameterPropertyDeclaration(param)) {
```
ajafff: prefer `isParameterProperty` from `tsutils` because it does a lot less work.

src/rules/memberAccessRule.ts
```diff
@@ -159,6 +171,8 @@ function typeToString(node: ts.ClassElement): string {
             return "get property accessor";
         case ts.SyntaxKind.SetAccessor:
             return "set property accessor";
+        case ts.SyntaxKind.Parameter:
+            return "class parameter property";
```
ajafff: I think you can omit "class" since parameter properties can only occur in classes.

src/rules/memberAccessRule.ts
```diff
@@ -114,19 +121,24 @@ function walk(ctx: Lint.WalkContext<Options>) {
...
-            const start = publicKeyword.end - "public".length;
-            ctx.addFailure(
-                start,
-                publicKeyword.end,
-                Rule.FAILURE_STRING_NO_PUBLIC,
-                Lint.Replacement.deleteFromTo(start, getNextToken(publicKeyword, ctx.sourceFile)!.getStart(ctx.sourceFile)),
-            );
+            const readonlyKeyword = getModifier(node, ts.SyntaxKind.ReadonlyKeyword);
+            // public is not optional for parameter property without the readonly modifier
+            const isPublicOptional = !ts.isParameterPropertyDeclaration(node) || readonlyKeyword !== undefined;
```
ajafff: you don't need to check for parameter properties here again. If a `ts.ParameterDeclaration` gets passed to this function, it's always a parameter property.
You can replace the method call with `node.kind !== ts.SyntaxKind.Parameter`

ajafff:
also enable the new option in `src/config/all.ts`
and make sure to run `yarn verify` afterwards and fix the new failures in this project caused by these stricter checks

cyrilgandon:
Great feedback, thank you!
There is no more error when I run `yarn verify`, because I added the configuration, and there is only `"member-access": true,` in the tslint.json, with a `// TODO: Enable stricter options for these`.
I think I will be able to apply a stricter check on `member-access` on another PR.

test/rules/member-access/no-public/test.ts.lint
```diff
@@ -6,6 +6,9 @@ class {
     public/*some comment*/ get y() {}
     ~~~~~~ [0]
     public() {}
+    constructor(public readonly param : number) {}
+                ~~~~~~ [0]
+    constructor(public param : number) {}
```
ajafff: add the same for `private` and `protected`, to make sure it won't regress

src/rules/memberAccessRule.ts
```diff
@@ -15,7 +15,7 @@
  * limitations under the License.
  */
 
-import { getChildOfKind, getModifier, getNextToken, getTokenAtPosition, isClassLikeDeclaration } from "tsutils";
+import * as utils from "tsutils";
```
ajafff: We usually tend to avoid namespace imports here. It's fine if the import list spans multiple lines.

Since this is only a matter of personal preference, I leave it up to you whether you want to change it or not.

test/rules/member-access/no-public/test.ts.lint
```diff
@@ -6,6 +6,9 @@ class {
     public/*some comment*/ get y() {}
     ~~~~~~ [0]
     public() {}
+    constructor(public readonly param : number) {}
```
ajafff: consider adding the `public` keyword to the `constructor` to make sure the rule finds both failures

src/rules/memberAccessRule.ts
cyrilgandon: I found both ways in other files, I had to choose between multiple line import or namespace import, but I think I like multiline import more.

test/rules/member-access/parameter-property/test.ts.lint
```diff
@@ -0,0 +1,14 @@
+class Members {
+    constructor(
+        readonly parameterPropertyR : number,
```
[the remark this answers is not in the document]
cyrilgandon: nice idea, didn't think about decorators at all!

ajafff:
Thanks a lot @cyrilgandon 

The generated files are no problem. #3310 removes them from the git repository and ignores that folder.

PR #3332, opened 2017-10-17
Async-ified writing the tslint.json file in --init

JoshuaKGoldberg:
The `runWorker` method is already `async`. Added a general purpose `async` wrapper in utils around `fs.writeFile`.

#### PR checklist

#### Overview of change:

Uses `fs.readFile` within an `async` wrapper under `utils` instead of `fs.readFileSync`.


JoshuaKGoldberg:
GitHub isn't seeing an updated commit for the PR. Very annoying. Will re-create PR...

Issue #3333, opened 2017-10-17
Convert fs.*Sync to async equivalents internally


JoshuaKGoldberg:
Spinning off of #2328: it's generally bad practice to use synchronous `fs` APIs, as they are blocking. Using `async` equivalents will make it easier to integrate with asynchronous code and I'd love to see some internals' file operations parallelized if possible for performance.

ajafff:
Switching to async IO in `runner.ts` and rule tests is definitely something we should do.
While rewriting the code we should try to parallelize IO. Your PRs tend to await the promise immediately, thus making everything run sequential again. That doesn't give any benefits as it only adds more overhead.

Regarding changes to `Linter`:
I don't see any benefit here. Linting is a CPU bound process. There's no IO except for fixes (and maybe some IO in Rules, but currently they cannot be async anyway). I guess users expect a (slightly) increased runtime while autofixing.
My main concern is about compatibility with API users. Some of them need a synchronous API.
I just looked at `tslint-language-service`. It seems the TypeScript language server requires plugins to implement sync APIs. /cc @egamma @angelozerr

We can probably reconsider this as soon as TypeScript starts to support async APIs

JoshuaKGoldberg:
> Your PRs tend to await the promise immediately

Agreed, this is sub-optimal. My hope was to trivially change the APIs to async/await to get the API modified, then rewrite.

> compatibility

I'm up for waiting - will try to remember to come back to this if and when the APIs allow async operations.

egamma:
@ajafff 

>My main concern is about compatibility with API users. Some of them need a synchronous API.
 I just looked at  tslint-language-service . It seems the TypeScript language server requires plugins to implement sync APIs. /cc @egamma @angelozerr

@ajafff is correct, the tslint-language-service runs inside the TypeScript server and the TypeScript host APIs are sync APIs. For example, computing the warnings is expected to be a sync operation, see https://github.com/angelozerr/tslint-language-service/blob/88791766ec122d0cf315af7502d38049549b9837/src/index.ts#L304. An async `lint` call would be breaking and we would also need a sync operation.

The situation is different in the vscode-tslint extension, there async APIs are supported as well and an async `lint` could be consumed, but the benefits are not clear to me yet.

PR #3336, opened 2017-10-17
Async-ified Linter.lint

JoshuaKGoldberg:
#### PR checklist

#### Overview of change:

Changed `Linter.lint` to be `async` and read/write files using the asynchronous `fs` APIs. Necessitated also changing `applyAllFixes` and `applyFixes` within `Linter`, `runTests` and `runTest` within `tests`, and the runner logic in `ruleTestRunner`.

#### CHANGELOG.md entry:

[api] Breaking change: `Linter.lint` is now `async` (returns a `Promise`)


src/test.ts
```diff
@@ -54,18 +54,19 @@ export interface TestResult {
...
+export async function runTests(patterns: string[], rulesDirectory?: string | string[]): Promise<TestResult[]> {
     const files: string[] = [];
     for (let pattern of patterns) {
         if (path.basename(pattern) !== "tslint.json") {
             pattern = path.join(pattern, "tslint.json");
         }
         files.push(...glob.sync(pattern));
     }
-    return files.map((directory: string): TestResult => runTest(path.dirname(directory), rulesDirectory));
+    return await Promise.all(
```
ajafff: Don't `return await` promises, just return the Promise. There are several cases thoughout this PR.

There's already a PR to add a rule to prevent this.

ajafff:
Currently I'm -0 on this one.
I'm not really convinced ot the benefits of the change.
And I see some compatibilty problems for API users that need a sync result.

But let's continue this discussion in the issue you just created.

src/test.ts
JoshuaKGoldberg: Interesting! I remember anecdotally hearing some preference for `return await` for standardization of where execution context results during errors... but all I can find now to back it up is https://stackoverflow.com/questions/38708550/difference-between-return-await-promise-and-return-promise (and that doesn't). Will change.

PR #3337, opened 2017-10-17
Async-ified writing the tslint.json file in --init

JoshuaKGoldberg:
The `runWorker` method is already `async`. Added a general purpose `async` wrapper in utils around `fs.writeFile`.

#### PR checklist

#### Overview of change:

Uses `fs.readFile` within an `async` wrapper under `utils` instead of `fs.readFileSync`.

### DIFF
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -15,6 +15,8 @@
  * limitations under the License.
  */
 
+import * as fs from "fs";
+
 /**
  * Enforces the invariant that the input is an array.
  */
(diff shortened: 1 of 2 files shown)


src/utils.ts
```diff
@@ -213,3 +215,15 @@ export function detectBufferEncoding(buffer: Buffer, length = buffer.length): En
 export function denormalizeWinPath(path: string): string {
     return path.replace(/\\/g, "/");
 }
+
+export async function writeFileAsync(fileName: string, data: string): Promise<void> {
+    await new Promise<void>((resolve, reject) => {
```
ajafff: you can simply return the Promise without awaiting

src/utils.ts
```diff
@@ -213,3 +215,15 @@ export function detectBufferEncoding(buffer: Buffer, length = buffer.length): En
...
+        fs.writeFile(fileName, data, (error?: Error) => {
+            if (error !== null) {
```
ajafff: the mixed types here don't make sense. The optional paramter indicates that `error` may be undefined, but you compare it to `null`.
I guess consistently using `null` here is the right thing to do.

JoshuaKGoldberg:
> I'm -1 on this.

Works for me.

Issue #3183, opened 2017-08-29
--project option outputs absolute paths


ajafff:
### Bug Report

- __TSLint version__: 5.7.0
- __TypeScript version__: 2.5.1
- __Running TSLint via__: CLI

#### Actual behavior
`tslint --project .` outputs absolute paths.

#### Expected behavior
Since `--outputAbsolutePaths` is not used, tslint should print relative paths.

ajafff:
I looked into this. When using `--project` we pass the `ts.Program` to the linter, which contains the source files with their absolute paths. These paths are used by the formatter to format the output.

Possible solution:
* Always use absolute paths in `runner.ts`. (not strictly necessary, but would simplify a lot of other stuff, too)
* Somehow pass an option to the `Formatter` whether it should print relative or absolute paths.
  * requires a change for custom formatters
  * option needs to pass though `Linter`

akaztp:
This absolute paths bug prevents from using tslint on VSCode running from a task, in the scenario where the tooling is ran in WSL and the editor in Windows. The output file paths are linux like paths, but the VSCode in Windows expects Windows formatted paths.

ajafff:
@akaztp that's probably better fixed in the formatter. What formatter are you using?

akaztp:
I solved my problem using a beta extension of tslint that uses the TypeScript language server plugin, so I don't need the output of the tslint as a task, anymore.
Nonetheless, I wondered around the Verbose `Formatter` code to evaluate the option to remove the current folder from the absolute paths. But that smells bad to me... And the bug reported in this issue would still be unresolved.

JoshuaKGoldberg:
### 💀 _**It's time!**_ 💀

**TSLint is being deprecated** and no longer accepting pull requests for major new changes or features. See #4534. 😱

If you'd like to see this change implemented, you have two choices:
* **Recommended:** Check if this is available in [ESLint](https://eslint.org) + [typescript-eslint](https://typescript-eslint.io) ✅
* _Not Recommended: Fork TSLint locally_ 🤷‍♂️

👋 It was a pleasure open sourcing with you!

_If you believe this message was posted here in error, please comment so we can re-open the issue!_

JoshuaKGoldberg:
🤖 Beep boop! 👉  **TSLint is deprecated** 👈 _(#4534)_  and you should **switch to [typescript-eslint](https://typescript-eslint.io)**! 🤖 

[...]

PR #3339, opened 2017-10-18
relativizes file names if 'rootDir' is provided in tsconfig.json

bowenni:
Run TSLint without "-p" makes file names in failures relative. Run TSLint with a tsconfig.json and "rootDir" makes file names in failures absolute. (I think it's because of this bug fix https://github.com/palantir/tslint/pull/2688)

To reproduce the problem:
```
~/workspace/tslint: cat repro/repro.ts 
let a = 1
~/workspace/tslint: cat repro/tslint.json 
{
    "rules": {
        "semicolon": [true, "always"]
    }
}
~/workspace/tslint: cat repro/tsconfig.json 
{
    "compilerOptions": {
        "rootDir": "/usr/local/home/bowenni/workspace/tslint",
        "outDir": "/usr/local/home/bowenni/workspace/tslint"
    },
    "files": [
        "repro.ts"
    ]
}
~/workspace/tslint: ./bin/tslint -c repro/tslint.json repro/repro.ts
ERROR: repro/repro.ts[1, 10]: Missing semicolon
~/workspace/tslint: ./bin/tslint -c repro/tslint.json -p repro/tsconfig.json
ERROR: /usr/local/home/bowenni/workspace/tslint/repro/repro.ts[1, 10]: Missing semicolon
```

This relativizes the file names in failures.
```
~/workspace/tslint: ./bin/tslint -c repro/tslint.json -p repro/tsconfig.json
ERROR: repro/repro.ts[1, 10]: Missing semicolon
```

#### PR checklist

  - [ ] Includes tests
- [ ] Documentation update

### DIFF
--- a/src/linter.ts
+++ b/src/linter.ts
@@ -131,6 +131,15 @@ class Linter {
             throw new Error(`formatter '${formatterName}' not found`);
         }
 
+        if (this.program) {
+            const rootDir = this.program.getCompilerOptions().rootDir;
+            if (rootDir) {
+                for (const f of this.failures) {
+                    f["fileName"] = path.relative(rootDir, f.getFileName());
+                }
+            }
+        }
+
         const output = formatter.format(this.failures, this.fixes);
 
         const errorCount = this.failures.filter((failure) => failure.getRuleSeverity() === "error").length;
--- a/src/linter.ts
+++ b/src/linter.ts
@@ -131,10 +131,11 @@ class Linter {
             throw new Error(`formatter '${formatterName}' not found`);
         }
 
-        if (this.program) {
+        if (this.program !== undefined) {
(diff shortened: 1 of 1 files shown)


ajafff:
I already looked into this, see #3183.
Making paths relative unconditional is not the right thing to do, because of the `--outoutAbsolutePaths` option.
Besides that, the path should be relative to cwd, not the rootDir from tsconfig.json. 

And please don't modify typescript internals. Nobody knows what will break if you change the fileName of a SourceFile. I suspect that it will break updating the program after fixes were applied.

bowenni:
Thanks for looking into this.
I think your solution of `Somehow pass an option to the Formatter whether it should print relative or absolute paths.` is much better than changing the actual fileNames. Closing this PR in favor of https://github.com/palantir/tslint/issues/3183

Issue #3334, opened 2017-10-17
callable-types auto fix produces invalid results


mrweiss:
### Bug Report

- __TSLint version__: 5.7.0
- __TypeScript version__: 2.5.2
- __Running TSLint via__: VSCode

#### TypeScript code being linted

```ts
export function issue(foo: any | { (param: any): any }) { foo = 'bar'; }
```

with `tslint.json` configuration:

```json
{
  "extends": ["tslint:recommended", "tslint-react"],
  "rulesDirectory": [
    "../../../../../../../../node_modules/tslint-eslint-rules/dist/rules",
    "../../../../../../../../node_modules/tslint-microsoft-contrib",
    "../../../../../../../../node_modules/tslint-misc-rules/rules"
  ],
  "rules": {
    "quotemark": [true, "single"],
    "ter-indent": ["error", 2, { "SwitchCase": true }],
    "object-curly-spacing": ["error", "always"],
    "member-access": [false],
    "ordered-imports": [false],
    "object-literal-sort-keys": false,
    "trailing-comma": [false],
    "arrow-parens": [false],
    "object-literal-key-quotes": [false],
    "variable-name": [false],
    "no-multi-spaces": ["error", "always"],
    "no-console": [false],
    "no-bitwise": false,
    "no-empty": false,
    "no-reference": false,
    "interface-over-type-literal": false,
    "react-tsx-curly-spacing": [true, "never", { "allowMultiline": true }],
    "array-bracket-spacing": [true, "never"],
    "block-spacing": [true, "always"],
    "no-unexpected-multiline": "error",
    "curly": true,
    "jsx-alignment": true,
    "jsx-wrap-multiline": true,
    "jsx-curly-spacing": ["never"],
    "jsx-no-multiline-js": false,
    "jsx-no-lambda": false,
    "class-method-newlines": true,
    "no-consecutive-blank-lines": [true],
    "max-classes-per-file": [
      false
    ]
  }
}
```

#### Actual behavior
```
severity: 'Error'
message: 'Type literal has only a call signature — use `(param: any) => any` instead. (callable-types)'
at: '2,36'
```

on auto fix changes to:
```ts
export function issue(foo: any | (param: any) => any) { foo = 'bar'; }
```

which is not valid:
```
severity: 'Error'
message: 'missing whitespace (one-line)'
at: '2,46'
source: 'tslint'

severity: 'Error'
message: ''{' or ';' expected.'
at: '2,47'
source: 'ts'

[...]

PR #3342, opened 2017-10-18
[bugfix] callable types suggestion forgot parens on unions and intersections

cyrilgandon:
#### PR checklist

- [x] Addresses an existing issue: #3334 
  - [x] Includes tests
- [ ] Documentation update

#### CHANGELOG.md entry:

[bugfix] `callable-types` auto fix produces invalid results

### DIFF
--- a/src/rules/callableTypesRule.ts
+++ b/src/rules/callableTypesRule.ts
@@ -79,17 +79,16 @@ function noSupertype(node: ts.InterfaceDeclaration): boolean {
     return isIdentifier(expr) && expr.text === "Function";
 }
 
-function renderSuggestion(
-    call: ts.CallSignatureDeclaration,
-    parent: ts.InterfaceDeclaration | ts.TypeLiteralNode,
-    sourceFile: ts.SourceFile): string {
+function renderSuggestion(call: ts.CallSignatureDeclaration,
+                          parent: ts.InterfaceDeclaration | ts.TypeLiteralNode,
+                          sourceFile: ts.SourceFile): string {
 
     const start = call.getStart(sourceFile);
     const colonPos = call.type!.pos - 1 - start;
     const text = sourceFile.text.substring(start, call.end);
 
     let suggestion = `${text.substr(0, colonPos)} =>${text.substr(colonPos + 1)}`;
-    if (parent.parent && shouldWrapSuggestion(parent.parent)) {
+    if (parent.parent !== undefined && shouldWrapSuggestion(parent.parent)) {
         suggestion = `(${suggestion})`;
     }
     if (parent.kind === ts.SyntaxKind.InterfaceDeclaration) {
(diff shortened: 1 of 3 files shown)


src/rules/callableTypesRule.ts
```diff
@@ -82,11 +82,15 @@ function noSupertype(node: ts.InterfaceDeclaration): boolean {
...
                           parent: ts.InterfaceDeclaration | ts.TypeLiteralNode,
                           sourceFile: ts.SourceFile): string {
+
     const start = call.getStart(sourceFile);
     const colonPos = call.type!.pos - 1 - start;
     const text = sourceFile.text.substring(start, call.end);
-    const suggestion = `${text.substr(0, colonPos)} =>${text.substr(colonPos + 1)}`;
 
+    let suggestion = `${text.substr(0, colonPos)} =>${text.substr(colonPos + 1)}`;
+    if (parent.parent !== undefined && shouldWrapSuggestion(parent.parent)) {
```
ajafff: `.parent` can never be undefined here. you can simply assert that it is not undefined with `shouldWrapSuggestion(parent.parent!)`

test/rules/callable-types/test.ts.lint
```diff
@@ -32,6 +32,16 @@ var fn: {(): void;};
          ~~~~~~~~~ [type % ('() => void')]
 function f(x: { (): void }): void;
                 ~~~~~~~~ [type % ('() => void')]
+function f(x: string | { (): void }): void;
```
ajafff: also add a test where the type is already wrapped in parentheses, e.g. `: string | ({ (): void })`

src/rules/callableTypesRule.ts
```diff
@@ -46,7 +46,7 @@ export class Rule extends Lint.Rules.AbstractRule {
 function walk(ctx: Lint.WalkContext<void>) {
     return ts.forEachChild(ctx.sourceFile, function cb(node: ts.Node): void {
         if ((isInterfaceDeclaration(node) && noSupertype(node)
-             || isTypeLiteralNode(node))
+            || isTypeLiteralNode(node))
```
ajafff: the alignment here was actually intentional because of the parens. it makes more sense to just remove the line break

test/rules/callable-types/test.ts.lint
cyrilgandon: The funny thing is that it still works and doesn't add parens in this case. I guess that it there is already parens, parent.parent doesn't reference the Union, be reference the node with the extra parens.

ajafff: That's right, `parent.parent` is `ts.ParenthesizedType` in that case. That's why I wanted this test case. To verify that we don't add parens if they are already present.

ajafff:
Thanks @cyrilgandon

Issue #3235, opened 2017-09-18
New rule: no-implicit-package-dependencies


adidahiya:
I'd like a new rule to raise a lint error when I import a module that is not explicitly declared in the `dependencies` block of `package.json`. Like this eslint plugin: https://github.com/lennym/eslint-plugin-implicit-dependencies.

Motivation: this problem is an unresolved question when using dependency hoisting in monorepos: https://github.com/yarnpkg/rfcs/blob/master/implemented/0000-workspaces-install-phase-1.md#unresolved-questions

We should consider adding a linter configuration option to specify the location of `package.json` similar to `--project` (maybe `--manifest`). Alternatively, we could walk up the folder tree until we find a `package.json` for each linted file/folder (potentially expensive).

ajafff:
:-1: for `--manifest`
Currently there is no way for rules to access anything but *their* config. We should not introduce some special behavior just for one rule.

Specifiying the path in `tslint.json` as rule option won't work, because we cannot resolve relative paths without knowing where the config comes from.

:+1: for using the findup approach. Although there might be a performance hit, we only have to do this once for every file that contains imports from `node_modules`.
Caching comes to mind, but that's not really possible as we don't know when to clear the cache. Ideally the host system caches these recurring file system operations and we don't need to worry.

adidahiya:
Fair points, findup seems fine to me.

ajafff:
Do we need to handle path mappings specified in tsconfig.json?
https://www.typescriptlang.org/docs/handbook/module-resolution.html#path-mapping

ajafff:
The rule should have a `"dev"` option to also look at `devDependencies`.
That option could then be enabled for tests.

PR #3343, opened 2017-10-18
new rule 'no-implicit-dependencies'

ajafff:
#### PR checklist

- [x] Addresses an existing issue: #3235
  - [x] Includes tests
- [x] Documentation update

#### Overview of change:
[new-rule] `no-implicit-dependencies`
Fixes: #3235

### DIFF
--- a/package.json
+++ b/package.json
@@ -38,6 +38,7 @@
   },
   "dependencies": {
     "babel-code-frame": "^6.22.0",
+    "builtin-modules": "^1.1.1",
     "chalk": "^2.1.0",
     "commander": "^2.9.0",
     "diff": "^3.2.0",
@@ -45,7 +46,7 @@
     "minimatch": "^3.0.4",
     "resolve": "^1.3.2",
     "semver": "^5.3.0",
-    "tslib": "^1.7.1",
+    "tslib": "^1.8.0",
     "tsutils": "^2.12.1"
   },
   "peerDependencies": {
@@ -76,7 +77,7 @@
     "ts-node": "^3.3.0",
     "tslint": "^5.7.0",
     "tslint-test-config-non-relative": "file:test/external/tslint-test-config-non-relative",
-    "typescript": "~2.5.1"
+    "typescript": "^2.6.0-dev.20171015"
   },
   "license": "Apache-2.0",
   "engines": {
--- a/package.json
+++ b/package.json
@@ -46,7 +46,7 @@
     "minimatch": "^3.0.4",
     "resolve": "^1.3.2",
     "semver": "^5.3.0",
-    "tslib": "^1.8.0",
+    "tslib": "^1.7.1",
     "tsutils": "^2.12.1"
   },
   "peerDependencies": {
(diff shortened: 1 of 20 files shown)


adidahiya:
@ajafff is it possible to decouple this from the TS 2.6 upgrade? is `ts. isExternalModuleNameRelative` available in TS 2.5 as a private util we can cast? I would really love to get this rule released in TSLint 5.8 without waiting for the TS release.

ajafff:
@adidahiya sure, I just need to add an assertion. We already use it in `no-submodule-imports`.

ajafff:
@adidahiya it now works with ts@2.5, PTAL

adidahiya:
thanks!

balassy:
I'm afraid this new rule triggers a false alarm when someone imports an interface from a type information package, for example in `package.json`:
```
"devDependencies": {
    "@types/aws-lambda": "0.0.17"
}
```
In a `.ts` code file:
```
import { APIGatewayEvent } from 'aws-lambda'; 
```
Thank you for creating and maintaining this package!

adidahiya:
@balassy are you not also using the package `aws-lambda` in that project? Seems strange to only declare its type definitions as a dependency. If this use case is prevalent then maybe we can add a whitelist of package names as part of the rule config.

alokito:
@adidahiya I am facing a similar but not directly relevant problem. When writing a AWS lambda function, the aws-sdk dependency is provided by the lambda container. It is not necessary to add it as an explicit dependency of the function. http://docs.aws.amazon.com/lambda/latest/dg/current-supported-versions.html. Ideally we could disable the implicit package check just for this case.

balassy:
I have to admit this is a very special situation: my project contains AWS Lambda functions implemented in TypeScript, which are basically callback functions, that will be called by AWS. They don't have any dependencies, so I don't have any `aws-lambda` package in my project. However, I have the `@types/aws-lambda` imported as a devDependency, because it provides type information for the callback function signature as it was defined by AWS.

So basically I am creating a plugin, and what I import is only the input-output definition of the interface defined by the hosting environment.

(Actually the `@types/aws-lambda` package has no relation to the `aws-lambda` package, they are completely independent, but that's another story.)

ajafff:
@balassy I would expect the `@types/aws-lambda` package to contain global types. That's how other declaration packages like `@types/jquery` work for example. They assume that the thing you are referencing is just there, no import required.

Can you to please report that in the DefinitelyTyped repo (and send a PR)?

---

It **should** be possible to detect type-only imports without the type checker. They will be elided during compilation and can be excluded from the check.
That's quite a difficult task and I wonder if it's worth it. I'll open a separate issue to track this and await more feedback.

balassy:
I've checked the `@types/aws-lambda` package, and it contains global types:
https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/aws-lambda/index.d.ts

Unfortunately I was not able to figure out how to let the TypeScript compiler aware of the type definitions in that package instead of using `import`.

Thanks, Klaus!

timocov:
We have the project structure like this:
```
root/
    package.json (project's root package.json with all deps)
    component1/
        package.json
        component1.ts
        ...
    component2/
        package.json
        component2.ts
        ...
```

`package.json` of a component just has a `private: true`, `main` and `typings` section to have short import like this `import { Component1Class } from 'component1';` (not `from 'component1/component1'`) and seems that this rule does not handle this case and trigger errors on each component.

PR #3348, opened 2017-10-19
vsoFormatter: don't duplicate output for fixed failures

ajafff:
#### PR checklist

  - [x] Includes tests
- [ ] Documentation update

#### Overview of change:
[bugfix] `vso` formatter no longer duplicates output for fixed failures

I noticed this while looking through the formatters to find out how the `fixes` parameter is used. Obviously this formatter got it wrong...

### DIFF
--- a/test/formatters/vsoFormatterTests.ts
+++ b/test/formatters/vsoFormatterTests.ts
@@ -45,7 +45,27 @@ describe("VSO Formatter", () => {
             getFailureString(TEST_FILE, 2, 12, "mid failure", "mid-name") +
             getFailureString(TEST_FILE, 9,  2,  "last failure", "last-name");
 
-        const actualResult = formatter.format(failures, failures);
+        const actualResult = formatter.format(failures);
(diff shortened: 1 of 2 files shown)


test/formatters/vsoFormatterTests.ts
```diff
@@ -45,7 +45,7 @@ describe("VSO Formatter", () => {
             getFailureString(TEST_FILE, 2, 12, "mid failure", "mid-name") +
             getFailureString(TEST_FILE, 9,  2,  "last failure", "last-name");
 
-        const actualResult = formatter.format(failures);
+        const actualResult = formatter.format(failures, failures);
```
adidahiya: this is confusing to read -- the second argument is `fixes` but it's called `failures` here. you might want to write a separate test to validate that passing in fixes as a second argument doesn't create excess formatter output, or at least leave a comment here

ajafff: I created a separate test where I duplicate the failures array to pass them as `fixes`

PR #3346, opened 2017-10-19
add skip-blank-lines option to no-trailing-whitespace

cyrilgandon:
#### PR checklist

- [x] Addresses an existing issue: #481
- [x] New feature, bugfix, or enhancement
  - [x] Includes tests
- [ ] Documentation update

#### Is there anything you'd like reviewers to focus on?

Should the option be called `skip-blank-lines` like [eslint](https://eslint.org/docs/rules/no-trailing-spaces), or `ignore-blank-lines`, like every other option of the rule?

#### CHANGELOG.md entry:

[new-rule-option] [`no-trailing-whitespace`](https://palantir.github.io/tslint/rules/no-trailing-whitespace/) : `ignore-blank-lines`

PR #3351, opened 2017-10-20
fix CI failures with typescript@next

ajafff:
#### PR checklist

- [x] Addresses an existing issue: https://github.com/palantir/tslint/pull/3346#issuecomment-338125279
  - [ ] Includes tests
- [ ] Documentation update

#### Overview of change:
typescript@next changes how JSDoc is associated with variable declarations if there are 2 or more variables in the VariableDeclarationStatement.
I don't know if this is a bug or intended. However, it breaks our tests on CI.

#### CHANGELOG.md entry:

[no-log]

### DIFF
--- a/test/rules/deprecation/test.ts.lint
+++ b/test/rules/deprecation/test.ts.lint
@@ -58,7 +58,9 @@ const A = 1, B = 2;
 
 A + B;
 ~     [err % ('A')]
+#if typescript < 2.7.0
     ~ [err % ('B')]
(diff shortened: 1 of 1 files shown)


ajafff:
For posterity: I asked if the change is intentional in the PR that caused this: https://github.com/Microsoft/TypeScript/pull/19355#issuecomment-338151850
I'd like to land this regardless of the answer. We can still revert later.

Issue #803, opened 2015-11-16
Tautology Checker in Conditions


JKillian:
I was writing code to check if two objects were equivalent and wrote code like this:

``` typescript
function areEqual(obj1: SomeClass, obj2: SomeClass) {
return obj1 != null
    && obj2 != null
    && obj1.foo === obj1.foo
    && obj1.foo === obj2.bar
    && obj1.baz === obj2.baz
}
```

The error in the code above is comparing `obj1.foo === obj1.foo`. In very few cases is an obviously true condition like this actually what the developer intended. While tslint probably doesn't want to be attempting to evaluate expressions on each side of a condition, it could do a simple textual comparison to catch mistakes like this.

noamyogev84:
Hi @JKillian ,I'd like to tackle this one.
Can you share some code pointers?

PR #3352, opened 2017-10-20
[new-rule] no-constant-binary-expression, aka no-tautology allowed

cyrilgandon:
#### PR checklist

- [x] Addresses an existing issue: #803
- [x] **New feature**, bugfix, or enhancement
  - [x] Includes tests
- [x] Documentation update

#### Overview of change:
A new rule to check if both side of a binary expression are the sames.  
I really love this idea, that's a rule extremely simple, but can be very powerful and can catch the most trivial error that the compiler or other rules can't catch.

English is not my first langage, so I don't really know how to name things!  

Moreover, this pass the rule: `(a - b)  ==  (a  - b)`, because of the extra space. I just check the raw text, I don't trim anything, should I? Is there utilities already existing that allow comparing node text content recursively?

#### CHANGELOG.md entry:

[new-rule] `no-self-compare`: Sides of a comparison operator cannot be the same


src/rules/noConstantBinaryExpressionRule.ts
```diff
@@ -0,0 +1,52 @@
...
+    public static FAILURE_STRING = "Both sides of a binary expression can't be the same";
+
+    public apply(sourceFile: ts.SourceFile): Lint.RuleFailure[] {
+        return this.applyWithFunction(sourceFile, walk);
+    }
+}
+
+function walk(ctx: Lint.WalkContext<void>) {
+    return ts.forEachChild(ctx.sourceFile, function cb(node: ts.Node): void {
+        if (isBinaryExpression(node)) {
```
ajafff: you need to be careful here. BinaryExpressions is everything where you have `<expression> <operator> <expression>`.
you need to check the kind of the operator to avoid errors on `foo + foo`, `foo -= foo`, `foo(), foo()` and probably a lot more.

If this rule is only intended for comparisons your are interested in `EqualsEqualsToken`, `EqualsEqualsEqualsToken`, `ExclamationEqualsToken`, `ExclamationEqualsEqualsToken`.
If you also want to show an error on `a = a`, you need `EqualsToken` as well

src/rules/noConstantBinaryExpressionRule.ts
```diff
@@ -0,0 +1,52 @@
...
+
+import { isBinaryExpression } from "tsutils";
+import * as ts from "typescript";
+
+import * as Lint from "../index";
+
+export class Rule extends Lint.Rules.AbstractRule {
+    /* tslint:disable:object-literal-sort-keys */
+    public static metadata: Lint.IRuleMetadata = {
+        ruleName: "no-constant-binary-expression",
```
ajafff: Right now you only test comparisions. A better name would be `no-self-compare` or similar.

src/rules/noConstantBinaryExpressionRule.ts
```diff
@@ -0,0 +1,52 @@
+/**
+ * @license
+ * Copyright 2013 Palantir Technologies, Inc.
```
ajafff: 2017

ajafff:
> Moreover, this pass the rule: (a - b) == (a - b), because of the extra space. I just check the raw text, I don't trim anything, should I? Is there utilities already existing that allow comparing node text content recursively?

We also use the raw text compare in other rules: `no-duplicate-switch-case`, `prefer-conditional-expression`, `prefer-switch` and maybe more.

I already thought about implementing utility functions to compare AST nodes. But I cannot find the motivation and time to do this.

src/rules/noConstantBinaryExpressionRule.ts
cyrilgandon: I would love to have a rule for a=a. Should it be in the same rule? This is an assignment, not a comparaison, the rule should have a different name so? Or have a different rule? `no-self-assignement`?

ajafff: Makes sense to split this into two separate rules unless someone can come up with a name that covers both cases.

cyrilgandon: Not to myself: this should error on `a < a` and so on (`a<=a, a>a, a>=a`)

cyrilgandon:
What about `y += y;`, which can be written as `y = 2*y`.
And `y -= y`, which can be written as `y = 0`.
Should it be linted? in the same rule?

ajafff:
Looking at https://github.com/SonarSource/SonarTS/blob/HEAD/sonarts-core/docs/rules/no-identical-expressions.md I don't think it makes sense to split this rule. There are too many other bad patterns that would need to be banned, like `a / a`.

I actually like the name `no-identical-expressions` they chose for their rule. Problem is if we give this rule the same name, it will have precedence.

PR #3371, opened 2017-10-22
Add 'dryRun' option to Linter

### LINKED ISSUE #1760: Add dry-run to Linter API fix option
The current implementation in tslint 4.0 for the lint method has the following:

```javascript
Linter.prototype.lint = function (fileName, source, configuration) {
        if (configuration === void 0) { configuration = configuration_1.DEFAULT_CONFIG; }
        var enabledRules = this.getEnabledRules(fileName, source, configuration);
        var sourceFile = this.getSourceFile(fileName, source);
        var hasLinterRun = false;
        var fileFailures = [];
        if (this.options.fix) {
            this.fixes = [];
            for (var _i = 0, enabledRules_1 = enabledRules; _i < enabledRules_1.length; _i++) {
                var rule = enabledRules_1[_i];
                var ruleFailures = this.applyRule(rule, sourceFile);
                var fixes = ruleFailures.map(function (f) { return f.getFix(); }).filter(function (f) { return !!f; });
                source = fs.readFileSync(fileName, { encoding: "utf-8" });
                if (fixes.length > 0) {
                    this.fixes = this.fixes.concat(ruleFailures);
                    source = rule_1.Fix.applyAll(source, fixes);
                    fs.writeFileSync(fileName, source, { encoding: "utf-8" });
                    // reload AST if file is modified
                    sourceFile = this.getSourceFile(fileName, source);
                }
                fileFailures = fileFailures.concat(ruleFailures);
            }
            hasLinterRun = true;
        }
```

If in the configuration I set `fix` to true then  `source = fs.readFileSync(fileName, { encoding: "utf-8" });` gets called. This is great for the CLI but when writing my custom code I usually don't pass the fileName of something that exists. I simply pass the source code. Now, in this case I wanted to try to see if I could see the results of the fixer but it failed at the line I just mentioned.

Could we get pass in an extra option in order to prevent the lint method from reading and writing the file? A dry run option would be great, for instance:

```ts
export interface ILinterOptions {
    fix: boolean;
    dryRun?: boolean;
    formatter?: string | Function;
    formattersDirectory?: string;
}
```

By default we can assume that dryRun is false so that the same behaviour stays the same.

The main goal I want to achieve is to get all the failures and the output of the fixer.

azz:
#### PR checklist

- [x] Addresses an existing issue: Fixes #1760
  - [x] Includes tests
- [x] Documentation update (updated type definitions)

#### Overview of change:

Adds a `dryRun` option to the `Linter` constructor. When set, during the fix process, no files will be written to disk. Instead, the result is returned as `fixedSources` in `linter.getResult()`.

```ts
const linter = new Linter({ fix: true, dryRun: true }, program);
linter.lint(filepath, code, config);
const fixed = linter.getResult().fixedSources[filepath];
```

Supports multiple calls to `lint()`, but fixes in other files are ignored. Is that OK?

#### CHANGELOG.md entry:

[api] add `dryRun` to `Linter`.

### DIFF
--- a/src/linter.ts
+++ b/src/linter.ts
@@ -52,7 +52,7 @@ class Linter {
 
     private failures: RuleFailure[] = [];
     private fixes: RuleFailure[] = [];
-    private fixedSources: Record<string, string> = {};
+    private fixedSources = new Map<string, string>();
 
     /**
      * Creates a TypeScript program object from a tsconfig.json file path and optional project directory.
@@ -165,7 +165,7 @@ class Linter {
                 const fixableFailures = updatedFailures.filter((f) => f.hasFix());
                 this.fixes = this.fixes.concat(fixableFailures);
                 source = this.applyFixes(sourceFileName, source, fixableFailures);
-                this.fixedSources[sourceFileName] = source;
+                this.fixedSources.set(sourceFileName, source);
                 sourceFile = this.getSourceFile(sourceFileName, source);
             }
         }
--- a/src/linter.ts
+++ b/src/linter.ts
@@ -52,6 +52,7 @@ class Linter {
 
     private failures: RuleFailure[] = [];
     private fixes: RuleFailure[] = [];
+    private fixedSources: Record<string, string> = {};
 
     /**
      * Creates a TypeScript program object from a tsconfig.json file path and optional project directory.
@@ -137,6 +138,7 @@ class Linter {
         return {
             errorCount,
             failures: this.failures,
+            fixedSources: this.fixedSources,
             fixes: this.fixes,
             format: formatterName,
             output,
@@ -163,6 +165,7 @@ class Linter {
                 const fixableFailures = updatedFailures.filter((f) => f.hasFix());
                 this.fixes = this.fixes.concat(fixableFailures);
                 source = this.applyFixes(sourceFileName, source, fixableFailures);
+                this.fixedSources[sourceFileName] = source;
                 sourceFile = this.getSourceFile(sourceFileName, source);
             }
         }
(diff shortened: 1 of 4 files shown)


palantirtech:
Thanks for your interest in [palantir/tslint](https://github.com/palantir/tslint), @azz! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

src/runner.ts
```diff
@@ -41,6 +41,11 @@ export interface Options {
     config?: string;
 
     /**
+     * When running with fix: true, do not write files to disk.
+     */
+    dryRun?: boolean;
```
ajafff: This is only relevant for the CLI. And this doesn't make sense as CLI option

src/runner.ts
```diff
@@ -211,6 +216,7 @@ async function doLinting(
     const possibleConfigAbsolutePath = options.config !== undefined ? path.resolve(options.config) : null;
     const linter = new Linter(
         {
+            dryRun: !!options.dryRun,
```
ajafff: should always be false when running from CLI

src/linter.ts
```diff
@@ -176,15 +179,17 @@ class Linter {
...
             if (path.resolve(filePath) === path.resolve(sourceFilePath)) {
                 source = Replacement.applyFixes(source, fileFixes);
                 fileNewSource = source;
-            } else {
+            } else if (!this.options.dryRun) {
                 const oldSource = fs.readFileSync(filePath, "utf-8");
                 fileNewSource = Replacement.applyFixes(oldSource, fileFixes);
             }
-            fs.writeFileSync(filePath, fileNewSource);
+            if (!this.options.dryRun && typeof fileNewSource === "string") {
```
ajafff: this basically reverts #2864 when `dryRun` is true. you don't want that

src/linter.ts
```diff
@@ -52,6 +52,7 @@ class Linter {
 
     private failures: RuleFailure[] = [];
     private fixes: RuleFailure[] = [];
+    private fixedSources: Record<string, string> = {};
```
ajafff: prefer `Map`

ajafff:
@azz can you please share your use case in #1760?

src/linter.ts
azz: `updateProgram` is called regardless of `dryRun`. Not sure what I'm missing?

ajafff: updateProgram reads the file from disk. With dryRun it is never updated

azz: Oh, wow. That's unfortunate. Know any way around that? I think using    `CompilerHost#getSourceFile` could work.

havenchyk:
@ajafff @azz dear contributors, could you please share the status of this PR, what's need to be done here?

JKillian:
https://github.com/palantir/tslint/issues/1760#issuecomment-338742362 and https://github.com/palantir/tslint/issues/1760#issuecomment-326729563 explain some of the reasons here why `dryRun` probably isn't the right abstraction. Some possible ways in #1760 how we could possibly improve things though

Issue #2587, opened 2017-04-16
Consider switching most executable tests to using runner.ts


JoshuaKGoldberg:
Debugging the CLI mocha tests locally takes a little while. Since pretty much all of the logic there was moved to the runner, would it make sense to move tests to directly calling it? Doing so would make debugging them smoother.

nchen63:
agree

PR #3374, opened 2017-10-22
Refactor executableTests to use runner.ts where possible

ajafff:
#### PR checklist

- [x] Addresses an existing issue: #2587
  - [x] Includes tests
- [ ] Documentation update

#### Overview of change:
Converts most uses of `execCLI` to `execRunner`. This avoids spawning new processes and initializing node with all modules from scratch. It also allows for easier debugging.
Fixes: #2587
This reduces the time taken by `test:mocha` from 35s to 13s on my machine.

A few tests are untouched. They test the CLI directly, need to run in another working directory or with environment variables.

During refactoring I noticed ``exits with code 0 if `--test` and `-r` flags are used with custom rule`` tested with the wrong directory so the test made no sense. I fixed that by the way.

I needed to modify `test.ts` to not print directly to console. Otherwise its output would interfere with the mocha output. We also test failing rule tests. Seeing their output would be misleading.


src/runner.ts
```diff
@@ -98,7 +98,7 @@ export interface Options {
     /**
      * That TSLint produces the correct output for the specified directory.
```
adidahiya: Can you  update the JSDoc here? It's not immediately clear what this field does

test/executable/executableTests.ts
```diff
@@ -233,230 +215,180 @@ describe("Executable", function(this: Mocha.ISuiteCallbackContext) {
...
-            execCli(
-                ["-c", "test/files/tsconfig-extends-relative/tslint-ok.json", "-p",
-                 "test/files/tsconfig-extends-relative/test/tsconfig.json"],
-                (err) => {
-                    assert.isNull(err, "process should exit without an error");
-                    done();
-                });
-        });
+        it("can extend `tsconfig.json` with relative path", async () => {
+            let status = await execRunner(
```
adidahiya: minor: prefer `const status1`, `const status2` variables instead of reassigning this one

Issue #3088, opened 2017-08-01
no-shadowed-variable: issue with interface extension


sinedied:
### Bug Report

- __TSLint version__: 5.5.0
- __TypeScript version__: 2.4.2
- __Running TSLint via__: angular-cli

#### TypeScript code being linted

```ts
import { RequestOptionsArgs } from '@angular/http';

export enum HttpCachePolicy {
  Never = <any>false,
  Always = <any>true,
  Update = <any>'update'
}

// RequestOptionArgs is declared in a re-exported module, so we have to extend the original module to make it work
// properly (see  https://github.com/Microsoft/TypeScript/issues/13897)
declare module '@angular/http/src/interfaces' {

  // Use this interface to extend requests options if needed
  export interface RequestOptionsArgs {
    cache?: boolean|HttpCachePolicy;
    skipErrorHandler?: boolean;
  }

}
```

with `tslint.json` configuration:

```json
{
  "rulesDirectory": [
    "node_modules/codelyzer"
  ],
  "rules": {
    "arrow-return-shorthand": true,
    "callable-types": true,
    "class-name": true,
    "comment-format": [
      true,
      "check-space"
    ],
    "curly": true,
    "eofline": true,
    "forin": true,
    "import-blacklist": [
      true,
      "rxjs"
    ],
    "import-spacing": true,
    "indent": [
      true,
      "spaces"
    ],
    "interface-over-type-literal": true,
    "jsdoc-format": true,
    "label-position": true,
    "max-line-length": [
      true,
      120
    ],
    "member-access": false,
    "member-ordering": [
      true,
      "public-before-private",
      "static-before-instance",
      "variables-before-functions"
    ],
    "no-arg": true,
    "no-bitwise": true,
    "no-console": [
      true,
      "debug",
      "info",
      "time",
      "timeEnd",
      "trace"
    ],
    "no-construct": true,
    "no-debugger": true,
    "no-duplicate-variable": [
      true,
      "check-parameters"
    ],
    "no-duplicate-super": true,
    "no-empty": false,
    "no-empty-interface": true,
    "no-eval": true,
    "no-inferrable-types": [
      true,
      "ignore-params"
    ],
    "no-shadowed-variable": true,
    "no-string-literal": false,
    "no-string-throw": true,
    "no-switch-case-fall-through": true,
    "no-trailing-whitespace": true,
    "no-unnecessary-initializer": true,
    "no-unused-expression": true,
    "no-use-before-declare": true,
[...]

PR #3387, opened 2017-10-24
no-shadowed-variable: ignore declaration files and ambient modules

ajafff:
#### PR checklist

- [x] Addresses an existing issue: #3088 
  - [x] Includes tests
- [ ] Documentation update

#### Overview of change:
[bugfix] `no-shadowed-variable` now excludes declaration files and ambient modules
Fixes: #3088

### DIFF
--- a/test/rules/no-shadowed-variable/default/ambient.d.ts.lint
+++ b/test/rules/no-shadowed-variable/default/ambient.d.ts.lint
@@ -1,0 +1,7 @@
+// no errors in declaration files
+declare type T = any;
+declare class C<T> {}
+declare namespace ns {
+    type T = C;
+    class C<T = any> {}
+}
--- a/src/rules/noShadowedVariableRule.ts
+++ b/src/rules/noShadowedVariableRule.ts
@@ -16,8 +16,8 @@
  */
 
 import {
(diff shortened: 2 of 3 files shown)


test/rules/no-shadowed-variable/default/ambient.d.ts.lint
```diff
@@ -0,0 +1,7 @@
+// no errors in declaration files
```
adidahiya: i'm not sure about this -- why should we skip declaration files entirely? Is the idea that declaration files have to adhere to all the idiosyncrasies of their corresponding JS source and therefore can't help it if it has shadowed names, so it's impossible to avoid the shadowing unless the JS is fixed as well?

ajafff: Exactly. We skip declartation files for the same reason we ignore module augmentations: they model something that already exists. Like you said:

> declaration files have to adhere to all the idiosyncrasies of their corresponding JS source and therefore can't help it if it has shadowed names, so it's impossible to avoid the shadowing unless the JS is fixed as well?

adidahiya: cool, we're on the same page.

adidahiya:
merging because this passes the old status checks (before we split into multiple builds)

Issue #3078, opened 2017-07-26
"no-shadowed-variable": vars shouldn't be shadowed by later declared vars


benny-medflyt:
### Bug Report

- __TSLint version__: 5.5.0
- __TypeScript version__: 2.4.2
- __Running TSLint via__: CLI

#### TypeScript code being linted

```ts
export function foo(wam: boolean) {
    if (wam) {
        const now = new Date();
        return now.getTime();
    }

    const now = new Date();
    return now.getTime() + 1000;
}
```

with `tslint.json` configuration:

```json
{
    "no-shadowed-variable": true
}
```

#### Actual behavior

```
ERROR: 3:15     no-shadowed-variable  Shadowed name: 'now'
```

#### Expected behavior

Even though the inner `now` variable is inside an inner scope where a variable of the same name resides, I don't believe that there is actually any shadowing.

Indeed the inner block cannot access the outer `now` variable. If we try (by commenting out line 3) then we get this tsc compile error:
```
error TS2448: Block-scoped variable 'now' used before its declaration
```

Code in the style above is pretty common IMHO and is readable, and should not trigger "no-shadowed-variable" (as I have shown there is technically no shadowing)

ajafff:
You're right that it doesn't make much sense in your example.

As soon as you have a closure in your block, that a different story:
```ts
export function foo(wam: boolean) {
    if (wam) {
        const now = new Date(); // if you remove this line, the next line will use `now` from the outer scope
        setTimeout(() => console.log(now));
    }

    const now = new Date();
    setTimeout(() => console.log(now));
}
```

If there is enough demand from the community, we could add an option to ignore variables that are shadowed in their [temporal dead zone](https://stackoverflow.com/questions/33198849/what-is-the-temporal-dead-zone), i.e. shadowed before the declaration.

benny-medflyt:
If you remove that line, then the code won't pass tslint due to "no-use-before-declare"

shelacek:
I think, this is valid concern. Please add option that take temporal dead zone into account. We have such great keywords like `let` and `const`, but they are devalued by this rule.

adidahiya:
> If you remove that line, then the code won't pass tslint due to "no-use-before-declare"

Note that we don't really take this rule into account when making TSLint design decisions; it's not very useful with `let` / `const` syntax, it's slow to compute, and it's not enabled in the built-in configuration presets.

domoritz:
Is this fixed in 5.9?

PR #3389, opened 2017-10-24
no-shadowed-variable: add new option 'temporalDeadZone'

ajafff:
#### PR checklist

- [x] Addresses an existing issue: #3078
  - [x] Includes tests
- [x] Documentation update

#### Overview of change:
Adds a new option to ignore variable shadowing when it occurs in the temporal dead zone of the shadowed block scoped variable.
Setting this option to `true` will restore the pre-v5.5.0 behavior, but only for block scoped declarations (parameters, classes, enums, let, const).
Fixes: #3078

#### CHANGELOG.md entry:

[new-rule-option] `"temporalDeadZone"` for `no-shadowed-variable` to ignore shadowing in the temporal dead zone of classes, parameters, enums and variables declared with `let` or `const`


src/rules/noShadowedVariableRule.ts
```diff
@@ -35,6 +35,26 @@ export class Rule extends Lint.Rules.AbstractRule {
             and \`"typeParameter"\`. Just set the value to \`false\` for the check you want to disable.
             All checks default to \`true\`, i.e. are enabled by default.
             Note that you cannot disable variables and parameters.
+
+            The option \`"temporalDeadZone"\` defaults to \`false\`. When set to \`true\` parameters, classes, enums and variables declared
```
adidahiya: I think it should default to `true` to align with the other options -- they are all `true` by default, which indicates that a certain type of check is _enabled_. Here, we intend to enable checking within temporal dead zones by default; requiring the option to be `false` would make sense to opt-out of the check -- this would make the rule less strict (similar to setting `false` on the other options).

test/rules/no-shadowed-variable/temporal-dead-zone/test.ts.lint
```diff
@@ -0,0 +1,89 @@
+['a', 'b', 'c'].map((path) => path.resolve(path))
+                     ~~~~ [err % ('path')]
+
+import * as path from 'path'
+
+export function foo(wam: boolean) {
+    if (wam) {
+        const now = new Date();
```
adidahiya: would you mind marking this up with comments to indicate where the option is taking effect? similar to your example in the docs, just a simple `// not an error with tdz disabled`

src/rules/noShadowedVariableRule.ts
ajafff: So you basically want me to swap true and false here and in all conditions? Sounds reasonable

adidahiya: Yes I would like you to swap true and false in all the new code you've added (leave the existing options as they were)

ajafff:
Updated as suggested

Issue #3396, opened 2017-10-26
TSLint fails to compile with Typescript 2.6


sandersn:
### Bug Report

- __TSLint version__: master
- __TypeScript version__: typescript@next (failures are the same with 2.6, though)
- __Running TSLint via__: not running, just compiling

#### Actual behavior
Two compile errors:
1. [tslib needs to be updated](https://github.com/Microsoft/TypeScript/wiki/What's-new-in-TypeScript) for the new emit helper `__makeTemplateObject`, used for template literals.
2. 2.6's improved checking of computed property names in object literals found a typo in a constant in completedDocsRule.ts.

```ts
export const TAGS_FOR_EXISTENCE = "exists";
// should be
export const TAGS_FOR_EXISTENCE = "existence";
```

#### Expected behavior
No compile errors.

ajafff:
@sandersn I already have a PR open to address these problems

sandersn:
#3320 differs in two main ways:

1. Doesn't address the failure in completedDocsRule.
2. Upgrades to Typescript 2.6.

#3397 can be merged ahead of the actual upgrade to 2.6. I'm not sure which is the better process; you can easily add completedDocsRule fix if you want, and I'll close #3397.

ajafff:
@sanders You're right, it makes sense to fix this now. I thought you were about to open a PR to upgrade to 2.6.
Thanks for fixing.

PR #3397, opened 2017-10-26
Fix compile errors on TS 2.6

sandersn:
#### PR checklist

- [x] Addresses an existing issue: #3396

#### Overview of change:
1. [tslib needs to be updated](https://github.com/Microsoft/TypeScript/wiki/What's-new-in-TypeScript) for the new emit helper `__makeTemplateObject`, used for template literals.
2. 2.6's improved checking of computed property names in object literals found a typo in a constant in completedDocsRule.ts.

1. Did completedDocsRule *ever* work? Why doesn't `src/rules/completed-docs/tagExclusion.ts` contain the `TAGS_FOR_*` constants instead of completedDocsRule.ts? I made the smallest fix possible, but the right thing might be to move those constants.
1. tslib 1.8 works fine on TS 2.5 and 2.6. I haven't tried it on earlier versions, but the additional helpers shouldn't change anything.

### DIFF
--- a/package.json
+++ b/package.json
@@ -46,7 +46,7 @@
     "minimatch": "^3.0.4",
     "resolve": "^1.3.2",
     "semver": "^5.3.0",
-    "tslib": "^1.7.1",
+    "tslib": "^1.8",
     "tsutils": "^2.12.1"
   },
   "peerDependencies": {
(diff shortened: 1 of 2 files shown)


adidahiya:
Thanks for making the minimal change. You're right that the constants should probably be moved, but we'll do that in a separate PR.

PR #3398, opened 2017-10-26
no-duplicate-imports: Allow duplicate imports from separate ambient module declarations

andy-ms:
#### PR checklist

- [X] Addresses an existing issue: Microsoft/dtslint/issues/79
  - [X] Includes tests
- [ ] Documentation update

#### Overview of change:

Allows two *ambient* modules to have separate imports:
```ts
declare module "a" {
    import foo from "foo";
}
declare module "b" {
    import foo from "foo";
}
```
Preserves existing behavior for module augmentations.

@ajafff It would be nice to reuse more code from `findImports` in tsutils but it doesn't seem possible given that it currently takes the whole source file as input, so I can't run it on a single `declare module`.

#### CHANGELOG.md entry:

[bugfix] no-duplicate-imports: Allow duplicate imports from separate ambient module declarations

### DIFF
--- a/src/rules/noDuplicateImportsRule.ts
+++ b/src/rules/noDuplicateImportsRule.ts
@@ -15,7 +15,7 @@
  * limitations under the License.
  */
 
-import { isLiteralExpression, isModuleBlock, isModuleDeclaration } from "tsutils";
+import { isImportDeclaration, isLiteralExpression, isModuleDeclaration } from "tsutils";
 import * as ts from "typescript";
 import * as Lint from "../index";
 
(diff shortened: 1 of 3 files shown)


src/rules/noDuplicateImportsRule.ts
```diff
@@ -44,13 +44,38 @@ export class Rule extends Lint.Rules.AbstractRule {
...
+        const im = tryGetImportSpecifier(statement);
+        if (im !== undefined && isLiteralExpression(im)) {
+            const { text } = im;
+            if (seen.has(text)) {
+                ctx.addFailureAtNode(im.parent!, Rule.FAILURE_STRING(text));
+            }
             seen.add(text);
         }
+
+        if (isModuleDeclaration(statement) && statement.body !== undefined && isModuleBlock(statement.body)) {
```
ajafff: you could add an additional check here if `statement.name.kind === ts.SyntaxKind.StringLiteral`, because namespaces cannot contain import statements.
and then you probably don't need the `isModuleBlock` check.

src/rules/noDuplicateImportsRule.ts
```diff
@@ -44,13 +44,38 @@ export class Rule extends Lint.Rules.AbstractRule {
...
+        }
+    }
+}
+
+function tryGetImportSpecifier(statement: ts.Statement): ts.Expression | undefined {
+    switch (statement.kind) {
+        case ts.SyntaxKind.ImportDeclaration:
+            return (statement as ts.ImportDeclaration).moduleSpecifier;
+        case ts.SyntaxKind.ImportEqualsDeclaration:
+            const ref = (statement as ts.ImportEqualsDeclaration).moduleReference;
```
ajafff: It didn't check ImportEqualsDeclaration before your change. That's not a problem, but should be mentioned in the changelog.

andy-ms: It was right before because you might want named imports from a module in addition to importing it as a function.

ajafff:
thanks @andy-ms

mxl:
@ajafff Could you make a release with these changes?

ajafff:
@mxl I could prepare a release PR, but I cannot do the actual release.
In addition there are other PRs that should be included in the next release. Seems like we just need to wait :confused:

Issue #3401, opened 2017-10-26
Output code warnings in yellow instead of red


jez9999:
Right now code warnings with the `codeFrameFormatter` output warning and error messages in red.  It's useful to distinguish between them by colour.

ildoc:
yep, I said the same thing here -> https://github.com/palantir/tslint/issues/2569#issuecomment-339780676

warning and errors should be color-distinguished by default

ajafff:
Fixed by #3402

PR #3402, opened 2017-10-26
Output code warnings in yellow instead of red

jez9999:
Addresses issue #3401.

### DIFF
--- a/src/formatters/codeFrameFormatter.ts
+++ b/src/formatters/codeFrameFormatter.ts
@@ -67,7 +67,7 @@ export class Formatter extends AbstractFormatter {
             }
 
             let failureString = failure.getFailure();
-            failureString = chalk.red(failureString);
+            failureString = failure.getRuleSeverity() === "warning" ? chalk.yellow(failureString) : chalk.red(failureString);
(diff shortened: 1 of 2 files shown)


src/formatters/codeFrameFormatter.ts
```diff
@@ -67,7 +67,7 @@ export class Formatter extends AbstractFormatter {
             }
 
             let failureString = failure.getFailure();
-            failureString = chalk.red(failureString);
+            failureString = failure.getRuleSeverity() === "warning" ? chalk.yellow(failureString) : chalk.red(failureString);
```
ajafff: can you please add a test for this in `test/formatters/codeFrameFormatterTests.ts`?

jez9999: How can I add another commit to this PR?  Because it was created by editing a file in the web browser I just see `merge 1 commit into palantir:master from jez9999:patch-2` rather than merging from a proper fork.

jez9999:
Added test.

ajafff:
thanks @jez9999

Issue #3272, opened 2017-10-03
Handle destructuring in 'object-literal-shorthand'?


ajafff:
### Feature request / Bug report / Discussion

The current implementation of `object-literal-shorthand` only checks properties of object literals.
Destructuring assignments are also checked:
```ts
// the code doesn't make much sense, but that's not the point
// there's an error in both object literals - that's expected
({foo: foo} = {foo: foo});
```

And then there are variable or parameter declarations using destructuring. They look pretty much the same. But they are currently not checked by the rule.

```ts
let {foo: foo} = obj; // I'd expect an error here
```

Is there a reason for the current behavior or was it just forgotten?

IMO adding this additional check qualifies as bugfix.

rjmunro:
I have the opposite problem; in general I don't want shorthand. I have the TSLint config `"object-literal-shorthand": [true, "never"]` but in the case of a destructuring assignment, I don't want to double everything:

```
({foo, bar} = {foo: 'string 1', bar: 'string 2'});
```

I think we need to control destructuring separately from normal object-literal-shorthand.

JoshuaKGoldberg:
### 💀 _**It's time!**_ 💀

**TSLint is deprecated** and no longer accepting pull requests other than security fixes. See #4534. ☠️ 
We recommend you instead use **[typescript-eslint](https://typescript-eslint.io)** to lint your TypeScript code with **ESLint**. ✅

👋 It was a pleasure open sourcing with you!

PR #3408, opened 2017-10-27
object-literal-shorthand: check destructuring

ajafff:
#### PR checklist

- [x] Addresses an existing issue: #3272
  - [x] Includes tests
- [ ] Documentation update

#### Overview of change:
[enhancement] `object-literal-destructuring` added check for object destructuring
Fixes: #3272

Do we need another error message specifically for object destructuring?

### DIFF
--- a/src/rules/objectLiteralShorthandRule.ts
+++ b/src/rules/objectLiteralShorthandRule.ts
@@ -18,6 +18,7 @@
 import {
     getChildOfKind,
     hasModifier,
+    isBindingElement,
     isFunctionExpression,
     isIdentifier,
     isMethodDeclaration,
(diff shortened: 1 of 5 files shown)


adidahiya:
the distinction here is "creating objects" (object literal) vs. "creating variables" (destructuring)

ajafff:
Fair enough, but the rule already checks destructuring assignments.
From the linked issue:
```ts
({foo: foo} = {foo: foo}); // error on LHS and RHS
```

The LHS doesn't create an object and doesn't declare a variable. One could argue if this is considered an object literal. 
We probably want to leave the existing check unchanged, don't we?

adidahiya:
Ah, that's a little unfortunate. I don't think we necessarily need to keep the existing check unchanged if we moved it to a new rule (I'd call it `destructuring-assignment-shorthand`) and tell users to enable the new rule. It'd be a small break but it wouldn't cause builds to fail since it just loosens some lint checks.

Issue #3413, opened 2017-10-29
`noRedundantJsdocRule` emits Error: `Unexpected tag kind`


emilio-martinez:
### Bug Report

- __TSLint version__: 5.8.0
- __TypeScript version__: 2.5.3
- __Running TSLint via__: CLI

#### TypeScript code being linted

```ts
/**
 * This is my class
 * @class This is a `class` test
 */
export class MyClass {}
```

with `tslint.json` configuration:

```json
{
  "defaultSeverity": "error",
  "rules": {
    "no-redundant-jsdoc": true
  }
}
```

#### Actual behavior

```
Error: Unexpected tag kind: JSDocClassTag
    at checkTag (/Users/emilio.martinez/Documents/emilio/is-datatype/node_modules/tslint/lib/rules/noRedundantJsdocRule.js:99:23)
    at cb (/Users/emilio.martinez/Documents/emilio/is-datatype/node_modules/tslint/lib/rules/noRedundantJsdocRule.js:62:25)
    at Array.forEach (native)
    at walk (/Users/emilio.martinez/Documents/emilio/is-datatype/node_modules/tslint/lib/rules/noRedundantJsdocRule.js:55:34)
    at Rule.AbstractRule.applyWithFunction (/Users/emilio.martinez/Documents/emilio/is-datatype/node_modules/tslint/lib/language/rule/abstractRule.js:39:9)
    at Rule.apply (/Users/emilio.martinez/Documents/emilio/is-datatype/node_modules/tslint/lib/rules/noRedundantJsdocRule.js:35:21)
    at Linter.applyRule (/Users/emilio.martinez/Documents/emilio/is-datatype/node_modules/tslint/lib/linter.js:177:29)
    at /Users/emilio.martinez/Documents/emilio/is-datatype/node_modules/tslint/lib/linter.js:119:85
    at Object.flatMap (/Users/emilio.martinez/Documents/emilio/is-datatype/node_modules/tslint/lib/utils.js:151:29)
    at Linter.getAllFailures (/Users/emilio.martinez/Documents/emilio/is-datatype/node_modules/tslint/lib/linter.js:119:32)
```

#### Expected behavior

It seems like in the source code for `noRedundantJsdocRule` there's a non-exhaustive `switch` statement. `JSDocClassTag` may not be the only missing case, but that's the one I happened to stumble onto. See https://github.com/palantir/tslint/blob/master/src/rules/noRedundantJsdocRule.ts#L100

[...]

PR #3414, opened 2017-10-29
no-redundant-jsdoc: handle class tag

ajafff:
#### PR checklist

- [x] Addresses an existing issue: #3413 
  - [x] Includes tests
- [ ] Documentation update

#### Overview of change:
[bugfix] `no-redundant-jsdoc` fixed crash on unhandled tag
Fixes: #3413

@andy-ms I also added `JSDocTypeLiteral` to the always redundant tags. Unfortunately I don't even know what exactly that is. Can you confirm that it is expected to always show an error on this tag?

### DIFF
--- a/src/rules/noRedundantJsdocRule.ts
+++ b/src/rules/noRedundantJsdocRule.ts
@@ -107,6 +107,7 @@ function walk(ctx: Lint.WalkContext<void>): void {
 const redundantTags = new Set([
     "abstract",
     "access",
+    "class",
     "constant",
     "constructs",
     "default",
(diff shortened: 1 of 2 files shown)


andy-ms:
@ajafff JSDocTypeLiteral isn't a tag so it should never appear in the `tags` list. It will only appear inside of a type expression in jsdoc.

ajafff:
@andy-ms makes sense. I was confused by `LastJSDocTagNode = JSDocTypeLiteral`. Is that a bug in `types.ts`

sandersn:
Yeah, JSDocTypeLiteral should be before the tag kinds. It probably doesn’t manifest any buggy behaviour but the organisation is wrong.

PR #3422, opened 2017-10-30
Fix CI - use newer versions of node

ajafff:
CircleCI updated to yarn 1.x.x. This caused our builds to fail because we tested with node versions that didn't support the `shell` option in `child_process.spawn`.
To fix this we need to use at least 4.8.0, 5.7.0 or 6.0.0 and above. See https://github.com/yarnpkg/yarn/issues/4337 for more details.

While fixing this I noticed that we always used node 4.2.6 on every node. Even executing `nvm use <version>` didn't change this, see https://discuss.circleci.com/t/node-npm-version-build-issues/11471. I fixed that by using `nvm alias default <version>`

[no-log]

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -5,7 +5,7 @@ general:
 dependencies:
   pre:
     - node ./scripts/assertMinCircleNodes.js $CIRCLE_NODE_TOTAL
-    - case $CIRCLE_NODE_INDEX in 0) nvm use 4.1.2 ;; 1) nvm use 5.7 ;; [2-3]) nvm use 6.1 ;; esac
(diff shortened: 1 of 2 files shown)


adidahiya:
We should probably also update `engines` in package.json.

Otherwise lgtm.

ajafff:
@adidahiya this is only relevant for development / CI. Using tslint on these versions should still work. 
Is your intention to signal to users that we don't test with older versions and therefore do not support them?

adidahiya:
> Is your intention to signal to users that we don't test with older versions and therefore do not support them?

yep, that's what I'd like to signal. no guarantees with versions lower than what we test with.

Issue #3426, opened 2017-10-31
object-literal-sort-keys doesn't flag errors on Windows starting in version 5.8.0


interpfister:
### Bug Report

- __TSLint version__: 5.8.0
- __TypeScript version__:2.5.3
- __Running TSLint via__: CLI and VSCode

#### TypeScript code being linted

```ts
  const key = {
    accountType: '3B',
    uid: '456',
    isEditable: true,
  };
```

with `tslint.json` configuration:

```json
{ "object-literal-sort-keys": true }
```

#### Actual behavior
No error

#### Expected behavior
```The key 'isEditable' is not sorted alphabetically```

Downgrading to version 5.7.0 fixes the issue. This code succeeds on a Mac with 5.8.0.  It's possible there's something else at play besides Mac vs. Windows here but that's the only difference we've been able to find. Sticking with 5.7.0 for now.

ajafff:
#3191 added the ability to group your properties by a blank line.
This probably introduced a bug where `\r\n` is treated as two line breaks on windows.

ajafff:
#3427 will fix this

PR #3427, opened 2017-10-31
object-literal-sort-keys: don't consider \r\n as two line breaks

ajafff:
#### PR checklist

- [x] Addresses an existing issue: #3426 
  - [x] Includes tests
- [ ] Documentation update

#### Overview of change:
[bugfix] `object-literal-sort-keys` fixed regression that effectively disabled the rule with `\r\n` line breaks
Fixes: #3426


adidahiya:
thanks for the quick regression fix!

PR #3429, opened 2017-10-31
fix condition for deprecation of typeof-compare

ajafff:
The condition was off. #3286 actually deprecated the rule only for typescript@2.1.
Well, not really, because `ts.versionMajorMinor` was introduced in a later release.
That means nobody gets a deprecation warning right now.

This PR fixes both issues.


karfau:
Can somebody please help me to understand this?
The check will only add the `deprecationMessage` to typescript versions `v2.1.*`, right?
Why should it not also be displayed when it would be enabled and used with a more recent typescript version?
When requiring the rule and looking at the metadata the `deprecationMessage` is an empty string.
How is this helpful?

PR #3445, opened 2017-11-02
no-inferred-empty-object-type: really check for inferred empty object

### LINKED ISSUE #3416: no-inferred-empty-object-type not detecting all generics
### Bug Report

- __TSLint version__: 5.8.0
- __TypeScript version__: 2.5.3
- __Running TSLint via__: (pick one) CLI

```ts
    {
        const wrapper = <L, A>(l: L) => ({ l });
        const x = wrapper(1); // expected error, no error
        // The second generic, A, is inferred as an empty object
    }
```

```json
{
    "rules": {
        "no-inferred-empty-object-type": true
    }
}
```

/cc @rsxdalv

ajafff:
#### PR checklist

- [x] Addresses an existing issue: Fixes #3416
  - [x] Includes tests
- [ ] Documentation update

#### Overview of change:
The old implementation of the rule only really checked the return type. I don't know why this ever worked.
However, it did not check type parameters that are not the direct return type. See the comments in the linked issue for more details.

The TypeChecker unfortunately offers no API to get the inferred type of the type parameters. But I noticed that hovering the call in VSCode shows the inferred type parameters. After digging into the typescript language service, I found that `SymbolDisplayBuilder` is used. Unfortunately this API is deprecated since 2.6.1. The replacement `TypeChecker#symbolToString` is not available in ts@2.1.x. I needed to use `SymbolDisplayBuilder` to polyfill `symbolToString`.
Since `symbolToString` returns text for human consumption, we need to parse it.

Now this rule finds all type parameters inferred as empty object type. Unfortunately this also applies to optional parameters:

```ts
declare function fn<T>(obj?: T, cb?: (obj: T) => void): void;
fn(); // error here, because T is inferred as {}
```

I'd say the error is correct, but sometimes unexpected. This is also the cause for changing the calls to `ts.forEach*CommentRanges` .

Another thing to note is the different inference result in older typescript versions with regard to contextual types:

```ts
const wrap: Wrapper<string> = new Wrapper();
```

The above is an error for typescript<2.4.0 and works fine for newer versions.
Note that this is not a problem for types like `Set` or `Map`, because they instead return `Set<any>` and `Map<any, any>`

#### CHANGELOG.md entry:

[bugfix] `no-inferred-empty-object-type` now checks all type parameters


OliverJAsh:
Thanks for fixing this!

src/rules/noInferredEmptyObjectTypeRule.ts
```diff
@@ -42,56 +41,128 @@ export class Rule extends Lint.Rules.TypedRule {
...
-                this.addFailureAtNode(node, Rule.EMPTY_INTERFACE_INSTANCE);
-            }
-        }
-    }
-
-    private checkCallExpression(node: ts.CallExpression): void {
+    private checkExpression<T extends ts.CallExpression | ts.NewExpression>(
+        node: T,
+        shouldCheck: (signature: ts.Signature, checker: ts.TypeChecker, node: T) => boolean,
+        failure: string,
```
adidahiya: nit: `failureMessage: string`

src/rules/noInferredEmptyObjectTypeRule.ts
```diff
@@ -42,56 +41,128 @@ export class Rule extends Lint.Rules.TypedRule {
...
+                    if (level === 0) {
+                        return false;
+                    }
+                    --level;
+            }
+            token = scanner.scan();
+        }
+    }
+
+    /** Compatibility wrapper for typescript@2.1 */
```
adidahiya: can you add a `TODO: remove in next major version` comment line here?

KeeganShay:
Is this going to be merged in anytime soon?

ajafff:
@KeeganShay I don't think so. I'm no longer actively involved in this project. This PR would need to be updated for recent TypeScript versions. In addition it uses a very unreliable algorithm to detect empty object types (there might be quite a few false positives). And then there's the need to resolve semantic merge conflicts...

If you are interested in a working version of this rule, take a look at my [Fimbullinter project](https://github.com/fimbullinter/wotan/blob/master/packages/wotan/README.md). It has a [core rule](https://github.com/fimbullinter/wotan/blob/master/packages/mimir/README.md#rules) named `no-inferred-empty-object` which is superior to TSLint's `no-inferred-empty-object-type` in all regards.

PR #3240, opened 2017-09-20
inherit defaultSeverity onto child configurations

FelschR:
#### PR checklist

- [x] Addresses an existing issue: #2569
- [X] New feature, bugfix, or enhancement
  - [ ] Includes tests
- [ ] Documentation update

#### Overview of change:
As reported in #2569 setting the defaultSeverity in a configuration that extends another configuration (e.g. `tslint:recommended`) leaves the severity of the extended rules unchanged.

The changes in this pull-request make the defaultSeverity inherited from the top-most configuration to all underlying ones (by `extends` & `ruleDirectory`).

#### Is there anything you'd like reviewers to focus on?

This is mainly a quick for me personally but I'm open to improvement.
E.g. this could be activated only by a flag (e.g. `--inheritDefaultSeverity`).
Also some people in #2569 mentioned they'd like to completely override all defaults. This could also be added as a 2nd flag (e.g. `--forceDefaultSeverity`).

Depending on the feedback I can adjust the code and I'll also create new unit tests if necessary.

#### CHANGELOG.md entry:

[enhancement] `inherit defaultSeverity onto child configurations`

Issue #2569, opened 2017-04-13
Overriding rule severity of recommended config preset is cumbersome and verbose


FelschR:
### Bug Report

- __TSLint version__: 5.1.0
- __TypeScript version__: 2.2.2
- __Running TSLint via__: CLI & VSCode

`tslint.json` configuration:

```json
{
    "defaultSeverity": "warning",
    "extends": [
        "tslint:recommended"
    ],
    "jsRules": {},
    "rules": {},
    "rulesDirectory": []
}
```

#### Actual behavior
All issues are reported as "ERROR".
Examples of those are: "Missing trailing comma", "Missing trailing comma", "Missing trailing comma"

#### Expected behavior
Depending on the issue they should have different types. I expected most of them to be warnings.

adidahiya:
This is intended behavior. We want TSLint failures to fail our builds, so we treat them as errors.

You can override any of the severities without changing the rule configuration inherited from `tslint:recommended`. And then you can even make your own reusable configuration preset from that.

adidahiya:
Example to illustrate my above comment:

```json
{
  "extends": "tslint:recommended",
  "rules": {
    "trailing-comma": { "severity": "warning" }
  }
}
```

Bnaya:
On that way i need to have a reference to every rule i'm extending on my inheriting file

MrAndersen1:
I have the same "issue". I want errors to be reserved for actual errors, so I find myself adding all tslint-rules to my file that reports things as errors, because to my knowledge there is no override for this. But if that is the way it is supposed to be then I'll just keep adding :) still love tslint!

adidahiya:
@Bnaya @MrAndersen1 ah, sorry about that, I missed this nuance. I think we should allow `defaultSeverity` to override all the configs you are extending. See my comment here https://github.com/palantir/tslint/issues/345#issuecomment-294582882

I'll update this thread with a more descriptive title

nchen63:
I was planning on addressing this with either another property or changing the `defaultSeverity` options. The reason it is implemented this way (affects only the current file) was to allow for these future scenarios:

- Scenario A: The extended rule set has a curated set of severities that the user wants to retain, and uses `defaultSeverity` to affect only the current file.
- Scenario B: The extended rules have a mix of default/warning/error severities and the user wants `defaultSeverity` to affect the inherited rules with severity=`default` AND `defaultSeverity` is set to `default` in the extended config.
[...]

PR #3449, opened 2017-11-03
Override defaultSeverity defined in extended configs 

M0ns1gn0r:
#### PR checklist

- [x] Addresses an existing issue: #2569
- [x] Enhancement
  - [x] Includes tests
- [x] Documentation update

#### Overview of change:
Helping @FelschR to finish the implementation he started in the PR #3240 (AFAIK it's not possible to add commits to somebody else's pull request). I took into account @ajafff's feedback and added the test case.

#### CHANGELOG.md entry:
[enhancement] Override `defaultSeverity` defined in extended configs (#2569)


palantirtech:
Thanks for your interest in [palantir/tslint](https://github.com/palantir/tslint), @M0ns1gn0r! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

test/configurationTests.ts
```diff
@@ -251,6 +251,22 @@ describe("Configuration", () => {
             const actualConfig = extendConfigurationFile(baseConfig, extendingConfig);
             assertConfigEquals(actualConfig, expectedConfig);
         });
+
+        it("overrides defaultSeverity of base configs", () => {
+            const config = loadConfigurationFromPath("./test/config/tslint-extends-default-severity.json");
+            assert.equal<RuleSeverity | undefined>(
+                "warning",
+                config.rules.get("default-severity-unspecified")!.ruleSeverity,
+                "did not apply defaultSeverity to base config with no defaultSeverity");
```
ajafff: I was a bit confused because all other tests use the message the other way around. They describe what the test expects, not what caused it to fail. For example: `assert.isNull(err, "process should exit without an error");`

There's only one other test in this file that provides a message for assertions. Please change them too while you are here.

test/config/tslint-extends-default-severity.json
```diff
@@ -0,0 +1,7 @@
+{
+  "extends": [ "./tslint-default-severity-unspecified.json", "./tslint-default-severity-error.json" ],
+  "defaultSeverity": "warning",
+  "rules": {
+    "default-severity-warning": { "severity": "default" }
+  }
+}
```
ajafff: Please add another config that has no defaultSeverity, that extends this one.
The test then checks that rules specified in the newly added extending config get a severity of "error" while all 3 rules coming from this config are warnings.

M0ns1gn0r: Please suggest the name for this file. I could come up only with `tslint-extends-default-severity-only-in-extended.json`.

ajafff: Just take the name you suggested. I cannot come up with a better one.

ajafff:
Code looks good to me.

I just want to make sure this is what we want. Ping @adidahiya 
To summarize the current behavior:
* `b.json` extends `a.json`
  * `b` has a `defaultSeverity`
  * `b` overrides the `defaultSeverity` of `a`
* `d.json` extends `c.json`
  * only `c` has a `defaultSeverity`
  * `d` does not inherit the `defaultSeverity` of `c`
  * instead it defaults to "error" while everything from `c` has the `defaultSeverity` of `c` applied

src/configuration.ts
```diff
@@ -25,7 +25,7 @@ import { arrayify, hasOwnProperty, stripComments } from "./utils";
 
 export interface IConfigurationFile {
     /**
-     * The severity that is applied to rules in _this_ config with `severity === "default"`.
+     * The severity that is applied to rules in this and _extended_ configs with `severity === "default"`.
```
adidahiya: This comment should probably be more verbose: `The severity that is applied to rules in this config file as well as rules in any inherited config file which have their severity set to "default"`

test/configurationTests.ts
```diff
@@ -251,6 +251,26 @@ describe("Configuration", () => {
...
+                config.rules.get("default-severity-error")!.ruleSeverity,
+                "should override defaultSeverity defined in base config");
+            assert.equal<RuleSeverity | undefined>(
+                "warning",
+                config.rules.get("default-severity-warning")!.ruleSeverity,
+                "should apply defaultSeverity to extending config");
+            assert.equal<RuleSeverity | undefined>(
+                "error",
+                config.rules.get("default-severity-only-in-extended")!.ruleSeverity,
+                "should not inherit defaultSeverity from base configs");
```
adidahiya: I don't think this is quite right. Let's say I inherit from a shared config file where defaultSeverity is set to `"warning"`.

Then I change one of the rule configs to enable some specific rule option.

```json
{
  "extends": "config-where-failures-are-warnings",
  "rules": {
    "comment-format": {
      "options": ["check-space", "check-lowercase"]
    }
  }
}
```

now, comment-format produces errors. this is mildly confusing UX. While I think people could generally live with it, I think defaultSeverity should ideally be inherited from the `extendingConfig` if it is not specified in the current config.

ajafff: That's actually a bit more complicated to achive. We have a circular dependency here: we need to call `parseConfigFile` to know extended configs, but `parseConfigFile` would already need to know the `defaultSeverity` of the extended config file.
We need a bigger refactor of `configuration.ts` to make it work. I think this would result in a breaking change for API users.

Btw. which `defaultSeverity` is chosen when extending two or more config files? The last one?

adidahiya: > which defaultSeverity is chosen when extending two or more config files? The last one?

yes, the last one in the `extends` list.

M0ns1gn0r: Do we agree that this change is better to be done in a separate PR?

ajafff: @M0ns1gn0r I agree that this may be too much for a first-time contributor. Though the necessary changes may not be too big.
We should definitely add the missing bits shortly after landing this PR and before cutting a release.

@adidahiya we could make it work as you described by moving the extending of the config file from `loadConfigurationFromPath` into `parseConfigFile`. That way we could know the `defaultSeverity` of the parent config before it is used in `parseRuleOptions`.
IMO the best way to implement this backward compatible is by adding an optional parameter to `parseConfigFile`. That parameter is a function to load a config file by path. If this parameter is passed, `parseConfigFile` resolves all extended configs and reduces them. Otherwise the config is returned without touching `extends`.

src/configuration.ts
```diff
@@ -25,7 +25,8 @@ import { arrayify, hasOwnProperty, stripComments } from "./utils";
 
 export interface IConfigurationFile {
     /**
-     * The severity that is applied to rules in _this_ config with `severity === "default"`.
+     * The severity that is applied to rules in this config file as well as rules
+     * in any inherited config files which have their severity set to "default".
      * Not inherited.
      */
     defaultSeverity?: RuleSeverity;
```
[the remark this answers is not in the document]
M0ns1gn0r: It is used on lines 70 and 78:

    export const DEFAULT_CONFIG: IConfigurationFile = {
        defaultSeverity: "error",

    export const EMPTY_CONFIG: IConfigurationFile = {
        defaultSeverity: "error",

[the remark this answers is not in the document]
ajafff: Yeah, I know. But the value is never read. In addition `parseConfigFile` and `extendConfigurationFile` simply don't set this property. That means you can basically never use it.
Since this could (very unlikely) break some users of the API, we should keep it around until the next major version. So this is not immediately actionable. I'll open a separate PR for that.

While you are already touching this code, can you please add `@deprecated property is never set` in the JSDoc comment? This could lead to lint errors on the two lines you mentioned above. You can just disable the error with `// tslint:disable-line:deprecation`.

M0ns1gn0r:
Is there anything left that prevents merging of this PR?

ajafff:
@M0ns1gn0r currently there's nothing you can do. I'd like @adidahiya to sign off the changes I proposed so I know how to proceed after merging this.

docs/usage/configuration/index.md
```diff
@@ -30,7 +30,7 @@ A path to a directory or an array of paths to directories of [custom rules][2].
   - Any rules specified in this block will override those configured in any base configuration being extended.
   - [Check out the full rules list here][3].
 * `jsRules?: any`: Same format as `rules`. These rules are applied to `.js` and `.jsx` files.
-* `defaultSeverity?: "error" | "warning" | "off"`: The severity level used when a rule specifies a default warning level. If undefined, "error" is used. This value is not inherited and is only applied to rules in this file.
+* `defaultSeverity?: "error" | "warning" | "off"`: The severity level used when a rule specifies a default warning level. If undefined, "error" is used. If specified, overrides the values defined in the configuration files which are extended by this configuration.
```
adidahiya: can you copy over the edit I suggested in `configuration.ts` to here? that would make it:

```
`defaultSeverity?: "error" | "warning" | "off"`: The severity level that is applied to rules in this config file as well as rules in any inherited config files which have their severity set to "default". If undefined, "error" is used as the defaultSeverity.
```

test/configurationTests.ts
adidahiya: @ajafff your proposal sounds reasonable to me. We might want to do this work on a feature branch so that `master` is not left in an in-between state (the Github UI supports re-targeting PR branch destinations).

ajafff:
@M0ns1gn0r to fix CI you need to merge the latest master into your PR

I'll open a PR based on this one that adds the missing functionality. We can then merge this PR and mine right after that.

Issue #3452, opened 2017-11-04
Chalk library typings are broken


jez9999:
Chalk has now been changed to be used as a default export and bundled with its own typings; the existing DefinitelyTyped typings are therefore broken.

ajafff:
Duplicate of  #3431

jez9999:
Why not use my PR for the duplicate issue?  I fixed it :-)

PR #3453, opened 2017-11-04
Use built-in chalk typings and fix references.

jez9999:
Addresses issue #3452 .


ajafff:
Thank you for the PR. Unfortunately this is a duplicate of #3435

ajafff:
For future reference: always install / update / remove dependencies with `yarn`. Otherwise `yarn.lock` is not updated. That's what makes CI fail for this PR.

Closing as a duplicate of #3435

Issue #3457, opened 2017-11-06
VS Code F5 ruleTestRunner tests don't show test names in its console


JoshuaKGoldberg:
### Bug Report

- __TSLint version__: 5.8
- __TypeScript version__: 2.6
- __Running TSLint via__: VSCode

#### Actual behavior

```
 Passed
 Passed
 Passed
 Passed
```

![image](https://user-images.githubusercontent.com/3335181/32456693-53594554-c2db-11e7-9bad-cfbc4a1d99da.png)

#### Expected behavior

I'd like to see the test names as a way of tracking progress.

ajafff:
There's only one difference: the test name is written to `process.stdout` while the test status (in this case `Passed`) is written using `console.log()`. The latter should eventually also end up writing to `process.stdout`. I don't know who's to blame here :confused:

PR #3458, opened 2017-11-06
Added test names to VS Code console logs for ruleTestRunner

JoshuaKGoldberg:
#### PR checklist

- [x] Addresses an existing issue: #3457

#### Overview of change:

Because some of the logs are to `process.stdout`, the VS Code F5 test running just shows a lot of "Passed" messages. Not very informative. This adds a `folder > folder`-style descriptor for the test if running with `--verboseResultHandler`, which is passed by `launch.json`.

Alternately, it could just print everything using the console instead of `process.stdout`...?

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -47,7 +47,7 @@
             "request": "launch",
             "program": "${workspaceRoot}/test/ruleTestRunner.ts",
             "stopOnEntry": false,
-            "args": ["run", "test"],
+            "args": ["--verboseResultHandler"],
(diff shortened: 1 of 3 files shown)


JoshuaKGoldberg:
Per https://github.com/palantir/tslint/issues/3457#issuecomment-342241640, switched it to always use `console.log` except for the diffs.

image
image

ajafff:
#3374 tries to get rid of all `console.log` calls and only use `process.stdout.write`.

PR #3466, opened 2017-11-08
Converted completed-docs rule to use a function instead of a walker

JoshuaKGoldberg:
No rule logic is changed: just the types of scoping used to call it.

### DIFF
--- a/src/rules/completedDocsRule.ts
+++ b/src/rules/completedDocsRule.ts
@@ -19,9 +19,8 @@ import { isVariableDeclarationList, isVariableStatement } from "tsutils";
 import * as ts from "typescript";
 
 import * as Lint from "../index";
-import { Exclusion } from "./completed-docs/exclusion";
 import { IInputExclusionDescriptors } from "./completed-docs/exclusionDescriptors";
-import { ExclusionFactory } from "./completed-docs/exclusionFactory";
+import { ExclusionFactory, ExclusionsMap } from "./completed-docs/exclusionFactory";
 
 export const ALL = "all";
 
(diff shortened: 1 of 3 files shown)


src/rules/completedDocsRule.ts
```diff
@@ -307,76 +304,67 @@ export class Rule extends Lint.Rules.TypedRule {
     }
 }
 
-class CompletedDocsWalker extends Lint.ProgramAwareRuleWalker {
-    private static modifierAliases: { [i: string]: string } = {
-        export: "exported",
-    };
+const modifierAliases: { [i: string]: string } = {
```
ajafff: `: string | undefined` to be precise

JoshuaKGoldberg:
Unrelated build failures:

```
yarn add v1.1.0
[1/5] Validating package.json...
[2/5] Resolving packages...
⠁ ⠂ tslint-test-config-non-relative@file:test/external/tslint-test-config-non-rela⠄ typescript@2.2⡀ typescript@2.2⢀ typescript@2.2[3/5] Fetching packages...
warning There appears to be trouble with your network connection. Retrying...
warning There appears to be trouble with your network connection. Retrying...
warning There appears to be trouble with your network connection. Retrying...
warning There appears to be trouble with your network connection. Retrying...
error An unexpected error occurred: "https://registry.yarnpkg.com/chalk/-/chalk-1.1.3.tgz: ESOCKETTIMEDOUT".
```

ajafff:
@JoshuaKGoldberg please hit the "Rebuild" button in CircleCI and ping me when it's done. After CI is green I can merge this.

JoshuaKGoldberg:
@ajafff could you? I don't have permissions to.

> You need write permissions to trigger builds.

ajafff:
Neither do I. Maybe if you log into CircleCI again.
Otherwise you need to push some changes to trigger a new build. You can just amend the last commit without any additional changes and force push it.

tslint.json
```diff
@@ -4,6 +4,8 @@
   },
   "extends": "tslint:all",
   "rules": {
+    "linebreak-style": false,
```
ajafff: This was probably not intended to be commited. Please revert this change.

JoshuaKGoldberg: Agh..

ajafff:
Thanks @JoshuaKGoldberg

PR #3469, opened 2017-11-09
[breaking] remove defaultSeverity from IConfigurationFile

ajafff:
The property is never set and also never used. It's basically dead code.
Ref: https://github.com/palantir/tslint/pull/3449#discussion_r149764581

[api] removed `IConfigurationFile.defaultSeverity` property


adidahiya:
lgtm, but don't merge yet

christiangirl41:
I ❤️ GitHub!!

JoshuaKGoldberg:
Removing the breaking change label per #4811!

...however, it looks like the head repository has been deleted, so I'll close this. LMK if it should be reopened!

PR #3475, opened 2017-11-10
[new-fixer] file-header rule

JKillian:
#### PR checklist

  - [x] Includes tests

#### Overview of change:

Adds a fixer to `file-header`. The rule will now insert text from a new second option as the header of your file if that option is provided and `--fix` is enabled. 

If the rule finds a header comment that's incorrect, it's replaced with the specified header. If the rule doesn't find a header comment at all, the specified header is inserted at the top of the document.

Is the newline handling legit? 

#### CHANGELOG.md entry:

[new-fixer] `file-header`

### DIFF
--- a/src/rules/fileHeaderRule.ts
+++ b/src/rules/fileHeaderRule.ts
@@ -57,19 +57,19 @@ export class Rule extends Lint.Rules.AbstractRule {
 
         // ignore shebang if it exists
         let offset = text.startsWith("#!") ? text.indexOf("\n") : 0;
-        // returns details about the first comment or undefined
-        const commentText  = ts.forEachLeadingCommentRange(
+        // returns the text of the first comment or undefined
+        const commentText = ts.forEachLeadingCommentRange(
             text,
             offset,
             (pos, end, kind) => text.substring(pos + 2, kind === ts.SyntaxKind.SingleLineCommentTrivia ? end : end - 2));
 
         if (commentText === undefined || !headerFormat.test(commentText)) {
-            const errorAtStart = offset === 0;
-            if (!errorAtStart) {
+            const isErrorAtStart = offset === 0;
+            if (!isErrorAtStart) {
                 ++offset; // show warning in next line after shebang
             }
-            const leadingNewlines = errorAtStart ? 0 : 1;
-            const trailingNewlines = errorAtStart ? 2 : 1;
+            const leadingNewlines = isErrorAtStart ? 0 : 1;
+            const trailingNewlines = isErrorAtStart ? 2 : 1;
 
             const fix = textToInsert !== undefined
                 ? Lint.Replacement.appendText(offset, this.createComment(sourceFile, textToInsert, leadingNewlines, trailingNewlines))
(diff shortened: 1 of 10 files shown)


src/rules/fileHeaderRule.ts
```diff
@@ -37,19 +54,49 @@ export class Rule extends Lint.Rules.AbstractRule {
...
+        if (commentDetails === undefined) {
+            const fix = textToInsert !== undefined
+                ? Lint.Replacement.appendText(offset, this.createComment(sourceFile, textToInsert))
+                : undefined;
+            return [new Lint.RuleFailure(sourceFile, offset, offset, Rule.FAILURE_STRING, this.ruleName, fix)];
+        } else {
+            const { pos, end, kind } = commentDetails;
+            const commentText = text.substring(pos, kind === ts.SyntaxKind.SingleLineCommentTrivia ? end : end - 2);
+            if (!headerFormat.test(commentText)) {
+                const isFirstCommentAHeader = ONLY_WHITESPACE.test(text.substring(offset, pos));
```
ajafff: the substring could contain line breaks.The regexp probably needs the `m` flag to make this work as expected.

src/rules/fileHeaderRule.ts
```diff
@@ -37,19 +54,49 @@ export class Rule extends Lint.Rules.AbstractRule {
...
+                const fix = textToInsert !== undefined
+                    ? isFirstCommentAHeader
+                        ? Lint.Replacement.replaceFromTo(pos, end, this.createComment(sourceFile, textToInsert, 0))
```
ajafff: Is removing the existing comment really expected? It could contain JSDoc or something similar useful.

src/rules/fileHeaderRule.ts
```diff
@@ -37,19 +54,49 @@ export class Rule extends Lint.Rules.AbstractRule {
...
         }
         return [];
     }
+
+    private createComment(sourceFile: ts.SourceFile, commentText: string, trailingNewlines = 2) {
+        const maybeCarriageReturn = sourceFile.text[sourceFile.getLineEndOfPosition(0)] === "\r" ? "\r" : "";
+        const lineEnding = `${maybeCarriageReturn}\n`;
+        return [
+            "/*",
+            ...commentText.split(lineEnding).map((line) => ` * ${line}`),
```
ajafff: consider splitting at `/\r?\n/g`. That allows users to only use `\n` in their config while using `\r\n` in source files.

src/rules/fileHeaderRule.ts
JKillian: Hmm, it's a tricky call. If it's a malformed header, you'd want it to be removed, otherwise you probably wouldn't want it to be. I don't think there's any way to automatically determine which is the case, so we'll want to just pick the option that is most correct in most cases.

I think I agree with your thoughts that replacing the header might be unexpected more often than desired. (Plus it'll be simpler code w/o that feature.) I guess we could make the behavior configurable in the future if desired

JKillian:
Updated to no longer replace the first comment it finds. I also modified the insertion code a little bit to handle insert newlines in (hopefully) the most expected manner.

emanuelpinho:
Are you expecting to include this in the next version? Do you already have a release date for that?

PR #3476, opened 2017-11-11
deprecate utils

ajafff:
Deprecate most of the utility functions in `src/language/utils.ts`. Some of them are no longer used.
The other deprecated functions can be easily replaced by functions from `tsutils`. 
I think we should get rid of these functions to avoid maintaining them. For example `forEachToken` (with `skipTrivia` false) and `forEachComment` will no longer function correctly with the newly added `JsxFragment`.

[deprecation] several utility functions have been deprecated

I'm in favor of making the remaining utilities in `src/utils.ts` and `src/language/utils.ts` internal in the next major version. Most of them are not meant to be exposed as public API. WDYT?


adidahiya:
> I'm in favor of making the remaining utilities in src/utils.ts and src/language/utils.ts internal in the next major version. Most of them are not meant to be exposed as public API. WDYT?

Yeah, i'm generally on board. We can stop exporting them from the root of the package and use direct imports to the utils modules wherever necessary (instead of `Lint.Utils`).

src/language/walker/scopeAwareRuleWalker.ts
```diff
@@ -107,6 +107,6 @@ export abstract class ScopeAwareRuleWalker<T> extends RuleWalker {
     }
 
     protected isScopeBoundary(node: ts.Node): boolean {
-        return isScopeBoundary(node);
+        return isScopeBoundary(node); // tslint:disable-line:deprecation
```
adidahiya: can you switch to using the `tsutils` function here?

src/language/walker/blockScopeAwareRuleWalker.ts
```diff
@@ -21,13 +21,15 @@ import { IOptions } from "../rule/rule";
 import { isBlockScopeBoundary } from "../utils";
 import { ScopeAwareRuleWalker } from "./scopeAwareRuleWalker";
 
+// tslint:disable:deprecation
```
adidahiya: can you comment here explaining why we're disabling the rule for the file?

ajafff: done

src/language/walker/scopeAwareRuleWalker.ts
ajafff: There are some differences between the two that _could_ break some custom rules. Since this is deprecated, I don't think we need to replace all deprecated utils.

Issue #3493, opened 2017-11-17
curly rule --fix problem for multiline statements starting on same line


rhys-vdw:
### Bug Report

- __TSLint version__: 5.8.0
- __TypeScript version__: 2.6.1
- __Running TSLint via__: (pick one) CLI

#### TypeScript code being ~linted~ **`--fix`ed**

```ts
  if (process.env.DEBUG) throw new TypeError(
    `Unrecognized test step type: ${question.type}`
  )
```

with `tslint.json` configuration:

(I assume it's the `curly` rule)
```json
{
  "extends": [
    "tslint:recommended",
    "tslint-eslint-rules",
    "tslint-react"
  ],
  "defaultSeverity": "error",
  "rules": {
    "array-bracket-spacing": [true, "never"],
    "curly": [true, "ignore-same-line"],
    "indent": [true, "spaces", 2],
    "interface-name": false,
    "jsx-curly-spacing": [true, "never"],
    "jsx-no-multiline-js": false,
    "jsx-boolean-value": [true, "never"],
    "max-line-length": [true, 120],
    "member-access": [true, "no-public"],
    "no-console": [true, "warn", "error"],
    "no-constant-condition": true,
    "no-debugger": true,
    "no-unused-variable": [true, "check-parameters", { "ignorePattern": "^_" }],
    "no-var-keyword": true,
    "no-var-requires": false,
    "object-curly-spacing": [true, "always"],
    "ordered-imports": false,
    "prefer-const": true,
    "quotemark": [true, "double", "jsx-double", "avoid-escape"],
    "semicolon": [true, "never"],
    "trailing-comma": [true, {
      "multiline": {
        "objects": "always",
        "arrays": "always",
        "functions": "ignore",
        "typeLiterals": "always",
        "imports": "always",
        "exports": "always"
      },
      "singleline": "never"
    }],
    "variable-name": {
      "options": [
        "allow-leading-underscore",
        "ban-keywords",
        "check-format",
        "allow-pascal-case"
      ]
    }
  },
  "linterOptions": {
    "exclude": ["./node_modules/**/*"]
  }
}
```

#### Actual behavior

```ts
  if (process.env.DEBUG) throw new TypeError( { { {
    `Unrecognized test step type: ${question.type}`
  )
  }
  }
  }
```

#### Expected behavior

Not sure. **My preference would be to consider the above to be a single-line statement that doesn't require braces.**

Otherwise I think this makes sense:

```ts
  if (process.env.DEBUG) {
[...]

PR #3494, opened 2017-11-17
Add regression test for #3493

ajafff:
Issue was already fixed by #3473
Fixes: #3493

[no-log]


ajafff:
CI failure is causes by a parse error in typescript@next. Reported in https://github.com/Microsoft/TypeScript/issues/20096

Issue #3479, opened 2017-11-11
completed-docs doesn't cover Accessors


nixxquality:
### Bug Report

- __TSLint version__: 5.8.0
- __TypeScript version__: 2.5.3
- __Running TSLint via__: CLI / tslint-language-service

#### TypeScript code being linted

```ts
class Foo {
    private innerbar: string;

    public get bar(): string {
        return innerbar;
    }
    public set bar(value: string) {
        this.innerbar = value;
    }
}

```

with `tslint.json` configuration:

```json
{
    "rules": {
        "completed-docs": [
            true, "classes", "enums", "enum-members", "functions", "interfaces", "methods", "namespaces", "properties", "types", "variables" // everything
        ]
    }
}
```

#### Actual behavior
Only foo and innerbar get warnings about missing documentation

#### Expected behavior
bar should also get a warning.

reduckted:
If I remember correctly, when I encountered this same problem about 12 months ago, I discovered that this is because TypeScript doesn't expose (?) those properties when walking the syntax tree. Or something like that. Maybe it's changed since then.

I recall finding an issue about this at the time (I think it was in the TypeScript repo), but I'm not having any luck finding it now. 😞

ajafff:
Now that we no longer rely on `SyntaxWalker` do visit nodes, we can simply add a condition for `isAccessorDeclaration(node)` and check for JSDoc.
In addition there is some special handling needed, because you probably don't want to enforce JSDoc on both accessors of a property.

reduckted:
I'm looking at this at the moment, and the change is very simple. [This `switch` statement](https://github.com/palantir/tslint/blob/cbdcfb5b7e82eacb4d3c97e06e678d1169684030/src/rules/completedDocsRule.ts#L315) just needs two extra case statements to handle `GetAccessor` and `SetAccessor` nodes.

Currently, because we are using the type checker, the documentation comments from the getter and setter are merged, so we don't need special handling for the getter having comments and the setter not (for example). That will change when we get rid of the type checker in #3376.

If we're happy to fix this while still using the type checker, then I'll go ahead and make a PR.

ajafff:
@reduckted Go for it. Changing that when dropping the type checker should be reasonably simple.

I guess we just show an error on both accessor if none of them has a documentation comment?

reduckted:
> I guess we just show an error on both accessor if none of them has a documentation comment?

Yep, that's exactly what it does.

ajafff:
[...]

PR #3497, opened 2017-11-18
completed-docs: Added option to check get and set accessors.

reduckted:
#### PR checklist

- [x] Addresses an existing issue: #3479
  - [x] Includes tests
- [x] Documentation update

#### Overview of change:
The `"properties"` for the `completed-docs` rule now checks `GetAccessor` and `SetAccessor` node types.

#### CHANGELOG.md entry:

[enhancement] `"properties"` option for `completed-docs` rule now checks getter and setter accessors.

### DIFF
--- a/src/rules/completedDocsRule.ts
+++ b/src/rules/completedDocsRule.ts
@@ -356,11 +356,8 @@ function walk(context: Lint.WalkContext<ExclusionsMap>, typeChecker: ts.TypeChec
                 break;
 
             case ts.SyntaxKind.GetAccessor:
-                checkNode(node as ts.GetAccessorDeclaration, ARGUMENT_PROPERTIES);
-                break;
-
             case ts.SyntaxKind.SetAccessor:
-                checkNode(node as ts.SetAccessorDeclaration, ARGUMENT_PROPERTIES);
+                checkAccessor(node as ts.NamedDeclaration);
         }
 
         return ts.forEachChild(node, cb);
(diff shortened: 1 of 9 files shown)


ajafff:
> Added an "accessor" option to the completed-docs rule that causes GetAccessor and SetAccessor node types to be checked for JSDoc comments.

To the consumer it doesn't matter if it's an accessor or a "real" property. That's an implementation detail.
I think checking accessors should be enabled by enabling `"properties"`.

reduckted:
Not sure why the "testNext" configuration is failing. The "trailing-comma" rule has failing tests, but my changes have nothing to do with that. 😕

ajafff:
@reduckted it's a bug in typescript@next, see https://github.com/palantir/tslint/pull/3494#issuecomment-345191452

reduckted:
@ajafff Ah, good catch. Accessors in object literals were expected to have documentation comments. I've made the rule skip over them now. 

Properties on object literals didn't require documentation. I'm not too familiar with the entire TypeScript syntax tree, but I'm guessing properties in object literals and properties in classes don't use the same `SyntaxKind`.

ajafff:
@reduckted

> I'm guessing properties in object literals and properties in classes don't use the same SyntaxKind.

That's correct. But `MethodDeclaration` is the same for both.

ajafff:
thanks @reduckted 

I'll open a follow-up PR to exclude method declarations in object literals

Issue #2806, opened 2017-05-21
Suggestion: `import-path` rule


andy-hanson:
```ts
import { x } from "foo/index";
import { x } from "foo.js";
```

This rule would warn when `/index` or `.js` are unnecessarily added on to the end of an import.
I noticed that `"../index"` is used a lot in this repository; I would really prefer just `".."` though.

vladimiry:
Is it not enough to simply forbid "index"-like imports using existing rules? `importBlacklistRule` is about to get the regexp patterns support https://github.com/palantir/tslint/pull/3504 or there is [no-import-zones](https://github.com/vladimiry/tslint-rules-bunch) rule that supports 
[micromatch](https://github.com/micromatch/micromatch) patterns.

JoshuaKGoldberg:
This seems difficult. What if the following all exist within a directory?

* `foo/index/index.js`
* `foo/index.js`
* `foo.js`
* `foo.js/index.js`

JoshuaKGoldberg:
### 💀 _**It's time!**_ 💀

**TSLint is being deprecated** and no longer accepting pull requests for major new changes or features. See #4534. 😱

If you'd like to see this change implemented, you have two choices:
* **Recommended:** Check if this is available in [ESLint](https://eslint.org) + [typescript-eslint](https://typescript-eslint.io) ✅
* _Not Recommended: Fork TSLint locally_ 🤷‍♂️

👋 It was a pleasure open sourcing with you!

_If you believe this message was posted here in error, please comment so we can re-open the issue!_

PR #3504, opened 2017-11-21
update importBlacklistRule to use regular expressions

jbsingh:
#### PR checklist

  - [x] Includes tests
- [x] Documentation update

#### Overview of change:

The current importBlacklistRule only allows exact match blacklisting.  To allow greater flexibility, would like to update this rule to use regular expression pattern match blacklisting.  Existing users of the importBlacklistRule would need to make a small update to retain expected behavior, specifically they'd need to place ^ at the start and $ at the end of their existing blacklist entries as shown in the updated documentation.

#### CHANGELOG.md entry:

[rule-change] `import-blacklist` updated to use regular expression patterns for blacklisting. Place ^ at the start and $ at the end of existing blacklist entries to retain old behavior.


src/rules/importBlacklistRule.ts
```diff
@@ -24,25 +24,32 @@ export class Rule extends Lint.Rules.AbstractRule {
...
-            Instead only sub modules may be imported from that module.`,
+            Disallows importing the specified modules directly via \`import\` and \`require\`.`,
         rationale: Lint.Utils.dedent`
-            Some libraries allow importing their submodules instead of the entire module.
-            This is good practise as it avoids loading unused modules.`,
-        optionsDescription: "A list of blacklisted modules.",
+            Some projects may wish to exclude imports matching certain patterns.
+
+            Examples:
+            ^loadsh$ Blacklist any import of the entire lodash module.
```
mikehaas763: typo in the example for lodash

jbsingh: Thanks Mike, I updated the example.

mikehaas763:
just noticed the label, maybe there is a way to do this in a backward compatible manner?

jbsingh:
Thanks Mike.  When originally coding this, I was hoping to make it backward compatible to eliminate impacts to any existing configuration.  Ultimately, I favored the approach chosen for the following reasons:

1. exact matching can be accomplished by regular expressions so it felt cleaner to use a single more powerful configuration option
2. supporting both existing exact match configuration in a non-breaking fashion as well as regular expression configuration would introduce some complexity in the code and configuration
3) the upgrade path is straightforward (anyone using the rule just needs to add anchors ^$ to retain existing functionality, which would be stated in the release notes that users should consult when upgrading)

If there’s strong community sentiment toward introducing the regular expression capability without a breaking change, I can definitely update the code in this PR.  Anyone have any thoughts about this?

adidahiya:
I'd prefer to do this in a backwards compatible manner. [Regex search is slower than `indexOf`](https://jsperf.com/regexp-vs-indexof) and I think the extra complexity of supporting both configuration types is worth it here.

jbsingh:
I made changes so that the solution is backwards compatible, but have not pushed the updates. I wanted to first find out what the protocol is. Since the changes were previously approved, should I wait for additional feedback or proceed with the backwards compatible solution?

krawetko:
merge?

ericanderson:
I'm sorry this took us so long to get to. It seems to be a bit at odds with import-blacklist as it exists now. Can you update it?

JoshuaKGoldberg:
Tangentially, this should also fix #2806

jbsingh:
I should be able to push an update within the next day or two.

leachiM2k:
Hello guys, thank you for merging this PR. This is exactly what we've needed.
But I have a strange issue with this PR. I can't find the implemented option (and docs) in the current 5.12.1 release on npm.

adidahiya:
@leachiM2k it wasn't released with 5.12.1. You can find out if a commit is in a release by clicking on the hash in the merge message:

image

and then looking at the tags list here:

image

there are no tags, so it is unreleased. this will be released in the next minor version.

Issue #3489, opened 2017-11-15
space-within-parens should allow exceptions


mmakrzem:
I should be allowed to specify that I want a space inside parentheses but no space if there is nothing being passed in.  ie:

Foo( 2 ); //good
Bar(); //also good

as it is now, I have to place a space inside the second case. ie Bar( ); //boo I don't want to have to insert a space between the empty parentheses

See how eslint handles spaces inside parentheses here: https://eslint.org/docs/rules/space-in-parens

ajafff:
I'd like to limit the scope to always allow '()' if there is nothing between the parens.
Exceptions like in ESLint can be implemented later on.

ajafff:
Fixed by #3513

PR #3513, opened 2017-11-24
update spaceWithinParensRule to always allow empty parens

jbsingh:
#### PR checklist

- [x] Addresses an existing issue: #3489
  - [x] Includes tests
- [x] Documentation update

#### Overview of change:

Updated spaceWithinParensRule to always allow empty parentheses ().  Also removed use of a hard coded character code.

#### CHANGELOG.md entry:

[enhancement] `space-within-parens` updated to always allow empty parentheses ().


src/rules/spaceWithinParensRule.ts
```diff
@@ -73,9 +73,13 @@ class SpaceWithinParensWalker extends Lint.AbstractWalker<Options> {
     public walk(sourceFile: ts.SourceFile) {
         forEachToken(sourceFile, (token: ts.Node) => {
             if (token.kind === ts.SyntaxKind.OpenParenToken) {
-                this.checkOpenParenToken(token);
+                if (sourceFile.text.charAt(token.end) !== ts.tokenToString(ts.SyntaxKind.CloseParenToken)) {
+                    this.checkOpenParenToken(token);
+                }
             } else if (token.kind === ts.SyntaxKind.CloseParenToken) {
-                this.checkCloseParenToken(token);
+                if (sourceFile.text.charAt(token.end - 2) !== ts.tokenToString(ts.SyntaxKind.OpenParenToken)) {
```
ajafff: `ts.tokenToString(ts.SyntaxKind.OpenParenToken)` just returns `(`. Consider comparing with the string directly.
Same for CloseParenToken above

src/rules/spaceWithinParensRule.ts
```diff
@@ -116,10 +120,10 @@ class SpaceWithinParensWalker extends Lint.AbstractWalker<Options> {
             currentChar = this.sourceFile.text.charCodeAt(currentPos);
         }
         /**
-         * Number 40 is open parenthese char code, we skip this cause
-         * it's already been caught by `checkOpenParenToken`
+         * Skip the open parenthesis character because it
+         * has already been caught by `checkOpenParenToken`
          */
-        if (!ts.isLineBreak(currentChar) && currentChar !== 40) {
+        if (!ts.isLineBreak(currentChar) && this.sourceFile.text.charAt(currentPos) !== ts.tokenToString(ts.SyntaxKind.OpenParenToken)) {
```
ajafff: I don't see a reason to replace the char code. If you don't like the magic number, you can add a `const enum` (because they are inlined by the compiler) with the charcode.

test/rules/space-within-parens/force-two-spaces/test.ts.fix
```diff
@@ -17,6 +17,9 @@ if (  x === y  ) {
     return (  {result: true, error: (  x + y  )}  );
 }
 
-// be well
+// expected number of spaces provided
 new Foo(  );
 
+// empty parens are always allowed
+new Foo();
```
ajafff: The current implementation adds an error for `new Foo( )`, because 1 space !== 2 spaces.
If that is intended, please add a test

src/rules/spaceWithinParensRule.ts
jbsingh: Thanks for the feedback, I updated to compare directly with the corresponding strings.

src/rules/spaceWithinParensRule.ts
jbsingh: Thanks for the feedback, I put back the previous code.

test/rules/space-within-parens/force-two-spaces/test.ts.fix
jbsingh: Thanks for the feedback.  My understanding is that the existing test was checking for no spaces i.e. new Foo() and the corresponding error "Needs 2 whitespaces..." was appearing.  Since the empty parens should no longer error I just originally removed that error expectation and added a test to make sure 2 spaces work as expected.  There were existing tests covering the missing single and double space as well as extra spaces, so I didn't originally add more missing space checks.  Based on the review feedback I have added a test for new Foo( ); that covers 1 missing space.

ajafff:
Thanks @jbsingh

Issue #3350, opened 2017-10-19
prefer-conditional-expression: checks deeply nested else if


ajafff:
### Bug Report

- __TSLint version__: 5.7.0
- __TypeScript version__: 2.5.1
- __Running TSLint via__: CLI

#### TypeScript code being linted

```ts
if (length === 0) {
    result = "foo";
} else if (length === 1) {
    result = "bar";
} else if (length === 2) { // error on this line
    result = "something else";
} else {
    result = "unknown";
}
```

with `tslint.json` configuration:

```json
"prefer-conditional-expression": true
```

#### Actual behavior

Error for the last `if else` nested inside `if else if else`. Note that the `if else` count needs to be even for this behavior to occur.

#### Expected behavior

No error, because `"check-else-if"` is not enabled.

adidahiya:
let's try to get this in v5.9

PR #3528, opened 2017-11-29
prefer-conditional-expression: avoid error on nested if statement

ajafff:
#### PR checklist

- [x] Addresses an existing issue: #3350
  - [x] Includes tests
- [ ] Documentation update

#### Overview of change:
[bugfix] `prefer-conditional-expression`: don't repeat error on nested if statements
Fixes: #3350

To fix the above issue, only the change under the `// Be careful not to fail again for the "else if"` comment is necessary.
While adding tests, I noticed that `check-else-if` added the error on the inner `if` when nesting with blocks. I fixed that along the way which leads to a bigger refactor.

### DIFF
--- a/test/rules/prefer-conditional-expression/check-else-if/test.ts.lint
+++ b/test/rules/prefer-conditional-expression/check-else-if/test.ts.lint
@@ -51,4 +51,56 @@ if (true) {
     foo(bar).baz = 1;
 }
 
+if (length === 0) {
+~~ [err % ('x')]
+    x = "foo";
+} else if (length === 1) {
+    x = "bar";
+} else if (length === 2) {
+    x = "something else";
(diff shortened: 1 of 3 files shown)


test/rules/prefer-conditional-expression/check-else-if/test.ts.lint
```diff
@@ -51,4 +51,56 @@ if (true) {
...
+}
+
+if (length === 0) {
+~~ [err % ('x')]
+    x = "foo";
+} else {
+    if (length === 1) {
+        x = "bar";
+    } else {
+        if (length === 2) {
```
ajafff: The error used to be reported on this line. I fixed that while I was already touching the code.

PR #3529, opened 2017-11-29
docs update - Add `hasFix` metadata for the indent rule

chrismbarr:
#### PR checklist

  - [ ] Includes tests
- [x] Documentation update

#### Overview of change:

Add the `hasFix` property to the metadata of the `indent` rule.  

The [docs for this rule](https://palantir.github.io/tslint/rules/indent/) mention auto-fixing, but the green "**🔧 Has Fixer**" badge doesn't appear, which could lead people to assume (like I did) that this rule has no fixer.

#### Future Suggestions:
Perhaps make the `hasFix` property required instead of optional so that rule authors are required to correctly set this properly.  Further more, maybe add a check to see if `hasFix: false` is set but the rule contains a call on the `Lint.Replacement`, or `Lint.RuleFailure` has something passed into the 6th parameter for a fix.

### DIFF
--- a/src/rules/indentRule.ts
+++ b/src/rules/indentRule.ts
@@ -66,6 +66,7 @@ export class Rule extends Lint.Rules.AbstractRule {
             [true, OPTION_USE_SPACES, OPTION_INDENT_SIZE_4],
             [true, OPTION_USE_TABS, OPTION_INDENT_SIZE_2],
         ],
+        hasFix: true,
         type: "maintainability",
         typescriptOnly: false,
     };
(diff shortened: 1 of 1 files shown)


adidahiya:
I think the docs description is misleading then. We need #3395 before declaring this rule as having a proper fixer.

chrismbarr:
I mean... [it *does* have a fixer right now](https://github.com/palantir/tslint/blob/master/src/rules/indentRule.ts#L124-L133) technically... but I see what you mean since the current one doesn't work in 100% of cases.

The docs do mention that it only works when certain options are set though, which seems sufficient enough to add this badge.


My use case & thought process: I work on a project with mixed tabs/spaces and I want to normalize them.
>I enable this rule to see all the violations... wow there are a lot!  I check the docs in hope of an auto-fix... nope, I guess I have to do all these fixes manually!  This sucks... oh wait, it **does** have an auto fix.  This is just a documentation issue,

chrismbarr:
@adidahiya Check out my latest changes, would adding text like that be sufficient?

adidahiya:
@chrismbarr looks good, but linting failed

PR #3530, opened 2017-11-29
 Inherit defaultSeverity and apply it to preceding base configs

ajafff:
Based on #3449. Only my commits after the merge are relevant for this PR.

This adds the missing piece to inherit the `defaultSeverity` from the last base config that specifies it. This inherited default is then applied to all base configs.

It turned out to be a bit complicated to do this while still maintaining a backwards compatible API.


test/config/tslint-extends-default-severity-precendence.json
```diff
@@ -0,0 +1,8 @@
+{
```
adidahiya: nit: file name has a typo. "precedence". I'll fix this in a follow-up.

Issue #3275, opened 2017-10-04
no-duplicate-super rule does not validate when using a ternary operator


Nysosis:
### Bug Report

- __TSLint version__: 5.3.2
- __TypeScript version__: 2.3.4
- __Running TSLint via__: CLI

#### TypeScript code being linted

```ts
class Base {
	constructor(props?: any) { }
}

class IfElse extends Base {
	constructor(props?: any) {
		if (props) {
			super(props)
		} else {
			super()
		}
	}
}

class Ternary extends Base {
	constructor(props?: any) {
		props ? super(props) : super()
	}
}

```

with `tslint.json` configuration:

```json
{
	"rules": {
		"no-duplicate-super": true
	}
}

```

#### Actual behavior
Class named `Ternary` is failing to pass the `no-duplicate-super` rule

#### Expected behavior
Class named `Ternary` to pass the `no-duplicate-super` rule

ajafff:
Fixed by #3544

PR #3544, opened 2017-12-03
fix no-duplicate-super validates with ternary operator

mateuszwitkowski:
#### PR checklist

- [x ] Addresses an existing issue: #3275
  - [x ] Includes tests
- [ ] Documentation update

#### Overview of change:
Added case for ternary operator in no-duplicate-super rule.

#### CHANGELOG.md entry:

[bugfix] using ternary operator for calling super() now passes no-duplicate-super rule.

### DIFF
--- a/src/rules/noDuplicateSuperRule.ts
+++ b/src/rules/noDuplicateSuperRule.ts
@@ -81,8 +81,13 @@ function walk(ctx: Lint.WalkContext<void>): void {
                     : Kind.NoSuper;
 
             case ts.SyntaxKind.ConditionalExpression: {
-                const { whenTrue, whenFalse } = node as ts.ConditionalExpression;
-                return worse(getSuperForNode(whenTrue), getSuperForNode(whenFalse));
+                const { condition, whenTrue, whenFalse } = node as ts.ConditionalExpression;
+                const inCondition = getSuperForNode(condition);
+                const inBranches = worse(getSuperForNode(whenTrue), getSuperForNode(whenFalse));
+                if (typeof inCondition !== "number" && typeof inBranches !== "number") {
+                    addDuplicateFailure(inCondition.node, inBranches.node);
+                }
+                return worse(inCondition, inBranches);
             }
 
             case ts.SyntaxKind.IfStatement: {
--- a/src/rules/noDuplicateSuperRule.ts
+++ b/src/rules/noDuplicateSuperRule.ts
@@ -80,6 +80,11 @@ function walk(ctx: Lint.WalkContext<void>): void {
                     ? { node: node.parent! as ts.CallExpression, break: false }
                     : Kind.NoSuper;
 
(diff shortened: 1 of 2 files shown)


src/rules/noDuplicateSuperRule.ts
```diff
@@ -80,6 +80,11 @@ function walk(ctx: Lint.WalkContext<void>): void {
                     ? { node: node.parent! as ts.CallExpression, break: false }
                     : Kind.NoSuper;
 
+            case ts.SyntaxKind.ConditionalExpression: {
+                const { whenTrue, whenFalse } = node as ts.ConditionalExpression;
+                return worse(getSuperForNode(whenTrue), whenFalse !== undefined ? getSuperForNode(whenFalse) : Kind.NoSuper);
```
ajafff: `whenFalse` can't be undefined

mateuszwitkowski: You're right of course. I'll fix it. There's another issue though. If someone would use nested ternary operators super() will be called twice:

```class Ternary extends Base {
	constructor(props?: any) {
        props ? super(props) : super() ? super() : super(); // if no props passed super() is called twice
	}
}
```
Do you think such a case should be covered? If so can you give me a tip how to handle it?

src/rules/noDuplicateSuperRule.ts
```diff
@@ -80,6 +80,11 @@ function walk(ctx: Lint.WalkContext<void>): void {
                     ? { node: node.parent! as ts.CallExpression, break: false }
                     : Kind.NoSuper;
 
+            case ts.SyntaxKind.ConditionalExpression: {
+                const { whenTrue, whenFalse } = node as ts.ConditionalExpression;
+                return worse(getSuperForNode(whenTrue), getSuperForNode(whenFalse));
```
ajafff: to also handle `super()` calls inside the condition:
```ts
const inCondition = getSuperForNode(node.condition);
const inBranches = worse(getSuperForNode(whenTrue), getSuperForNode(whenFalse));
if (typeof inCondition !== "number" && typeof inBranches !== number) {
    addDuplicateFailure(inCondition.node, inBranches.node);
}
return worse(inCondition, inBranches);
```

*code is untested and written only in GitHub comment editor.

test/rules/no-duplicate-super/test.ts.lint
```diff
@@ -259,5 +259,14 @@ declare const n: number;
     }
 }
 
+// With ternary operator
+{
+    class {
```
ajafff: btw. this is not a valid class declaration. Class declarations need a name.

mateuszwitkowski: I thought about it but that's how it's done in all the other test cases for that rule. I've figured out there's some reason for it and followed it. I'll fix it in all test cases with updated PR.

ajafff:
Thanks @mateuszwitkowski

PR #3557, opened 2017-12-06
Removed use of type checker from completed-docs

### LINKED ISSUE #3365: New completed-docs "tags" property doesn't work with single-line doc comments
### Bug Report

- __TSLint version__: 5.8.0
- __TypeScript version__: 2.5.3
- __Running TSLint via__: CLI

```ts
...
/** @see Class */
export const balls: string = 'balls';
...
    /** @inheritDoc */
    public async doSomething () : Promise<string> {
        return balls;
    }
...
```

```json
...
"completed-docs": [true, {
	"classes": {"visibilities": ["exported"]},
	"enums": {"visibilities": ["exported"]},
	"functions": {"visibilities": ["exported"]},
	"interfaces": {"visibilities": ["exported"]},
	"methods": {"privacies": ["protected", "public"]},
	"namespaces": {"visibilities": ["exported"]},
	"properties": {"privacies": ["protected", "public"]},
	"types": {"visibilities": ["exported"]},
	"variables": {"visibilities": ["exported"]},
	"tags": {
		"content": {"see": [".*"]},
		"exists": ["inheritDoc"]
	}
}],
...
```

(Also tested with `tags` copied and pasted as a nested property of all of its sibling properties.)

#### Actual behavior

```
...
ERROR: /path/to/file.ts[12, 14]: Documentation must exist for exported variables.
ERROR: /path/to/file.ts[43, 2]: Documentation must exist for public,async methods.
...
```

#### Expected behavior

No errors.

### LINKED ISSUE #3376: Stop using the type checker in completed-docs
As per @ajafff's feedback in #2415: 

> If you want to get the jsdoc comment, you can simply do this:
> 
> const [firstChild] = documentationNode.getChildren();
> if (firstChild.kind === ts.SyntaxKind.JSDocComment) {
>     // firstChild is ts.JSDoc
>     // do stuff with the comment
> }
> For more info on JSDoc, see https://github.com/Microsoft/TypeScript/blob/master/src/compiler/types.ts#L2065
> You also don't need to parse it manually with the regex magic below. Typescript's parser already did that for you

> @JoshuaKGoldberg SyntaxWalker#visitNode will never visit any JSDoc nodes, that's why I pointed you to node.getChildren() which does some magic to include jsdoc as the first entries if available.
> However, I don't consider this a blocker for the PR. I'm fine with merging this without that change and tweaking it afterwards.

This should also fix #3365.

JoshuaKGoldberg:
Uses tsutil's `getJsDoc` instead.

#### PR checklist

- [x] Addresses an existing issue: Fixes #3376. Fixes #3365.
  - [x] Includes tests
- [x] Documentation update

#### Overview of change:

Makes `completed-docs` an untyped rule by removing usages of the TSLint program. It instead uses `tsutils`'s `getJsDoc`.

#### CHANGELOG.md entry:

[enhancement] Made completed-docs an untyped rule
[bugfix] Fixed single-line comment parsing for completed-docs


test/rules/completed-docs/accessors/test.ts.lint
```diff
@@ -45,7 +45,8 @@ export class CommentsOnGetOnly {
      */
     get myAccessor(): string { return this.myField; }
 
-    set myAccessor(value: string) { this.myField = value; } // Comments for the setter are inherited from the getter.
+    set myAccessor(value: string) { this.myField = value; }
+    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~        [Documentation must exist for properties.]
```
ajafff: I would actually avoid adding a failure here. The property is documented. Why duplicate the comment?

src/rules/completedDocsRule.ts
```diff
@@ -307,7 +307,22 @@ const modifierAliases: { [i: string]: string } = {
     export: "exported",
 };
 
-function walk(context: Lint.WalkContext<ExclusionsMap>, typeChecker: ts.TypeChecker) {
+/**
+ * @remarks See https://github.com/ajafff/tsutils/issues/16
+ */
+const getApparentJsDoc = (node: ts.Node): ts.JSDoc[] => {
+    if (node.kind === ts.SyntaxKind.VariableDeclaration) {
+        const parent = node.parent as ts.VariableDeclarationList;
```
ajafff: this assumtion is not always true... well, it might be in the context of this rule

src/rules/completedDocsRule.ts
```diff
@@ -307,7 +307,22 @@ const modifierAliases: { [i: string]: string } = {
...
+
+        return node === parent.declarations[0]
+            ? tsutils.parseJsDocOfNode(parent)
+            : tsutils.parseJsDocOfNode(node);
```
ajafff: this is not what you want to do.

there are a few things to consider here:
* if the `VariableDeclaration` is not the first of the declaration list, TypeScript won't attach any JSDoc to that variable. Neither should this rule.
* If you have a `VariableDeclarationList`, you need to get the JSDoc from the containing `VariableStatement`. `parseJsDocOfNode` will fail if you have `/**foo*/ export let foo;` for example.

You may want to look at `deprecationRule.ts:200`. We already do something similar there.

src/rules/completedDocsRule.ts
JoshuaKGoldberg: I think it is for this rule. Just to be safe, I kept the function internal to this file instead of making it more globally available (pending tsutils having its own).

src/rules/completedDocsRule.ts
```diff
@@ -366,43 +366,66 @@ function walk(context: Lint.WalkContext<ExclusionsMap>, typeChecker: ts.TypeChec
...
+    function checkAccessorNode(node: ts.AccessorDeclaration): void {
+        if (nodeIsExcluded(node, ARGUMENT_PROPERTIES, node) || nodeHasDocs(node)) {
+            return;
+        }
+
+        const parent = node.parent as ts.ClassDeclaration | ts.ClassExpression;
+        const correspondingKind = node.kind === ts.SyntaxKind.GetAccessor
+            ? ts.SyntaxKind.SetAccessor
+            : ts.SyntaxKind.GetAccessor;
+        const correspondingAccessor = getNodeOfKind(parent.members, correspondingKind);
```
ajafff: This will return the first accessor of the class, not the corresponding accessor.
You also need to check the name. Note that typescript doesn't merge documentation of accessors with computed names. 
You can use `getPropertyName` to get the name as string or `undefined` if it's computed.

JoshuaKGoldberg:
_(will look at version-specific failures soon - will also add test case for mismatched names in accessors)_

suchanlee:
whats the status of this PR?

JoshuaKGoldberg:
I _think_ I've addressed all of @ajafff's feedback. It's been through a few merges.

Also, some amusing errata: this also fixes the TS 2.8 issue with the `[Symbol.iterator]() {}` doc comment.

buu700:
Any recent updates on this? (Very eager to have back my completed-docs that hasn't worked since December 2016!) Would it be worth merging #3841 as a stopgap until this is completely ready to go?

PR #3561, opened 2017-12-07
fix typo

AlexxNica:
#### PR checklist

- [ ] Includes tests
- [X] Documentation update

```diff
-paramter
+parameter
```

@adidahiya Take a look. Is this the correct file?

If it's not and the correct one is `master/src/rules/noDuplicateVariableRule.ts`, then idk, because it's already without the typo:

image

### DIFF
--- a/rules/no-duplicate-variable/index.html
+++ b/rules/no-duplicate-variable/index.html
@@ -9,7 +9,7 @@ rationale: |-
 
   A variable can be reassigned if necessary -
   there's no good reason to have a duplicate variable declaration.
-optionsDescription: You can specify `"check-parameters"` to check for variables with the same name as a paramter.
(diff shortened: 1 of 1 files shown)


rules/no-duplicate-variable/index.html
```diff
@@ -9,7 +9,7 @@
 
   A variable can be reassigned if necessary -
   there's no good reason to have a duplicate variable declaration.
-optionsDescription: You can specify `"check-parameters"` to check for variables with the same name as a paramter.
+optionsDescription: You can specify `"check-parameters"` to check for variables with the same name as a parameter.
```
ajafff: This one is also just generated from the source file on the master branch. As I already said, this is already fixed. The generated docs are updated during the next release.

PR #3573, opened 2017-12-13
Added no-dynamic-delete rule

### LINKED ISSUE #3553: Port tslint-microsoft-contrib's no-delete-expression

JoshuaKGoldberg:
Equivalent to tslint-microsoft-contrib's no-delete-expression. Renamed because all deletes are technically of node kind DeleteExpression, though they may sometimes contain an expression as the argumentExpression.

Allowed string literals of invalid property accesses, numeric literals, and negative numeric literals because those can't be simplified.
Suggested fixes for converting valid string literals to direct lookups and simplified unnecessarily positive numbers (like `+7` to `7`).

Fixes #3553.

#### CHANGELOG.md entry:

[new-rule] `no-dynamic-delete`

### DIFF
--- a/src/rules/noDynamicDeleteRule.ts
+++ b/src/rules/noDynamicDeleteRule.ts
@@ -41,7 +41,7 @@ export class Rule extends Lint.Rules.AbstractRule {
 
 function walk(context: Lint.WalkContext<void>) {
     function checkDeleteAccessExpression(node: ts.Expression | undefined): void {
-        if (node === undefined || !ts.isElementAccessExpression(node)) {
+        if (node === undefined || !tsutils.isElementAccessExpression(node)) {
             return;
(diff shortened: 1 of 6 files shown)


src/rules/noDynamicDeleteRule.ts
```diff
@@ -0,0 +1,102 @@
...
+ */
+
+import * as tsutils from "tsutils";
+import * as ts from "typescript";
+
+import * as Lint from "../index";
+
+export class Rule extends Lint.Rules.AbstractRule {
+    public static metadata: Lint.IRuleMetadata = {
+        description: "Enforces not deleting computed expressions.",
```
adidahiya: "Bans using the delete operator with computed key expressions"

Issue #3576, opened 2017-12-14
completed-docs documentation for tags is incorrect


reduckted:
### Bug Report

- __TSLint version__: 5.8.0

The docs for `completed-docs` state:

> All types may also provide "tags" with members specifying tags that allow the docs to not have a body.
> * "content": Object mapping tags to RegExp bodies content allowed to count as complete docs.
> * "exists": Array of tags that must only exist to count as complete docs.

It says the property is `exists`, but according to the code, the property name is `existence`: 

https://github.com/palantir/tslint/blob/676a2066f34c08448a78eb3f4632b55b5a2a6ff1/src/rules/completed-docs/tagExclusion.ts#L38-L41

In addition to this, the example that uses the `tags` property is wrong:

```json
"completed-docs": [
  true,
  {
    "enums": true,
    "functions": {"visibilities": ["exported"]},
    "methods": {"locations": "instance", "privacies": ["public", "protected"]},
    "tags": {"content": {"see": ["#.*"]}, "exists": ["inheritdoc"]}
  }
]
```

It lists `tags` as a sibling to `methods`, but it should be, for example, a sibling to the `locations` and `privacies` properties of the `methods` object.

ajafff:
Fixed by #3578

PR #3578, opened 2017-12-14
completed-docs: corrected example config for tags

reduckted:
#### PR checklist

- [x] Addresses an existing issue: #3576
- [x] Documentation update

#### Overview of change:

Corrected the example config that uses `tags`.

Also, the published documentation says (incorrectly) that there is an "exists" property, but the committed docs correctly states that it's called "existence", so I guess the latest documentation just hasn't been pushed to the website yet.

#### CHANGELOG.md entry:

[no-log] completed-docs: fixed example config for tags


ajafff:
Thanks @reduckted 

> the committed docs correctly states that it's called "existence", so I guess the latest documentation just hasn't been pushed to the website yet.

That was already fixed. The docs are usually published as part of a new release.

Issue #3354, opened 2017-10-20
[enhancement request] fixer for newline-before-return rule


aervin:
Maybe there's a good reason this rule does not have a fixer?

PR #3584, opened 2017-12-15
[enhancement] fixer for newline-before-return rule

aervin:
#### PR checklist

- [X] Addresses an existing issue: [#3354](https://github.com/palantir/tslint/issues/3354)
  - [X] Includes tests
- [X] Documentation update

The failing test case. Are there better ways to check for indentation?


src/rules/newlineBeforeReturnRule.ts
```diff
@@ -51,6 +52,17 @@ class NewlineBeforeReturnWalker extends Lint.AbstractWalker<void> {
...
 
+    private getFixer(returnNode: ts.ReturnStatement, previousNode: ts.Node): Lint.Replacement {
+        const prevText = previousNode.getFullText(this.sourceFile);
+        const returnText = returnNode.getFullText(this.sourceFile);
+        const indentationOfPrevStmt = prevText.substring(0, prevText.search(/\S/)).match(/ /g)!.length;
+        return Lint.Replacement.replaceFromTo(
+                                    returnNode.getFullStart(),
+                                    returnNode.getEnd(),
+                                    `\n\n${" ".repeat(indentationOfPrevStmt) + returnText.trim()}`,
+                                );
```
ajafff: Instead of detecting any indentation, I would just insert an additional line (or two) break at `returnNode.pos`.
This probably needs some special handling for comments.
It should also use `\r\n` if that's already used in the file. See `eoflineRule.ts` for example.

aervin: So my initial approach was:
```ts
return Lint.Replacement.appendText(returnNode.pos, NEWLINE_TYPE);
``` 
but this causes the fixer for the following test case to fail:
```ts
function foo(fn) {
    fn(); return;
          ~nil [0]
}
```

Actual `tslint` output being:
```ts
function foo(fn) {
    fn();
 return; /* leading-space here */
}
```
rather than
```ts
function foo(fn) {
    fn();
    return;
}
```

Perhaps I'm missing a simple solution?

ajafff: That's the only case that would need special handling. I don't think it's worth the effort because the code is poorly formatted before the fix, too.
The resulting code can be automatically fixed by the `align` (or eventually `indent`) rule.

Otherwise you need to worry about the correct indentation char here. Currently you assume blank only.

aervin:
Not sure how many rules run into the `\n` vs `\r\n` check, but may make sense to add it as a utility function somewhere?

ajafff:
There are quite a few rules that would benefit from a utility function:

* eofline
* curly
* file-header
* ordered-imports
* newline-per-chained-method (does it have a fixer?)

aervin:
Then I can open a separate PR for that change. Would the utility function typically be added via `tsutils`? I see a `utils` dir, but also commit messages about deprecating it.

ajafff:
> Would the utility function typically be added via tsutils? I see a utils dir, but also commit messages about deprecating it.

I'd accept a PR to tsutils.
Regarding the deprecation of utils.ts: we plan to deprecate and remove all functions that are available through tsutils. Eventually we also don't want to export these utils anymore, making them local to this project and not available for custom rules.

aervin:
Okay, I can do that. 

Regarding the fixer for `newline-per-chained-method`, I was observing the progression of this [PR](https://github.com/palantir/tslint/pull/3395) because I thought it would make the fixer easier to reason about. And where will that rule end up once rules are split out into more specific packages? Maybe the fixer is unnecessary?

ajafff:
@aervin I just released tsutils@2.14.0 with a function `getLineBreakStyle` that returns either `\n` or `\r\n`

listepo:
Hi @ajafff any news?

Issue #2099, opened 2017-01-22
Support linting vue/html file


nrip-monotype:
### Feature Request

- __TSLint version__: `4.3.1`
- __TypeScript version__: `2.1.5`
- __Running TSLint via__: Node.js API

#### TypeScript code being linted

```html
<template>
  <q-layout>
    <div slot="header" class="toolbar">
      <q-toolbar-title :padding="0">
        Quasar Framework v{{quasarVersion}}
      </q-toolbar-title>
    </div>

    <!--
      Replace following "div" with
      "<router-view class="layout-view">" component
      if using subRoutes
    -->
    <div class="layout-view">
      <div class="logo-container non-selectable no-pointer-events">
        <div class="logo" :style="position">
          <img src="~assets/quasar-logo.png">
          <p class="caption text-center">
            <span class="desktop-only">Move your mouse!!!.</span>
            <span class="touch-only">Touch screen and move.</span>
          </p>
        </div>
      </div>
    </div>
  </q-layout>
</template>

<script lang="ts">
  import * as Quasar from 'quasar';
  import { Utils } from 'quasar';
  import * as Vue from 'vue';
  import Component from 'vue-class-component';

  const moveForce = 30;
  const rotateForce = 40;

  @Component({
  })
  export default class Index extends Vue {
    rotateX: number;
    rotateY: number;
    moveY: number;
    moveX: number;
    quasarVersion: string;
   ...  
 }
</script>

<style lang="stylus">
....
</style>
```

with `tslint.json` configuration:

```json
{
  "rules": {
    "class-name": true,
    "curly": true,
    "eofline": false,
    "expr" : true,
    "forin": true,
    "indent": [true, "spaces"],
    "label-position": true,
    "label-undefined": true,
    "max-line-length": [true, 140],
    "no-arg": true,
    "no-bitwise": true,
    "no-console": [true,
      "debug",
      "info",
      "time",
      "timeEnd",
      "trace"
    ],
    "no-construct": true,
    "no-debugger": true,
    "no-duplicate-key": true,
    "no-duplicate-variable": true,
    "no-empty": true,
    "no-eval": true,
    "no-string-literal": false,
    "no-switch-case-fall-through": true,
    "no-trailing-comma": true,
    "no-trailing-whitespace": true,
    "no-unused-expression": false,
    "no-unused-variable": true,
    "no-unreachable": true,
    "no-use-before-declare": true,
    "one-line": [true,
      "check-open-brace",
      "check-catch",
[...]

PR #3596, opened 2017-12-20
[WIP] Add plugin support to customize linting (vue support)

Toilal:
#### PR checklist

- [x] Addresses an existing issue: #2099
- [ ] Includes tests
- [ ] Documentation update

#### Overview of change:

Add plugin support. It's meant to be used with https://github.com/Toilal/tslint-plugin-vue which is a plugin that implements #2099, but it may open other possibilities too ...

This is a work in progress, but feel free to try. You can add the plugin with `npm install git+https://https://github.com/Toilal/tslint-plugin-vue` and enable it by adding `plugins: 'vue'` into `tslint.json`.

Currently it doesn't work with `-p/--project` option, you have to use `-c/--config` option and it doesn't support type checking.

I also plan plugins to be able to yield additional files that are not reported by typescript program. (would be useful for vue plugin to work with `-p/--project`).

I'll add test and docs later.

### DIFF
--- a/src/configuration.ts
+++ b/src/configuration.ts
@@ -50,6 +50,11 @@ export interface IConfigurationFile {
     }>;
 
     /**
+     * An array of plugins to use.
+     */
+    plugins: string[];
+
+    /**
      * Directories containing custom rules. Resolved using node module semantics.
      */
     rulesDirectory: string[];
@@ -76,6 +81,7 @@ export const DEFAULT_CONFIG: IConfigurationFile = {
     defaultSeverity: "error",
     extends: ["tslint:recommended"],
     jsRules: new Map<string, Partial<IOptions>>(),
+    plugins: [],
     rules: new Map<string, Partial<IOptions>>(),
     rulesDirectory: [],
 };
@@ -84,6 +90,7 @@ export const EMPTY_CONFIG: IConfigurationFile = {
     defaultSeverity: "error",
     extends: [],
     jsRules: new Map<string, Partial<IOptions>>(),
+    plugins: [],
     rules: new Map<string, Partial<IOptions>>(),
     rulesDirectory: [],
 };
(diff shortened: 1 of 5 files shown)


palantirtech:
Thanks for your interest in [palantir/tslint](https://github.com/palantir/tslint), @Toilal! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

ajafff:
Let's start with some high-level feedback on the approach:

* not working with `--project` is probably a big deal (I know this is a quite complex thing to accomplish)
* destroying your source files with `--fix` is an even bigger deal (see below)
* error locations are wrong (if you add or remove characters, like your plugin does)

Toilal:
Thanks for first review.

I'm currently working on `--project` support by creating a custom ts program. @prograhammer has already [done this here](https://github.com/Realytics/fork-ts-checker-webpack-plugin/pull/77/files), so it should be possible to port this here.

As I understand, for `--fix` option to work, locations must be exact ? So the `// tslint:enable` + `// tslint:disable` trick is not an option ... vue-template-compiler supports {pad: 'space'} option that would make the file locations exact, but we have to tell tslint to skip all blanks from the beginning of the file to avoid raising false positive rules related to those empty lines.

ajafff:
For `--fix` to work, the error locations need to be exact. For usable error reporting, they also need to be exact. So this needs to be fixed anyway.

After linting, before fixing, the plugin needs to map the error locations of each Failure to the correct location in the original file.

There's another problem with fixing: after the fixes are applied to the **original** source, this fixed original source is used to create the new SourceFile. There needs to be yet another hook for the plugin.

Hooking plugins into the creation of the TypeScript Program requires even more effort.

Putting it all together requires a complete refactor of `runner.ts` and `linter.ts` (AFAICT).

ajafff:
Possibly relevant if you want to hook into the file lookup while parsing tsconfig.json: https://github.com/Microsoft/TypeScript/issues/20878

nicocevallos:
@Toilal: Sorry, only a little typo issue. Check you wrote `https://` twice in the `npm install` instruction.
Waiting for add plugins to tslint and improve the work experience with Vue.js

Toilal:
Sorry, I don't have time to work on this, so I close the PR

PR #3607, opened 2017-12-23
[new-rule-option] [object-literal-sort-key] Add shorthand-first option

### LINKED ISSUE #3606: Conflict between object-shorthand-properties-first and object-literal-sort-key
### Bug Report

- __TSLint version__: ^5.8.0
- __TypeScript version__: ^2.6.2
- __Running TSLint via__: VSCode

```ts
const a = 'a'
const c = 'c'
const obj = {
  a,
  b: 'b',
  c,
}
```

```json
{
  "extends": ["tslint:recommended", "tslint-config-airbnb", "tslint-react"]
}
```

#### Actual behavior

The rules conflict. There isn't a single field ordering that matches both rules, resulting on a non-fixable warning.

##### a, b, c
object-shorthand-properties-first

##### a, c, b
object-literal-sort-keys

#### Expected behavior

Not conflict.

_PS: I see this was mentioned on #2040 by @ajafff_

_Also I noticed `object-shorthand-properties-first` is from another repo but the fix makes more sense to be inside `object-literal-sort-key`_

brunolemos:
Fix #3606

Similar to PR #2831 but different implementation and focused only on shorthand-first

#### PR checklist

- [x] Addresses an existing issue: #3606
- [x] New feature, bugfix and enhancement 🤔
  - [x] Includes tests
- [x] Documentation update

#### CHANGELOG.md entry:

[new-rule-option] [object-literal-sort-key] Add shorthand-first option

### DIFF
--- a/test/rules/object-literal-sort-keys/shorthand-first/test.ts.lint
+++ b/test/rules/object-literal-sort-keys/shorthand-first/test.ts.lint
@@ -1,0 +1,79 @@
+const a = 'a'
+const c = 'c'
+const x = {}
+
+const fail1 = {
+  a,
+  b: 'b',
+  c,
+  ~  [shorthandErr % ('c')]
+  d: 'd',
+}
+
+const fail2 = {
+  c,
+  a,
+  ~  [alphaErr % ('a')]
+  b: 'b',
+  d: 'd',
+}
+
+const fail3 = {
+  a,
+  c,
+  d: 'd',
+  b: 'b',
+  ~       [alphaErr % ('b')]
+}
+
+const fail4 = {
+  c,
+  ...x,
+  a,
+  d: 'd',
+  b: 'b',
+  ~       [alphaErr % ('b')]
+}
+
+const pass1 = {
+  a,
+  c,
+  b: 'b',
+  d: 'd',
+}
+
+const pass2 = {
+  ...x,
+  a,
+  c,
+  b: 'b',
+  d: 'd',
+}
+
+const pass3 = {
+  a,
+  c,
+  ...x,
+  b: 'b',
+  d: 'd',
+}
+
+const pass4 = {
+  c,
+  ...x,
+  a,
+  b: 'b',
+  d: 'd',
+}
+
+const pass5 = {
+  d: 'd',
+  ...x,
+  c,
+  ...x,
+  a,
+  b: 'b',
+}
+
+[alphaErr]: The key '%s' is not sorted alphabetically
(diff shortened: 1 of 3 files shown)


palantirtech:
Thanks for your interest in [palantir/tslint](https://github.com/palantir/tslint), @brunolemos! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

brunolemos:
@ajafff any thing I should do before merge? thanks

ajafff:
Thanks @brunolemos

brunolemos:
@ajafff awesome, thanks 🙌

brunolemos:
Just tested on a real project, works great.

In case someone get here, I had to add these two rules to my `tslint.json`:

```json
"object-literal-sort-keys": [true, "shorthand-first"],
"object-shorthand-properties-first": false,
```

...to avoid duplicated error report, because `object-shorthand-properties-first` was being added by one of these: `"extends": ["tslint:recommended", "tslint-react", "tslint-config-airbnb", "tslint-config-prettier"],`

test/rules/object-literal-sort-keys/shorthand-first/test.ts.lint
```diff
@@ -0,0 +1,79 @@
+const a = 'a'
+const c = 'c'
+const x = {}
+
+const fail1 = {
+  a,
+  b: 'b',
+  c,
+  ~  [shorthandErr % ('c')]
```
adidahiya: what happens in the following example? I expect no errors because a `...x` spread should reset the sorting.

```ts
const fail1 = {
  a,
  b: 'b',
  ...x,
  c,
  d: 'd',
};
```

I think there needs to be a test case like this; I don't see one currently in this file.

brunolemos: @adidahiya Yes this would give no error. The [pass5](https://github.com/brunolemos/tslint/blob/7b1223ecac22a43235d3cbd42f79b52a8ba8ac1a/test/rules/object-literal-sort-keys/shorthand-first/test.ts.lint#L69-L76) seems to be the closest to that for now:

```js
const pass5 = {
  d: 'd',
  ...x,
  c,
  ...x,
  a,
  b: 'b',
}
```

adidahiya: ah, I missed pass5; that looks sufficient.

Issue #3563, opened 2017-12-08
[bug] unnecessary space before parameter after running fixer for arrow-parens


mateuszwitkowski:
### Bug Report

- __TSLint version__: 5.8.0
- __TypeScript version__: 2.5.3
- __Running TSLint via__: VSCode

#### TypeScript code being linted

```ts
somearray.map((element) => doSomething(element));
```

with `tslint.json` configuration:

```json
"rules": {
    "arrow-parens": [
      true,
      "ban-single-arg-parens"
    ]

```

#### Actual behavior
After running fixer:
```ts
somearray.map( element => doSomething(element));
```
There`s unnecessary space before parameter.
#### Expected behavior
```ts
somearray.map(element => doSomething(element));
```
No space before parameter.

ajafff:
Fixed by #3618

PR #3618, opened 2018-01-01
fix no unnecessary space before parameter from arrow-parens fixer

mateuszwitkowski:
#### PR checklist

- [x] Addresses an existing issue: #3563
  - [x] Includes tests
- [ ] Documentation update

#### Overview of change:
Fixer for ban-single-arg-parens doesn't add unnecessary whitespace before argument if parent node is a call expresion.

#### CHANGELOG.md entry:
[bugfix] no unnecessary whitespace before argument in callback functions fixed with arrow-parens fixer

### DIFF
--- a/src/rules/arrowParensRule.ts
+++ b/src/rules/arrowParensRule.ts
@@ -19,6 +19,7 @@ import { getChildOfKind, isArrowFunction } from "tsutils";
 import * as ts from "typescript";
 
 import * as Lint from "../index";
+import { isCallExpression } from 'typescript';
 
 const BAN_SINGLE_ARG_PARENS = "ban-single-arg-parens";
 
@@ -72,8 +73,10 @@ function walk(ctx: Lint.WalkContext<Options>) {
                 }
             } else if (ctx.options.banSingleArgParens) {
                 const closeParen = getChildOfKind(node, ts.SyntaxKind.CloseParenToken)!;
+                const parentNode = node.parent!;
+                const replaceValue = isCallExpression(parentNode) ? "" : " ";
                 ctx.addFailureAtNode(node.parameters[0], Rule.FAILURE_STRING_EXISTS, [
-                    Lint.Replacement.replaceFromTo(openParen.pos, node.parameters[0].getStart(ctx.sourceFile), " "),
+                    Lint.Replacement.replaceFromTo(openParen.pos, node.parameters[0].getStart(ctx.sourceFile), replaceValue),
                     Lint.Replacement.deleteFromTo(node.parameters[0].end, closeParen.end),
                 ]);
             }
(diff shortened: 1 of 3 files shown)


test/rules/arrow-parens/ban-single-arg-parens/test.ts.fix
```diff
@@ -38,3 +40,4 @@ let foo = bar => bar;
 
 let foo = async bar => bar;
 
+arr.map(a => b);
```
ajafff: this will probably fail for `foo(async (a) => b)` by fixing it to `foo(asyca => b)`

src/rules/arrowParensRule.ts
```diff
@@ -72,8 +72,10 @@ function walk(ctx: Lint.WalkContext<Options>) {
                 }
             } else if (ctx.options.banSingleArgParens) {
                 const closeParen = getChildOfKind(node, ts.SyntaxKind.CloseParenToken)!;
+                const parentNode = node.parent!;
+                const replaceValue = ts.isCallExpression(parentNode) ? "" : " ";
```
ajafff: This should not be limited to CallExpressions. This is also relevant for NewExpression and probably many other.
I propose to remove the blank all the time if there is no alpha character at `openParen.pos - 1`. That includes `async (a) => b` and `export default (a) => b`.
Another rule will probably complain about a missing blank, but that's not the business of this rule.

mateuszwitkowski:
@ajafff I've updated my PR. Apart from checking for alpha characters fixer checks also assignemnt character. Otherwise it would fix `var a = (a) => {}` to `var a =a => {}`

src/rules/arrowParensRule.ts
```diff
@@ -72,8 +72,10 @@ function walk(ctx: Lint.WalkContext<Options>) {
                 }
             } else if (ctx.options.banSingleArgParens) {
                 const closeParen = getChildOfKind(node, ts.SyntaxKind.CloseParenToken)!;
+                const charBeforeOpenParen = ctx.sourceFile.text.substring(openParen.pos - 1, openParen.pos);
+                const replaceValue = charBeforeOpenParen.match(/[a-z=]/i) !== null ? " " : "";
```
ajafff: I don't think this rule should do anything related to code formatting. It adds the space only where strictly necessary to avoid invalid syntax.
Everything else assumes a certain style preference of the user, that this rule cannot know.

Anyways, if you really want to add a space after `=` you should probably also add one after
* `,` as in `foo(bar, (a) => b)`
* `?` and `:` as in `foo ? (a) => b : (a) => c`
* `&` and `|` as in `someFn || (a) => b`
* maybe more

test/rules/arrow-parens/ban-single-arg-parens/test.ts.lint
```diff
@@ -51,4 +53,9 @@ let foo = async (
     ~~~ [0]
 ) => bar;
 
+arr.map((a) => b);
+         ~ [0]
+
+foo(async (a) => b)
```
ajafff: can you please add some more tests like
```ts
function *gen() { yield (a) => b; }
export default (a) => b;
throw (a) => b; // hopefully nobody ever does this
```

src/rules/arrowParensRule.ts
mateuszwitkowski: I think you're right about unnecessary code formatting here. I'll change it and adjust test cases accordingly.

ajafff:
Thanks @mateuszwitkowski

Issue #3011, opened 2017-07-06
no-use-before-declare doesn't handle class members


NN---:
### Bug Report

- __TSLint version__: 5.5.0
- __TypeScript version__: 2.4.1
- __Running TSLint via__: gulp

#### TypeScript code being linted

```ts
class A { b: B = new B(); }
class B { }
```

with `tslint.json` configuration:

```json
no-use-before-declare: true
```

#### Actual behavior
variable B is used before declaration

#### Expected behavior
No error

Patrik-Lundqvist:
I also experience the same issue inside class members functions.

```ts
class A {
  test() {
      const b = new B(); // variable 'B' used before declaration
  }
}

class B {}
```

ajafff:
This request is not valid. Classes are not hoisted and can therefore not be used before declaration.

The following code would fail at runtime:

```ts
class A { b: B = new B(); }
new A();
class B { }
```

The current behavior of the rule is correct.

kodicat:
What about the code

```
export class One {
    toSecond() {
        return new Two();
    }
}

export class Two {
    toFirst() {
        return new One();
    }
}
```

It will not fail at runtime.

PR #3651, opened 2018-01-13
no-use-before-declare rule ignores creating instances of a class

mateuszwitkowski:
#### PR checklist

- [x ] Addresses an existing issue: #3011
- [ x] New feature, bugfix, or enhancement
  - [x ] Includes tests
- [ ] Documentation update

#### Overview of change:
Rule no-use-before-declare ignores creating instance of a class before class declaration.

#### CHANGELOG.md entry:
[enhancement]:  no-use-before-declare ignores using class before its declaration

### DIFF
--- a/src/rules/noUseBeforeDeclareRule.ts
+++ b/src/rules/noUseBeforeDeclareRule.ts
@@ -79,7 +79,8 @@ function walk(ctx: Lint.WalkContext<void>, checker: ts.TypeChecker): void {
         const declaredBefore = declarations.some((decl) => {
             switch (decl.kind) {
                 case ts.SyntaxKind.FunctionDeclaration:
-                    // Functions may be declared later.
+                case ts.SyntaxKind.ClassDeclaration:
+                    // Functions and classes may be declared later.
                     return true;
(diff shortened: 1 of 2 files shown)


src/rules/noUseBeforeDeclareRule.ts
```diff
@@ -79,7 +79,8 @@ function walk(ctx: Lint.WalkContext<void>, checker: ts.TypeChecker): void {
         const declaredBefore = declarations.some((decl) => {
             switch (decl.kind) {
                 case ts.SyntaxKind.FunctionDeclaration:
-                    // Functions may be declared later.
+                case ts.SyntaxKind.ClassDeclaration:
+                    // Functions and classes may be declared later.
```
ajafff: This is wrong. Classes are not hoisted and can therefore not be used before their declaration.

mateuszwitkowski: Yeah, you're right of course. Can you take down the 'Accepting PRs' label from this issue: #3011 ? Or close it since it's not a valid request?

PR #3673, opened 2018-01-24
Fixed "manging" typo in no-this-assignment

JoshuaKGoldberg:
It's been there since Jun 2017 :(


JoshuaKGoldberg:
CircleCI failures are _definitely_ not from this change :)

PR #3688, opened 2018-02-02
added no-implicit-constructor-visibility rule

mauriedo:
#### PR checklist

  - [x] Includes tests
- [x] Documentation update

#### Overview of change:
Adds the `no-implicit-constructor-visibility` rule, which forces to specify visibility modifiers for constructors and not rely in the default `public` modifier. The fix for it just prepends the `public` keyword at the beginning of the constructor statement.

I am being very restrictive with the implementation and detecting cases where constructors have no modifiers at all. If this is a concern I am happy to modify it to constructors that don't have `public`/`private`/`protected` modifiers, specifically.

Enabling this rule has made me fix almost 20 files from the source, I am also happy to add `//disable` statements instead of the fixes if that is what's preferred, sorry unaware of the conventions here as this is my first PR against the project.

#### CHANGELOG.md entry:
[new-rule] `arrow-return-shorthand`

### DIFF
--- a/src/configs/all.ts
+++ b/src/configs/all.ts
@@ -202,6 +202,7 @@ export const rules = {
     "no-angle-bracket-type-assertion": true,
     "no-boolean-literal-compare": true,
     "no-consecutive-blank-lines": true,
+    "no-implicit-constructor-visibility": true,
     "no-parameter-properties": true,
     "no-redundant-jsdoc": true,
     "no-reference-import": true,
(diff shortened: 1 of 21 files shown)


palantirtech:
Thanks for your interest in [palantir/tslint](https://github.com/palantir/tslint), @mauriedo! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

mauriedo:
Sorry, ignore. Just found this is already implemented as part of the 'member-access' rule :(

PR #3701, opened 2018-02-07
[docs] no-string-throw: add missing examples.

BB9z:
#### PR checklist

  - [ ] Includes tests
- [x] Documentation update

#### Overview of change:

This PR is really simple. no-string-throw rule miss an example.

https://palantir.github.io/tslint/rules/no-string-throw/


aervin:
@BB9z The ability to add code examples just landed in master. Will you please consider adding an example for this rule? I know it seems like a straightforward rule, but...

If you're willing, please see the `curly` rule for an example.

BB9z:
@aervin Just merge and check the latest code. no-string-throw rule has not options, so just `[true]` is enough.

aervin:
@BB9z The code examples would still be helpful for some people I think:

Passes:
```ts
try {
    // ...
} catch (e) {
    throw e;
}
```

Fails:
```ts
try {
    // ...
} catch {
    throw 'There was a problem.';
}
```

BB9z:
@aervin I see. The code examples have add into the rationale.

aervin:
@BB9z Looking good! But please see these files for the best way to add code examples to rules:

* [curly code examples file](https://github.com/palantir/tslint/blob/master/src/rules/code-examples/curly.examples.ts)
* [curly rule file](https://github.com/palantir/tslint/blob/master/src/rules/curlyRule.ts#L75)

BB9z:
@aervin done. Seems this is a new feature added two days ago :]

PR #3721, opened 2018-02-16
added prefer-empty-constructor rule

mauriedo:
#### PR checklist

  - [x] Includes tests
- [x] Documentation update

#### Overview of change:
Added a new rule to require constructors that only assign class members with their parameters to be empty instead, as the compiler does that initialisation for you.

I've had to jump through some hoops to deal with the fact that `node.body` can be undefined, I would like extra eyes on that bit. Also of course if there are other constructor shapes that wouldn't be covered by this.

#### CHANGELOG.md entry:
[new-rule] `prefer-empty-constructor`

### DIFF
--- a/src/configs/all.ts
+++ b/src/configs/all.ts
@@ -226,6 +226,7 @@ export const rules = {
         "named-imports-order": "case-insensitive",
         "module-source-path": "full",
     }],
+    "prefer-empty-constructor": true,
     "prefer-function-over-method": true,
     "prefer-method-signature": true,
     "prefer-object-spread": true,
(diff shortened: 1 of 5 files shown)


JoshuaKGoldberg:
There's a separate rule about not allowing parameter properties. Sometimes that's good, sometimes that's bad. It looks like this rule encourages the opposite. How would that rule and this one work when extending from `tslint:all`?

IMO it would be better to have more configuration on a single rule, choosing between something like the following three options:
* Always use constructor parameter properties
* Use parameter properties if all parameters become properties
* Never use constructor parameter properties

Issue #3562, opened 2017-12-07
Option for file-header to enforce a blank line after the header


JoshuaKGoldberg:
Bad (in some style preferences):
```typescript
/**
 * @copyright My Corporation.
 */
import * as React from "react";
```

Good (in some style preferences):
```typescript
/**
 * @copyright My Corporation.
 */

import * as React from "react";
```

If turned on by default, this extra setting could be auto-fixed easily (if the file header does match).

JoshuaKGoldberg:
This was fixed by #3740. 🎉

PR #3740, opened 2018-02-27
Enforce Blank Line after File Header

rwaskiewicz:
#### PR checklist

- [X] Addresses an existing issue: #3562
  - [X] Includes tests
- [X] Documentation update

#### Overview of change:
- Enforces new lines after file header via new configuration key to be provided in TSLint file


rwaskiewicz:
I don't _think_ CircleCI failures are a result of these changes (but could always be mistaken 🙂 )

src/rules/fileHeaderRule.ts
```diff
@@ -79,6 +89,18 @@ export class Rule extends Lint.Rules.AbstractRule {
...
     }
 
+    private doesNewLineEndingViolationExist(text: string, offset: number): boolean {
+        let violationExists = false;
+        const entireComment = ts.forEachLeadingCommentRange(
+            text,
+            offset,
+            (pos, end) => text.substring(pos, end + 2));
+
+        violationExists = entireComment !== undefined && entireComment.match(/^.*((\r)?\n){2,}$/mg) !== null;
```
aervin: Maybe assign the regex to a constant w/ a more descriptive name and use `DESCRIPTIVE_NAME.test(entireComment)` instead of match?

src/rules/fileHeaderRule.ts
```diff
@@ -79,6 +89,18 @@ export class Rule extends Lint.Rules.AbstractRule {
...
+
+        return violationExists;
```
aervin: Looks like you could nix the `violationExists` var and in-line the above expression w/ the return statement.

src/rules/fileHeaderRule.ts
aervin: After testing this expression [here](https://regexr.com/), it seems it is not returning a reliable result. Maybe you could strip all spaces from `entireComment` and use `entireComment.lastIndexOf("\n") !== entireComment.length - 1`?

src/rules/fileHeaderRule.ts
```diff
@@ -63,7 +71,9 @@ export class Rule extends Lint.Rules.AbstractRule {
             offset,
             (pos, end, kind) => text.substring(pos + 2, kind === ts.SyntaxKind.SingleLineCommentTrivia ? end : end - 2));
 
-        if (commentText === undefined || !headerFormat.test(commentText)) {
+        const trailingNewLineViolation = enforceExtraTrailingLine ? this.doesNewLineEndingViolationExist(text, offset) : false;
```
aervin: I think this config option is only relevant when the file header exists and is otherwise valid. So:
```ts
const trailingNewLineViolation = 
    enforceExtraTrailingLine &&
    headerFormat.test(commentText!) &&
    this.doesNewLineEndingViolationExist(text, offset)
```

In fact, since we need to a return a `RuleFailure` when `trailingNewLineViolation` is `true`, you could just inline this expression in an `if` statement;

src/rules/fileHeaderRule.ts
```diff
@@ -63,7 +71,9 @@ export class Rule extends Lint.Rules.AbstractRule {
...
+
+        if (commentText === undefined || !headerFormat.test(commentText) || trailingNewLineViolation) {
```
aervin: If there's a `trailingNewLineViolation`, we should return (before this `if` statement) a `RuleFailure` with a single `Replacement`, which just appends a newline to the end of the valid file header. So this check for `trailingNewLineViolation` will become unnecessary.

test/rules/file-header/bad-newline/test.ts.lint
```diff
@@ -0,0 +1,11 @@
+/*
+~nil [missing file header]
+ * Bad header
```
aervin: Should be updated to a valid header like `Good header 1`

test/rules/file-header/bad-single-newline/test.ts.lint
```diff
@@ -0,0 +1,9 @@
+// Bad header
```
aervin: Should also be updated to a valid header value

src/rules/fileHeaderRule.ts
aervin: Also looks like this regular expression is not returning reliable results (I tested [here](https://regexr.com/).

rwaskiewicz:
Thanks for the feedback @aervin ! I'm in the process of working through it now.  Hope to have something ready by this weekend 🙂

johnwiseheart:
Hi @rwaskiewicz,

I just want to completely understand this:

Would
```typescript
/*!
 * Good header
 */
export class A {
    public x = 1;
     public B() {
        return 2;
    }
}
```

pass or fail? It should fail right? If so - can you please add a test for this case?

johnwiseheart:
Oh my bad, I see this is covered in `test/rules/file-header/bad-newline/test.ts.lint`.

LGTM then!

jukben:
Hey, @rafalwrzeszcz isn't this rule implemented by chance contrariwise? 
I looks like it checks for the newline BEFORE headline, not AFTER (and auto fix actually adds newline AFTER)

@JoshuaKGoldberg

PR #3807, opened 2018-04-04
promise-function-async: add options to specify function types to check.

blair:
#### PR checklist

  - [x] Includes tests
- [x] Documentation update

#### Overview of change:
If one wants to enable the `promise-function-async` warning, then one gets warnings for all functions, including arrow functions that are passed to Promise#then() and Promise#catch(). There's no point in making these async then then() and catch() already handle the given function throwing an exception.

This commit allows one to select which function types to check. A useful option is to only check function and method declarations.

Question: after this PR, all the warnings will be disabled by default. Should they all be enabled if there are no options set?

#### CHANGELOG.md entry:

[rule-change] `promise-function-async` add rules options to enable checks for the four different types of functions.


palantirtech:
Thanks for your interest in [palantir/tslint](https://github.com/palantir/tslint), @blair! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

src/rules/promiseFunctionAsyncRule.ts
```diff
@@ -29,10 +51,30 @@ export class Rule extends Lint.Rules.TypedRule {
...
+                    OPTION_FUNCTION_DECLARATION,
+                    OPTION_FUNCTION_EXPRESSION,
+                    OPTION_ARROW_FUNCTION,
+                    OPTION_METHOD_DECLARATION,
+                ],
+            },
+            minLength: 0,
+            maxLength: 4,
+        },
+        optionExamples: [[true, OPTION_FUNCTION_DECLARATION, OPTION_METHOD_DECLARATION]],
```
suchanlee: maybe we can add two examples here, the one that you added and the other `true`?

src/rules/promiseFunctionAsyncRule.ts
```diff
@@ -19,6 +19,28 @@ import { hasModifier } from "tsutils";
...
+const KIND_FOR_OPTION: { [arg: string]: number } = {
+    [OPTION_FUNCTION_DECLARATION]: ts.SyntaxKind.FunctionDeclaration,
+    [OPTION_FUNCTION_EXPRESSION]: ts.SyntaxKind.FunctionExpression,
+    [OPTION_ARROW_FUNCTION]: ts.SyntaxKind.ArrowFunction,
+    [OPTION_METHOD_DECLARATION]: ts.SyntaxKind.MethodDeclaration,
+};
+
+type EnabledSyntaxKinds = ReadonlySet<number>;
+
+function parseOptions(ruleArguments: string[]): EnabledSyntaxKinds {
```
suchanlee: i think if this rule is enabled we should default to all kinds. Or else what's the point of turning the rule on in the first place?

blair:
@suchanlee Please take another look, I made the requested changes.

BTW, I didn't know if I should have a new commit on top of the original one or replace it, I did the later with a forced push.

PR #3823, opened 2018-04-09
Improved "complete docs" docs

304NotModified:
#### PR checklist

- [x] Documentation update

#### Overview of change:
- It was unclear to me if it works for JSDocs (because in another rule "JSDoc comment" is used instead of "Documentation"

- Updated the style for the inline code

supersedes https://github.com/palantir/tslint/pull/3006

### DIFF
--- a/src/rules/completedDocsRule.ts
+++ b/src/rules/completedDocsRule.ts
@@ -181,9 +181,9 @@ export class Rule extends Lint.Rules.TypedRule {
     /* tslint:disable:object-literal-sort-keys */
     public static metadata: Lint.IRuleMetadata = {
         ruleName: "completed-docs",
-        description: "Enforces documentation for important items be filled out.",
+        description: "Enforces JSDoc comments for important items be filled out.",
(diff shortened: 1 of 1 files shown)


src/rules/completedDocsRule.ts
```diff
@@ -181,9 +181,9 @@ export class Rule extends Lint.Rules.TypedRule {
     /* tslint:disable:object-literal-sort-keys */
     public static metadata: Lint.IRuleMetadata = {
         ruleName: "completed-docs",
-        description: "Enforces documentation for important items be filled out.",
+        description: "Enforces JSDoc comments for important items be filled out.",
         optionsDescription: Lint.Utils.dedent`
-            \`true\` to enable for [${Object.keys(Rule.defaultArguments).join(", ")}]],
+            \`true\` to enable for `[${Object.keys(Rule.defaultArguments).join(", ")}]`,
```
johnwiseheart: Do you need to escape this backtick?

Issue #922, opened 2016-01-18
New rule to forbid access to commonly-confused ambient globals


adidahiya:
For example, `window.event` ([here](https://github.com/Microsoft/TypeScript/blob/31bbac3c11c82af6cb93fd26ccf1632da6a04353/src/lib/dom.generated.d.ts#L12686) in lib.d.ts) is often referenced in erroneous code like this:

``` ts
function broken(evt: Event) {
    // Meant to do something with `evt` but typed it incorrectly.
    Event.target;  // compiler error
    event.target;  // should be a lint failure
}
```

myitcv:
Also referencing [this thread](https://github.com/Microsoft/TypeScript/issues/1351#issuecomment-160416723) because I'm unclear how such a lint rule could even be defined given the number of such variables.

Perhaps the solution can be approached by using a modified [`lib.dom.d.ts`](https://github.com/Microsoft/TypeScript/blob/master/lib/lib.dom.d.ts) that does _not_ define the globals like `event` but instead relies on the fact that the same objects/functions are available via `window` (which would be one of a well defined, very small number of globals that would include core types). That way, people would be forced to do something like this:

```typescript
function Apple() {
  console.log(event); // error; event no longer defined as a global
  console.log(window.event); // ok
}
```

Thoughts?

Not tested this so this might break some type definitions...?

adidahiya:
@myitcv Is such an approach feasible without [lib.d.ts modularity](https://github.com/Microsoft/TypeScript/issues/494) in TS 2.0?

This issue is fairly similar to #327. Both would almost certainly involve using the type checker (but not necessarily #680 because we could just tell langauge services to consider `lib.d.ts` and the type checker will pick up those global types).

myitcv:
@adidahiya 

> Is such an approach feasible without lib.d.ts modularity in TS 2.0?

It is possible but perhaps not the cleanest solution.

We use custom `lib.core.d.ts` and `lib.dom.d.ts` via the [`--noLib` compiler option](https://github.com/Microsoft/TypeScript/wiki/Compiler-Options). The custom `lib.core.d.ts` and `lib.dom.d.ts` files we use are well defined transformations of the [TypeScript provided versions](https://github.com/Microsoft/TypeScript/tree/master/lib). The transformations we currently apply are:

* All ambient `var`'s in both type definitions are redefined as `const`'s
[...]

PR #3824, opened 2018-04-09
[new-rule] no-restricted-globals

Zzzen:
#### PR checklist

- [x] Addresses an existing issue: #922
  - [x] Includes tests
- [ ] Documentation update

#### Overview of change:
  prevents accidental access to variables that are declared in `lib.dom.d.ts`.

  1. a descriptive rule name
  2. corner cases
  3. performance


palantirtech:
Thanks for your interest in [palantir/tslint](https://github.com/palantir/tslint), @Zzzen! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

src/rules/banDomGlobalRule.ts
```diff
@@ -0,0 +1,92 @@
...
+ * limitations under the License.
+ */
+
+import * as ts from "typescript";
+
+import * as Lint from "../index";
+
+export class Rule extends Lint.Rules.TypedRule {
+    /* tslint:disable:object-literal-sort-keys */
+    public static metadata: Lint.IRuleMetadata = {
```
JoshuaKGoldberg: Could you add a rationale explaining why this rule is useful? It would be confusing as a newcomer to TSLint reading the rule page on palantir.github.io/tslint why this would be useful.

It should probably also recommend specifically not using the dom lib in your tsconfig if you're not writing a browser app.

Zzzen: Haven't you check out [New rule to forbid access to commonly-confused ambient globals](https://github.com/palantir/tslint/issues/922)? I write browser app primarily and have been bit by `name` in `lib.dom.d.ts`. Maybe I should rename this to `forbid-ambient-globals` and provides options like 
```json
[
  {
    "file": "lib.dom.d.ts",
    "variables": [
      "event",
      "name"
    ]
  },
  {
    "file": "@types/jquery/index.d.ts",
    "variables": [
      "$"
    ]
  }
]
```

Zzzen: Definition file names are redundant since we cannot use a global declared in lib A if it is also declared and used in lib B.

Zzzen:
@JoshuaKGoldberg any recommendation? thanks

src/rules/noRestrictedGlobalsRule.ts
```diff
@@ -0,0 +1,97 @@
...
+export class Rule extends Lint.Rules.TypedRule {
+    /* tslint:disable:object-literal-sort-keys */
+    public static metadata: Lint.IRuleMetadata = {
+        ruleName: "no-restricted-globals",
+        description: "Disallow specific global variables.",
+        descriptionDetails: Lint.Utils.dedent`
+            Disallowing usage of specific global variables can be useful if you want to allow
+            a set of global variables by enabling an environment, but still want to disallow
+            some of those. For instance, early Internet Explorer versions exposed the current
+            DOM event as a global variable event, but using this variable has been considered
```
JoshuaKGoldberg: Nit: escape the "event" in "global variable event" with code to make it a little more clear.

```text
global variable \`event\`, but using
```

src/rules/noRestrictedGlobalsRule.ts
```diff
@@ -0,0 +1,97 @@
...
+            as a bad practice for a long time. Restricting this will make sure this variable
```
JoshuaKGoldberg: Nit: just "considered a bad practice"

src/rules/noRestrictedGlobalsRule.ts
```diff
@@ -0,0 +1,97 @@
...
+            ],
+        ],
+        type: "functionality",
+        typescriptOnly: false,
+        requiresTypeInfo: true,
+    };
+    /* tslint:enable:object-literal-sort-keys */
+
+    public static FAILURE_STRING(name: string) {
+        return `Unexpected global variable '${name}'. Use local parameter instead.`;
```
JoshuaKGoldberg: Or a local variable, right?

`Use a local parameter or variable instead.`

src/rules/noRestrictedGlobalsRule.ts
```diff
@@ -0,0 +1,97 @@
...
+            return;
+        }
+
+        const symbol = checker.getSymbolAtLocation(node);
+        const declarations = symbol === undefined ? undefined : symbol.declarations;
+        if (declarations === undefined || declarations.length === 0) {
+            return;
+        }
+
+        const declaredInLibDom = declarations.some((decl) => decl.getSourceFile().fileName.endsWith(".d.ts"));
```
JoshuaKGoldberg: What if you declare something in an `externals.d.ts` file, or define your own `lib.dom.d.ts` with either that name or `my-lib.d.ts`?

Or what about a project that uses `namespace`s and has a `const name = "HA!";` somewhere?

Maybe the rule should get the list of files TS is compiling with _(can we do that?)_ along with its compiler options, and add `lib.d.ts`-like files as per `lib` and `--noLib` compiler settings?
Maybe the rule should keep a default list of regular expressions to match names against, like `lib*.d.ts`?

test/rules/no-restricted-globals/tslint.json
```diff
@@ -0,0 +1,10 @@
+{
+  "rules": {
+    "no-restricted-globals": [
```
JoshuaKGoldberg: Some more test cases to consider...

* `./lib.d.ts`
* `./my-types.d.ts`
* ./`has-global-const-name-declarations.d.ts`
* Functions, variables, interfaces, types, and classes with suspicious names like `name`

src/rules/noRestrictedGlobalsRule.ts
```diff
@@ -0,0 +1,97 @@
...
+            DOM event as a global variable event, but using this variable has been considered
+            as a bad practice for a long time. Restricting this will make sure this variable
+            isn’t used in browser code.
+        `,
+        optionsDescription: "This rule takes a list of strings, where each string is a global to be restricted.",
+        options: {
+            type: "list",
+            items: {type: "string"},
+        },
+        optionExamples: [
```
JoshuaKGoldberg: Can you provide these as a default list? The initial set of things not to touch should be pretty standard.

src/rules/noRestrictedGlobalsRule.ts
Zzzen: IMO, custom definition files should be considered the same as official `lib.dom.d.ts`. There is no need to get compiler options.

src/rules/noRestrictedGlobalsRule.ts
Zzzen: Updated. thanks

aervin:
@Zzzen this is awesome! The ability to add code examples to the documentation just landed on master--would you mind adding some for this rule? See the `curly` rule for an example.

Zzzen:
Definition files under tests do not seem to be included in compilation.  Because compiler cannot find declaration of `myGlobal` in `test/rules/no-restricted-globals/lib-dom/test.ts.lint` though I've declared it in `./custom-global.d.ts.lint`. Same goes to `test/rules/no-restricted-globals/namespace/lib.ts.lint`. What did I do wrong?

rafeememon:
Hey all, any word on when this PR can be merged?  I've been bitten by this issue a handful of times and would love to use this rule.  Thanks!

Zzzen:
Please take a look when you have time. @JoshuaKGoldberg

src/rules/noRestrictedGlobalsRule.ts
```diff
@@ -0,0 +1,116 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import * as ts from "typescript";
+
+import * as Lint from "../index";
+
+// TODO: add a default list
```
JoshuaKGoldberg: Err, is this TODO still relevant?

Zzzen: it's outdated, since `["event", "name", "length"]` are banned by default.

jdom:
Would love to see this in too, as I'm currently re-bundling updated lib declaration files just to ban a few of these global variables (which is very cumbersome)... Just to confirm, this is blocked only on the author providing more test cases, but the behavior/naming is acceptable?

JoshuaKGoldberg:
@jdom also pending review from a member of Palantir, yes.

bradleyayers:
This would be very useful, is there anything that can be done to move it along?

Zzzen:
🤔 I guess all requested changes already have been addressed?

test/rules/no-restricted-globals/misc.ts.lint
```diff
@@ -0,0 +1,21 @@
+function typeTest() {
```
JoshuaKGoldberg: A few missing test cases:

* `let { foo = event } = bar` - should be a lint complaint
* `let { foo = (() => event)() } = bar;` - should be a lint complaint
* Function with a parameter of a banned global, and references to that parameter
* `instanceof` and `typeof` usages

src/rules/noRestrictedGlobalsRule.ts
```diff
@@ -0,0 +1,115 @@
...
+            return [];
+        } else {
+            return this.applyWithFunction(sourceFile, walk, bannedGlobals, program.getTypeChecker());
+        }
+    }
+}
+
+function walk(ctx: Lint.WalkContext<Set<string>>, checker: ts.TypeChecker): void {
+    return ts.forEachChild(ctx.sourceFile, function recur(node: ts.Node): void {
+        switch (node.kind) {
```
JoshuaKGoldberg: Switching on `node.kind` to ignore some nodes is a bit unusual. Normally you'd want to do something like:

```typescript
if (tsutils.isIdentifier(node)) {
    checkIdentifier(node);
}
```

...is there a reason you didn't set up it this way? If not, please change to it.

Not recursing on some nodes can be problematic because identifiers can show up in the darndest places. For example, this would be ksipped because you don't check object binding patterns:

```typescript
let { foo = event } = bar;
let { foo = (() => event)() } = bar;
```

src/rules/noRestrictedGlobalsRule.ts
```diff
@@ -0,0 +1,115 @@
...
+            case ts.SyntaxKind.ObjectBindingPattern:
+            case ts.SyntaxKind.TypeReference:
+                return;
+            case ts.SyntaxKind.PropertyAccessExpression:
+                // Ignore `y` in `x.y`, but recurse to `x`.
+                return recur((node as ts.PropertyAccessExpression).expression);
+            case ts.SyntaxKind.Identifier:
+                return checkIdentifier(node as ts.Identifier);
+            default:
```
JoshuaKGoldberg: What about using a banned `typeof`?

```typescript
let foo: typeof event;
```

That should also be banned, right?

test/rules/no-restricted-globals/misc.ts.lint
Zzzen: You are really experienced!

Zzzen:
Code updated. You're really helpful. @JoshuaKGoldberg

src/rules/noRestrictedGlobalsRule.ts
```diff
@@ -0,0 +1,122 @@
...
+            return;
+        }
+
+        const symbol = checker.getSymbolAtLocation(node);
+        const declarations = symbol === undefined ? undefined : symbol.declarations;
+        if (declarations === undefined || declarations.length === 0) {
+            return;
+        }
+
+        const isAmbientGlobal = declarations.some((decl) => decl.getSourceFile().isDeclarationFile);
```
sandersn: This errors too often if the declaration comes from a declaration file that is *not* lib.dom.d.ts. Consider the case of adone's types in Definitely Typed:

```ts
// @Filename: adone-tests.ts
namespace eventsTests {
    namespace EventEmitter {
        namespace static {
            const a: number = adone.event.Emitter.listenerCount(new adone.event.Emitter(), "event");
            const b: number = adone.event.Emitter.defaultMaxListeners;
        }
// @Filename: events.d.ts
declare namespace adone {
    namespace event {
        class Emitter {
            static listenerCount(emitter: Emitter, event: string | symbol): number;
            static defaultMaxListeners: number;
            // ...
```

You will get bogus errors on `event`.

PR #3826, opened 2018-04-09
[add-rule] no focused tests

HyphnKnight:
#### PR checklist

  - [x] Includes tests
- [ ] Documentation update

#### Overview of change:
Add a rule that prevents people from accidentally committing focused tests to their repository resulting in accidentally skipped tests.

I'm worried that the walk function could be more optimized if there is a faster way to do what I'm doing let me know.

### DIFF
--- a/src/rules/noFocusedTestsRule.ts
+++ b/src/rules/noFocusedTestsRule.ts
@@ -22,10 +22,10 @@ import * as ts from "typescript";
 import * as Lint from "../index";
 
 const forbiddenFocusFunctions: string[] = [
-  'fdescribe',
-  'fit',
-  'ddescribe',
-  'iit',
+  "fdescribe",
+  "fit",
+  "ddescribe",
+  "iit",
 ];
 
 export class Rule extends Lint.Rules.AbstractRule {
--- a/src/rules/noFocusedTestsRule.ts
+++ b/src/rules/noFocusedTestsRule.ts
@@ -1,0 +1,70 @@
+
+/**
+ * @license
+ * Copyright 2013 Palantir Technologies, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { isCallExpression } from "tsutils";
+import * as ts from "typescript";
+
+import * as Lint from "../index";
+
+const forbiddenFocusFunctions: string[] = [
+  'fdescribe',
+  'fit',
+  'ddescribe',
+  'iit',
+];
+
+export class Rule extends Lint.Rules.AbstractRule {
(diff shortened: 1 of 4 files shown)


src/rules/noFocusedTestsRule.ts
```diff
@@ -0,0 +1,70 @@
...
+];
+
+export class Rule extends Lint.Rules.AbstractRule {
+    /* tslint:disable:object-literal-sort-keys */
+    public static metadata: Lint.IRuleMetadata = {
+        ruleName: "no-focused-tests",
+        description: "Disallows test focusing functions like `fdescribe` or `fit`.",
+        rationale: Lint.Utils.dedent`
+            \`fdescribe\`,\`fit\`,\`ddescribe\`, and \`iit\` are functions used for debugging tests,
+            by allowing users to focus on just one test however if commited they can result in tests
```
suchanlee: grammar nit: `...just one test. However,`
spelling nit: `commited` -> `committed`

src/rules/noFocusedTestsRule.ts
```diff
@@ -0,0 +1,70 @@
...
+            that the programmer wants to run not running. This will notify them and prevent accidental
+            ignoring of test cases.`,
+        optionsDescription: "Not configurable.",
+        options: null,
+        optionExamples: [true],
+        type: "functionality",
```
suchanlee: i would say that it better fits `maintainability`

src/rules/noFocusedTestsRule.ts
```diff
@@ -0,0 +1,70 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { isCallExpression } from "tsutils";
+import * as ts from "typescript";
+
+import * as Lint from "../index";
+
+const forbiddenFocusFunctions: string[] = [
```
suchanlee: if you use a `Set<string>`, checks can be done much faster

JoshuaKGoldberg:
Pretty similar to tslint-microsoft-contrib's `mocha-avoid-only`. That rule can probably be deprecated if this one goes in.

test/rules/no-focused-tests/tests.ts.lint
```diff
@@ -0,0 +1,15 @@
...
+  });
+});
+
+fdescribe('myClass',() => {
+~~~~~~~~~                   [Forbidden focused test fdescribe]
+  fit('should do myMethod', () => {
+  ~~~                               [Forbidden focused test fit]
+    // run my test
+  });
+});
```
JoshuaKGoldberg: Please also test...

* Correct uses of `describe("...", () => {...})` and `it("...", () => {...})`
* Variables and classes named `describe` and `it`
* 
    ```
    const it = "This is silly but technically valid";
    const describe = () => "This too!";

    describe();
    ```

aervin:
Could the ban rule be used to accomplish this?

suchanlee:
Let's just use the ban rule instead (which actually has focused tests in its example): https://palantir.github.io/tslint/rules/ban/

rms095:
I was about to post the same feature. Thank god I got this. Is it released or not?

suchanlee:
@rms095 you can use the ban rule for this, which has been released for a while: https://palantir.github.io/tslint/rules/ban/

rms095:
I am sorry but **ban rule** is not working for me.

package.json - 

**tslint version** - 5.10.0
**angular-core** - 4.4.3
**typescript** - 2.3.4

"lint": "node_modules/tslint/bin/tslint --project tsconfig.json -c tslint.json 'src/app/**/*.ts'"

For testing, I have added fdescribe and fit in specs file but it is passing lint test.

PR #3834, opened 2018-04-13
Updated max-line-length ignore-pattern example

janis91:
#### PR checklist

  - [ ] Includes tests
- [x] Documentation update

#### Overview of change:
The example uses:

```
class [a-zA-Z] implements 
```

but this only matches a class that has a single character as name. Even if it is an example, the rule should accept more than one character for the class name. I struggled with this ignore pattern in a project, because I didn't read the pattern itself completely. Therefore I want to give this feedback, such that other people don't have to search for the reason of a failing rule, like me.


palantirtech:
Thanks for your interest in [palantir/tslint](https://github.com/palantir/tslint), @janis91! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

Issue #3813, opened 2018-04-05
completed-docs erroring on CircleCI:testNext


JoshuaKGoldberg:
https://circleci.com/gh/JoshuaKGoldberg/tslint/437?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

The failure isn't expected, but is now showing up.

```
    // TODO: TypeScript API doesn't give us a symbol for this, so we must ignore it.
    // https://github.com/Microsoft/TypeScript/issues/14257
    [Symbol.iterator]() {}
    ~~~~~~~~~~~~~~~~~~~~~~ [Documentation must exist for methods.]
```

johnwiseheart:
I believe @suchanlee has a fix for this.

suchanlee:
Quite a few tests broke from TS2.8 and I’ve been working through those, but starting 2.8, the `ts.Diagnostic` block seems to have changed which broke a ton of tests for `no-unused-variable`.

Not sure if the fix here is to go through the tests and fix the tests or to actually change the code to handle this so that its output remains the same as <2.8. I'll do some investigation on what has changed.

JoshuaKGoldberg:
@suchanlee would you like me to just remove that part of the `completed-docs` tests?

PR #3853, opened 2018-04-25
Fix build

suchanlee:
#### PR checklist

- [x] Addresses an existing issue: https://github.com/palantir/tslint/issues/3813
  - [x] Includes tests
- [ ] Documentation update

#### Overview of change:
* Fixes tests that broke due to compiler changes in TS 2.8
* Quite a few changes in the logic for `no-unused-variables` rule due to changes (kind, count, text range) in what the compiler outputs for `preEmitDiagnostics`

### DIFF
--- a/src/rules/noUnusedVariableRule.ts
+++ b/src/rules/noUnusedVariableRule.ts
@@ -126,17 +126,19 @@ function walk(ctx: Lint.WalkContext<Options>, program: ts.Program): void {
             }
         }
 
-        if (kind === UnusedKind.VARIABLE_OR_PARAMETER) {
-            const importName = findImport(diag.start, sourceFile);
-            if (importName !== undefined) {
-                if (declaration && isImportUsed(importName, sourceFile, checker)) {
-                    continue;
-                }
+        if (kind === UnusedKind.VARIABLE_OR_PARAMETER || kind === UnusedKind.DECLARATION) {
+            const importNames = findImports(diag.start, sourceFile, kind);
+            if (importNames.length > 0) {
+                for (const importName of importNames) {
+                    if (declaration && isImportUsed(importName, sourceFile, checker)) {
+                        continue;
+                    }
 
-                if (importSpecifierFailures.has(importName)) {
-                    throw new Error("Should not get 2 errors for the same import.");
+                    if (importSpecifierFailures.has(importName)) {
+                        throw new Error("Should not get 2 errors for the same import.");
+                    }
+                    importSpecifierFailures.set(importName, failure);
                 }
-                importSpecifierFailures.set(importName, failure);
                 continue;
             }
         }
@@ -337,12 +339,14 @@ function forEachImport<T>(sourceFile: ts.SourceFile, f: (i: ImportLike) => T | u
     });
 }
 
-function findImport(pos: number, sourceFile: ts.SourceFile): ts.Identifier | undefined {
+function findImports(pos: number, sourceFile: ts.SourceFile, kind: UnusedKind): ReadonlyArray<ts.Identifier> {
     return forEachImport(sourceFile, (i) => {
+        if (!isInRange(i, pos)) {
+            return undefined;
+        }
+
         if (i.kind === ts.SyntaxKind.ImportEqualsDeclaration) {
-            if (i.name.getStart() === pos) {
-                return i.name;
-            }
+            return [i.name];
         } else {
             if (i.importClause === undefined) {
                 // Error node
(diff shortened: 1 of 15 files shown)


src/rules/noUnusedVariableRule.ts
```diff
@@ -337,12 +339,14 @@ function forEachImport<T>(sourceFile: ts.SourceFile, f: (i: ImportLike) => T | u
     });
 }
 
-function findImport(pos: number, sourceFile: ts.SourceFile): ts.Identifier | undefined {
+function findImports(pos: number, sourceFile: ts.SourceFile, kind: UnusedKind): ReadonlyArray<ts.Identifier> {
```
suchanlee: changing to an array since in the case of kind === DECLARATION we want to return all the imports in the node

src/rules/noUnusedVariableRule.ts
```diff
@@ -126,17 +126,19 @@ function walk(ctx: Lint.WalkContext<Options>, program: ts.Program): void {
...
 
-        if (kind === UnusedKind.VARIABLE_OR_PARAMETER) {
-            const importName = findImport(diag.start, sourceFile);
-            if (importName !== undefined) {
-                if (declaration && isImportUsed(importName, sourceFile, checker)) {
-                    continue;
-                }
+        if (kind === UnusedKind.VARIABLE_OR_PARAMETER || kind === UnusedKind.DECLARATION) {
+            const importNames = findImports(diag.start, sourceFile, kind);
+            if (importNames.length > 0) {
```
suchanlee: logic changes to handle the array return type

test/rules/no-unused-variable/check-parameters/test.ts.lint
```diff
@@ -66,7 +66,11 @@ export class DestructuringTests {
 }
 
 abstract class AbstractTest {
+#if typescript >= 2.8.0
+~~~~~~~~~~~~~~~~~~~~~~~~~~~ [err % ('AbstractTest')]
```
suchanlee: changes in compiler generated ranges

test/rules/no-unused-variable/default/false-positives.ts.lint
```diff
@@ -2,7 +2,11 @@
 const fs = require("fs");
 
 module Foo {
-       ~~~   [err % ('Foo')]
+#if typescript >= 2.8.0
```
suchanlee: changes in compiler generated ranges

test/rules/no-unused-variable/default/function.ts.lint
```diff
@@ -3,12 +3,16 @@ function func1(x: number, y: number) {
...
 
 var func2 = () => {
-    ~~~~~           [err % ('func2')]
+    ~~~~~ [err % ('func2')]
     //
 };
 
 function func3() {
-         ~~~~~     [err % ('func3')]
+#if typescript >= 2.8.0
```
suchanlee: changes in compiler generated ranges

test/rules/no-unused-variable/default/import.ts.lint
```diff
@@ -42,7 +42,11 @@ $(_.xyz());
 /// <reference path="../externalFormatter.test.ts" />
 
 module S {
+#if typescript >= 2.8.0
```
suchanlee: changes in compiler generated ranges

suchanlee:
Added builds for TS 2.7 and 2.8

suchanlee:
@JoshuaKGoldberg can you take a look at this?

src/rules/noUnusedVariableRule.ts
```diff
@@ -351,37 +355,71 @@ function findImport(pos: number, sourceFile: ts.SourceFile): ts.Identifier | und
...
                     }
                 }
             }
         }
         return undefined;
     });
+    return imports !== undefined ? imports : [];
+}
+
+function isInRange(range: ts.TextRange, pos: number): boolean {
```
JoshuaKGoldberg: Nit: might belong in some utils file somewhere? `isInTextRange`?

test/rules/completed-docs/defaults/test.ts.lint
```diff
@@ -12,9 +12,12 @@ export class Aaa {
     public set prop(value) { this.bbb = value; }
     ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~       [Documentation must exist for public properties.]
 
-    // TODO: TypeScript API doesn't give us a symbol for this, so we must ignore it.
+    // TypeScript API doesn't give a symbol for this before version 2.8.0
     // https://github.com/Microsoft/TypeScript/issues/14257
     [Symbol.iterator]() {}
+#if typescript >= 2.8.0
```
JoshuaKGoldberg: TIL this is possible. Is this documented anywhere? If not, should we file a followup issue for another PR?

test/rules/no-unused-variable/check-parameters/test.ts.lint
```diff
@@ -65,8 +65,15 @@ export class DestructuringTests {
     }
 }
 
+// Could be a bug with 2.9-dev but getPreEmitDiagnostics isn't
```
JoshuaKGoldberg: Can you link to the corresponding TypeScript issue?

JoshuaKGoldberg:
FYI @amcasey (I think you've been working in this space recently?)

test/rules/no-unused-variable/check-parameters/test.ts.lint
JKillian: 👍 or if there isn't one, file a ticket with them and link it to this PR!

test/rules/completed-docs/defaults/test.ts.lint
suchanlee: not sure if it has been documented anywhere, I found it while reading the tests. I think documenting it would be useful. Let's file a follow up issue

src/rules/noUnusedVariableRule.ts
suchanlee: hmm not sure where it would belong though

suchanlee:
So it wasn't an issue with TS, but the diagnoses were being ignored because the diagnosis code had changed in 2.9. Made changes to the code to handle those and fixed the tests accordingly.

But this solidifies the concern that I had, which is that this rule is complex and hard to maintain with a living and changing compiler. **I feel that we should deprecate this rule in the near future** for the following reasons:
* It is expensive to maintain
* The rule is complex
* It has many issues: https://github.com/palantir/tslint/issues?utf8=%E2%9C%93&q=is%3Aissue+no-unused-variables
* Most importantly, TS supports it very well. What this rule actually does is take what the TS compiler outputs and process them and to construct TSLint failures and fixes which it's surprisingly complex and slow. And I believe until 2.7 the failures that TS/TSLint output were similar, **but starting from 2.7, it has deviated quite a bit, both in the error messages and the erroring text range**.

Deprecating the rule was discussed: https://github.com/palantir/tslint/issues/1481 but the community decided to not deprecate it for two main reasons:
(1) TSLint errors don't affect compilation while TS errors do, and sometimes you just want to fast dev ignoring warnings
(2) TSLint autofixes

(1) is in discussion in https://github.com/Microsoft/TypeScript/issues/13408 and is already supported via https://github.com/Microsoft/vscode/pull/37616
(2) I don't think is supported yet.

suchanlee:
@JKillian @johnwiseheart @JoshuaKGoldberg ^^^ not that we should take any action now but something to keep track of in the future

Issue #3761, opened 2018-03-07
Error in function argument renamed destructuring with type


ghost:
### Bug Report

- __TSLint version__:  5.9.1
- __TypeScript version__: 2.7.2
- __Running TSLint via__: CLI

#### TypeScript code being linted

```ts
interface SessionMeta {
  // ...
}

this.sessionService
  .getSessions()
  .subscribe( ( { meta: newMeta }:  { meta: SessionMeta } ) => {
    //            |
    //      error points to here

    this.setCount(newMeta.match_count);
  });
```

with `tslint.json` configuration:

```json
{
  "no-use-before-declare": true,
}
```

#### Actual behavior
`ERROR: ...: variable 'meta' used before declaration`

#### Expected behavior
I think, the syntax is valid, the compiled code is good, the expected. You can see a similar example here:
https://github.com/Microsoft/TypeScript-Handbook/blob/master/pages/Variable%20Declarations.md#property-renaming

adrienboulle:
I also have this problem, syntax is fine, I think this is a false positive

JoshuaKGoldberg:
Looks like this was fixed by #3849. 🎉

Issue #3849, opened 2018-04-23
Add more function features


ghost:
The following features would be appreciated:

`max-function-line-count`:  A line limit for functions.

Add some parameters to `max-file-line-count` so that the line count can ignore lines containing comments, or empty lines, for example.

`single-return`:  Functions can only contain one return statement if this is set to true (some argue that long functions with multiple return statements can be hard to follow).

johnwiseheart:
Hi @seanhly, can you please make a separate issue for each rule that you would like?

thorn0:
There already exists https://github.com/palantir/tslint/issues/3549 about `max-function-line-count` / `max-func-body-length`.

JoshuaKGoldberg:
> `single-return`

-1 to a built-in preference for a single return. I find it easier to read code that has minimal nesting of conditionals.

```typescript
function sample(input) {
    if (isInvalid(input)) {
        return invalidValue;
    }

    const first = runFirst(input);
    const last = runLast(input, first);

    return processLast(last);
}
```

seanhly:
I agree, your example looks like fine code.

I think when I was writing this I was in the process of refactoring a method with dozens of exit points (not safe guards, just arbitrary return statements), and when I wanted to do something simple like "print this when the method returns", I had to edit a lot more than one line.  It seems clearer to me now the problem was with long functions rather than multiple return statements.

If a "single-return" rule or "max-return-statements" rule is added, some code perfectionist with quirks out there is bound to use it, which I find a little amusing.  TSLint in a team project is like the code implementation of passive-aggressive behavior, and always leads to interesting debates when someone notices their favorite funky way to do something doesn't even make it past the linter to compile.  But I digress, "single-return" may have no real use.

JoshuaKGoldberg:
Note: per #4534, this issue will be closed in less than a month if no PR is sent to add the rule. If you _really_ need the rule, custom rules are always an option and can be maintained outside this repo!

JoshuaKGoldberg:
### 💀 _**It's time!**_ 💀

**TSLint is being deprecated** and no longer accepting pull requests for new rules. See #4534. 😱

If you'd like to see this rule implemented, you have two choices:
* **Recommended:** Use [typescript-eslint](https://typescript-eslint.io) to lint your TypeScript code with ESLint ✅
* **Legacy:** Write it as a [custom TSLint rule](https://palantir.github.io/tslint/develop/custom-rules/) 🤷‍♂️

[...]

PR #3876, opened 2018-05-04
bugfix - no-use-before-declare - exception for destructuring w/ rename

aervin:
#### PR checklist

- [X] Addresses an existing issue: #3761 
  - [X] Includes tests
- [x] Documentation update

#### Overview of change:
The rule checks that identifiers have a declaration with a `pos` <= to the reference `pos`, which isn't the case when destructuring w/ a rename. 

```ts
var { x: y } = { x: 43 };
```

Any edge cases with the destructuring syntax that would require additional tests?

#### CHANGELOG.md entry:
[bugfix] [`no-use-before-declare`](https://palantir.github.io/tslint/rules/no-use-before-declare/) Fixes false positives when using the destructuring syntax (#3761)

### DIFF
--- a/test/rules/no-use-before-declare/test.ts.lint
+++ b/test/rules/no-use-before-declare/test.ts.lint
@@ -76,6 +76,8 @@ export {
 var undeclaredA = 42;
 var { undeclaredB } = { undeclaredB: 43 };
 var { undeclaredB: undeclaredA } = { undeclaredB: 43 };
+var { undeclaredB: undeclaredA, undeclaredC } = { undeclaredB: 43, undeclaredC: 'hello' };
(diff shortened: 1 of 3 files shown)


test/rules/no-use-before-declare/test.ts.lint
```diff
@@ -68,7 +74,8 @@ export {
 };
 
 var undeclaredA = 42;
-let { undeclaredB } = { undeclaredB: 43 };
+var { undeclaredB } = { undeclaredB: 43 };
+var { undeclaredB: undeclaredA } = { undeclaredB: 43 };
 const [ undeclaredC, [undeclaredD] ] = [ 1, [2] ];
```
suchanlee: Can we add test cases for multiple reassignments and reassignment and non-reassignment in the same destructuring?

JoshuaKGoldberg:
Looks like this was fixed by #3849. 🎉

PR #3624, opened 2018-01-06
Add the option allow-empty-functions in no-empty

markvincze:
#### PR checklist

- [x] Addresses an existing issue: #2107
- [x] New feature, bugfix, or enhancement
  - [x] Includes tests
- [x] Documentation update

#### Overview of change:

Introduces option to allow function declarations to be empty while using `no-empty`.

#### Is there anything you'd like reviewers to focus on?

This is my first PR to tslint, so any feedback is welcome.  
I was a bit unsure if I did the correct setup for the `options` property (an `array` with an `anyOf` of the two strings).

#### CHANGELOG.md entry:

[enhancement] Introduces option to allow function declarations to be empty while using `no-empty`.

PR #3897, opened 2018-05-09
no-empty: fix lack of methods handling

keradus:
#### PR checklist

- [x] Addresses an existing issue: #3624 => that is the feature PR that incorporate the bug
  - [x] Includes tests
- [ ] Documentation update

#### Overview of change:
I do like #3624 and I wanted to use it.
It already has support for classes, as it has special test case with constructor, which is handled differently.
Yet, class might need to have an empty method - and even if one is using `allow-empty-function`, then empty method is considered as violation.
When empty method may be valid? Class implements interface, want to provide empty implementation for abstract method from parent class or one is using ducktyping.

#### CHANGELOG.md entry:

`[bugfix] allow-empty-functions option of no-empty rule is now properly respecting empty methods`

### DIFF
--- a/src/rules/noEmptyRule.ts
+++ b/src/rules/noEmptyRule.ts
@@ -97,7 +97,8 @@ function isExcluded(node: ts.Node, options: Options): boolean {
     }
 
     if (options.allowEmptyFunctions &&
-        (node.kind === ts.SyntaxKind.FunctionDeclaration ||
+        (node.kind === ts.SyntaxKind.MethodDeclaration ||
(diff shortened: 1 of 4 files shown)


keradus:
Thank you :heart:  Really nice to hear! (especially when I do equivalent soft for PHP)

Also, big :+1: for quick reaction, it's appreciated :+1:
Now, awaiting for bugfix release ;)

keradus:
hi there @giladgray ! any chances for have this PR released ?

keradus:
ping @giladgray, any chances?

dh-tdoumtsis:
any news on that?

PR #3869, opened 2018-05-03
[docs] additional code examples for 12 rules

Shinigami92:
#### PR checklist

- [x] Addresses an existing issue: #3601
- [x] New feature, bugfix, or **enhancement**
  - [ ] Includes tests
- [x] Documentation update

#### Overview of change:

Added code examples to rules:

- noAny
- noEmptyInterface
- noSparseArrays
- noUnnecessaryCallbackWrapper
- objectLiteralSortKeys
- oneVariablePerDeclaration
- onlyArrowFunctions
- preferTemplate
- radix
- switchDefault
- typedef
- useIsnan

#### CHANGELOG.md entry:

See: "Overview of change"

PR #3925, opened 2018-05-23
Add "force-singleline" options to jsdoc-format rule

ianwremmel:
#### PR checklist

  - [x] Includes tests
- [ ] Documentation update

#### Overview of change:

This option requires (and fixes) JSDoc blocks like

```
/**
 * A comment that fits on one line
 */
```

to be formatted like

```
/** A comment that fits on one line */
```

1. I'm not thrilled by the option name, so I'm open to suggestions.
2. Is more documentation needed?
3. This relies exclusively on `\n` and not on character length. Is that appropriate?
4. Are there test scenarios I didn't consider?

#### CHANGELOG.md entry:

[new-rule-option] `jsdoc-format` adds `force-singleline` with fixer to make single-line doc blocks more compact


palantirtech:
Thanks for your interest in [palantir/tslint](https://github.com/palantir/tslint), @ianwremmel! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

ianwremmel:
@palantirtech, I signed the CLA, but the PR didn't update. Perhaps because I use a GitHub privacy email address in my commits?

suchanlee:
1. I'm not thrilled by the option name, so I'm open to suggestions.
-> I think this is fine, can't think of anything much better. "prefer-singleline"?

2. Is more documentation needed?
-> You can now add code examples. I think adding that would be very helpful. See this PR for examples: https://github.com/palantir/tslint/pull/3869

3. This relies exclusively on \n and not on character length. Is that appropriate?
-> I think that adding a configurable and optional character length would be helpful, and use only `\n` if unspecified?

4. Are there test scenarios I didn't consider?
-> i would add test cases to make sure there are no false positives

aervin:
@suchanlee Recent versions of tsutils have a method for determining the newline character of a sourcefile (`\n` vs `\r\n`). Not sure if that's necessary here but perhaps something to consider. I think it's called `getLineBreakStyle`.

src/rules/jsdocFormatRule.ts
```diff
@@ -21,6 +21,7 @@ import * as ts from "typescript";
 import * as Lint from "../index";
 
 const OPTION_CHECK_MULTILINE_START = "check-multiline-start";
+const OPTION_FORCE_SINGLELINE = "force-singleline";
```
giladgray: `prefer-singleline` makes sense, as there's no forcing

src/rules/jsdocFormatRule.ts
```diff
@@ -40,34 +41,45 @@ export class Rule extends Lint.Rules.AbstractRule {
         optionsDescription: Lint.Utils.dedent`
             You can optionally specify the option \`"${OPTION_CHECK_MULTILINE_START}"\` to enforce the first line of a
             multiline JSDoc comment to be empty.
+
+            You can optionally specify the option \`"${OPTION_FORCE_SINGLELINE}"\` to require JSDoc comments that can
+            fit on a single lin to do so.
```
giladgray: single lin[e]

src/rules/jsdocFormatRule.ts
```diff
@@ -40,34 +41,45 @@ export class Rule extends Lint.Rules.AbstractRule {
...
+            [true, OPTION_CHECK_MULTILINE_START, OPTION_FORCE_SINGLELINE],
+        ],
         type: "style",
         typescriptOnly: false,
     };
     /* tslint:enable:object-literal-sort-keys */
 
     public static ALIGNMENT_FAILURE_STRING = "asterisks in jsdoc must be aligned";
     public static FORMAT_FAILURE_STRING = "jsdoc is not formatted correctly on this line";
+    public static FORCE_SINGLE_LINE_FALURE = "singleline jsdoc should use compact docblock form";
```
giladgray: "compact docblock" ??

how about "format jsdoc comment on a single line"

src/rules/jsdocFormatRule.ts
```diff
@@ -85,6 +97,22 @@ function walk(ctx: Lint.WalkContext<Options>) {
             return;
         }
 
+        // This is a single-line comment using the multiline format
+        if (ctx.options.forceSingleline && lines.length === 3) {
+            // Add 3 to account for the leading /**
+            const endOfFirstLine = pos + firstLine.length + 3;
+            const endOfSecondLine = endOfFirstLine + lines[1].length + 1;
+            const endOfThirdLine = endOfSecondLine + lines[2].length + 1;
```
giladgray: this logic feels brittle, but i don't have alternatives to offer. would it handle trailing whitespace?

ianwremmel: Yea, it's quite brittle. I misunderstood exactly what `pos` and `fullText` referred to when I wrote this. (I'd been under the impression that `fullText` was the lines containing the comment block and `pos` was, effectively, the indent.) 

I actually ended up rewriting this addition in isolation from the rest of `jsdoc-format` so we could use it here until this PR is ready to go. In so doing, I gave up on a per-line marker and just mark the whole comment block which seems a lot more stable.

ianwremmel:
Thanks all for the feedback. I ended up implementing this change as a separate rule in our code base until this PR is ready. I found a bunch of places to improve things and learned I'd made some incorrect assumptions. I'll have updates in a few days that integrate all of your comments with what I've got in our code base.

For our purposes, we want to limit single line/compact doc blocks to fit with 80 characters, so I've written a lint/fix that checks that as well. I'll make the 80 configurable, of course.

One thing of note: I found debugging quite difficult: console output seems to be suppressed and there seem to be too many binaries in play to hook up a debugger. Does anyone have any recommendations (particularly in how to get console output)?

suchanlee:
@ianwremmel debugging is quite easy if you use VSCode, there are debug configurations in the repo that you can run and debug from the code editor.

screen shot 2018-06-07 at 4 05 03 pm

ianwremmel:
gotcha, thanks @suchanlee! I've gotten in the habit of not bothering with the vscode debugger for compiled variants of JS, so I didn't even think to look if y'all had configured anything.

src/rules/jsdocFormatRule.ts
giladgray: @ianwremmel would you kindly update this PR with your new implementation? be the change you want to see in the repo!

ianwremmel: haha, yea, hopefully today or tomorrow; work priorities shifted a bit and it's been hard to find the time.

src/rules/jsdocFormatRule.ts
```diff
@@ -40,44 +42,172 @@ export class Rule extends Lint.Rules.AbstractRule {
...
+    preferSingleline: boolean;
+    maxWidth: number;
+}
+
+function shouldBeSingleline(ctx: Lint.WalkContext<Options>, lines: string[], {pos, end}: ts.CommentRange): boolean {
+    if (!ctx.options.preferSingleline) {
+        return false;
+    }
+
+    if (lines.length === 3) {
```
suchanlee: is the max width not taken into account here?

src/rules/jsdocFormatRule.ts
```diff
@@ -40,44 +42,172 @@ export class Rule extends Lint.Rules.AbstractRule {
...
+        ctx.addFailure(pos, end, Rule.PREFER_SINGLE_LINE_FAILURE, [
+            Lint.Replacement.replaceFromTo(pos, end, `/** ${lines[1].trim().substr(1).trim()} */`),
+        ]);
+    }
+
+    return false;
+}
+
+function shouldBeMultiline(ctx: Lint.WalkContext<Options>, lines: string[], fullText: string, { pos, end }: ts.CommentRange): boolean {
```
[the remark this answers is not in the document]
ianwremmel: yea, i was struggling with that a bit. on the one hand, it's combining what should be an effect with a pure function, but on the other hand, separating them would require duplicating most of the expensive bits of code and executing them twice.

src/rules/jsdocFormatRule.ts
ianwremmel: As far as I can tell, it doesn't need to be. This function will fix it to a single-line docblock and then `shouldBeMultiline` will fix it back to multiline and break it up appropriately if fitting it on a single line is inappropriate.

src/rules/jsdocFormatRule.ts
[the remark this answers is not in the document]
ianwremmel: Actually, I take that back; when I started, there would have been a lot of duplicated expensive code, but in their current form, they can probably be split into test and effect pretty easily.

src/rules/jsdocFormatRule.ts
```diff
@@ -40,44 +42,172 @@ export class Rule extends Lint.Rules.AbstractRule {
...
 
 function walk(ctx: Lint.WalkContext<Options>) {
-    return utils.forEachComment(ctx.sourceFile, (fullText, {kind, pos, end}) => {
+    return utils.forEachComment(ctx.sourceFile, (fullText, range) => {
+        const { kind, pos, end } = range;
         if (kind !== ts.SyntaxKind.MultiLineCommentTrivia ||
             fullText[pos + 2] !== "*" || fullText[pos + 3] === "*" || fullText[pos + 3] === "/") {
-            return;
-        }
+                return;
```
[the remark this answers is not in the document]
ianwremmel: Looking at the full code, I think it's fine. Or, at least, as fine as prettier's opinions can make it :)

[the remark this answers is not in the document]
ianwremmel: hmm, prettier may not be in play. Which of these is appropriate?

```ts
        if (kind !== ts.SyntaxKind.MultiLineCommentTrivia ||
            fullText[pos + 2] !== "*" || fullText[pos + 3] === "*" || fullText[pos + 3] === "/") {
            return;
        }
```

```ts
        if (kind !== ts.SyntaxKind.MultiLineCommentTrivia ||
            fullText[pos + 2] !== "*" || fullText[pos + 3] === "*" || fullText[pos + 3] === "/") {
                return;
            }
```

ianwremmel:
It seems the version of tsutils pulled in by Circle CI doesn't have `getLineBreakStyle`. Should I update the tsutils version in package.json and if so, should I go to latest or to the old that includes `getLineBreakStyle`?

JoshuaKGoldberg:
> tsutils pulled in by Circle CI

TSLint specifies `"tsutils": "^2.27.2"` in `package.json`. https://github.com/ajafff/tsutils/blob/master/CHANGELOG.md#2140 says `getLinebreakStyle` was added in 2.14 now. Should be resolved

> Prettier

All files in `master` are updated to have Prettier formatting now. That shouldn't be an issue after one more merge/rebase from `master`.

JoshuaKGoldberg:
There are some structural changes necessary to have this accepted in. Closing for housekeeping for now. I did a quick merge from master; if anybody would like to update this PR (@ianwremmel - comment here if you'd like it reopened) or send a new one, that'd be great!

PR #3935, opened 2018-06-01
Implemented fixer for member-ordering and added corresponding tests.

### LINKED ISSUE #2023: Feature Request: correct member-ordering issues with the --fix flag
It would be fantastic if tslint could correct member-ordering issues with the `--fix` flag. For example, the following warnings are not corrected currently:

```bash
$ tslint file.d.ts --fix

file.d.ts[1067, 5]: Declaration of public static field not allowed after declaration of constructor. Instead, this should come at the beginning of the class/interface.
file.d.ts[1298, 5]: Declaration of public static field not allowed after declaration of public instance method. Instead, this should come at the beginning of the class/interface.
file.d.ts[1783, 5]: Declaration of public static method not allowed after declaration of public instance method. Instead, this should come after constructors.
```

NaridaL:
#### PR checklist

- [X] Addresses an existing issue: fixes #2023
- [X] Includes tests
- [ ] Documentation update

#### Overview of change:
New fixer. See checkMembers JSDoc for implementation details.

test/rules/member-ordering/alphabetize/ is failing, but the diff isn't showing anything. The `.fix` doesn't exactly mirror the error message in some cases, as many configurations have multiple correct orders; that shouldn't be an issue.

I'm currently using node.getFullStart() and node getEnd() to find the node source, however this doesn't match the typescript trivia ownership rules (see failing test test/rules/member-ordering/fix-trivia). I've seen ts.getLeadingCommentRanges, but that starts at the beginning of the comment and whitespace before the leading comments would get lost. It seems like the best way would be to split at the first line break after a node, but I haven't seen any corresponding utility functions. Anyone have any advice on the matter or a reference to another rule which does something similar?


palantirtech:
Thanks for your interest in [palantir/tslint](https://github.com/palantir/tslint), @NaridaL! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

NaridaL:
image

NaridaL:
Anyone available to review this PR?

listepo-alterpost:
@ajafff ?

src/rules/memberOrderingRule.ts
```diff
@@ -272,6 +294,52 @@ class MemberOrderingWalker extends Lint.AbstractWalker<Options> {
                 prevRank = rank;
             }
         }
+        if (failureExists) {
+            // const info = (x: Node)=>[x.name && nameString(x.name), this.memberRank(x), members.indexOf(x)]
+            const sortedMembers = members.slice().sort((a, b) => {
+                // first, sort by member rank
+                // const ai = info(a);
+                // const bi = info(b);
```
giladgray: no commented code

src/rules/memberOrderingRule.ts
```diff
@@ -487,3 +555,34 @@ function nameString(name: ts.PropertyName): string {
...
+ * Returns the last element of an array. (Or undefined).
+ */
+function arrayLast<T>(array: ArrayLike<T>): T {
+    return array[array.length - 1];
+}
+
+/**
+ * Array.prototype.findIndex, but the last index.
+ */
+function arrayFindLastIndex<T>(
```
giladgray: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/lastIndexOf ?

src/rules/memberOrderingRule.ts
```diff
@@ -487,3 +555,34 @@ function nameString(name: ts.PropertyName): string {
...
+    for (let i = array.length; i-- > 0;) {
+        if (predicate(array[i], i, array)) {
+            return i;
+        }
+    }
+    return -1;
+}
+
+/**
+ * Applies a Replacement to a part of the text which starts at offset.
```
giladgray: this comment does not match the impl. it returns a string??

src/rules/memberOrderingRule.ts
```diff
@@ -250,7 +270,8 @@ class MemberOrderingWalker extends Lint.AbstractWalker<Options> {
                     : "at the beginning of the class/interface";
                 const errorLine1 = `Declaration of ${nodeType} not allowed after declaration of ${prevNodeType}. ` +
                     `Instead, this should come ${locationHint}.`;
-                this.addFailureAtNode(member, errorLine1);
+                this.addFailureAtNode(member, errorLine1, []);
```
giladgray: why the addition of empty array? seems unnecessary

NaridaL: The empty array is needed so the Replacement can later be pushed (L241). The field itself is readonly. I copied this tidbit from the ordered-imports rule. https://github.com/palantir/tslint/blob/master/src/rules/orderedImportsRule.ts#L265

src/rules/memberOrderingRule.ts
NaridaL: Yeah, that doesn't support a custom predicate.

src/rules/memberOrderingRule.ts
NaridaL: It's just a slightly modified version of Replacement.apply https://github.com/palantir/tslint/blob/master/src/language/rule/rule.ts#L204, which also returns a string.

NaridaL:
@giladgray Thanks for the review, do you have anything to add re: the two points I mentioned in the first post?

NaridaL:
test/rules/member-ordering/alphabetize/ had a switched empty line and non-empty line, which wasn't showing up in the diff. 

I went ahead and implemented a function to figure out the next newline after a node.

This should be mergeable now.

giladgray:
@NaridaL merge latest master to fix failing test.

src/rules/memberOrderingRule.ts
giladgray: a code comment would clarify this for the future.

src/rules/memberOrderingRule.ts
```diff
@@ -272,6 +294,54 @@ class MemberOrderingWalker extends Lint.AbstractWalker<Options> {
                 prevRank = rank;
             }
         }
+        if (failureExists) {
+            // const info = (x: Node)=>[x.name && nameString(x.name), this.memberRank(x), members.indexOf(x)]
```
giladgray: remove commented code

src/rules/memberOrderingRule.ts
```diff
@@ -272,6 +294,54 @@ class MemberOrderingWalker extends Lint.AbstractWalker<Options> {
...
+            const sortedMembersText = sortedMemberIndexes.map((i) => {
+                const start = splits[i];
+                const end = splits[i + 1];
+                let nodeText = this.sourceFile.text.substring(start, end);
+                while (true) {
+                    // check if there are previous fixes which we need to merge into this one
+                    // if yes, remove it from the list so that we do not return overlapping Replacements
+                    const fixIndex = arrayFindLastIndex(
+                        this.fixes,
+                        ([, r]) => r.start >= start && r.start + r.length <= end,
```
giladgray: 🙊 didn't know you could destructure an array and skip indices!!

NaridaL:
I rebased and fixed the comments.

NaridaL:
@giladgray Are you sure this fixeds #3965 ?

giladgray:
@NaridaL you wrote this... are you sure that it fixes it?

NaridaL:
This doesn't actually do anything for #3965 . https://github.com/palantir/tslint/pull/3984 does, however.

Justinidlerz:
Wow, I'm look forward to the `member-ordering` feature long time, Could publish an new npm beta version? I want to using it now.

roman-petrov:
@Justinidlerz  same for our team. Member ordering is hard to fix manually in a big project. @NaridaL, thank you for great feature! looking forward to see next release soon

PR #3984, opened 2018-06-21
[bugfix] member-ordering includes getters and setters

### LINKED ISSUE #3965: member-ordering rule, alphabetize not working on getters and setters
### Bug Report

member-ordering rule seems not to work properly on getters and setters.

- __TSLint version__: 5.10.0 
- __TypeScript version__: 2.9.2
- __Node version__: v8.11.1
- __Running TSLint via__:  CLI / VSCode 

```ts
class Test {
  public static a;
  public static b;

  public method1(): any {
    //
  }

  public method2(): any {
    //
  }

  private get k() {  // no error here
    return 2;
  }

  public get e() { // no error here
      return 2;
  }

  public get c() { // no error here
      return 2;
  }

  public set f(value: any) {
    //
  }
}
```

```json
{
  "defaultSeverity": "error",
  "extends": [
    "tslint:recommended"
  ],
  "rules": {
    "member-ordering": [true, {
      "alphabetize": true,
      "order": "statics-first"
    }]
  }
}
```

#### Actual behavior
No error 
```
$ node_modules/.bin/tslint Test.ts
$ echo $?
0
```

#### Expected behavior
tslint should report errors like:
- 'c' should come alphabetically before 'e'
- Declaration of public instance method not allowed after declaration of private instance method.

Maybe this issue is somehow related to:
- member-access rule not working on getters and setters #750
- Fix #612 #801

NaridaL:
#### PR checklist

- [X] Addresses an existing issue: fixes #3965
- [X] Includes tests
- [ ] Documentation update

#### Overview of change:
Added categories "-accessor" for getters and setters. Added those categories to the presets. Based on https://github.com/palantir/tslint/pull/3935, merge that first.

#### CHANGELOG.md entry: 
[bugfix] `member-ordering` includes property accessors (getters and setters).


giladgray:
@NaridaL I just merged #3935 so go ahead and update this 👍

NaridaL:
Done. Let's see if the lint still fails...

NaridaL:
OK, yeah, for some reason running `npm i` locally resulted in me getting the error too.

```
ERROR: 193:5  member-ordering  Declaration of public instance method not allowed after declaration of private instance method. Instead, this should come after private instance fields.
```

The error doesn't make much sense, because tslint's tslint.json is using the old-style option "variables-before-functions", and accessors are neither. As this is going to be a breaking change, as I added the accessors to the presets, and if you didn't happen to put them in the same location you're going to get errors now. Therefore I suggest doing what the comment on line 534 suggests and remove the old-style options entirely and update tslint.json.

```
/**
 * Convert from undocumented old-style options.
 * This is designed to mimic the old behavior and should be removed eventually.
 */
```

NaridaL:
OK, I changed tslint's own config to use the current config type.

Can I remove the old-style option parsing entirely or do we need a version which outputs warnings first?

src/language/rule/rule.ts
```diff
@@ -165,6 +165,8 @@ export interface ReplacementJson {
     innerText: string;
 }
 export class Replacement {
+    constructor(readonly start: number, readonly length: number, readonly text: string) {}
```
giladgray: why are these constructors moved? i did not expect them to come first.

test/rules/member-ordering/alphabetize/test.ts.fix
```diff
@@ -3,6 +3,16 @@ class X {
     bar: number;
     foo: string;
 
+    get y() {
+        return 0;
+    }
+    set y(x: number) {}
+
+    set z(x: number) {}
+    get z() {
```
giladgray: why does it go `get set set get`? i would expect `get set get set`.

src/rules/memberOrderingRule.ts
```diff
@@ -27,34 +27,58 @@ const OPTION_ALPHABETIZE = "alphabetize";
...
 enum MemberKind {
     publicStaticField,
-    publicStaticMethod,
     protectedStaticField,
-    protectedStaticMethod,
     privateStaticField,
+
+    protectedStaticMethod,
     privateStaticMethod,
+    publicStaticMethod,
```
giladgray: wrong order, should be `public,protected,private` for each type

tslint.json
```diff
@@ -49,11 +49,7 @@
     ],
     "max-classes-per-file": false,
     "max-line-length": [true, 140],
-    "member-ordering": [true,
-      "public-before-private",
-      "static-before-instance",
-      "variables-before-functions"
-    ],
+    "member-ordering": [true, { "order": "fields-first" }],
```
giladgray: is the old options syntax still supported?

giladgray: oh right, you asked. we cannot ship a breaking change. please preserve the old syntax, or we'll have to talk about 6.0

src/language/rule/rule.ts
NaridaL: >tslint.json: use preset "fields-first" for member-ordering rule instead of previous old-style config.
>This seems to match previous config most closely, requiring only 5 changes.

I can recheck if one of the other presets matches more closely, otherwise, this seems the way to go...

test/rules/member-ordering/alphabetize/test.ts.fix
NaridaL: This was discussed in the underlying issue https://github.com/palantir/tslint/issues/3965

Currently the sort works by first sorting by category, then optionally sorting alphabetically. To get 

    get x
    set x
    get y
    set y
    // other members

would require sorting first by category, then by alphabet, and a third "pass" sorting get < set. I've left this out as it would be added complexity (maybe some people want set < get) for what would only be very local errors. 

Adding this it isn't out of the question if you feel it's important.

NaridaL:
@giladgray I've added new categories for accessors. I've added these to the presets. (Currently getters and setters are ignored entirely). Therefore, if a user is using one of the presets, this is going to be a breaking change one way or the other. We might as well add other breaking changes (such as removing old style options).

giladgray:
@NaridaL there's a big difference between a change in the lint result and an **API change** in the configuration. changing the configuration is a new API contract and _requires_ user intervention during the upgrade in a way that is not automate-able.

conversely, changing the lint result may have no effect (if it's auto-fixable, which it is in this case), or it fails their build and they can easily update the code accordingly. we do not consider lint style changes to be API breaking.

so adding new categories is not an actual API break: it's actually an enhancement. but changing the config options in a non-backward-compatible way is an API break.

NaridaL:
Ah, thanks for clearing up the distinction. I haven't removed the old-
style option handling code, so this pr isn't breaking by your definition. I
could add a console warning when using this option so that it could be
removed in a future release

src/language/rule/rule.ts
NaridaL: Yeah this preset is the closest.

test/rules/member-ordering/alphabetize/test.ts.fix
giladgray: can we interpret these as `x get, `x set`, `y get`, `y set`? that'll give the desired sorting.

giladgray:
no need to add a console warning as this would apply to _all_ rules

test/rules/member-ordering/alphabetize/test.ts.fix
NaridaL: Heh, I had the exact same idea https://github.com/palantir/tslint/issues/3965#issuecomment-398552202. I dismissed it would result in things like this:
```
get x() {}
"x lol"() {}
set x() {}
```

Rethinking it however, that seems like such an unlikely scenario that it might still be a good solution.

NaridaL:
@giladgray All rules?

I was thinking something like 

image

giladgray:
that is awesome! very clear message with migration steps 👌 
however, definitely does not belong in this PR as that looks like a rather complex feature in itself and deserves to be reviewed separately.

don't think we're ready to deprecate old options usage just yet, though i imagine that'll be a flagship feature for 6.0 (that's what i meant by all rules -- migrating to object syntax across the board)

NaridaL:
The rationale for adding the warning here is that the old style options don't handle accessors correctly and output unhelpful warnings. This rule already handles old style options by converting them to current ones, therefore outputting the above warning doesn't require any logic.

src/rules/memberOrderingRule.ts
```diff
@@ -469,7 +532,14 @@ function getOptionsJson(allOptions: any[]): { order: MemberCategoryJson[]; alpha
...
         // Undocumented direct string option. Deprecate eventually.
-        return { order: convertFromOldStyleOptions(allOptions), alphabetize: false }; // presume allOptions to be string[]
+        const order = convertFromOldStyleOptions(allOptions);
+        showWarningOnce(dedent`
+            Warning: member-ordering - Direct string option is deprecated and does not support accessors.
+            See also https://palantir.github.io/tslint/rules/member-ordering/
+            You should replace ${allOptions.map((o) => JSON.stringify(o)).join()}
+            with the following equivalent options and add -accessor categories as appropriate:\n` +
+            JSON.stringify(order, undefined, "  "));
+        return { order, alphabetize: false }; // presume allOptions to be string[]
```
NaridaL: I've pushed the warning for the previously stated reason. Let me know if you still want it in a separate PR...

NaridaL:
Failing tests are due to latest commit on master.

maiis:
@NaridaL I think you need to rebase this commit to fix the tests: 29fbace7bd67195c7e7f1f115a3170e092f4e78c

NaridaL:
rebased onto master

JoshuaKGoldberg:
Removing the `breaking change` label per #4811!

adidahiya:
thanks for the PR @NaridaL!

Issue #3987, opened 2018-06-22
[Bug][TS 2.9] Whitespace required between import and parens with the type import syntax


Tenga:
### Bug Report

- __TSLint version__: 5.10.0
- __TypeScript version__: 2.9.2
- __Running TSLint via__: CLI (also reproable with VS Code)

#### TypeScript code being linted

```ts
let test: typeof import("lodash/isString");
```

with `tslint.json` configuration:

```json
{
  "extends": "tslint:recommended"
}

```

#### Actual behavior

TSLint reports `missing whitespace` on column 24:

```
ERROR: /src/file.ts[1, 24]: missing whitespace
```
Basically asking for the line above to look like this:
```ts
let test: typeof import ("lodash/isString");
```

Fixer also fixes it to add the space in-between the `import` and parens, like illustrated above.

#### Expected behavior
TSLint should not require whitespace between `import` and parens in this specific case. 

Release notes for the [added type import syntax in TS 2.9](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-9.html#import-types) should anyone need a reference to it.

PR #3992, opened 2018-06-24
[Bugfix] Fix for whitespace rule requiring extra space in ImportType in TS 2.9

Tenga:
#### PR checklist

- [x] Addresses an existing issue: #3987 
  - [x] Includes tests
- [ ] Documentation update

#### Overview of change:

Fixes the bug where the new `ImportType` syntax required extra spacing when `whitespace` rule was enabled.

- Beware, my first dive into TSLint and the AST. 😄 
- Might have overdone or missed the test cases
- As this seems to be only relevant in TS 2.9 and above, not sure how you usually deal with cases like this if you want to support older TS versions?
  - I've done the first thing that came to mind and typecasted the `SyntaxKind` as `any` and let the `ImportKind` end up `undefined` to fail the check. 
  - Is there an standard way on how this is handled usually?
  - I've extracted the tests for this to a file which is only ran on TS 2.9, I'm assuming that's fine as the syntax for this to work should get struck down by the TS compiler for users of the older TS anyway?

#### CHANGELOG.md entry:

[bugfix] `whitespace` rule no longer requires extra spaces with the TS 2.9 import type syntax

### DIFF
--- a/src/rules/whitespaceRule.ts
+++ b/src/rules/whitespaceRule.ts
@@ -296,6 +296,10 @@ function walk(ctx: Lint.WalkContext<Options>) {
                 }
                 break;
             case ts.SyntaxKind.ImportKeyword:
+                if ((ts.SyntaxKind as any).ImportType !== undefined && parent.kind === (ts.SyntaxKind as any).ImportType) {
+                    return; // Don't check ImportType
+                }
+
                 if (parent.kind === ts.SyntaxKind.CallExpression &&
                     (parent as ts.CallExpression).expression.kind === ts.SyntaxKind.ImportKeyword) {
                     return; // Don't check ImportCall
--- a/test/rules/whitespace/all/import-type.ts.lint
+++ b/test/rules/whitespace/all/import-type.ts.lint
@@ -1,0 +1,31 @@
+[typescript]: >=2.9.0
+const foo: import("bar");
+const foo: import ("bar");
+
+function foo(a: import("bar")): import("baz") {}
+function foo(a: import ("bar")): import ("baz") {}
+
+const foo = (a: import("bar")): import("baz") => {};
+const foo = (a: import ("bar")): import ("baz") => {};
+
+type foo = { bar: import("baz") } & import("qux");
+type foo = { bar: import ("baz") } & import ("qux");
+
+interface Foo {
+    bar: import("baz");
+    qux: import ("quux");
+}
+
+class Foo {
+    bar: import("baz");
+    qux: import ("quux");
+}
+
+/**
+ * @param bar { import("qux") }
+ * @param baz { import ("qux") }
+ */
+function foo(bar, baz) { }
+
+type foo = Bar<import("baz")>;
+type foo = Bar<import ("baz")>;


src/rules/whitespaceRule.ts
```diff
@@ -296,6 +296,10 @@ function walk(ctx: Lint.WalkContext<Options>) {
                 }
                 break;
             case ts.SyntaxKind.ImportKeyword:
+                if (parent.kind === (ts.SyntaxKind as any).ImportType) {
```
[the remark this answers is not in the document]
Tenga: Sure, done. Tho it seems redundant to me in this case.

johnwiseheart:
Hi @Tenga, just wanted to check if you have signed the CLA as that appears to be the remaining test left to pass. The cla-bot has been a little confused lately, so if you're unsure what the CLA is, I've copied its message from another PR:

> Thanks for your interest in palantir/tslint, @Tenga! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

Tenga:
Hey @johnwiseheart. I'm aware of it, I'm just waiting for the go-ahead from my employer before signing. ⌚️

Tenga:
@johnwiseheart CLA has been signed.

src/rules/whitespaceRule.ts
```diff
@@ -296,6 +296,10 @@ function walk(ctx: Lint.WalkContext<Options>) {
                 }
                 break;
             case ts.SyntaxKind.ImportKeyword:
+                if ((ts.SyntaxKind as any).ImportType !== undefined && parent.kind === (ts.SyntaxKind as any).ImportType) {
```
johnwiseheart: Hmm, I don't like having to cast as any. What is the issue here - is it that `ts.SyntaxKind.ImportType` is a ts2.9+ thing?

Tenga: Yup.

johnwiseheart: @giladgray @suchanlee @JKillian any ideas on how to work around this?

JoshuaKGoldberg: Let's define a shared `isImportKeyword` in `src/utils.ts` similar to existing helper functions from tsutils.

JoshuaKGoldberg:
Looks like #4243 is the same as this PR but with the comment resolved (and great tests reflected). Thanks for sending this in @Tenga! Closing this now that #4243 went in, but still very much appreciated!

Tenga:
While I'm happy this fairly thin PR will finally be resolved, it leaves a bad taste in my mouth.

This PR has been, with my full understanding, sitting in the queue for months, with me waiting for feedback/advice on how you guys would like to resolve a style problem. The end result is merging a PR that openly rips 90% of the work from this PR, because it was hard to wait a couple of hours to amend this PR with a one-liner change?

¯\\\_(ツ)\_/¯

JoshuaKGoldberg:
@Tenga Thanks for the feedback. You're right, and I'm sorry for marginalizing your contribution here. I should have waited a few days before merging in the other PR.

For context on the sitting in the queue for months, TSLint hasn't had an active maintainer in those months, and I'm coming in now as a new maintainer to clean up the issues and PR queues. My goal here was to minimize the amount of pending work we have, but I went a bit overboard and didn't stop to think about this. 

If you send another PR, I can promise the same thing won't happen twice.

Tenga:
@JoshuaKGoldberg Appreciate the response. 👍

If I encounter an opportunity to contribute again, I'll certainly do so. 

Just to clarify as I probably should've probably phrased it better. I understand that OSS is hard. "Sitting in queue for months" was not meant to be a diss or dissatisfaction with that, just the understanding of the reality. (Contributing guidelines prepare you for that anyway. 😛)

PR #3996, opened 2018-06-27
[new-rule] No self assign

Zzzen:
#### PR checklist

- [x] Addresses an existing issue: #3976 
  - [x] Includes tests
- [ ] Documentation update


michaeljota:

Zzzen:
ping?

src/rules/noSelfAssignRule.ts
```diff
@@ -0,0 +1,210 @@
...
+
+interface Option {
+    props: boolean;
+}
+
+export class Rule extends Lint.Rules.AbstractRule {
+    /* tslint:disable:object-literal-sort-keys */
+    public static metadata: Lint.IRuleMetadata = {
+        ruleName: "no-self-assign",
+        description: "This rule is aimed at eliminating self assignments.",
```
JoshuaKGoldberg: This description doesn't add anything. "This rule is aimed at" is implied; "eliminating" is a longer form of "no"; "assignments" is abbreviated by "assign". Amusing that that's what the ESLint docs call it, but can we get a better description here? Maybe:

> Stops variables from being assigned to themselves.

src/rules/noSelfAssignRule.ts
```diff
@@ -0,0 +1,210 @@
...
+                },
+            },
+        },
+        optionExamples: [[true, { props: false }]],
+        type: "maintainability",
+        typescriptOnly: false,
+    };
+
+    public static FAILURE_STRING(name: string) {
+        return `variable '${name}' is assigned to itself.`;
```
JoshuaKGoldberg: This isn't always for variables. Either change it to say the type (variable, property, etc.) or just remove that part of the string.

src/rules/noSelfAssignRule.ts
```diff
@@ -0,0 +1,210 @@
...
+
+    if (isPropertyAccessExpression(leftExp) && isPropertyAccessExpression(rightExp)) {
+        return isSameMember(leftExp, rightExp);
+    }
+
+    return isIdentifier(leftExp) && isIdentifier(rightExp) && leftExp.text === rightExp.text;
+}
+
+function isSameProperty(left: ts.Node, right: ts.Node) {
+    if (isPropertyAccessExpression(left) && isPropertyAccessExpression(right)
```
JoshuaKGoldberg: Nit: just directly return the result of this expression. `return isPropertyAccessExpression(left) && ...`

test/rules/no-self-assign/default/test.ts.lint
```diff
@@ -0,0 +1,65 @@
+var a = a;
```
JoshuaKGoldberg: Could you please add some cases with three variables in array and object spreads? Seems like that could be a cause of bugs.

JoshuaKGoldberg: How about some test cases for assignments within operators?

```typescript
a = a, a = b, b = b;
a = a && a == a && b = a || c = a, a = b, b = b;
```

Zzzen:
Since this rule exists in eslint and now tslint is deprecated. I guess this pr should be closed.

PR #3583, opened 2017-12-15
Add fixer and new option to commentFormatRule

smoehrle:
#### PR checklist

- [x] Addresses an existing issue: fixes #649 
- [x] New feature, bugfix, or enhancement
  - [x] Includes tests
- [x] Documentation update

#### Overview of change:
- change existing rule to only emit one error
- add fixer
- add option allow-trailing-lowercase

#### CHANGELOG.md entry:

[new-rule-option] "allow-trailing-lowercase" option for comment-format
[new-fixer] comment-format

PR #4003, opened 2018-07-01
[new-rule-option] no-space-keywords for comment-format

Shinigami92:
#### PR checklist

- [x] Addresses an existing issue: #3951
  - [x] Includes tests
- [ ] Documentation update

#### Overview of change:

You can add a new object in `comment-format` to add some exceptions for checking special keywords that can ignore the check-space option

e.g.:
```json
{"comment-format": [true, "check-space", {"no-space-keywords": ["TODO", "FIXME"]}]}
```

```js
// some comment
//TODO: my todo
```

There is a check that allows an optional colon at the end of the keyword

- Using of `in` on line [183](https://github.com/Shinigami92/tslint/blob/84694f6ea84d904f9d935f09be632eb68104d2ea/src/rules/commentFormatRule.ts#L183)
- Line [241](https://github.com/Shinigami92/tslint/blob/84694f6ea84d904f9d935f09be632eb68104d2ea/src/rules/commentFormatRule.ts#L241) to [248](https://github.com/Shinigami92/tslint/blob/84694f6ea84d904f9d935f09be632eb68104d2ea/src/rules/commentFormatRule.ts#L248)

#### CHANGELOG.md entry:

- [new-rule-option] `no-space-keywords` in `comment-format`


Shinigami92:
The formatting in my last push was prettier #4012

michaeljota:

src/rules/commentFormatRule.ts
```diff
@@ -173,19 +239,37 @@ function walk(ctx: Lint.WalkContext<Options>) {
...
             return;
         }
 
+        if (ctx.options.noSpaceExceptions !== undefined) {
+            const match = ctx.options.noSpaceExceptions.exec(commentText);
+            if (match !== null) {
+                const lengthOfMatch = match[0].length;
+                commentText = commentText.substring(lengthOfMatch);
+                // Check optional colon after keyword
+                if (commentText[0] === ":") {
```
JoshuaKGoldberg: If folks want to ignore colons, shouldn't they put that in the regex? What if we don't want it or prefer ` - ` instead?

src/rules/commentFormatRule.ts
```diff
@@ -72,98 +77,159 @@ export class Rule extends Lint.Rules.AbstractRule {
...
+            [true, "check-space", { "no-space-keywords": ["TODO", "FIXME"] }],
+            [true, "check-lowercase", { "ignore-words": ["TODO", "HACK"] }],
+            [true, "check-lowercase", { "ignore-pattern": "STD\\w{2,3}\\b" }]
         ],
         type: "style",
         typescriptOnly: false,
-        hasFix: true,
+        hasFix: true
     };
     /* tslint:enable:object-literal-sort-keys */
```
JoshuaKGoldberg: Please also mention this in the `optionsDescription`. I was confused until I read the test cases 😊.

src/rules/commentFormatRule.ts
```diff
@@ -173,19 +239,37 @@ function walk(ctx: Lint.WalkContext<Options>) {
...
         // whitelist //#region and //#endregion and JetBrains IDEs' "//noinspection ..."
         if (/^(?:#(?:end)?region|noinspection\s)/.test(commentText)) {
             return;
         }
 
+        if (ctx.options.noSpaceExceptions !== undefined) {
+            const match = ctx.options.noSpaceExceptions.exec(commentText);
+            if (match !== null) {
+                const lengthOfMatch = match[0].length;
+                commentText = commentText.substring(lengthOfMatch);
```
JoshuaKGoldberg: Nit suggestion: you only use the `lengthOfMatch` variable once, so just say `commentText.substring(match[0].length)` to have fewer lines.

src/rules/commentFormatRule.ts
Shinigami92: Like my comment pointed out: it is only optional. You can type for example _TODO_ or _TODO:_, the second will then trigger the if statement

src/rules/commentFormatRule.ts
Shinigami92: Updated optionsDescription in commit https://github.com/palantir/tslint/pull/4003/commits/e75444d0285cb4bba08c29d74b1392ce35607634

src/rules/commentFormatRule.ts
JoshuaKGoldberg: Oh, I'm silly. Thanks!

src/rules/commentFormatRule.ts
JoshuaKGoldberg: More strictly, what if someone explicitly doesn't want `:`s to be allowed here? Let's say their coding standards are to have something like:

```typescript
// TODO 55 - enable x y z
```

Allowing a `:` by default, without a way to turn it off, would be disruptive for that user.

It's a minor point, but if this goes in, it would be a breaking change to disable it, which would also be disruptive.

src/rules/commentFormatRule.ts
Shinigami92: 😄

src/rules/commentFormatRule.ts
Shinigami92: OK I think I have to write some new test for this
Will programm it when I have more time

Do you want to have a regexp to do this or should I only remove the optional colon and then everything is open?

currently it is only there because you can not / should not place a space between the TODO and the colon

So if you write `// TODO 55 - ...` then the part after TODO is already the comment

JoshuaKGoldberg: +1 for a regexp!

Shinigami92: So, after I've looked at what I implemented then ~4 months ago, I came back to why regex here makes no sense.

This feature makes it possible to use keywords like TODO without a space between the comment marks and the beginning of the actual comment.
Therefore, it makes the option synonymous only and only sense in combination with `"check-space"`.

The optional colon is a common syntax and is not in conflict with any ' - ' behind the TODO.

Therefore, the following things are possible:

```js
//TODO my comment
//TODO: my comment
//TODO - my comment
//TODO: - my comment (dont think that anyone use this form ^^)
//TODO 55 - my comment (the comment starts after the TODO with 55...)
//TODO: 55 - my comment
```


The only option I have ever seen that could be "problematic" would be the following:

```js
//TODO(abc): comment (abc is an abbreviation)
```

But that comes from the c ++ style guide: https://google.github.io/styleguide/cppguide.html#TODO_Comments
In the example, again there is a space between the comment and the //

So if someone wants to force a comment style, he should use the other rule option ignore-pattern and even that's not really correct.

I only see two options:

1. It is not part of tslint to do anything to enforce a style of commenting
2. Implement a new rule / feature and open a proposal

JoshuaKGoldberg: Ok @Shinigami92, you've convinced me 😄 sounds good. The way it was before, with or without the `:`, seems fine.

Shinigami92:
Will look into the conflict in the next few days

JoshuaKGoldberg:
Hmm, looks like the changes in this PR are actually already resolved by the `ignore-pattern` option in #3583. There are some pretty substantial merge conflicts here that are otherwise nontrivial to resolve as well. Closing this PR for housekeeping. 

Please comment here or in #3951 if you think that's a mistake and you'd like this PR to be re-opened!

Shinigami92:
I'm with your desicion.

PR #4012, opened 2018-07-03
Prettier

danielreichhart:
#### PR checklist

  - [ ] Includes tests
- [ ] Documentation update

#### Overview of change:

- Change "npm run" to "yarn" in docs/develop/docs/index.md
- Added the code formatter prettier (https://prettier.io/)
- Added a precommit hook: To run prettier pre commit over every staged file

If the pull request is accepted, I will create an extra pull request, where prettier is runned over every file in the project.


palantirtech:
Thanks for your interest in [palantir/tslint](https://github.com/palantir/tslint), @danielreichhart! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

package.json
```diff
@@ -67,13 +70,17 @@
     "@types/semver": "^5.3.30",
     "chai": "^3.5.0",
     "github": "^8.2.1",
+    "husky": "^0.14.3",
     "json-stringify-pretty-compact": "^1.0.3",
     "mocha": "^3.2.0",
     "npm-run-all": "^4.0.2",
     "nyc": "^10.2.0",
+    "prettier": "^1.13.7",
```
Shinigami92: Like the documentation pointed out https://prettier.io/docs/en/install.html
Use exact version instead of ^ or ~
Only use 1.13.7

danielreichhart: Right. I changed it.

.gitignore
```diff
@@ -26,3 +26,6 @@ docs/formatters/*/index.html
 
 /coverage/
 /.nyc_output
+
+# ide
```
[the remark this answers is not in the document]
danielreichhart: 👍

r00ger:
Why is prettier needed? From my expirience it makes development less convenient (I'm talking only about typescript):

* in some cases it [reduces readabiility](https://prettier.io/playground/#N4Igxg9gdgLgprEAuEcAeAHCAnGACSKAZ3yLgBs4wYBJImgWwbmyIEsA3OAGQEMBPCAFcYSPAGUYOOOIpUp2ADwAjCBEq8oAPjwBePAAoAlHp0AzXuTIBuEABoQEDDDbQiyUL2zYIAdwAKXgjuKLwcEGwAJvYgyti8YADWcDDiGAlsUADmyDDYQnAOmWS4-vFZDLzIFlaFIABWRGgAQvFJKeK8zNyZcNWWZA7prCzIsbzK-OTQMRjYmTAA6lEwABbIAGwArEM+ZIvxGGNzcCVcMdhwAI5CbJdlvBVVSDWDIGQMbLn5dezZlABFIQQeD9WoOGATZaRNbIABMEPibHImSyAGEIExniAoNA+g4hGQACoTEKvOAAXwpQA)
* maked PR management more complicated because it puts arguments and imported members on a single line or multiple lines depending on line length. For example, if you just rename an argument and change its name's length, instead of changing one line, prettier will change several lines, as 

  ```ts
  function foo(
    does,
    not,
    fit,
  )
  // ↓
  function foo(do, fit, now)
  ```

danielreichhart:
Okay, that’s right. But in my opinion, it reduces the readability only in few edge cases. In general, it improves it. For the few cases, we could open an issue or a pull request at the prettier project. 

You can define the line length in the. prettierrc config file. The current print width is 120 chars. If it makes the PR management more complicated, we could increase the length.

johnwiseheart:
I agree that prettier is a good idea to add here: in the majority of cases, it simplified code style and makes it easier to review PRs - there no longer needs to be checks of code layout/formatting as prettier enforces this. @danielreichhart any idea why the tests are failing after this change? It might not be an issue with this PR but we should figure it out before merging.

danielreichhart:
The failing tests are fixed in this pull request https://github.com/palantir/tslint/pull/4019

johnwiseheart:
I merged the PR for the failing tests - could you merge master into your PR and re-push?

danielreichhart:
@johnwiseheart done

.prettierrc.json
```diff
@@ -0,0 +1,4 @@
+{
+    "tabWidth": 4,
+    "printWidth": 120
```
johnwiseheart: I'm not sure about having a print width of `120` - the prettier docs specifically mention why the default of 80 makes sense. @giladgray do you have any comments here? Maybe we could set it to `100` so its in the middle.

danielreichhart: @johnwiseheart @giladgray are there any news?

Shinigami92: IMO 100 is a good choice because you can see the changes in a commit side by side without scrolling (on a 1080p monitor)

danielreichhart: I’ve always worked with 120, but 100 is a good compromise.

Also, the point the commit changes without scrolling is a good point.

PR #4023, opened 2018-07-10
[new-rule] no-default-import

r00ger:
#### Overview of change:

As the `no-default-export`, the rule supposed to reduce usage of default imports/exports.

New rule supposed to narrow the scope of your changes in the case of medium-sized (and bigger) projects. Say, you have 20 packages and every removed default export from utility package would lead to changes in each package, which might be harder to get merged by various reasons (harder to get your code approved due to a number of required reviewers; longer build time due to a number of affected packages). That's why "requires too many changes elsewhere" is a reason why I see `no-default-export` ignored so often.

Unlike `no-default-export`, the rule requires you to make changes only in the package you work on and the package you import from (unless the member you try to import already exported as a named one).

It has a config option where you have to specify a regexp for packages you own, imports from which are to be checked. By default it checks relative imports, presuming, you own these files.

#### CHANGELOG.md entry:

[new-rule] `no-default-import`

### DIFF
--- a/src/configs/all.ts
+++ b/src/configs/all.ts
@@ -153,6 +153,7 @@ export const rules = {
     "max-file-line-count": [true, 1000],
     "max-line-length": [true, 120],
     "no-default-export": true,
+    "no-default-import": true,
     "no-duplicate-imports": true,
     "no-irregular-whitespace": true,
     "no-mergeable-namespace": true,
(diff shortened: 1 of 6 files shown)


johnwiseheart:
@r00ger, can you please fix the tests by adding your rule to `tslint:all`?

r00ger:
@johnhferland, ✅

michaeljota:

r00ger:
@johnwiseheart, please take a look at another attempt

r00ger:
Hi @johnwiseheart,

any chance to have this merged?

Issue #4033, opened 2018-07-12
no-invalid-this false positive for method declaration syntax


amacleay:
### Bug Report

- __TSLint version__: 5.10.0
- __TypeScript version__: 2.9.2
- __Running TSLint via__: (pick one) CLI 

#### TypeScript code being linted

```ts
function bindProperty(this: { someProperty: number }, fn: Function) {
  return fn.bind(this);
}

const anotherObj: Record<string, Function> = {
  bindProperty: function(this: { someProperty: number }, fn: Function) {
    return fn.bind(this);
  },
};

const someObj: Record<string, Function> = {
  bindProperty(this: { someProperty: number }, fn: Function) {
    return fn.bind(this);
  },
};
```

with `tslint.json` configuration:

```json
{
    "defaultSeverity": "error",
    "jsRules": {},
    "rules": {
      "no-invalid-this": [true]
    },
    "rulesDirectory": []
}
```

#### Actual behavior
Even though `someObj` and `anotherObj` produce identical javascript output because one is just a shorthand for the other, the method declaration style produces a tslint error
```$ tslint --project . --format stylish
/tmp/tmp.DcFq9uCAcp/index.ts:13:20
ERROR: 13:20  no-invalid-this  the "this" keyword is disallowed outside of a class body

```

#### Expected behavior
All of these examples should pass lint

#### Full repro example

`for x in $( git ls-files | grep -v yarn ); do echo;echo "# $x"; cat $x; done`

```

# .gitignore
*.js
node_modules

# index.ts
function bindProperty(this: { someProperty: number }, fn: Function) {
  return fn.bind(this);
}

const anotherObj: Record<string, Function> = {
  bindProperty: function(this: { someProperty: number }, fn: Function) {
    return fn.bind(this);
  },
};

const someObj: Record<string, Function> = {
  bindProperty(this: { someProperty: number }, fn: Function) {
    return fn.bind(this);
  },
};

# package.json
{
  "name": "tmp.dcfq9ucacp",
  "version": "1.0.0",
  "description": "",
  "main": "index.js",
  "scripts": {
    "test": "echo \"Error: no test specified\" && exit 1"
  },
  "author": "",
  "license": "UNLICENSED",
  "devDependencies": {
    "tslint": "^5.10.0",
    "typescript": "^2.9.2"
  }
}

# tsconfig.json
{
  "compilerOptions": {
    "target": "ES5",
    "strict": true,
    "strictNullChecks": true
  }
}

# tslint.json
[...]

PR #4034, opened 2018-07-12
no-invalid-this: false positive for method syntax

amacleay:
* Keep track of evaluation context: look to the closest parent function
  and decide if it is allowed to refer to `this`
* Fix a false negative in the tests from #3267

#### PR checklist

- [x] Addresses an existing issue: #4033
  - [X] Includes tests
- [ ] Documentation update

#### Overview of change:

Previously, the `no-invalid-this` rule only had shallow tracking of context. `this` should be allowed if any only iff the closest parent non-arrow function is either a) a class method, or b) a function with a `this` type parameter.

There were several drawbacks of not tracking context directly:
- The rule didn't treat method declaration syntax the same as function declaration syntax on object literals, causing a false positive on method declaration syntax in object literals (#4033)
- The rule could not distinguish between a correct function with a `this` type parameter and a function _inside_ a function with a `this` type parameter - in fact, there was a test checked in which enforced bad behavior: https://github.com/palantir/tslint/blob/b104e5ad8fdbbb75f1271a1016ff6272f1aafb9d/test/rules/no-invalid-this/enabled/test.ts.lint#L12 - the `this` on this line does _not_ inherit type parameter

#### CHANGELOG.md entry:

[bugfix] [`no-invalid-this`](https://palantir.github.io/tslint/rules/no-invalid-this/) rule fixes false positives on method-like syntax and false negatives on nested functions

### DIFF
--- a/src/rules/noInvalidThisRule.ts
+++ b/src/rules/noInvalidThisRule.ts
@@ -61,21 +61,18 @@ export class Rule extends Lint.Rules.AbstractRule {
     }
 }
 
+const enum parentType {
+    None,
+    Class,
+    ClassMethod,
+    BoundRegularFunction,
+    UnboundRegularFunction,
+}
+const thisAllowedParents = new Set([parentType.ClassMethod, parentType.BoundRegularFunction]);
+
 function walk(ctx: Lint.WalkContext<boolean>): void {
     const { sourceFile, options: checkFuncInMethod } = ctx;
 
-    const enum parentType {
-        None,
-        Class,
-        ClassMethod,
-        BoundRegularFunction,
-        UnboundRegularFunction,
-    }
-
-    function thisIsAllowed(parent: parentType): boolean {
-        return [parentType.ClassMethod, parentType.BoundRegularFunction]
-            .some((t) => parent === t);
-    }
     let currentParent: parentType = parentType.None;
     let inClass = false;
 
@@ -115,7 +112,7 @@ function walk(ctx: Lint.WalkContext<boolean>): void {
                 }
 
             case ts.SyntaxKind.ThisKeyword:
-                if (!thisIsAllowed(currentParent)) {
+                if (!thisAllowedParents.has(currentParent)) {
                     if (!inClass) {
                         ctx.addFailureAtNode(node, Rule.FAILURE_STRING_OUTSIDE);
                     } else if (checkFuncInMethod) {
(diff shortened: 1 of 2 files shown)


palantirtech:
Thanks for your interest in [palantir/tslint](https://github.com/palantir/tslint), @amacleay! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

amacleay:
cc @aervin  - I changed a line you added which I believe to be a mistake (https://github.com/palantir/tslint/pull/4034/files#diff-7062f9042aeae109e75e6f78d7430418R33) so my changes deserve extra scrutiny because I'm pretty often wrong!

amacleay:
Hi all! Feedback would be much appreciated if someone has an opportunity to provide it. I think this should be ready to go as-is if the changes seem appropriate to fix the bug (or, what I believe to be a bug).

michaeljota:

src/rules/noInvalidThisRule.ts
```diff
@@ -59,39 +59,64 @@ export class Rule extends Lint.Rules.AbstractRule {
 
 function walk(ctx: Lint.WalkContext<boolean>): void {
     const { sourceFile, options: checkFuncInMethod } = ctx;
+
+    const enum parentType {
```
JoshuaKGoldberg: Seems semantically odd to recreate this each time `walk` is called. We don't need that, so this should be moved outside of the `walk` function.

src/rules/noInvalidThisRule.ts
```diff
@@ -59,39 +59,64 @@ export class Rule extends Lint.Rules.AbstractRule {
...
+        None,
+        Class,
+        ClassMethod,
+        BoundRegularFunction,
+        UnboundRegularFunction,
+    }
+
+    function thisIsAllowed(parent: parentType): boolean {
```
JoshuaKGoldberg: Nit: instead of this three-line function, how about `const thisAllowedTypes = new Set([ParentType.ClassMethod, ParentType.BoundRegularFunction]);`, and use its `.has`?

amacleay:
Pushing a rebase to hopefully resolve the failure

src/rules/noInvalidThisRule.ts
```diff
@@ -61,41 +61,63 @@ export class Rule extends Lint.Rules.AbstractRule {
     }
 }
 
+const enum parentType {
```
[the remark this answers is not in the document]
amacleay: You got it: https://github.com/palantir/tslint/pull/4376

PR #4064, opened 2018-07-23
Change normalisation of ordered-imports

### LINKED ISSUE #4063: ordered-imports rule doesn't sort in the same order as Organize Imports in TypeScript
### Bug Report

- __TSLint version__: master
- __TypeScript version__: 2.9
- __Running TSLint via__: CLI

See https://github.com/Microsoft/TypeScript/issues/25114

```ts
import { foo, _bar } from 'baz';
```

```json
{
  "ordered-imports": true,
}
```

#### Actual behaviour
It sorts it to `import { _bar, foo }`, which is different to how TypeScript organises it.

#### Expected behaviour
No lint error.

The difference seems to be that TypeScript is normalising strings using `toUpperCase` where as `tslint` uses `toLowerCase`.

aboyton:
This makes it consistent with TypeScript's Organize Imports command

Fixes #4063

#### PR checklist

- [x] Addresses an existing issue: #0000
  - [x] Includes tests
- [ ] Documentation update

#### Overview of change:

This makes the sorting consistent with what TypeScript's Organise Imports ordering does, see https://github.com/Microsoft/TypeScript/issues/25114

Nothing in particular.

#### CHANGELOG.md entry:

[bugfix] Make `ordered-imports` consistent with TypeScript's Organise Imports ordering
[new-rule-option] `case-insensitive-legacy` for `ordered-imports` rule

### DIFF
--- a/src/rules/orderedImportsRule.ts
+++ b/src/rules/orderedImportsRule.ts
@@ -50,6 +50,7 @@ export class Rule extends Lint.Rules.AbstractRule {
             Possible values for \`"import-sources-order"\` are:
 
             * \`"case-insensitive'\`: Correct order is \`"Bar"\`, \`"baz"\`, \`"Foo"\`. (This is the default.)
+            * \`"case-insensitive-legacy'\`: Correct order is \`"Bar"\`, \`"baz"\`, \`"Foo"\`.
             * \`"lowercase-first"\`: Correct order is \`"baz"\`, \`"Bar"\`, \`"Foo"\`.
             * \`"lowercase-last"\`: Correct order is \`"Bar"\`, \`"Foo"\`, \`"baz"\`.
             * \`"any"\`: Allow any order.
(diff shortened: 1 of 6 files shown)


palantirtech:
Thanks for your interest in [palantir/tslint](https://github.com/palantir/tslint), @aboyton! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

johnwiseheart:
Your code itself looks like a pretty simple change, but I have some questions:

Consider the following:
```typescript
import { foo, _bar } from 'baz';
```

Before this PR, `fix`ing this code would result in:
```typescript
import { _bar, foo } from 'baz';
```
right? Because we were previously enforcing that underscored variables would be at the beginning, I think that makes this a breaking change. Can you confirm @giladgray or @suchanlee?

One fix for this might be to allow either way as an option to the rule, in which case in a future major release we can swap the default from being at the front to at the back.

aboyton:
Yes, technically this is a breaking change. It is enforcing a different ordering. You could provide an option to allow people to migrate, or you could tell people to run `--fix` to get the new ordering. Up to you.

johnwiseheart:
I think that the option is probably the best idea. Are you interested in implementing that? Unfortunately the maintainers don't have a lot of time to develop features other than trying to manage the PRs and issues.

aboyton:
I've added a option to restore the legacy behaviour as requested. This diff would be much smaller if #4214 is merged first.

eliasmeire:
Hi all, is there anything blocking this PR from being merged? Seems like it has been idle since Nov 2018. This fix would enable us to use this lint rule since a lot of devs use "Organize Imports" from TypeScript :smile:

aboyton:
If it needs rebasing I'm happy to do that.

eliasmeire:
Ok! thanks for the reply

JoshuaKGoldberg:
@adidahiya another one with the `clean-lockfile` CI freeze :o

adidahiya:
thanks for the PR @aboyton!

PR #4109, opened 2018-08-15
[new rule]: ternary-operator-spacing

arhontmw:
#### PR checklist

- [ ] Addresses an existing issue: #0000
- [x] New feature, bugfix, or enhancement
  - [x] Includes tests
- [ ] Documentation update

#### Overview of change:
New rule, which determines whether a space is required before and/or after question and colon tokens in ternary conditional operator

PR #4110, opened 2018-08-15
[bugfix][enhancement][rule-change] Get rid of check-type option in `whitespace` rule

arhontmw:
#### PR checklist

  - [x] Includes tests
- [ ] Documentation update

#### Overview of change:
Get rid of **check-type** option of `whitespace` rule because of its inconsistency: basing on code, rule's walker determines whether a previous token is a colon and then checks a need for whitespace. The problem is that the colon is used not only for type definition, but also, for example, for conditional operator.

I propose to use separate `typedef-whitespace` rule instead.

### DIFF
--- a/src/configs/all.ts
+++ b/src/configs/all.ts
@@ -265,7 +265,6 @@ export const rules = {
         "check-operator",
         "check-module",
         "check-separator",
-        "check-type",
         "check-typecast",
         "check-preblock",
         "check-type-operator",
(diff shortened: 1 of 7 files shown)


johnwiseheart:
As I mentioned in #4109, if theres not an issue for this it would be great if you could make one so we can discuss all these whitespace changes together.

michaeljota:

adidahiya:
Closing for housekeeping, see https://github.com/palantir/tslint/pull/4109#issuecomment-502409155

Issue #3279, opened 2017-10-04
strict-boolean-expressions should not fire for logical operators used in a non-boolean context


marcind:
### Bug Report

- __TSLint version__: 5.7
- __TypeScript version__: 2.5.3
- __Running TSLint via__: CLI

#### TypeScript code being linted

```ts
// code snippet
function valueOrDefault(a?: string) {
  return a || "the default";
}
```

with `tslint.json` configuration:

```json
{
  "defaultSeverity": "error",
  "extends": [
      "tslint:all"
  ]
}
```

#### Actual behavior

Getting report of 2 errors:
`strict-boolean-expressions  This type is not allowed in the operand for the '||' operator because it could be undefined. Only booleans are allowed.`
`strict-boolean-expressions  This type is not allowed in the operand for the '||' operator because it is always truthy. Only booleans are allowed.`

#### Expected behavior

No errors being reported. I'm using the logical `||` operator to provide a default value for the parameter (I know I could use a function parameter default declaration, the snippet is just to illustrate the general problem). There's no `boolean` in play anywhere: neither the input or the output of the expression.

marcind:
Thoughts? Prayers? Suggestions on where/how to get started?

IllusionMH:
@marcind can you provide examples on how it should behave differently from turning of the rule?

marcind:
Sorry, which part my initial submission is unclear?

I guess my point is that this rule should only run in a "boolean context", I can think of two cases:
1. Used in an `if`, `while`, or `for`
2. Assigned to a variable typed as `boolean`.

The rule should not run when I'm trying to provide a default value or short-circuit evaluation
1. `const a: string = potentiallyUndefinedString || "the default";`
2. `const a: string | undefined = potentiallyUndefinedObject && potentiallyUndefinedObject.getString()`

ajafff:
@marcind the examples from your last post would work if you use the `"allow-undefined-union"` option.

The other part of your request should be a separate rule named `strict-boolean-conditions` which only checks `if`, `for`, `while`, `do ... while` and conditional expressions (`x ? y : z`). 
I can imagine it would only check the type of the whole condition and not it's constituents:
```ts
function foo(a: boolean, b?: boolean) {
    if (b || a) {} // passes, result is always boolean
    if (b && a) {} // fails, result is boolean | undefined
    if (a || b) {} // fails, result is boolean | undefined
    if (a || !!b) {} // passes
}
```
[...]

Issue #4158, opened 2018-09-08
strict-boolean-expressions should ignore RHS of `&&` and `||`


dobesv:
Idiomatic JavaScript / TypeScript typically treat `||` or `&&` operators such that only the left operand is considered a "boolean expression" whereas the RHS might be any value that would be used as a result.

Currently an expression like `process.env.PORT || 8080` gives an error with `strict-boolean-expressions` because it rejects the RHS for being "always truthy".  This is a common enough pattern it should be accomodated by this rule.

This has already been discussed to some degree in https://github.com/palantir/tslint/issues/3279 but it doesn't have its own issue tracking it and I'm not sure what it's status is at this point.

I suggest an option to enable checking of the rhs of these operations, default would be not to check.

JoshuaKGoldberg:
Duplicate of #3279. Still accepting PRs! 😊

dobesv:
I did create a PR for this one, didn't I?

JoshuaKGoldberg:
Ha, yes _(thanks for that, by the way!)_. Just closing duplicate issues to keep our issue count down.

PR #4159, opened 2018-09-08
Add `ignore-rhs` option for `strict-boolean-expressions`

dobesv:
#### PR checklist

- [x] Addresses an existing issue: https://github.com/palantir/tslint/issues/4158
  - [x] Includes tests
- [x] Documentation update

#### Overview of change:

If specified, the right-hand operand of `&&` and `||` will not be checked
as a strict boolean expression.

#### CHANGELOG.md entry:

[new-rule-option] `strict-boolean-expressions` accepts `ignore-rhs` option to disable checking the right-hand side of the `&&` and `||` operators as strictly boolean.

### DIFF
--- a/src/rules/strictBooleanExpressionsRule.ts
+++ b/src/rules/strictBooleanExpressionsRule.ts
@@ -26,8 +26,9 @@ const OPTION_ALLOW_ENUM = "allow-enum";
 const OPTION_ALLOW_NUMBER = "allow-number";
 const OPTION_ALLOW_MIX = "allow-mix";
 const OPTION_ALLOW_BOOLEAN_OR_UNDEFINED = "allow-boolean-or-undefined";
+const OPTION_IGNORE_RHS = "ignore-rhs";
 
-// tslint:disable object-literal-sort-keys
+// tslint:disable object-literal-sort-keys no-bitwise
 
 export class Rule extends Lint.Rules.TypedRule {
     public static metadata: Lint.IRuleMetadata = {
@@ -64,6 +65,7 @@ export class Rule extends Lint.Rules.TypedRule {
               - Also allows \`true | false | undefined\`.
               - Does not allow \`false | undefined\`.
               - This option is a subset of \`${OPTION_ALLOW_UNDEFINED_UNION}\`, so you don't need to enable both options at the same time.
+            * \`${OPTION_IGNORE_RHS}\` ignores the right-hand operand of \`&&\` and \`||\'
         `,
         options: {
             type: "array",
@@ -76,10 +78,11 @@ export class Rule extends Lint.Rules.TypedRule {
                     OPTION_ALLOW_ENUM,
                     OPTION_ALLOW_NUMBER,
                     OPTION_ALLOW_BOOLEAN_OR_UNDEFINED,
+                    OPTION_IGNORE_RHS,
                 ],
             },
             minLength: 0,
-            maxLength: 5,
+            maxLength: 7,
         },
         optionExamples: [
             true,
@@ -116,6 +119,7 @@ interface Options {
     allowNumber: boolean;
     allowMix: boolean;
     allowBooleanOrUndefined: boolean;
+    ignoreRhs: boolean;
 }
 
 function parseOptions(ruleArguments: string[], strictNullChecks: boolean): Options {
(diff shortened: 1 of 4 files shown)


palantirtech:
Thanks for your interest in [palantir/tslint](https://github.com/palantir/tslint), @dobesv! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

dobesv:
Is there anything else I need to (or can) do for this PR to move forward?

krishan:
I would love for #4158 to get solved, as it's blocking me from using `strict-boolean-expresssions` in my project. 

So I'm wondering, what's up with this PR? 
What would be needed to move this forward?

Kinrany:
Has @palantirtech already updated this PR, or is it still waiting for the contributor license agreement to be signed?

dobesv:
It's just waiting for a team member review, I guess.  I haven't heard any feedback about it and I signed the CLA, so assuming I read it right it's just a matter of someone with write access reviewing and approving it.

dobesv:
Fixes https://github.com/palantir/tslint/issues/3279

dobesv:
@JoshuaKGoldberg can you review this one?

JoshuaKGoldberg:
Will do, soon!

Note: I'm new to reviewing & maintaining TSLint, so I won't merge it in soon if it passes review. I'll just leave comments so that if a more experienced maintainer to double-check. Same with other PRs.

rhys-vdw:
Any movement on this one?

JoshuaKGoldberg:
@rhys-vdw unfortunately no 🙁. The Palantir folks are looking for someone to maintain TSLint full time, but until then there's no SLA on PR reviews.

dobesv:
@adidahiya I reversed the option as requested.  Let me know if there's anything else I can do.

krishan:
Thank you, looking forward to using this 🎉

jmoseley:
How/when does this change get published?

lukewlms:
@dobesv This is useful.  The ideal for us would be to still disallow RHS non-boolean items in places that _require_ a boolean (e.g. ternary, if/else...) while allowing RHS non-boolean in any other locations (particularly React rendering - someBool && <MyComponent />).  Do you think this is a common enough case to address?

dobesv:
@lukewlms Are you trying to say that you want the rhs to be forced as a boolean in case like:

```
if(x && 1) return;  // <-- error: 1 is not a boolean
```

but still ignored in other expressions, like:

```
const whatever = { b: (Date.now() > 5000) && 1 }; // <-- OK because the surrounding expression is not a conditional
```

My change only ignores the rhs in the sense that it does not complain if the rhs or `&&` or `||` is not a boolean.  However, the overall expression itself may still not be a boolean, and it's likely an existing rule requiring the condition of `if()` or the ternary operator would still report a problem independantly of this change.

Have you verified that your case isn't already covered by the rule as it is ?

lukewlms:
@dobesv Yes, checked it - here's an example.  I'd like the first two to get flagged but the third to not be flagged - so in places where we absolutely must have a boolean, we're sure to get a strict one.

```ts
const func = (b: boolean, s: string) => {
  // Gets flagged with ignore-rhs on
  if (s) {
    // ...
  }

  // Does not get flagged with ignore-rhs on, but I'd like it to!
  if (b || s) {
    // ...
  }

  // This does not get flagged and that's good!
  return b || s;
}
```

dobesv:
I see.  That does seem like a bug or oversight.  I'd suggest filling a new
issue or a PR for it, since this one is closed.


On Apr 9, 2019 at 4:55 AM, <Luke Williams <notifications@github.com>>
wrote:

@dobesv <https://github.com/dobesv> Yes, checked it - here's an example.
I'd like the first two to get flagged but the third to not be flagged - so
in places where we absolutely must have a boolean, we're sure to get a
strict one.

const func = (b: boolean, s: string) => {
  // Gets flagged with ignore-rhs on
  if (s) {
    // ...
  }

  // Does not get flagged with ignore-rhs on, but I'd like it to!
  if (b || s) {
    // ...
  }

  // This does not get flagged and that's good!
  return b || s;
}

—
You are receiving this because you were mentioned.
Reply to this email directly, view it on GitHub
<https://github.com/palantir/tslint/pull/4159#issuecomment-481220290>, or mute
the thread

adidahiya:
I don't think we reached a consensus on whether boolean-coerced expressions in conditional statements should be treated differently from expressions in other positions in #3279. So I don't know if it's a "bug" per se. Please do file a new issue if you'd like to keep discussing your proposed rule behavior change.

lukewlms:
Logged https://github.com/palantir/tslint/issues/4671 for addressing ignoring RHS only when a boolean value is not required.

PR #4182, opened 2018-09-21
Fix typo in ban docs

pe8ter:
#### PR checklist

  - [ ] Includes tests
- [x] Documentation update

#### Overview of change:

Just a small typo.

#### CHANGELOG.md entry:

[docs] Fix docs typo

### DIFF
--- a/src/rules/banRule.ts
+++ b/src/rules/banRule.ts
@@ -50,7 +50,7 @@ export class Rule extends Lint.Rules.AbstractRule {
               * the name of the function in an array with one element: \`["functionName"]\`
               * an object in the following format: \`{"name": "functionName", "message": "optional explanation message"}\`
             * banning methods:
-              * an array with the object name, method name and optional message: \`["functionName", "methodName", "optional message"]\`
+              * an array with the object name, method name and optional message: \`["objectName", "methodName", "optional message"]\`
               * an object in the following format: \`{"name": ["objectName", "methodName"], "message": "optional message"}\`
                 * you can also ban deeply nested methods: \`{"name": ["foo", "bar", "baz"]}\` bans \`foo.bar.baz()\`
                 * the first element can contain a wildcard (\`*\`) that matches everything. \`{"name": ["*", "forEach"]}\` bans\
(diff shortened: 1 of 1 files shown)


pe8ter:
I'm not sure where all that auto-formatting came from. I'll tamp that down and force push a change.

pe8ter:
I'm at a loss here. It looks like the pre-commit hook is formatting my code, but CircleCI fails on the linting step for the changes the pre-commit hook made. How do I proceed?

src/rules/banRule.ts
```diff
@@ -133,10 +137,14 @@ function parseOptions(args: Array<string | string[] | OptionsInput>): Options {
...
                 case 1:
-                    functions.push({name: arg.name[0], message: arg.message});
+                    functions.push({ name: arg.name[0], message: arg.message });
                     break;
                 default:
-                    methods.push({name: arg.name[arg.name.length - 1], object: arg.name.slice(0, -1), message: arg.message});
+                    methods.push({
+                        name: arg.name[arg.name.length - 1],
+                        object: arg.name.slice(0, -1),
+                        message: arg.message
```
johnwiseheart: So just to confirm, if you move `message` to the beginning of this object, the commit hook moves it back to the end?

pe8ter: It doesn't and I see why. The formatter on my commit changes the line

```typescript
methods.push({name: arg.name[arg.name.length - 1], object: arg.name.slice(0, -1), message: arg.message});
```

into the snippet you posted.

If I then make another commit, the formatter runs again and move `message` up into the correct sorting position. It looks like the formatter can't make both edits in one commit.

pe8ter: What I can do is squash this second commit down and try pushing up again.

pe8ter: The formatter is not able to correct this file no matter how many times it runs.

pe8ter: I manually moved the `message` line up and CI passes now. I guess I don't understand how much the formatter is supposed to change and falsely believed it should correct the sort order.

src/rules/banRule.ts
```diff
@@ -50,7 +50,7 @@ export class Rule extends Lint.Rules.AbstractRule {
               * the name of the function in an array with one element: \`["functionName"]\`
               * an object in the following format: \`{"name": "functionName", "message": "optional explanation message"}\`
             * banning methods:
-              * an array with the object name, method name and optional message: \`["functionName", "methodName", "optional message"]\`
+              * an array with the object name, method name and optional message: \`["objectName", "methodName", "optional message"]\`
```
pe8ter: This is the only change that I explicitly made for this PR. The rest of the changes are auto-formatter changes.

src/rules/banRule.ts
johnwiseheart: Awesome. The autoformatter is a bit of a pain, but hopefully will let us work towards a codebase that is nicer without ruining everyones open PRs.

pe8ter:
Thanks!

Issue #4167, opened 2018-09-16
object-literal-sort-keys and accented words


rodrigojcmello:
### Bug Report

- __TSLint version__: 5.11.0
- __TypeScript version__: 3.0.3
- __Running TSLint via__:  VSCode

#### TypeScript code being linted

```ts
interface IAdicionarItem {
    autor: string; // these 3 items were sorted alphabetically by vscode (no tslint was used for this), here it is right.
    título: string; // i with accent 
    type: 'ADICIONAR_ITEM';
}

export const adicionarItem = (autor: string, título: string): IAdicionarItem => ({
    autor,
    título, // "i" with accent should still come before "y"
    type: 'ADICIONAR_ITEM', // [tslint] The key 'type' is not sorted alphabetically (object-literal-sort-keys)
});
```

with `tslint.json` configuration:

```json
{
    "defaultSeverity": "warning",
    "extends": [
        "tslint:recommended",
        "tslint-react"
    ],
    "jsRules": {},
    "rules": {
        "quotemark": [
            true,
            "single",
            "jsx-single"
        ],
        "jsx-no-multiline-js": false,
        "jsx-no-lambda": false,
        "no-console": false
    },
    "rulesDirectory": []
}
```

#### Actual behavior
Words with accents must be in the end so there is no error / warning in TSLint

#### Expected behavior
Items should be sorted alphabetically considering accents without any TSLint error / warning

<img width="1280" alt="captura de tela 2018-09-16 as 11 18 32" src="https://user-images.githubusercontent.com/5174526/45597590-fae6b980-b9a4-11e8-8877-3834a84e232c.png">

pe8ter:
Looks like JavaScript has some options for an easy win:

https://davidwalsh.name/sorting-strings-accented-characters
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/localeCompare

I could make a PR if this. I'd like to know if using `localeCompare()` for all sorting would be an acceptable solution, or if adding an option to use this as the sort function would be preferable.

johnwiseheart:
I think its probably preferable to add an option for this to avoid throwing errors on code that was previously "correct". In the next major version we can turn on the option by default.

pe8ter:
@rodrigojcmello Feature released in https://github.com/palantir/tslint/releases/tag/5.12.0.

johnwiseheart:
Since it’s been released, I think we can close this issue.

PR #4193, opened 2018-09-28
Add object literal sort option to account for special characters

pe8ter:
#### PR checklist

- [x] Addresses an existing issue: #4167 
- [x] Enhancement
  - [x] Includes tests
- [x] Documentation update

#### Overview of change:

The current behavior of the `object-literal-sort-keys` rule is to sort any accented characters after the non-accented "z" character, which is undesirable in accent-heavy languages like Spanish, French, and German. This change gives users of this rule the option to sort accented characters next to their non-accented counterpart.

* This PR makes no attempt to allow the user to configure which language they want to target. The default behavior of the `localeCompare()` function is to make a "best fit" guess at which locale to use.
* The bulk of this PR is auto-formatting changes so I noted the logic change explicitly with a comment.
* I wasn't sure how to format the _CHANGELOG.md_ message.

Reference links:
* https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/localeCompare
* https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl#Locale_negotiation

#### CHANGELOG.md entry:

[new-rule-option] `object-literal-sort-keys`: `locale-compare`


src/rules/objectLiteralSortKeysRule.ts
```diff
@@ -173,12 +210,25 @@ function walk(ctx: Lint.WalkContext<Options>, checker?: ts.TypeChecker): void {
...
-                        property.name.kind === ts.SyntaxKind.StringLiteral) {
-                        const key = ignoreCase ? property.name.text.toLowerCase() : property.name.text;
+                    if (
+                        property.name.kind === ts.SyntaxKind.Identifier ||
+                        property.name.kind === ts.SyntaxKind.StringLiteral
+                    ) {
+                        const key = ignoreCase
+                            ? property.name.text.toLowerCase()
+                            : property.name.text;
                         // comparison with undefined is expected
```
pe8ter: Here is where the logic change starts for this PR.

pe8ter:
I'm not sure what's happening. I resolved the conflicts using GitHub's built-in tool and committed my changes, but this page still shows that there's a merge conflict and the commit doesn't appear in this PR's history. If I try to resolve the conflicts again, the tool refuses to save my changes due to an unspecified error.

pe8ter:
I'm not sure what's happening. GitHub seems to be having issues right now. I fixed my conflicts, but they're not appearing in this PR. I already tried adding a comment stating this, but the comment disappeared mysteriously.

test/rules/object-literal-sort-keys/local-compare/test.ts.lint
```diff
@@ -0,0 +1,10 @@
+const a = {
+    familia: 1,
+    fútbol: 2
+}
+
+const b = {
+    fútbol: 2,
+    familia: 1
+    ~~~~~~~ [The key 'familia' is not sorted alphabetically]
```
ericanderson: I copied your test to my checkout on master and it still seems to work without your code change. Can you write a test that would fail under the prior conditions but pass under the current ones please?

ericanderson:
Apologies for the delay in getting to this PR.

test/rules/object-literal-sort-keys/local-compare/test.ts.lint
pe8ter: - [x] Fixed.

pe8ter:
Thank you for your feedback. I made all requested changes:

- [x] Use data set where tests fail in master, but pass in feature branch.
- [x] Rename _local-compare_ → _locale_compare_.

PR #4198, opened 2018-10-03
Added function-constructor rule

### LINKED ISSUE #3554: Port tslint-microsoft-contrib's no-function-constructor-with-string-args

> `no-function-constructor-with-string-args` can now done as an optionally typed rule: without type info it only flags string literals. It should probably also include `setInterval` and `setTimeout`, with a config to add other built-in functions that allow strings as functions.

JoshuaKGoldberg:
#### PR checklist

- [x] Addresses an existing issue: #3554
  - [x] Includes tests
- [x] Documentation update

#### Overview of change:

Blocks using the `Function()`/`new Function()` methods to create new functions.

Fixes #3554.

#### CHANGELOG.md entry:

[new-rule] `function-constructor`

### DIFF
--- a/src/configs/all.ts
+++ b/src/configs/all.ts
@@ -29,19 +29,25 @@ export const rules = {
     "ban-types": {
         options: [
             ["Object", "Avoid using the `Object` type. Did you mean `object`?"],
-            ["Function", "Avoid using the `Function` type. Prefer a specific function type, like `() => void`."],
+            [
+                "Function",
+                "Avoid using the `Function` type. Prefer a specific function type, like `() => void`."
+            ],
             ["Boolean", "Avoid using the `Boolean` type. Did you mean `boolean`?"],
             ["Number", "Avoid using the `Number` type. Did you mean `number`?"],
             ["String", "Avoid using the `String` type. Did you mean `string`?"],
-            ["Symbol", "Avoid using the `Symbol` type. Did you mean `symbol`?"],
-        ],
+            ["Symbol", "Avoid using the `Symbol` type. Did you mean `symbol`?"]
+        ]
     },
     "ban-ts-ignore": true,
     "member-access": [true, "check-accessor", "check-constructor", "check-parameter-property"],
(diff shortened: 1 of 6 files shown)


src/configs/all.ts
```diff
@@ -29,19 +29,25 @@ export const rules = {
     "ban-types": {
```
JoshuaKGoldberg: Err. I think my prettier plugin did this automatically. Will revert...

ericanderson:
@JoshuaKGoldberg Since I'm new to maintaining this, I want to make sure we are doing the right things. Doesn't changing "recommended" constitute a break?

JoshuaKGoldberg:
Heya @ericanderson - technically yes, but historically it's had additive changes in patch versions. For example, `5.11.0` added the `no-reference-import` rule to recommended. IMO this is fine because the `recommended`/`all` lists don't come with a guarantee that nothing will be added, as with TypeScript's `--strict` flag. It'd also be pretty restrictive if those rulesets could only be changed in major versions.

If this were removing a rule from `recommend` or breaking behavior of an existing rule, I'd feel differently.

Does that sit right with you? I can remove the changed to `recommended.ts` if you want. 😊

adidahiya:
@JoshuaKGoldberg @ericanderson actually, we shouldn't be making `tslint:recommended` more strict across minor versions, only very obvious bug fixes should cause a change to that rule set. This is documented under "Configuration presets" here: https://palantir.github.io/tslint/usage/configuration/

> This configuration follows semver, so it will not have breaking changes across minor or patch releases.

`tslint:latest` is the rule set which does not follow semver:

> Using this config may introduce breaking changes across minor releases as new rules are enabled which cause lint failures in your code.

JoshuaKGoldberg:
Darn, thanks for the clarification @adidahiya!

ericanderson:
I wanted to make it less strict :)

adidahiya:
@ericanderson I agree with that overall goal in v6.0, through #4305

PR #4204, opened 2018-10-06
Add Exclusion Names for Rule file-name-casing

rwaskiewicz:
#### PR checklist

- [X] Addresses an existing issue: fixes #4050 
- [X] New feature, bugfix, or enhancement
  - [X] Includes tests
- [ ] Documentation update

#### Overview of change:
- Add exclusion option for one or more files
- Add test files

#### Is there anything you'd like reviewers to focus on?
N/A

Issue #4071, opened 2018-07-27
[Feature Request] Support pattern matching for rule "file-name-casing"


Ky6uk:
### Feature request

Many projects can have different styles for naming files. For example React projects can have `pascal-case` rule for component files (`Component.tsx`, `Some/Other/Component.tsx`) and `kebab-case` rule for other files (`utils/formatter.ts`, `helper/check-string.ts`).

So I think we should extend rule `file-name-casing` to array of styles with filename pattern as one of arguments. For example: 

```yaml
file-name-casing:
  severity: 'error'
  options:
    - [/\.tsx$/, 'pascal-case']
    - [/\.style\.ts$/, 'pascal-case']
    - [/\.ts$/, 'kebab-case']
```

cheeZery:
I'll take a shot at this one!

PR #4206, opened 2018-10-07
Added possibility to enforce different file name casings

cheeZery:
#### PR checklist

- [x] Addresses an existing issue: #4071
- [x] Enhancement
  - [x] Includes tests
- [ ] Documentation update

#### Overview of change:
Enhanced file-name-casing rule by adding possibility to enforce casing per file name match (case-insentive RegExp).

Since this is my first contribution to this repo, I'd like you to be very thoroughly in your review :-) I also wasn't familiar with the syntax for the `options` object, so that might be improvable.
And I only now see, that there's another PR (#4204, @rwaskiewicz) which also enhances this very rule by adding an exclusion option. Maybe that feature and this one can be combined by adding an additional "ignore" option to the file matcher argument. For example:
```
"file-name-casing": [true, [
    "index\.ts": "ignore",
    "\.ts$": "camel-case",
]];
```
By the way, I simply defined the RegExp as being case-insensitive but this can of course easily be changed. The check will currently also only consider the file name and its extension. I thought about taking the whole path in consideration but I guess that would have been more complicated due to two reasons: the different file separators (`/` vs. `\`), and that we probably should use the relative path from the project base dir or the tslint.json file

In addition to that please consider the fact, that there are now some cases were the rule could produce false-positives for invalid rule arguments.
For example when the rule's arguments are `[true, "camelCase"]` (instead of "camel-case") the linter will now ignore this option. Prior to my change this was somewhat of a false-negative, because TSLint printed the error "File name must be undefined" (for every file).
And when the user now enters an invalid RegExp as a file match, that entry will be ignored.

Can and should the user be informed about those things with some kind of warning?

(Last but not least I stumbled upon a type in the `launch.json` and fixed it, hope that was fine)

#### CHANGELOG.md entry:

[new-rule-option] File name whitelist for `file-name-casing`


palantirtech:
Thanks for your interest in [palantir/tslint](https://github.com/palantir/tslint), @cheeZery! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

src/rules/fileNameCasingRule.ts
```diff
@@ -92,15 +140,53 @@ export class Rule extends Lint.Rules.AbstractRule {
...
+            return true;
+        } catch (e) {
+            return false;
+        }
+    }
+
+    private static findApplicableCasing(
+        fileBaseName: string,
+        fileNameCasings: FileNameCasings
+    ): Casing | null {
```
JoshuaKGoldberg: > `| null`

Why so many `null`s instead of `undefined`? It's generally preferable in JavaScript/TypeScript to prefer using `undefined`, as it's the default value. TSLint typically goes by that preference.

Unless you have a real need for using `null`, please switch to `undefined` as defaults.

src/rules/fileNameCasingRule.ts
```diff
@@ -25,47 +25,95 @@ enum Casing {
...
             [true, Casing.CamelCase],
             [true, Casing.PascalCase],
             [true, Casing.KebabCase],
             [true, Casing.SnakeCase],
+            [
+                true,
+                [
+                    [".style.ts$", Casing.KebabCase],
+                    [".tsx$", Casing.PascalCase],
+                    [".*", Casing.CamelCase]
```
JoshuaKGoldberg: It's generally convention to use objects for key/value pairs. Unless you have a specific need, please switch to something like:

```typescript
{
    ".style.ts$": Casing.KebabCase,
    ".tsx$": Casing.PascalCase,
    ".*": Casing.CamelCase,
}
```

One benefit, in addition to being semantically more clear, is that you get rid of duplicate keys easily.

src/rules/fileNameCasingRule.ts
```diff
@@ -25,47 +25,95 @@ enum Casing {
...
-        typescriptOnly: false,
+        typescriptOnly: false
     };
     /* tslint:enable:object-literal-sort-keys */
 
     private static FAILURE_STRING(expectedCasing: Casing): string {
         return `File name must be ${Rule.stylizedNameForCasing(expectedCasing)}`;
     }
 
+    private static isValidCasingOption(casing: string) {
```
JoshuaKGoldberg: Nit: instead of this dedicated function, you can keep a `const validCasingOptions = new Set([Casing.CamelCase, ...]);` outside of the class and use its `.has`. A little less code.

src/rules/fileNameCasingRule.ts
```diff
@@ -92,15 +140,53 @@ export class Rule extends Lint.Rules.AbstractRule {
...
+    }
+
+    private static findApplicableCasing(
+        fileBaseName: string,
+        fileNameCasings: FileNameCasings
+    ): Casing | null {
+        const applicableCasing = fileNameCasings.find(fileNameCasing => {
+            const fileNameMatch = fileNameCasing[0];
+            return (
+                Rule.isValidRegExp(fileNameMatch) && RegExp(fileNameMatch, "i").test(fileBaseName)
```
JoshuaKGoldberg: ...so if an invalid regular expression is provided, the rule will fail silently? 

We should prefer to fail verbosely for invalid configurations. Either remove the `isValidRegExp` check _(+1)_ or `console.warn` if it fails.

There are a lot of ways to provide invalid configurations, and it's near-impossible to guard against all of them. TSLint is moving towards standardized config options but it'll be a while before they're ready.

JoshuaKGoldberg:
> Can and should the user be informed about those things with some kind of warning?

A `console.warn` seems like a good step in this direction, though agreed that there's not much precedent here.

src/rules/fileNameCasingRule.ts
cheeZery: My bad, in my team I'm still fighting for `strictNullChecks`, so as of now we don't really explicitly work with either `undefined` or `null`. But we're getting there 😊 While changing the code to use `undefined` I realized that's way more intuitive, thanks for the suggestion!

src/rules/fileNameCasingRule.ts
cheeZery: Will switch to object, thanks!
I'm embarassed to admit how long it took me to realize that the syntax in `options` is actually JSON schema (since I never really worked with that). After that I optimized the declaration a bit, looking forward to your opinion after I pushed my changes.
Maybe the comment for `options` in `IRuleMetadata` should begin with "_JSON_ schema ..." for people like me 😄

src/rules/fileNameCasingRule.ts
cheeZery: Yeah, there already were cases like this in the current implementation, see [here](https://github.com/palantir/tslint/blob/master/src/rules/fileNameCasingRule.ts#L96) for example: if the mandatory casing wasn't provided simply return an empty array 🤷‍♂️
I would like to fail more verbosely as you suggested, hence my question that you already answered below.
Currently there don't seem to be any `console.warn`'s in the rules, so maybe there's a better way? In [objectLiteralSortKeys](https://github.com/palantir/tslint/blob/master/src/rules/objectLiteralSortKeysRule.ts#L103) for example an actual `Error` is thrown.
But I will go with what you guys suggest 🌝

cheeZery: @JoshuaKGoldberg I just realized that it might be not a good idea to either throw an error or use console.warn, since this would happen for every linted file in the project. Using `showWarningOnce` on the other should work, some rules even use this already. Any other suggestions?

JoshuaKGoldberg: Agreed. Filed #4280 to tackle this more broadly.

In the meantime, feel free to either use the `showWarningOnce` approach or not validate. IMO it shouldn't be each rule's responsibility to validate options _ideally_, since they already specify their format in their metadata... but since there isn't yet a method for that, it seems up to you.

cheeZery: Thanks for the issue! (There's a small typo in my mention 😌)

src/rules/fileNameCasingRule.ts
```diff
@@ -35,37 +36,73 @@ export class Rule extends Lint.Rules.AbstractRule {
...
-            * \`${Casing.PascalCase}\`: File names must be Pascal-cased: \`FileName.ts\`.
-            * \`${Casing.KebabCase}\`: File names must be kebab-cased: \`file-name.ts\`.
-            * \`${Casing.SnakeCase}\`: File names must be snake-cased: \`file_name.ts\`.`,
+            * In both cases the casing string must be one of the options:
+            ** \`${Casing.CamelCase}\`: File names must be camel-cased: \`fileName.ts\`.
+            ** \`${Casing.PascalCase}\`: File names must be pascal-cased: \`FileName.ts\`.
+            ** \`${Casing.KebabCase}\`: File names must be kebab-cased: \`file-name.ts\`.
+            ** \`${Casing.SnakeCase}\`: File names must be snake-cased: \`file_name.ts\`.
+
+            * The array again consists of array with two items. The first item must be a case-insenstive
```
JoshuaKGoldberg: > The array again

Should this be updated to mention that it takes in an object?

cheeZery:
@JoshuaKGoldberg, I also now saw the duplicate PR #4280. Both PR pretty much do the same, in #4280 there are a more functions "outsourced" and a bit more types, here you have some additional checks and console warnings. I'm not sure which one should be merged 🤷‍♂️

yordis:
> here you have some additional checks and console warnings.

I love this, and I would propose to merge that part of the PR into my version (with your commits sign off @cheeZery I wouldn't like it otherwise to be honest) also the extra testings.

Also the other PR do not have extra changes unrelated to the issue so should be simple to review.

The functions could be move into the class definition if you want, but I would say I prefer to have the concept of the different validators as I did. Specially that reading each body of the functions you could map your head around it easier.

`validate` exists because it is a factory function basically responsable for deciding what validatio you should use.

Technically we are doing also the same but I decided to move different sections into functions separating the responsabilities.

cheeZery:
@yordis I too like the idea to get the best of both PRs. I guess I even like your approach, to define the functions outside of the class scope better!
I wouldn't mind if your PR would be merged first and after that we take a look and figure out what of my PR is still needed and merge that afterwards :-)

JoshuaKGoldberg:
Great, let's do that! Thanks you two 😍

In summary, I'll:
* Merge in #4284
* Resolve merge conflicts on this PR
* Re-review this one, which should just be the added rule option validation

cheeZery:
Perfect!

JoshuaKGoldberg:
@cheeZery does this merge look good to you?

cheeZery:
@JoshuaKGoldberg it does indeed look good to me, I agreed with @yordis 👬 Thanks to the both of you!

JoshuaKGoldberg:
Awesome, thanks everyone!

Issue #4085, opened 2018-07-31
unknown support for 'array-type'


ThomasdenH:
### Bug Report

- __TSLint version__: 5.11.0
- __TypeScript version__: 3.0.1
- __Running TSLint via__: CLI

#### TypeScript code being linted

```ts
const a: unknown[] = [];
```

with `tslint.json` configuration:

```json
{
  "rules": {
    "array-type": [true, "array-simple"],
  }
}
```

#### Actual behavior
TSLint recommends `Array<unknown>` instead of `unknown[]`.

#### Expected behavior
`unknown[]` is considered a simple type and thus be preferred over `Array<unknown>`.

johnwiseheart:
We should definitely accept the `unknown` type as a simple type. Happy to review a PR for this.

ThomasdenH:
Since `unknown` was added in Typescript 3.0.0, how to add this rule with backwards compatibility?

PR #4219, opened 2018-10-12
Add support for unknown to array-type rule.

Retsam:
#### PR checklist

- [X] Addresses an existing issue: #4085
  - [X] Includes tests
- [ ] Documentation update

#### Overview of change:

Currently, when used with TS 3.0, the `array-simple` rule will error on `unknown[]`, instead enforcing that it should be written as `Array<unknown>`.  A small change was needed for `unknown` to be considered a simple type with TS 3.0.

#### Backwards and forwards compatibility

This change works with both Typescript 2.x and 3.x.

The tests are backwards compatible with TS 2.x, because in `unknown` is already treated as a simple type, in that case.  (It's treated as a non-generic type reference)  

The implementation is backwards compatible with TS 2.x, because `ts.SyntaxKind.UnknownKeyword:` will evaluate to `undefined:` and so that case of the switch statement will just never match with `nodeType.kind`, which should never be undefined.

#### CHANGELOG.md entry:

[enhancement] `array-type` rule handles Typescript 3.0's unknown type properly


palantirtech:
Thanks for your interest in [palantir/tslint](https://github.com/palantir/tslint), @Retsam! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

JoshuaKGoldberg:
👋 @Retsam - I just merged a similar PR that didn't have all the Prettier noise. You should still get the "contributor" badge for you work on this one so I'll clean up the conflicts here and try to merge it in. Thanks for the PR!

PR #4227, opened 2018-10-16
Enabled strict-type-predicates internally

JoshuaKGoldberg:
As per the TODO list in `tslint.json`


JoshuaKGoldberg:
The PR queue is still pretty huge and this is low priority; closing for housekeeping.

PR #4258, opened 2018-10-31
Update Configuration Documentation

### LINKED ISSUE #3694: Confused by the configuration document
According to the [configuration document](https://palantir.github.io/tslint/usage/configuration):

> `rules?: { [name: string]: RuleSetting }`: A map of rule names to their configuration settings.
> - These rules are applied to `.ts` and `.tsx` files.
> - Each rule is associated with an object containing:
> - `options?: any`: An array of values that are specific to a rule.
> - `severity?: "default" | "error" | "warning" | "off"`: Severity level. Level "error" will cause exit code 2.
> - A boolean value may be specified instead of the above object, and is equivalent to setting no options with default severity.
> - Any rules specified in this block will override those configured in any base configuration being extended.
> - [Check out the full rules list here](https://palantir.github.io/tslint/rules).

To my understanding, the document says a rule should be associated with an object like this:

```json
{
    "options": ["option1", "option2", "option3"],
    "serverity": "some value"
}
```

Or a boolean:

```json
true
```

Or another boolean:

```json
false
```

Is my understanding correct? If so, does specifying `true` has the same meaning as specifying `false`?

The document says the value of `"options"` is an array, for a rule like [no-empty](https://palantir.github.io/tslint/rules/no-empty/) whose type is a `string`, should I enclose the string with an array?

Also, I saw [many](https://palantir.github.io/tslint/rules/indent/) [rule](https://palantir.github.io/tslint/rules/interface-name/) [examples](https://palantir.github.io/tslint/rules/max-classes-per-file/) specifying an array with a leading `true`:

```json
"indent": [true, "spaces", 4]
```

```json
"interface-name": [true, "never-prefix"]
```

```json
"max-classes-per-file": [true, 5, "exclude-class-expressions"]
```

I couldn‘t find anything about this in the document. What does the leading `true` mean?

rwaskiewicz:
Fixes #3694 

#### PR checklist

- [X] Addresses an existing issue: #3694
  - [ ] Includes tests
- [x] Documentation update

#### Overview of change:
- Add section on array syntax
- Mark boolean syntax as legacy

#### CHANGELOG.md entry:
[documentation] Update configuration documentation to include array syntax

### DIFF
--- a/docs/usage/configuration/index.md
+++ b/docs/usage/configuration/index.md
@@ -11,29 +11,31 @@ configure which rules get run and each of their options.
 
 `tslint.json` or `tslint.yaml` files can have the following fields specified:
 
-* `extends?: string | string[]`:
-The name of a built-in configuration preset (see built-in presets below), or a path or
-array of paths to other configuration files which are extended by this configuration.
-This value is handled using node module resolution semantics.
-For example, a value of `"tslint-config"` would tell TSLint to try and load the main file of a module
-named "tslint-config" as a configuration file. Specific files inside node modules can also be
-specified, eg. `"tslint-config/path/to/submodule"`. Relative paths to JSON files or JS modules
-are also supported, e.g. `"./tslint-config"`.
-* `rulesDirectory?: string | string[]`:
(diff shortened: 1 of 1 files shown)


rwaskiewicz:
CircleCI is reporting:
> test/rules/object-literal-key-quotes/consistent-as-needed/test.ts.lint: Failed!

But this PR only affects a markdown file - I don't _think_ this change causes it.  Please let me know if I'm wrong here 🙂

PR #4284, opened 2018-11-12
Add support for complex matching to file-name-casing-rule

### LINKED ISSUE #4071: [Feature Request] Support pattern matching for rule "file-name-casing"

### LINKED ISSUE #4220: Improve file-name-casing rule
### Rule Suggestion

Working on the React community is common to find two patterns when it comes to the naming convention.

1. Component definition files are use PascalCase
2. Otherwise, use camelCase

Add a configuration to `file-name-casing` rule where allow us to define the rules apply for this cases

```json
"file-name-casing" [true, { "jsx": "pascal-case", "all": "camel-case" }]
```

**Additional context**

yordis:
fixes #4220. Fixes #4071

#### PR checklist

- [x] Addresses an existing issue: #4220
  - [x] Includes tests
- [x] Documentation update

#### Overview of change:

Following #4220 we could pass a key-value pair where the key represents a regex that will be used
for matching against the names.

A little caveat, the regex is always a termination regex.

For example:

`.component.tsx` is actually `.component.tsx$` (notice the `$`).

Allowing arbitrary Regex is a rabbit hole, at least this way we ware safe to remove the last part of the Regex from the string and be able to check the remaining name, since we do not want to check validate the regex section of the name.

I am missing the correct `options` configuration, I am not sure what is the correct config.

#### CHANGELOG.md entry:

```
[enhancement] file-name-casing-rule supports regex expression for matching file
```

### DIFF
--- a/src/rules/fileNameCasingRule.ts
+++ b/src/rules/fileNameCasingRule.ts
@@ -78,13 +78,10 @@ const validateWithSimpleConfig: Validator<Casing> = (sourceFile, casingConfig) =
     return isValid ? undefined : casingConfig;
 };
 
-const validate = (sourceFile: ts.SourceFile, casingConfig: Config): ValidationResult => {
-    const validator =
-        typeof casingConfig === "string" ? validateWithSimpleConfig : validateWithRegexConfig;
-
-    // @ts-ignore https://github.com/Microsoft/TypeScript/issues/7294
-    return validator(sourceFile, casingConfig);
-};
+const validate = (sourceFile: ts.SourceFile, casingConfig: Config): ValidationResult =>
+    casingConfig === "string"
+        ? validateWithSimpleConfig(sourceFile, casingConfig as SimpleConfig)
+        : validateWithRegexConfig(sourceFile, casingConfig as RegexConfig);
 
 export class Rule extends Lint.Rules.AbstractRule {
     /* tslint:disable:object-literal-sort-keys */
--- a/src/rules/fileNameCasingRule.ts
+++ b/src/rules/fileNameCasingRule.ts
@@ -28,6 +28,74 @@ enum Casing {
     SnakeCase = "snake-case",
 }
 
+type RegexConfig = Record<string, Casing>;
+
+type SimpleConfig = Casing;
+
+type Config = SimpleConfig | RegexConfig;
+
+interface ValidatorSuccessfulResult {
+    valid: true;
+}
+
+interface ValidatorFailureResult {
+    valid: false;
+    failedCasing: Casing;
+}
+
+type ValidatorResult = ValidatorSuccessfulResult | ValidatorFailureResult;
+
+type Validator<T> = (casing: T) => (sourceFile: ts.SourceFile) => ValidatorResult;
+
+const rules = [Casing.CamelCase, Casing.PascalCase, Casing.KebabCase, Casing.SnakeCase];
+
+function isCorrectCasing(fileName: string, casing: Casing): boolean {
+    switch (casing) {
+        case Casing.CamelCase:
+            return isCamelCased(fileName);
+        case Casing.PascalCase:
+            return isPascalCased(fileName);
+        case Casing.KebabCase:
+            return isKebabCased(fileName);
+        case Casing.SnakeCase:
+            return isSnakeCased(fileName);
+    }
+}
+
+const RegexValidator: Validator<RegexConfig> = casingConfig => (sourceFile): ValidatorResult => {
+    const fileName = path.parse(sourceFile.fileName).base;
+    const config = Object.keys(casingConfig).map(key => ({
+        casing: casingConfig[key],
+        regex: RegExp(`${key}$`),
+    }));
+
+    const match = config.find(c => c.regex.test(fileName));
+
+    if (!match) {
+        return { valid: true };
+    }
+
+    const normalizedFileName = fileName.replace(match.regex, "");
+
+    return isCorrectCasing(normalizedFileName, match.casing)
+        ? { valid: true }
+        : { valid: false, failedCasing: match.casing };
+};
+
+const SimpleValidator: Validator<Casing> = casingConfig => (sourceFile): ValidatorResult => {
+    const fileName = path.parse(sourceFile.fileName).name;
+    const isValid = isCorrectCasing(fileName, casingConfig);
+
+    return isValid ? { valid: true } : { valid: false, failedCasing: casingConfig };
+};
+
+const validate = (sourceFile: ts.SourceFile, casingConfig: Config): ValidatorResult => {
+    const validator = typeof casingConfig === "string" ? SimpleValidator : RegexValidator;
+
+    // @ts-ignore https://github.com/Microsoft/TypeScript/issues/7294
+    return validator(casingConfig)(sourceFile);
+};
+
 export class Rule extends Lint.Rules.AbstractRule {
     /* tslint:disable:object-literal-sort-keys */
     public static metadata: Lint.IRuleMetadata = {
@@ -43,12 +111,27 @@ export class Rule extends Lint.Rules.AbstractRule {
             * \`${Casing.SnakeCase}\`: File names must be snake-cased: \`file_name.ts\`.`,
         options: {
             type: "array",
-            items: [
-                {
-                    type: "string",
-                    enum: [Casing.CamelCase, Casing.PascalCase, Casing.KebabCase, Casing.SnakeCase],
-                },
-            ],
+            items: {
+                anyOf: [
+                    {
+                        type: "array",
+                        items: [
+                            {
+                                type: "string",
+                                enum: rules,
+                            },
+                        ],
+                    },
+                    {
+                        type: "object",
+                        // TODO: I am not sure how to setup the type
+                        // properties: {
+                        //     type: "string",
+                        //     enum: rules,
+                        // },
+                    },
+                ],
+            },
         },
         optionExamples: [
             [true, Casing.CamelCase],
@@ -79,32 +162,24 @@ export class Rule extends Lint.Rules.AbstractRule {
         }
     }
 
-    private static isCorrectCasing(fileName: string, casing: Casing): boolean {
-        switch (casing) {
-            case Casing.CamelCase:
-                return isCamelCased(fileName);
-            case Casing.PascalCase:
-                return isPascalCased(fileName);
-            case Casing.KebabCase:
-                return isKebabCased(fileName);
-            case Casing.SnakeCase:
-                return isSnakeCased(fileName);
-        }
-    }
-
     public apply(sourceFile: ts.SourceFile): Lint.RuleFailure[] {
         if (this.ruleArguments.length !== 1) {
             return [];
         }
 
-        const casing = this.ruleArguments[0] as Casing;
-        const fileName = path.parse(sourceFile.fileName).name;
-        if (!Rule.isCorrectCasing(fileName, casing)) {
-            return [
-                new Lint.RuleFailure(sourceFile, 0, 0, Rule.FAILURE_STRING(casing), this.ruleName),
-            ];
-        }
+        const casingConfig = this.ruleArguments[0] as Config;
+        const validationResult: ValidatorResult = validate(sourceFile, casingConfig);
 
-        return [];
+        return validationResult.valid
+            ? []
+            : [
+                  new Lint.RuleFailure(
+                      sourceFile,
+                      0,
+                      0,
+                      Rule.FAILURE_STRING(validationResult.failedCasing),
+                      this.ruleName,
+                  ),
+              ];
     }
 }
--- a/src/rules/fileNameCasingRule.ts
+++ b/src/rules/fileNameCasingRule.ts
@@ -95,7 +95,14 @@ export class Rule extends Lint.Rules.AbstractRule {
             * \`${Casing.CamelCase}\`: File names must be camel-cased: \`fileName.ts\`.
             * \`${Casing.PascalCase}\`: File names must be Pascal-cased: \`FileName.ts\`.
             * \`${Casing.KebabCase}\`: File names must be kebab-cased: \`file-name.ts\`.
-            * \`${Casing.SnakeCase}\`: File names must be snake-cased: \`file_name.ts\`.`,
+            * \`${Casing.SnakeCase}\`: File names must be snake-cased: \`file_name.ts\`.
+
+            Or an object, where the key reprents a regular expression that
+            matches the file name, and the value is the file name rule from
+            the previous list.
+
+            * { ".tsx": ${Casing.PascalCase}, ".ts": ${Casing.CamelCase} }
+        `,
         options: {
             type: "array",
             items: {
--- a/src/rules/fileNameCasingRule.ts
+++ b/src/rules/fileNameCasingRule.ts
@@ -45,7 +45,7 @@ interface ValidatorFailureResult {
 
 type ValidatorResult = ValidatorSuccessfulResult | ValidatorFailureResult;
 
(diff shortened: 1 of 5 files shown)


yordis:
Linter is failing related to an issue I encounter, to be honest, I tried to fix it but it seems that TypeScript is the issue.

I need help on that one.

src/rules/fileNameCasingRule.ts
```diff
@@ -28,6 +28,74 @@ enum Casing {
...
+}
+
+interface ValidatorFailureResult {
+    valid: false;
+    failedCasing: Casing;
+}
+
+type ValidatorResult = ValidatorSuccessfulResult | ValidatorFailureResult;
+
+type Validator<T> = (casing: T) => (sourceFile: ts.SourceFile) => ValidatorResult;
```
JoshuaKGoldberg: > `Validator<T> = (casing: T) => ...`

If you really want to be type safe here, you could consider `Validator<T extends Config>`.

Then again, TypeScript is pretty good at inferencing local variable types. You could get rid of `Validator` and explicitly type validator methods. That seems slightly preferable to get rid of this extra line.

src/rules/fileNameCasingRule.ts
```diff
@@ -28,6 +28,74 @@ enum Casing {
...
+interface ValidatorSuccessfulResult {
+    valid: true;
+}
+
+interface ValidatorFailureResult {
+    valid: false;
+    failedCasing: Casing;
+}
+
+type ValidatorResult = ValidatorSuccessfulResult | ValidatorFailureResult;
```
JoshuaKGoldberg: Type unions are nice but you could get the same behavior with `Casing | undefined` on the validators. If a `Casing` is returned, it would indicate a failure.

src/rules/fileNameCasingRule.ts
```diff
@@ -28,6 +28,74 @@ enum Casing {
...
+        case Casing.PascalCase:
+            return isPascalCased(fileName);
+        case Casing.KebabCase:
+            return isKebabCased(fileName);
+        case Casing.SnakeCase:
+            return isSnakeCased(fileName);
+    }
+}
+
+const RegexValidator: Validator<RegexConfig> = casingConfig => (sourceFile): ValidatorResult => {
```
JoshuaKGoldberg: Naming convention for these would be pascalCase, like `regexValidator`.

Nit: consider also changing them to something like `validateWithRegex` and `validateWithSimpleConfig`. A bit more clear on what they're doing.

src/rules/fileNameCasingRule.ts
```diff
@@ -43,12 +111,27 @@ export class Rule extends Lint.Rules.AbstractRule {
...
+                        items: [
+                            {
+                                type: "string",
+                                enum: rules,
+                            },
+                        ],
+                    },
+                    {
+                        type: "object",
+                        // TODO: I am not sure how to setup the type
```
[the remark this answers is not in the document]
JoshuaKGoldberg: 🤷‍♀️ me neither... there's no validation for these _(yet: #1950)_, so it seems fine to use what you commented out. There are almost certainly dozens of other rules with invalid configs.

src/rules/fileNameCasingRule.ts
```diff
@@ -43,12 +111,27 @@ export class Rule extends Lint.Rules.AbstractRule {
...
+                        // properties: {
+                        //     type: "string",
+                        //     enum: rules,
+                        // },
+                    },
+                ],
+            },
         },
         optionExamples: [
```
JoshuaKGoldberg: Please update these and the plain-text description in the metadata to include your addition. It'd be a shame if folks didn't know about it!

src/rules/fileNameCasingRule.ts
```diff
@@ -28,6 +28,74 @@ enum Casing {
     SnakeCase = "snake-case",
 }
 
+type RegexConfig = Record<string, Casing>;
+
+type SimpleConfig = Casing;
```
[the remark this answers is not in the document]
yordis: Readability matters, By using `SimpleConfig` in the source code we know what we are referring to. I prefer to bind `Casing` to `SimpleConfig` that having `RegexConfig` (notice the naming convention of `Config` and `Casing`.

P.S: I used to have it with just `Casing`

src/rules/fileNameCasingRule.ts
```diff
@@ -28,6 +28,74 @@ enum Casing {
...
+    const fileName = path.parse(sourceFile.fileName).name;
+    const isValid = isCorrectCasing(fileName, casingConfig);
+
+    return isValid ? { valid: true } : { valid: false, failedCasing: casingConfig };
+};
+
+const validate = (sourceFile: ts.SourceFile, casingConfig: Config): ValidatorResult => {
+    const validator = typeof casingConfig === "string" ? SimpleValidator : RegexValidator;
+
+    // @ts-ignore https://github.com/Microsoft/TypeScript/issues/7294
```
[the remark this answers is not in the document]
yordis: Don't break my beautiful lambda 😭 

Yeah I did it this way because on testing (or usage) I could apply the same `Config` to different files but yeah, too fancy without needs I guess 😅

src/rules/fileNameCasingRule.ts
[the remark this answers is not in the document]
yordis: Would leave it comment it out, maybe improve the comment

src/rules/fileNameCasingRule.ts
yordis: > Then again, TypeScript is pretty good at inferencing local variable types. You could get rid of Validator and explicitly type validator methods. That seems slightly preferable to get rid of this extra line.

I am not following this, I improved the types but what do you want me to remove?!

src/rules/fileNameCasingRule.ts
yordis: 🤷‍♂️ Hopefully we do not need to return anything else from inside the validator in the future.

src/rules/fileNameCasingRule.ts
yordis: @JoshuaKGoldberg yeah sorry, I forgot to ask where do I need to update the documentation since I didn't understand the bullet point

src/rules/fileNameCasingRule.ts
JoshuaKGoldberg: You can remove the `type Validator<T> = > ...` line, and instead of this:

```typescript
const RegexValidator: Validator<RegexConfig> = casingConfig => (sourceFile): ValidatorResult => {
```

...you can do this:

```typescript
const regexValidator = (casingConfig: Casing, sourceFile: ts.SourceFile):  ValidatorResult => {
```

_(along with whatever other names / return types are changed)_

It's a little less code overall.

yordis: @JoshuaKGoldberg oh got it.

Well, I prefer to have a type that forces the validators to implement that type.

Right now, it will require us to add the same types to `simpleConfigValidator` and` regexValidator` so multiple places.

Also, I believe is more boilerplate code overall actually, since you need to add the types for all the variables and return instead of just doing `Validator<T>`

yordis:
@JoshuaKGoldberg finished the reviews except one that I need your response about it

src/rules/fileNameCasingRule.ts
JoshuaKGoldberg: > prefer to have a type

Interesting. Why? Normally, manually declared types such as these are meant for when you have to pass variables around, and need to restrict external (other file, function, etc.) constructs from allowing the wrong type. You only ever use the `Validator` functions in one place, and if you were to type them incorrectly, TypeScript would give a compiler error.

> more boilerplate code

Not really. Instead of this:

```typescript
type Validator<T extends Config> = (sourceFile: ts.SourceFile, casing: T) => ValidationResult;
const validateWithRegexConfig: Validator<RegexConfig> = (sourceFile, casingConfig) => {
const validateWithSimpleConfig: Validator<Casing> = (sourceFile, casingConfig) => {
```

You'd get this:

```typescript
const validateWithRegexConfig: (sourceFile: ts.SourceFile, casingConfig: RegexConfig) => {
const validateWithSimpleConfig: (sourceFile: ts.SourceFile, casingConfig: Casing) => {
```

...which is 2 lines instead of 3, and about the same length on the `const` lines.

This isn't blocking for the PR going in. More of a preference and a curiosity.

src/rules/fileNameCasingRule.ts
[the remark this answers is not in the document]
JoshuaKGoldberg: > if you really feel strongly about it

I do, because it was confusing reading through it and seeing multiple types referring to the same and trying to wrap my head around why there was a different before realizing there wasn't any difference 😅.

In general, TSLint rules try to favor simplicity in their implementations. If there's no reason for added complexity (extra type definitions or aliases, etc.), it shouldn't be added.

src/rules/fileNameCasingRule.ts
```diff
@@ -43,18 +101,46 @@ export class Rule extends Lint.Rules.AbstractRule {
...
         optionExamples: [
             [true, Casing.CamelCase],
             [true, Casing.PascalCase],
             [true, Casing.KebabCase],
             [true, Casing.SnakeCase],
+            [
+                true,
+                {
+                    ".tsx": "pascal-case",
+                    ".ts": "camel-case",
```
JoshuaKGoldberg: Nit: `"tsx": Casing.SnakeCase` and `".ts": Casing.CamelCase` please, per the convention in the other examples.

src/rules/fileNameCasingRule.ts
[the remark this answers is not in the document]
JoshuaKGoldberg: Similarly, would you mind also removing `ValidationResult`? It's just another level of misdirection that makes it harder to parse through later lines of code.

src/rules/fileNameCasingRule.ts
```diff
@@ -28,6 +28,64 @@ enum Casing {
...
+
+    return isValid ? undefined : casingConfig;
+};
+
+const validate = (sourceFile: ts.SourceFile, casingConfig: Config): ValidationResult => {
+    const validator =
+        typeof casingConfig === "string" ? validateWithSimpleConfig : validateWithRegexConfig;
+
+    // @ts-ignore https://github.com/Microsoft/TypeScript/issues/7294
+    return validator(sourceFile, casingConfig);
```
JoshuaKGoldberg: All right, reasoning for the TS error: it's a lack of type inference in the type system that `validator`, either when explicitly typed as a `Validator<Config>` or as the result of the ternary expression, is not allowed to be called this way. In this line, TypeScript doesn't know that `validator` takes in a simple config if `casingConfig === "string"` or a regexp config otherwise.

To be clear: your code is correct, but TypeScript doesn't have the ability to understand that.

Solution: remove the `validator` variable:

```typescript
return casingConfig === "string"
    ? validateWithSimpleConfig(sourceFile, casingConfig)
    : validateWithRegexConfig(sourceFile, casingConfig);
```

You can look for issues on TypeScript if you'd like to see if this has already been reported. Fun stuff!

JoshuaKGoldberg: _(and once you fix that, the TSLint error should go away)_

src/rules/fileNameCasingRule.ts
yordis: @JoshuaKGoldberg well, you are missing `ValidationResult` from the second example, unless you are assuming that TypeScript infers the return types for you, which fair enough but if you do that then anything that I return from those functions are valid, I am not forced to return `Casing | undefined` which is what you must return.

When my read the type alone, it has more meaning, which is the same reason why I cast one type `Casing` into another type `SimpleConfig` and why I created simple types like `ValidationResult`, 

I shaped the application design into meaniful types that helps others to read and reason about the code without having to understand the internals.

This is very beneficial when you are looking one type at the same, as you would do in your IDE/Editors. 

If you use the definition suggested, fair enough, you still have the type checking and system but the types itself are not that meaningful.

When you read `(sourceFile: ts.SourceFile, casing: T) => {};`.

- OK, a function that takes `X, and Y` and no clue about what returns.

You don't think about `Validators` since nothing tells you what that is , **except if you know the source code for sure.** or have the best naming ever 😄 

Which for people that are not familiar with the source code will have a harder time dealing with it, they will need to map too many things in their head at once for be able to understand.

The more code you have, the harder the task become.

But when you read `Validator<RegexConfig>` at least the type system is telling you what is going on for that function, the intentation of it and tracing the types you could understand many things without even looking at the source code of specific functions

Or you could create new validators as long as they follow the same shape (logically is another history).

Without that `Validator` type, you are copy-pasting code, but you do not have a defined boundery, you are hoping that the copy-pasting from the other functions are right 😅, but you have no answers of what is the corelation between them.

Why do they always expect those two params 🤷‍♂️ 

That is my thinking process when it comes to types, I used them for documentation, and a mechanism to help others to understand the code.

If you add `tsdoc` with documentation this become even more valuable.

When you document about `Validators` your documentation will be inlined with the type (the same for `SimpleConfig` bind)

Anyway, habit, as the person that will be using this, I prefer to have more meaniful types than just having type checking.

JoshuaKGoldberg:
Oh dear, I just realized there's already PR for this. Sorry @yordis, I didn't realize the linked #4220 is a duplicate of #4071. Funny how two PRs for about the same feature look so different.

I don't want to close this PR, in case the other goes in, but let's hold off on merging this in out of respect for @cheeZery having already sent one in.

src/rules/fileNameCasingRule.ts
yordis: I don't understand what you mean by `per the convention in the other examples`.

The other examples are about a single value, this is showing the example for a complex setup.

src/rules/fileNameCasingRule.ts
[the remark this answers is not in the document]
yordis: As I commented on the other comment. The type systems is being used to understand the source code without merging contexts.

When we talk about the possible configurations we would use `SimpleConfig` and when we talk about the enum we use `Casing`.

I don't see any issue at all by casting one type into another one, specially when it helps to understand the types and their usage.

src/rules/fileNameCasingRule.ts
```diff
@@ -40,21 +95,50 @@ export class Rule extends Lint.Rules.AbstractRule {
             * \`${Casing.CamelCase}\`: File names must be camel-cased: \`fileName.ts\`.
             * \`${Casing.PascalCase}\`: File names must be Pascal-cased: \`FileName.ts\`.
             * \`${Casing.KebabCase}\`: File names must be kebab-cased: \`file-name.ts\`.
-            * \`${Casing.SnakeCase}\`: File names must be snake-cased: \`file_name.ts\`.`,
+            * \`${Casing.SnakeCase}\`: File names must be snake-cased: \`file_name.ts\`.
+
+            Or an object, where the key reprents a regular expression that
```
JoshuaKGoldberg: > reprents 

Just noticed this: "represents"

yordis:
@JoshuaKGoldberg do you know how long it would take to get this feature released?

JoshuaKGoldberg:
Unfortunately no. The Palantir folks are looking into getting some to maintain TSLint full time and I don't have those rights.

PR #4327, opened 2018-11-29
Improve JUnit formatter

iMarv:
#### PR checklist

  - [x] Includes tests
- [ ] Documentation update

#### Overview of change:

This updates the JUnit formatter to use the required property `classname`.
This change is related to an issue for [Gitlab-CE](https://gitlab.com/gitlab-org/gitlab-ce/issues/50964) which has problems with displaying JUnit results if they are missing the `classname` property.

Not sure if we should add other properties which are required, but cannot be filled with a proper value like `time` or `tests` and `errors`. These could be filled with a dummy value, but I would rather leave them out.

#### CHANGELOG.md entry:
[enhancement] `Improve JUnit formatter`

### DIFF
--- a/src/formatters/junitFormatter.ts
+++ b/src/formatters/junitFormatter.ts
@@ -32,8 +32,8 @@ export class Formatter extends AbstractFormatter {
         <?xml version="1.0" encoding="utf-8"?>
         <testsuites package="tslint">
           <testsuite name="myFile.ts">
-            <testcase name="Line 1, Column 14: semicolon">
-              <failure type="warning">Missing semicolon</failure>
+            <testcase name="semicolon" classname="myFile.ts">
+              <failure type="warning">Missing semicolon Line 1, Column 14</failure>
             </testcase>
           </testsuite>
         </testsuites>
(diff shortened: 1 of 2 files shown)


palantirtech:
Thanks for your interest in [palantir/tslint](https://github.com/palantir/tslint), @iMarv! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

iMarv:
Hey @JoshuaKGoldberg , thank you for looking into it. This is the [XML-Schema](https://github.com/windyroad/JUnit-Schema/blob/master/JUnit.xsd) for JUnit.

You are right regarding an issue, would definitly create one for bigger changes

mikcsabee:
As far as I see, the major difference is the testcase name field:
`<testcase name="Line 1, Column 14: semicolon">`
vs
`<testcase name="semicolon" classname="myFile.ts">`
The reason why I used the `lineAndCharacter + rule` instead of the rule only, is because makes the testcases looks a bit unique and it gives a better overview. If you have the same error multiple times, (like 'semicolon') you have to open the failure for every single testcase to get the line and character information. Some applications are not really good to present this case.

nfm:
@iMarv I'm experimenting with using the JUnit formatter on GitLab CI (and saw you commented on [this GitLab issue](https://gitlab.com/gitlab-org/gitlab-ce/issues/50964)) but it seems to be generating absolute paths in the `classname` for me, which it looks like doesn't play nicely with the merge request summary.

I see this (note the lack of filenames!):

image

And the generated JUnit XML is:

```xml
<testsuites package="tslint"><testsuite name="/absolute/path/to/legal-document/index.tsx"><testcase name="no-console" classname="/absolute/path/to/legal-document/index.tsx"><failure type="error">Calls to 'console.log' are not allowed. Line 36, Column 5</failure></testcase></testsuite></testsuites>
```

Have you run into this, or are your name and classname values relative (and behaving in the GitLab UI)?

Thanks :heart:

iMarv:
I'll be honest @nfm, I copied the format from the related issue in [eslint](https://github.com/eslint/eslint/issues/11068) which as you said does not seem to be working correctly in gitlab. I did not catch this in time as I did not come around to add this to our pipelines.

Nevertheless, I still think that the format I added with this merge request fits the [JUnit Schema](https://github.com/windyroad/JUnit-Schema/blob/master/JUnit.xsd), which means I am going to ~~blame gitlab for parsing it incorrectly~~ will mention it in the [issue for gitlab-ce](https://gitlab.com/gitlab-org/gitlab-ce/issues/50964)

PR #4344, opened 2018-12-05
Add quotemark type validation to allow early return (#4310)

### LINKED ISSUE #4310: quotemark: Invalid regular expression: /\\)/: Unmatched ')'
### Bug Report

- __TSLint version__: 5.11.0
- __Running TSLint via__: CLI

```ts
('a)
```

```json
{ "rules": { "quotemark": true } }
```

#### Actual behavior

```
The 'quotemark' rule threw an error in 'test.ts':
SyntaxError: Invalid regular expression: /\\)/: Unmatched ')'
    at new RegExp (<anonymous>)
    at cb (...\tslint\lib\rules\quotemarkRule.js:106:33)
    ...
```

#### Expected behavior

No error.

Looks like the same issue as #3073 and #4181 but those issues got closed as there was no reproducible case.

isabelleweber:
#### PR checklist

- [x] Addresses an existing issue: fixes #4310 
    - [x] Includes tests
- [ ] Documentation update

#### Overview of change:

Add quotemark type validation to allow early return if not one of \`, `'` or `"`.

**Before:**

When linting something like `(a')`... 

`SyntaxError: Invalid regular expression: /\\)/: Unmatched ')'`

**After:**

No error

#### CHANGELOG.md entry:

[bugfix] additional `quotemark` validation


palantirtech:
Thanks for your interest in [palantir/tslint](https://github.com/palantir/tslint), @isabelleweber! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

test/rules/quotemark/double/test.ts.lint
```diff
@@ -8,6 +8,7 @@ var tabNewlineWithinSingle = 'tab\tNewline\nWithinSingle';
                              ~~~~~~~~~~~~~~~~~~~~~~~~~~~~  [' should be "]
 'escaped\'quotemark';
 ~~~~~~~~~~~~~~~~~~~~ [' should be "]
+var unbalancedSingle = ('a)
```
JoshuaKGoldberg: Adding this test case here means the rest of the file might not get linted properly, since following characters across lines will be considered part of the same (invalid) string. Plopping this file into astexplorer.net is showing that everything here is being considered a TaggedTemplateExpression.

Could you move this to a different file? Even if you moved this to the last line of this file, there's no guarantee later folks wouldn't add lines after it.

src/rules/quotemarkRule.ts
```diff
@@ -37,6 +37,10 @@ interface Options {
     avoidTemplate: boolean;
 }
 
+function isQuoteMark(value: QUOTE_MARK | string): value is QUOTE_MARK {
```
JoshuaKGoldberg: Nit: you can just put `value: string`, since `QUOTE_MARK` is a subset of `string`.

test/rules/quotemark/double/test.ts.lint
isabelleweber: That's what I assumed as well, until the tests actually pass regardless of where I list this example (at the start vs. here).

I can still move it to its own tests directory, just seemed like overkill for this one test case. Could name it something like `unbalanced-quotes`?

JoshuaKGoldberg: Sounds good! Would you mind adding tests for `"` and <code>`</code as well? On second glance I think that's valuable to test.

JoshuaKGoldberg:
👋 @isabelleweber just a heads up - I merged in `master` and resolved the test feedback. This'll be available in the next TSLint release!

Issue #4345, opened 2018-12-06
Configuration tests don't properly test rules


mogzol:
### Bug Report

- __TSLint version__: 5.11.0
- __TypeScript version__: 2.9.2
- __Running TSLint via__: N/A

#### Actual behavior
Several of the tests in `configurationTests.ts` make calls like this:
```
assertConfigEquals(config.jsRules, expectedConfig.rules);
```
This doesn't actually work, since `assert.deepEqual` doesn't work on `Map` objects. Because of this, the configs can differ and the assertion will still pass.

#### Expected behavior
When comparing actual rules to expected ones, the assertion should fail if the rules differ.

I created a fix [here](https://github.com/mogzol/tslint/commit/dc857843b181fb7ac5b710e969135dac4d755745), I'll put up a PR once I create this issue

mogzol:
Fixed with #4346

PR #4346, opened 2018-12-06
Fix configuration tests when comparing rules

mogzol:
#### PR checklist

- [x] Addresses an existing issue: #4345
  - [x] Includes tests
- [x] Documentation update (N/A)

#### Overview of change:

Several of the tests in configurationTests.ts make comparisons on config rules using assert.deepEqual, which doesn't work on Maps. Because of this, configs that differed could still pass the tests. This commit fixes that by calling demap on the maps before the comparisons.

This is strictly fixing the tests. One quirk I noticed is boolean configurations result in `[]` as the rule arguments, while normal configurations result in `undefined`. If this behavior is not desired, it should probably be fixed in a separate PR.

### DIFF
--- a/test/configurationTests.ts
+++ b/test/configurationTests.ts
@@ -428,53 +428,59 @@ describe("Configuration", () => {
 
         it("extends with package", () => {
             const config = loadConfigurationFromPath("./test/config/tslint-extends-package.json");
-            const expectedConfig = getEmptyConfig();
-            expectedConfig.rules.set("rule-one", { ruleSeverity: "error" });
-            expectedConfig.rules.set("rule-two", { ruleSeverity: "off" });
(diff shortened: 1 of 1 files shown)


palantirtech:
Thanks for your interest in [palantir/tslint](https://github.com/palantir/tslint), @mogzol! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

JoshuaKGoldberg:
Nice catch, thanks for this @mogzol!

> If this behavior is not desired, it should probably be fixed in a separate PR.

Yeah the difference seems odd, likely unintended. Would you mind filing an issue for that please? Seems like an interesting quirk to discuss.

PR #4351, opened 2018-12-07
Fix semicolon: Cannot read property 'initializer' of undefined

### LINKED ISSUE #4309: semicolon: Cannot read property 'initializer' of undefined
### Bug Report

- __TSLint version__: 5.11.0
- __Running TSLint via__: CLI

```ts
const
```

```json
{ "rules": { "semicolon": [true, "never"] } }
```

#### Actual behavior

```
The 'semicolon' rule threw an error in 'test.ts':
TypeError: Cannot read property 'initializer' of undefined
    at SemicolonNeverWalker.checkVariableStatement (...\tslint\lib\rules\semicolonRule.js:309:51)
    ...
```

#### Expected behavior

No error. The code should handle the case where `declarations` is an empty array.

rrogowski:
#### PR checklist

- [x] Addresses an existing issue: #4309 
  - [x] Includes tests
- [ ] Documentation update

#### Overview of change:

`SemicolonNeverWalker.checkVariableStatement` handles the case where `declarations` is an empty array.

1. Is there any documentation that needs to be updated as a result of this change?
2. Unsure if better to export one of the existing utility functions:
  - `last` (`memberOrderingRule.ts`)
  - `arrayLast` (`switchFinalBreakRule.ts`).

### DIFF
--- a/src/rules/semicolonRule.ts
+++ b/src/rules/semicolonRule.ts
@@ -357,8 +357,8 @@ class SemicolonNeverWalker extends SemicolonWalker {
     }
 
     private checkVariableStatement(node: ts.VariableStatement) {
-        const declarations = node.declarationList.declarations;
-        if (declarations[declarations.length - 1].initializer === undefined) {
+        const declaration = last(node.declarationList.declarations);
+        if (declaration === undefined || declaration.initializer === undefined) {
             // variable declaration does not continue on the next line if it has no initializer
             return this.checkSemicolonOrLineBreak(node);
         }
@@ -377,4 +377,11 @@ class SemicolonNeverWalker extends SemicolonWalker {
             this.checkSemicolonOrLineBreak(member);
         }
     }
+}
+
+/**
+ * Returns the last element of an array (or undefined if the array is empty).
+ */
+function last<T>(array: ArrayLike<T>): T | undefined {
+    return array[array.length - 1];
 }
--- a/src/rules/semicolonRule.ts
+++ b/src/rules/semicolonRule.ts
@@ -382,6 +382,6 @@ class SemicolonNeverWalker extends SemicolonWalker {
 /**
  * Returns the last element of an array (or undefined if the array is empty).
  */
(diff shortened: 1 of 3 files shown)


palantirtech:
Thanks for your interest in [palantir/tslint](https://github.com/palantir/tslint), @rrogowski! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

src/rules/semicolonRule.ts
```diff
@@ -357,8 +357,8 @@ class SemicolonNeverWalker extends SemicolonWalker {
     }
 
     private checkVariableStatement(node: ts.VariableStatement) {
-        const declarations = node.declarationList.declarations;
-        if (declarations[declarations.length - 1].initializer === undefined) {
+        const declaration = last(node.declarationList.declarations);
+        if (declaration === undefined || declaration.initializer === undefined) {
```
JoshuaKGoldberg: Declaring a new `last<T>` is a bit of overhead for this. How about:

```typescript
if (declarations.length !== 0 && declarations[declarations.length - 1].initializer === undefined)
```

test/rules/semicolon/never/test.ts.lint
```diff
@@ -269,4 +269,7 @@ var foo = () => {};
 var bar; var foo = function() {};
 []
 
+const
```
JoshuaKGoldberg: Let's also check `var` and `let` on their own.

src/rules/semicolonRule.ts
JoshuaKGoldberg: 🙂 not particularly. [`tslint.json`](https://github.com/palantir/tslint/blob/master/tslint.json) has it set to 140 characters, so this shouldn't trigger it.

JoshuaKGoldberg:
Looks great, thanks @rrogowski! Congrats on the first contribution here! 🎉

PR #4362, opened 2018-12-11
Allow unknown in 'no-object-literal-type-assertion' rule.

### LINKED ISSUE #4355: no-object-literal-type-assertion should allow unknown type
### Feature request

**Is your feature request related to a problem? Please describe.**
There is 'as any' exception for the rule. #2716
With the new unknown type there should be exception for 'as unknown' as well.

**Describe the solution you'd like**
Add exception for 'as unknown'.

NN---:
Fix #4355

### DIFF
--- a/src/rules/noObjectLiteralTypeAssertionRule.ts
+++ b/src/rules/noObjectLiteralTypeAssertionRule.ts
@@ -61,7 +61,7 @@ function walk(ctx: Lint.WalkContext<void>): void {
             // Compare with UnknownKeyword if using TS 3.0 or above
             (!!(ts.SyntaxKind as any).UnknownKeyword ?
              (node.type.kind !== (ts.SyntaxKind as any).UnknownKeyword) :
-             true) &&
+             node.type.getText(ctx.sourceFile) !== "unknown") &&
             isObjectLiteralExpression(
                 isParenthesizedExpression(node.expression)
                     ? node.expression.expression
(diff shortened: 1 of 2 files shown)


palantirtech:
Thanks for your interest in [palantir/tslint](https://github.com/palantir/tslint), @NN---! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

src/rules/noObjectLiteralTypeAssertionRule.ts
```diff
@@ -58,6 +58,10 @@ function walk(ctx: Lint.WalkContext<void>): void {
         if (
             isAssertionExpression(node) &&
             node.type.kind !== ts.SyntaxKind.AnyKeyword &&
+            // Compare with UnknownKeyword if using TS 3.0 or above
+            (!!(ts.SyntaxKind as any).UnknownKeyword ?
+             (node.type.kind !== (ts.SyntaxKind as any).UnknownKeyword) :
+             true) &&
```
JoshuaKGoldberg: @NN--- if you change the `true` to `node.type.getText(ctx.sourceFile) !== "unknown"` the `<unknown> {};` line will be properly linted, so you can remove the expected error from it.

NN---: Just to make sure, in TS 2 there is no unknown keyword and this rule can be abused.
I think these cases are rare.

```ts
type unknown = {};
{} as unknown // Not what we wanted.
```

NN---: My point is that you could make anything in TS 2.0 as unknown is not a keyword.
Perhaps not a clever idea but still :)

JoshuaKGoldberg: Yes understood - and I think that's ok for this rule because, as you noted, it's not a clever idea.

PR #4383, opened 2018-12-15
Fix [ban]: wrong logic matching deeply nested methods

zhouhaoyu:
#### PR checklist

  - [x] Includes tests
- [ ] Documentation update

#### Overview of change:

Fix a bug where `a.b.c()` may also match with `a.c()`.

#### CHANGELOG.md entry:

[bugfix] `ban`: Fix a false positive which would occur when banning method calls nested inside objects. Previously, banning `["a", "b", "c"]` would trigger lint failures on the syntax `b.c()`, which was not the intent of this rule.

### DIFF
--- a/src/rules/banRule.ts
+++ b/src/rules/banRule.ts
@@ -94,6 +94,7 @@ export class Rule extends Lint.Rules.AbstractRule {
                 { name: ["it", "only"], message: "don't focus tests" },
                 { name: ["chai", "assert", "equal"], message: "Use 'strictEqual' instead." },
                 { name: ["*", "forEach"], message: "Use a regular for loop instead." },
+                { name: ["*", "_id", "toString"], message: "Use 'toHexString' instead." },
             ],
         ],
         type: "functionality",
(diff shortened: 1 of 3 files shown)


palantirtech:
Thanks for your interest in [palantir/tslint](https://github.com/palantir/tslint), @zhouhaoyu! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

JoshuaKGoldberg:
Note: this would be a breaking change, as some users may be accidentally relying on the old _(technically incorrect)_ behavior. This PR will have to wait until TSLint is ready to merge in PRs for a next major version.

Generally it'd be better to have an issue filed before a PR for changes, but this seems small enough to discuss here. Great catch @zhouhaoyu!

alexluecke:
Any idea when this will be released?

adidahiya:
We usually treat fixes to rule behavior like this as non-breaking, even though users might be relying on the old faulty behavior. I just want to confirm that the false positive existed before this change and then I'll merge this.

PR #4404, opened 2018-12-20
Revert breaking change to tslint:recommended, update tslint:latest

adidahiya:
Addresses the issue I brought up in https://github.com/palantir/tslint/pull/4198#issuecomment-449134147

Also enables some v5.12 rules in `tslint:latest`


johnwiseheart:
Seems reasonable. Do you think its worth releasing a patch including this?

adidahiya:
@johnwiseheart yes

Issue #4402, opened 2018-12-19
Getting warning ` should be ` (backtick should be backtick)


bnfrnz:
### Bug Report

- __TSLint version__: 5.12.0
- __TypeScript version__: 2.6.2
- __Running TSLint via__:  CLI (via ionic-app-scripts)

#### TypeScript code being linted

```ts
this.toastCtrl.present(`Please tick "Yes, I'm sure" to confirm`);
```
Note that all three ` and " and ' are being used for the string.

with `tslint.json` configuration:

```json
"quotemark": [true, "single", "avoid-escape", "avoid-template"]
```

#### Actual behavior
Getting warning: \` should be \`
Note that both of these are backticks. So it asks me to replace a backtick with a backtick. This makes no sense to me. Either I'm missing something here or this is a bug.

#### Expected behavior
Expected no warning. Never got a warning. This just started appearing I after updated tslint yesterday.

bnfrnz:
I tried StackBlitz but can't seem to get tslint to work. Not sure if I'm too thick but as soon as I add tslint the repo stops working. Which other cloud tool is good to share the minimal project?

This is my local vscode:

![screen shot 2019-01-10 at 13 35 16](https://user-images.githubusercontent.com/26401091/50948215-aa0f4980-14dc-11e9-8912-963692ff2673.png)

Same warning appears in cli...

adidahiya:
[...]

PR #4408, opened 2018-12-22
Fix quotemark avoid-template issues

johnwiseheart:
#### PR checklist

- [X] Addresses an existing issue: #4402
  - [ ] Includes tests

#### Overview of change:

Add an additonal check to make sure we never render a warning of the format `" should be "` or ``` ` should be ` ```.

Whats the best test for this? I couldn't find a way to write a test that failed before making this fix, but I feel like we shouldn't merge this without one.

#### CHANGELOG.md entry:

[bugfix] fix some `quotemark` issues when using backticks with the `avoid-template` option

### DIFF
--- a/src/rules/quotemarkRule.ts
+++ b/src/rules/quotemarkRule.ts
@@ -27,12 +27,12 @@ const OPTION_JSX_DOUBLE = "jsx-double";
 const OPTION_AVOID_TEMPLATE = "avoid-template";
 const OPTION_AVOID_ESCAPE = "avoid-escape";
 
-type QUOTE_MARK = "'" | '"' | "`";
-type JSX_QUOTE_MARK = "'" | '"';
+type QUOTEMARK = "'" | '"' | "`";
+type JSX_QUOTEMARK = "'" | '"';
 
 interface Options {
-    quoteMark: QUOTE_MARK;
-    jsxQuoteMark: JSX_QUOTE_MARK;
+    quotemark: QUOTEMARK;
+    jsxQuotemark: JSX_QUOTEMARK;
     avoidEscape: boolean;
     avoidTemplate: boolean;
 }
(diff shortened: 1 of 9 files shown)


src/rules/quotemarkRule.ts
```diff
@@ -155,6 +155,10 @@ function walk(ctx: Lint.WalkContext<Options>) {
                 }
             }
 
+            if (actualQuoteMark === fixQuoteMark) {
```
JoshuaKGoldberg: Does the `this.toastCtrl.present(`Please tick "Yes, I'm sure" to confirm`);` from the issue not work as a test?

adidahiya:
the fix in the first commit doesn't work; I added some better test cases. there are issues around the rule behavior when both `avoid-escape` and `avoid-template` are enabled. I'll fix this up

src/rules/quotemarkRule.ts
```diff
@@ -27,12 +27,12 @@ const OPTION_JSX_DOUBLE = "jsx-double";
 const OPTION_AVOID_TEMPLATE = "avoid-template";
 const OPTION_AVOID_ESCAPE = "avoid-escape";
 
-type QUOTE_MARK = "'" | '"' | "`";
-type JSX_QUOTE_MARK = "'" | '"';
+type QUOTEMARK = "'" | '"' | "`";
```
adidahiya: I renamed "quote mark" to "quotemark" everywhere to be consistent

src/rules/quotemarkRule.ts
```diff
@@ -114,108 +115,98 @@ function walk(ctx: Lint.WalkContext<Options>) {
...
+                Rule.FAILURE_STRING(actualQuotemark, fixQuotemark),
+                new Lint.Replacement(start, node.end - start, fixQuotemark + text + fixQuotemark),
             );
         }
         ts.forEachChild(node, cb);
     });
 }
 
-function getQuotemarkPreference(args: any[]): QUOTE_MARK {
-    type QUOTE_PREF = typeof OPTION_SINGLE | typeof OPTION_DOUBLE | typeof OPTION_BACKTICK;
```
adidahiya: also simplified these two helper functions to reduce duplication and the extra typedefs

adidahiya:
thanks @JoshuaKGoldberg

PR #4413, opened 2018-12-26
Mark RuleWalker and ProgramAwareRuleWalker as deprecated

### LINKED ISSUE #4298: Mark RuleWalker, ProgramAwareRuleWalker, and ScopeAwareRuleWalker as deprecated
Recapping #2522: `RuleWalker`-based rules are slow, so it's generally preferable to use `applyWithFunction` instead of `applyWithWalker`.

if folks shouldn't use `RuleWalker`, it would be good to actively encourage new PRs not to. TSLint could mark them as `@deprecated` in their doc comments and put `// tslint:disable-next-line:deprecation` with links to issue(s) tracking converting them to `walk` functions.

Edit: ~~Two~~Three thumbs up from over a month - marking as accepting PRs 😊

Nazanin1369:
Fixes #4298

#### PR checklist

- [ x] Addresses an existing issue: #4298
- [ ] Includes tests
- [ ] Documentation update

### DIFF
--- a/src/language/walker/programAwareRuleWalker.ts
+++ b/src/language/walker/programAwareRuleWalker.ts
@@ -20,6 +20,7 @@ import * as ts from "typescript";
 import { IOptions } from "../rule/rule";
 import { RuleWalker } from "./ruleWalker";
 
+// tslint:disable-next-line:deprecation
 export class ProgramAwareRuleWalker extends RuleWalker {
     private readonly typeChecker: ts.TypeChecker;
 
--- a/src/language/walker/programAwareRuleWalker.ts
+++ b/src/language/walker/programAwareRuleWalker.ts
@@ -26,7 +26,7 @@ import { RuleWalker } from "./ruleWalker";
  * so it's generally preferable to use applyWithFunction instead of applyWithWalker.
  * @see https://github.com/palantir/tslint/issues/2522
  */
-export class ProgramAwareRuleWalker extends RuleWalker {
+export class ProgramAwareRuleWalker extends RuleWalker { // tslint:disable-line deprecation
     private readonly typeChecker: ts.TypeChecker;
 
     constructor(
--- a/src/language/walker/programAwareRuleWalker.ts
+++ b/src/language/walker/programAwareRuleWalker.ts
@@ -20,7 +20,12 @@ import * as ts from "typescript";
 import { IOptions } from "../rule/rule";
 import { RuleWalker } from "./ruleWalker";
 
-// tslint:disable-next-line:deprecation
+/**
+ * @deprecated
+ * RuleWalker-based rules are slow,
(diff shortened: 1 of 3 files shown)


palantirtech:
Thanks for your interest in [palantir/tslint](https://github.com/palantir/tslint), @Nazanin1369! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

src/language/walker/programAwareRuleWalker.ts
```diff
@@ -20,6 +20,7 @@ import * as ts from "typescript";
 import { IOptions } from "../rule/rule";
 import { RuleWalker } from "./ruleWalker";
 
+// tslint:disable-next-line:deprecation
 export class ProgramAwareRuleWalker extends RuleWalker {
```
JoshuaKGoldberg: Oh, I'd forgotten this exists! 😄

Could you please also mark this as deprecated @Nazanin1369? Its runtime works the same as `RuleWalker` _(but with the type checker and program)_, so it should be deprecated for the same reasons. I'll update the linked issue.

Nazanin1369: Sure

Nazanin1369: scopeAwareRuleWalker.ts already has the deprecated annotation.

Nazanin1369:
@JoshuaKGoldberg  Looks like this is a breaking change with TS 3.0+. Do I need to change anything here?

JoshuaKGoldberg:
Good catch - yes 😄

adidahiya:
I don't consider this a breaking change, it's just a deprecation. Users might get errors with the `deprecation` lint rule but that's ok. What do you mean about TS 3.0+?

JoshuaKGoldberg:
@adidahiya interesting, is it not? I would have thought build breaks from new `deprecation` rule violations to be considered breaking?

At any rate, I'd interpreted "TS 3.0+" to refer to the next major version of TSLint.

adidahiya:
@JoshuaKGoldberg I see how it might be treated as breaking, but I've generally developed libraries without the constraint that deprecations can only occur at a major version change -- instead, they can occur at any minor version with the indication that the functionality will be removed in the next major version. At least, that's how we deprecate things in our internal libraries at Palantir. I find that TSLint sometimes introduces slightly stricter checks for rules across minor versions, so, in practice, users will sometimes see build breaks upon upgrading if they have lint failures reported as errors instead of warnings. You might argue that we shouldn't be afraid of major version bumps, we could just release 6.0 with this "break", move on to 7.0, etc, but given the current release cadence of TSLint, I'm fine with merging this change in the 5.x range.

PR #4420, opened 2018-12-29
Add grouped imports by default in tslint:all

VincentLanglet:
#### Overview of change:

The tslint:all config seems to be used in order to activate all the rules and to be the strictest possible.
`"grouped-imports":true` is stricter than `"grouped-imports":false`

#### CHANGELOG.md entry:

[enhancement] add grouped import in the all config


VincentLanglet:
BTW, since ordered imports is used in tslint:recommended.
Should we add the grouped imports option ?

test/ruleLoaderTests.ts
```diff
@@ -18,9 +18,11 @@
 import { assert } from "chai";
 import * as fs from "fs";
 import * as path from "path";
+
 import { rules as allRules, RULES_EXCLUDED_FROM_ALL_CONFIG } from "../src/configs/all";
+import { IOptions } from "../src/language/rule/rule";
 import { camelize } from "../src/utils";
-import { IOptions } from "./../src/language/rule/rule";
+
```
adidahiya: was this newline added by the grouped-imports option? what's the logic here? it seems to separate the different levels of imports into groups in some places, but not in others (e.g. in `exclusionFactory.ts`, `../../` imports are in the same group as `../` imports). I haven't been following this rule closely, but that behavior is a little surprising to me and I'd like to revisit it before enabling this rule in `tslint:all`.

VincentLanglet: Yes. There is 3 différents levels: libraries, parents folder and same folder.
```
import something from "module"

import foo from "../../foo"
import bar from "../bar"

import baz from "./baz"
```

adidahiya:
ok, the documentation for `"grouped-imports"` is pretty light, so I'd like to see it expanded before merging this, at the very least for other TSLint contributors so they know what to expect from this rule when working on the TSLint code base.

VincentLanglet:
@adidahiya 
The first time I read this rules I found it pretty clear, there is a similar rule for [eslint](https://github.com/benmosher/eslint-plugin-import/blob/master/docs/rules/order.md). I can add a commit with more explanation for the rule, but what did you not understood with the existing following explanation ?

> you can enforce a grouping of third-party, parent directories and the current directory.

> Group source imports by "bar", "../baz", "./foo".

adidahiya:
with #4134 merged, the docs are now in a better place. I'll merge this and resolve any conflicts on master

Issue #4089, opened 2018-08-01
quotemark rule should not apply to JSON files


JamesMessinger:
### Bug Report

- __TSLint version__:  5.11.0
- __TypeScript version__:  3.0.1
- __Running TSLint via__: CLI

#### TypeScript code being linted

**sample.ts**
```ts
// TypeScript now allows you to import JSON files
import data from 'data.json';
console.log(data)
```

**data.json**
```json
{
    "foo": "bar"
}
```

**tslint.yaml**
```yaml
rules:
  quotemark:
    - true
    - single
```

#### Actual behavior

When running `tslint --fix`, the double quotes in the `data.json` file are replaced with single-quotes due to the `quotemark` rule.  But single-quotes are not valid in JSON files.  Only double quotes are valid JSON.

#### Expected behavior

The `quotemark` rule should not apply to JSON files.

adidahiya:
Let's skip linting JSON files altogether rather than address this issue for only one rule.  #4426

adidahiya:
Just to clarify, you are importing JSON files using [the TS 2.9+ `--resolveJsonModule` option](https://github.com/Microsoft/TypeScript/wiki/What%27s-new-in-TypeScript#new---resolvejsonmodule), right?

JamesMessinger:
Yes, that's correct.

adidahiya:
should be fixed in v5.12.1, which no longer tries to lint JSON files at all. let me know if that works for you

JamesMessinger:
Awesome!  thanks!

PR #4423, opened 2019-01-01
quotemark rule does not apply to JSON files

aprilrd:
#### PR checklist

- [x] Addresses an existing issue: #4089
  - [x] Includes tests
- [x] Documentation update

#### Overview of change:

quotemark rule does not apply to JSON files

I am not sure if using filename is the best way to detect json files.

#### CHANGELOG.md entry:

[bugfix] quotemark rule does not apply to JSON files


src/rules/quotemarkRule.ts
```diff
@@ -91,6 +93,15 @@ export class Rule extends Lint.Rules.AbstractRule {
         const quoteMark = getQuotemarkPreference(args);
         const jsxQuoteMark = getJSXQuotemarkPreference(args);
 
+        if (sourceFile.fileName.endsWith(".json")) {
+            return this.applyWithFunction(sourceFile, walk, {
+                avoidEscape: true,
+                avoidTemplate: true,
+                jsxQuoteMark: '"',
+                quoteMark: '"',
+            });
```
JoshuaKGoldberg: Shouldn't TypeScript already give errors for invalid JSON? Can the rule skip these files altogether?

I'm also wondering if all rules should skip JSON files, but that's a question for another issue...

aprilrd: Created a separate issue for all rules: https://github.com/palantir/tslint/issues/4426

JoshuaKGoldberg: Thanks!

adidahiya:
I'd rather fix https://github.com/palantir/tslint/issues/4426 (skip all JSON file linting) than implement this band aid. I will try to get that in for v5.12.1.

PR #4444, opened 2019-01-07
Fix strict-type-predicate for unknown

### LINKED ISSUE #4107: unknown and strict-type-predicate
### Bug Report

- __TSLint version__: 5.11.0
- __TypeScript version__: 3.0.1
- __Running TSLint via__: CLI

```ts
const body: unknown = 'test';
if (typeof body === 'object') // Expression is always true.
  console.log('a');
```

with `tslint.yaml` configuration:

```yaml
---
rules:
  strict-type-predicates: true
```

#### Actual behavior
tslint falsely warns that the expression is always true.

#### Expected behavior
Since the object is of type `unknown` (and actually a `string`), the expression could be both true and false and thus tslint should not warn.

ThomasdenH:
#### PR checklist

- [x] Addresses an existing issue: Closes #4107
  - [x] Includes tests
- [ ] Documentation update: N/a

#### Overview of change:
Fixes the issue where a comparison with unknown would falsly give an error. Instead, comparison with unknown should always be allowed.

#### CHANGELOG.md entry:

[bugfix] fix `strict-type-predicate` with `unknown`

### DIFF
--- a/src/rules/strictTypePredicatesRule.ts
+++ b/src/rules/strictTypePredicatesRule.ts
@@ -244,8 +244,6 @@ function getTypePredicateForKind(kind: string): Predicate | undefined {
             return flagPredicate(ts.TypeFlags.ESSymbol);
         case "function":
             return isFunction;
-        case "unknown":
-            return flagPredicate(ts.TypeFlags.Unknown);
         case "object":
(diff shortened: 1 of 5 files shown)


ThomasdenH:
Does a maintainer have time to review this?

PR #4469, opened 2019-01-19
[new-rule-option] [switch-default] Add `allow-exhaustive` option

mikeyhew:
#### PR checklist

- [x] Addresses an existing issue: #2104
  - [x] Includes tests
- [x] Documentation update

#### Overview of change:

This builds upon #3370, and adds a new option to the "switch-default" rule to allow you to omit the default clause, as long as all possible cases have been handled. If there are cases missing, the error message will say which cases are missing. This option requires type information, and if there is no type checker available, the rule is not be applied and a warning is printed.

The exhaustiveness checking works as long as the switch expression is a union of the following types: `string` literal, `number` literal, `bigint` literal, `true`, `false`, `boolean`, `null`, or `undefined`. If any other types are included, it falls back to the existing behaviour and requires a `default` case.

A possible extension would be to add another option that reports a failure if there is an unnecessary `default` clause.

- decide if you're OK using `IntrinsicType.intrinsicName` to tell the difference between `true` and `false`, even though it's marked `@internal`.
- this is my first PR in any TypeScript project, so I would appreciate any feedback
- there's lots of code, good luck!

#### CHANGELOG.md entry:

[new-rule-option] `switch-default` Add `allow-exhaustive` option


mikeyhew:
@JoshuaKGoldberg I imported the functions from `tsutils` as you suggested. It only works on TypeScript >= 2.4 because the `value` field on `ts.LiteralType` had a different name in earlier versions. Is that OK with you?

I added a check when the rule is applied that throws an error if the version number is < 2.4 and the option is enabled. I can change it to a warning or remove the check altogether if you want.

JoshuaKGoldberg:
> because the `value` field on `ts.LiteralType` had a different name in earlier versions

Huh, spooky. Good catch!

TSLint still has to support some TypeScript versions < 2.4 unfortunately. You should be able to check if the `value` field exists, and if it doesn't, try the previous field name.

mikeyhew:
@JoshuaKGoldberg  Does support for old versions mean that even new rules and rule options have to be supported? I assumed we just had to avoid breaking TSLint for existing configurations.

Anyway I added the fix you suggested, so the new rule option should work with TypeScript > 2.0

src/rules/switchDefaultRule.ts
```diff
@@ -15,41 +15,216 @@
...
+    // will be the same, too).
+    return possibleValues.filter(value => caseValues.indexOf(value) === -1);
+}
+
+// ts.PseudoBigInt was added in Typescript 3.2, and is a possible `LiteralType` value.
+// Since it is not available in earlier versions, we copy its definition and use it here.
+interface PseudoBigInt {
+    negative: boolean;
+    base10Value: string;
+}
```
mikeyhew: As far as I can tell, tslint is always compiled using the version of typescript listed under `devDependencies` in package.json. So if we upgrade that dependency to typescript 3.2, I can remove this definition and use `ts.PseudoBigInt` directly.

JoshuaKGoldberg: SGTM!

src/rules/switchDefaultRule.ts
```diff
@@ -15,41 +15,216 @@
...
     /* tslint:enable:object-literal-sort-keys */
 
-    public static FAILURE_STRING = "Switch statement should include a 'default' case";
+    public static FAILURE_STRING = "Switch statement should include a `default` case";
+
+    public static failureStringForMissingCases(cases: SingletonValue[]): string {
+        let casesString: string;
+
+        if (cases.length === 0) {
+            throw Error("There should be no failure if there are no missing cases");
```
JoshuaKGoldberg: If this is never able to happen, could you leave this error throwing out?

src/rules/switchDefaultRule.ts
```diff
@@ -15,41 +15,216 @@
...
+export class Rule extends Lint.Rules.OptionallyTypedRule {
     /* tslint:disable:object-literal-sort-keys */
     public static metadata: Lint.IRuleMetadata = {
         ruleName: "switch-default",
         description: "Require a `default` case in all `switch` statements.",
-        optionsDescription: "Not configurable.",
-        options: null,
-        optionExamples: [true],
+        optionsDescription: Lint.Utils.dedent`
+            One argument may be optionally provided:
```
JoshuaKGoldberg: TSLint is moving towards having options provided in objects, which makes it easier for rules to move towards more advanced configurations. See `trailingCommaRule.ts` for an example.

src/rules/switchDefaultRule.ts
```diff
@@ -15,41 +15,216 @@
...
+        // ts.IntrinsicType is marked as @internal, but AFAIK this is the only way to tell the
+        // difference between `true` and `false`.
+        const name = (type as ts.IntrinsicType).intrinsicName;
+        switch (name) {
+            case "true":
+                return { value: true };
+            case "false":
+                return { value: false };
+            default:
+                throw Error(
```
JoshuaKGoldberg: Nit: should be `throw new Error` (here and another place or two)

src/rules/switchDefaultRule.ts
```diff
@@ -15,41 +15,216 @@
...
+type SingletonValue = string | number | PseudoBigInt | true | false | null | undefined;
+
+// if `type` is a singleton type (a literal, `true`, `false`, `null`, or `undefined`),
+// returns an object whose `value` field is the single value that occupies that type.
+// Otherwise, returns `undefined`.
+// The object wrapper is to distinguish between `undefined` and `{value: undefined}`.
+function singletonValueOfType(type: ts.Type): { value: SingletonValue } | undefined {
+    if (isTypeFlagSet(type, ts.TypeFlags.Null)) {
+        return { value: null };
+    } else if (isTypeFlagSet(type, ts.TypeFlags.Undefined)) {
```
JoshuaKGoldberg: Nit: you can remove the `else` if the preceding `if` block `return`s. Makes it a bit easier to read 🙂

src/rules/switchDefaultRule.ts
```diff
@@ -15,41 +15,216 @@
...
+                } else if (uncovered.length > 0) {
+                    ctx.addFailureAtNode(node, Rule.failureStringForMissingCases(uncovered));
+                }
+            }
         }
         return ts.forEachChild(node, cb);
     });
 }
+
+// returns a (possibly empty) array of missing values, or the string 'non-exhaustive' if the
```
JoshuaKGoldberg: Nit: should be JSDoc-formatted `/* @returns` so that IDEs can surface this in hover information for the method. `//` comments won't be (at least in VS Code).

src/rules/switchDefaultRule.ts
```diff
@@ -15,41 +15,216 @@
...
+    if (isTypeFlagSet(type, ts.TypeFlags.Null)) {
+        return { value: null };
+    } else if (isTypeFlagSet(type, ts.TypeFlags.Undefined)) {
+        return { value: undefined };
+    } else if (isTypeFlagSet(type, ts.TypeFlags.BooleanLiteral)) {
+        // ts.IntrinsicType is marked as @internal, but AFAIK this is the only way to tell the
+        // difference between `true` and `false`.
+        const name = (type as ts.IntrinsicType).intrinsicName;
+        switch (name) {
+            case "true":
```
JoshuaKGoldberg: Nit/suggestion: instead of this larger switch statement, use:

```ts
return { value: name === "true" };
```

...since we know it's a boolean literal and the name is either `"true"` or `"false"`

test/rules/switch-default/allow-exhaustive/test.ts.lint
```diff
@@ -0,0 +1,79 @@
...
+~~~~~~~~~~~~~~~~~~~
+    case 1: return;
+~~~~~~~~~~~~~~~~~~~
+  }
+~~~ [Switch statement should include a `default` case]
+}
+
+type LotsOfCases = 'foo' | 'bar' | 1 | true | false | null | undefined;
+
+function missingTwoCases(value: LotsOfCases) {
```
JoshuaKGoldberg: Missing a test case on `bigint` (once you update the dev dependency).

See `no-submodule-imports/dynamic-imports`' use of `[typescript]: >=2.4.0` for an existing separate set of tests that require a minimum TS version.

JoshuaKGoldberg:
Additionally, how should this handle `null` and `undefined` in non-strict mode? Should the `allow-exhaustive` flag also require strict null checking?

```ts
let value: "red" | "blue" = null;

switch (value) {
    case "red": return 0;
    case "blue": return 1;
}
```

JoshuaKGoldberg:
Closing this PR for housekeeping, given the lack of activity since January. Please do comment here or send a new PR if you'd to push this forward!

Note: #4534 tracks the deprecation of TSLint overall.

Issue #3443, opened 2017-11-02
2 tests fail on just cloned repo


M0ns1gn0r:
### Bug Report

- __TSLint version__: master d1caf11
- __TypeScript version__: 2.6.1
- __Running TSLint via__: CLI (yarn test)

#### Actual behavior
I cloned the repo and followed the instructions in the readme, i.e. executed `yarn compile` and then `yarn test`. Two tests keep failing:

      ...

	  137 passing (58s)
	  2 failing

	  1) Executable Files exits with code 0 if correct file is passed:
		 Uncaught AssertionError: process should exit without an error: expected [Error: Command failed: X:\tslint\test\executable\npm-like-executable.cmd src/configuration.ts
	Warning: The 'promise-function-async' rule requires type information.
	Warning: The 'await-promise' rule requires type information.
	Warning: The 'no-floating-promises' rule requires type information.
	Warning: The 'no-for-in-array' rule requires type information.
	Warning: The 'no-inferred-empty-object-type' rule requires type information.
	Warning: The 'no-unsafe-any' rule requires type information.
	Warning: The 'restrict-plus-operands' rule requires type information.
	Warning: The 'strict-boolean-expressions' rule requires type information.
	Warning: The 'use-default-type-parameter' rule requires type information.
	Warning: The 'deprecation' rule requires type information.
	Warning: The 'match-default-export-name' rule requires type information.
	Warning: The 'no-boolean-literal-compare' rule requires type information.
	Warning: The 'no-unnecessary-type-assertion' rule requires type information.
	Warning: The 'return-undefined' rule requires type information.
	] to equal null
		  at Function.assert.isNull (node_modules\chai\lib\chai\interface\assert.js:407:32)
		  at build\test\executable\executableTests.js:52:31
		  at build\test\executable\executableTests.js:482:9
		  at ChildProcess.exithandler (child_process.js:213:5)
		  at maybeClose (internal/child_process.js:877:16)
		  at Process.ChildProcess._handle.onexit (internal/child_process.js:226:5)

	  2) Executable Files exits with code 0 if several files passed without `-f` flag:
		 Uncaught AssertionError: process should exit without an error: expected [Error: Command failed: X:\tslint\test\executable\npm-like-executable.cmd src/configuration.ts src/formatterLoader.ts
	Warning: The 'promise-function-async' rule requires type information.
	Warning: The 'await-promise' rule requires type information.
	Warning: The 'no-floating-promises' rule requires type information.
[...]

PR #4470, opened 2019-01-19
Added no-tautology-expressions rule

noamyogev84:
#### PR checklist

- [x] Addresses an existing issue: #803 
- [x] New feature
  - [x] Includes tests
- [ ] Documentation update

#### Overview of change:
Add a basic rule that validates that two equal literals are not being compared.


palantirtech:
Thanks for your interest in [palantir/tslint](https://github.com/palantir/tslint), @noamyogev84! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

src/rules/noTautologyExpressionRule.ts
```diff
@@ -0,0 +1,58 @@
...
+        return this.applyWithFunction(sourceFile, walk);
+    }
+}
+
+function walk(context: Lint.WalkContext<void>) {
+    const cb = (node: ts.Node): void => {
+        if (tsutils.isBinaryExpression(node)) {
+            if (isLiteral(node.left) && isLiteral(node.right)) {
+                // Simple textual comparison of both sides
+                if (node.left.getText() === node.right.getText()) {
```
JoshuaKGoldberg: You should be able to use `.text` since you've determined they're literals. On the `isLiteral` function, have it return `node is ts.StringLiteral | ts.isNumericLiteral` or some equivalent type.

Note that won't hold true if you start also being able to check things like `obj1.foo === obj1.foo`, but for literals it should.

src/rules/noTautologyExpressionRule.ts
```diff
@@ -0,0 +1,58 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import * as tsutils from "tsutils";
+import * as ts from "typescript";
+
+import * as Lint from "../index";
+
+const TAUTOLOGY_DISCOVERED_ERROR_STRING = "Expression is a tautology. Comparison is redundant.";
```
JoshuaKGoldberg: This rule is missing the ability to do the logic mentioned in the original issue: `obj1.foo === obj1.foo`. Additionally, it could be made quite a bit more powerful by becoming an optionally-type-checked rule, so it could check cases like:

```ts
const seven = 7;

if (seven === 7) { }
```

I think the `obj1.foo === obj1.foo` case should be a built-in part of the rule when it first goes in, but it's fine to file a followup issue for the optional type checking. It'd be a shame to have the `obj1.foo` case be a breaking change after the rule is enabled... but other maintainers would have a better knowledge of what the preference is here?

`unnecessaryBindRule.ts` is a good example to go off of.

src/rules/noTautologyExpressionRule.ts
```diff
@@ -0,0 +1,58 @@
...
+export class Rule extends Lint.Rules.AbstractRule {
+    public static metadata: Lint.IRuleMetadata = {
```
JoshuaKGoldberg: Please add some code examples to help make the reasoning for this rule more clear. `arrowReturnShorthandRule.ts` is a good reference for how to add those in.

noamyogev84:
Good to see you here @JoshuaKGoldberg . 😄 
I'll continue to work on this.

noamyogev84:
@JoshuaKGoldberg,
I have these 2 errors when running `yarn test`:
1. `1) Executable Files exits with code 0 if correct file is passed:`
2. `2) Executable Files exits with code 0 if several files passed without `-f` flag:`
tried to follow #3443 with no success. Do you have any idea what's the problem?

JoshuaKGoldberg:
☹ that's unfortunate @noamyogev84. For now, you can always delete those tests locally and/or only run `yarn run test:rules`, since your changes to add this rule don't affect those tests.

If that doesn't work for you, ping me on [Gitter](https://gitter.im/palantir/tslint)?

noamyogev84:
Hi @JoshuaKGoldberg, sorry for the long wait. 🙏 
Did some rework. Waiting for your review.

noamyogev84:
H @JoshuaKGoldberg. Pinging you one more time... 🙏

src/rules/noTautologyExpressionRule.ts
```diff
@@ -0,0 +1,86 @@
...
+        return ts.forEachChild(node, cb);
+    };
+    return ts.forEachChild(context.sourceFile, cb);
+}
+
+function isLiteral(node: ts.Node): node is ts.StringLiteral | ts.NumericLiteral {
+    return (node as ts.StringLiteral) !== undefined || (node as ts.NumericLiteral) !== undefined;
+}
+
+function isRelationalOrOrLogicalOperator(operator: ts.BinaryOperatorToken): boolean {
```
JoshuaKGoldberg: Nit: name typo, two `Or`s? :)

test/rules/no-tautology-expression/test.ts.lint
```diff
@@ -0,0 +1,108 @@
...
+        ~~~~~~~~~~~~~~~~~~  [Expression is either a tautology or a contradiction. Binary expression is redundant.]
+        return true;
+    }
+}
+
+const someVar1 = 3 + 3;
+const someVar2 = 3 - 3;
+const someVar3 = 3 * 3;
+const someVar4 = 3 % 3;
+const someVar5 = 3 / 3;
```
JoshuaKGoldberg: Were you planning on using these or the following blank lines for something? If not, let's remove them as unused.

test/rules/no-tautology-expression/test.ts.lint
```diff
@@ -0,0 +1,108 @@
+
```
JoshuaKGoldberg: Missing test cases for the property access expression cases.

src/rules/noTautologyExpressionRule.ts
```diff
@@ -0,0 +1,86 @@
...
+        ) {
+            if (isLiteral(node.left) && isLiteral(node.right)) {
+                if (node.left.text === node.right.text) {
+                    context.addFailureAtNode(node, TAUTOLOGY_DISCOVERED_ERROR_STRING);
+                }
+            } else if (
+                tsutils.isPropertyAccessExpression(node.left) &&
+                tsutils.isPropertyAccessExpression(node.right)
+            ) {
+                if (node.left.name.text === node.right.name.text) {
```
JoshuaKGoldberg: You'll also need to check that each side's `.expression` is an `Identifier` with the same name.

```ts
left.same === right.same
```

src/rules/noTautologyExpressionRule.ts
```diff
@@ -0,0 +1,86 @@
...
+                }
+            }
+        }
+        return ts.forEachChild(node, cb);
+    };
+    return ts.forEachChild(context.sourceFile, cb);
+}
+
+function isLiteral(node: ts.Node): node is ts.StringLiteral | ts.NumericLiteral {
+    return (node as ts.StringLiteral) !== undefined || (node as ts.NumericLiteral) !== undefined;
```
JoshuaKGoldberg: The `as` operator is a development-time construct and will not change the value of `node`. Using typescriptlang.org/play, this method compiles to:

```js
function isLiteral(node) {
    return node !== undefined || node !== undefined;
}
```

Use `isLiteralExpression` from `tsutils` instead.

src/rules/noTautologyExpressionRule.ts
```diff
@@ -0,0 +1,86 @@
...
+ * limitations under the License.
+ */
+
+import * as tsutils from "tsutils";
+import * as ts from "typescript";
+
+import * as Lint from "../index";
+
+const TAUTOLOGY_DISCOVERED_ERROR_STRING =
+    "Expression is either a tautology or a contradiction. Binary expression is redundant.";
```
JoshuaKGoldberg: > either a tautology or a contradiction

It'd be nice to indicate why the error was detected. How about this phrasing?
```suggestion
    "Both sides of this equality comparison are the same, so the expression is either a tautology or a contradiction.";
```

JoshuaKGoldberg:
Looking through the CircleCI failures, your rule has some kind of logic problem, because it's flagging acceptable binary expressions in TSLint's source as tautologies.

test/rules/no-tautology-expression/test.ts.lint
noamyogev84: These are meant for testing that arithmetic binary expressions are not being checked by the rule.
Still wanna remove them?

adidahiya:
friendly ping @noamyogev84, do you expect to have time to return to this PR? thanks

noamyogev84:
Hi @adidahiya. I'll do some work this weekend hopefully. 🕵️

noamyogev84:
Hi @JoshuaKGoldberg , @adidahiya .
added fixes and ready for you review. 🍻

src/rules/noTautologyExpressionRule.ts
```diff
@@ -0,0 +1,89 @@
...
+                    }
+                }
+            }
+        }
+        return ts.forEachChild(node, cb);
+    };
+    return ts.forEachChild(context.sourceFile, cb);
+}
+
+function isLiteral(node: ts.Node): node is ts.StringLiteral | ts.NumericLiteral {
```
JoshuaKGoldberg: There are few literals that this doesn't capture:

* `kind` `ts.SyntaxKind.NullKeyword`
* `kind` `ts.SyntaxKind.FalseKeyword`
* `kind` `ts.SyntaxKind.TrueKeyword`

_(sorry for not mentioning this till now; I didn't think to ask for these before)_

src/rules/noTautologyExpressionRule.ts
JoshuaKGoldberg: Ehh, let's just file a followup issue. Don't want to block this from coming in.

noamyogev84:
I think we're done 😃 
@JoshuaKGoldberg

src/rules/noTautologyExpressionRule.ts
noamyogev84: Unfortunately `tsutils` lacks some functionality regarding to what you've added.
Hope i'm not mistaken here 🙏

JoshuaKGoldberg: @noamyogev84 you can directly check the `.kind` on `node` to see if it's a particular `ts.SyntaxKind`.

noamyogev84: @JoshuaKGoldberg yeah that's what I did..
Just mentioned it here....

JoshuaKGoldberg: Oh, sorry, I was looking at the outdated diff... :)

src/rules/noTautologyExpressionRule.ts
```diff
@@ -0,0 +1,101 @@
...
+                tsutils.isPropertyAccessExpression(node.right)
+            ) {
+                if (node.left.expression.getText() === node.right.expression.getText()) {
+                    if (node.left.name.text === node.right.name.text) {
+                        context.addFailureAtNode(node, TAUTOLOGY_DISCOVERED_ERROR_STRING);
+                    }
+                }
+            } else if (
+                (isBooleanLiteral(node.left) && isBooleanLiteral(node.right)) ||
+                (isNullLiteral(node.left) && isNullLiteral(node.right))
```
JoshuaKGoldberg: Continuing the previous conversation: this doesn't capture numeric literals or `undefined`.

noamyogev84: numeric and string literals are covered earlier.
what about undefined? what do you mean by covering this ?

JoshuaKGoldberg: Oh, wow, I should review when fully awake. This looks great. I had forgotten `undefined` is an identifier.

JoshuaKGoldberg:
@noamyogev84 you've persisted through multiple rounds of code review and slightly inaccurate questioning on my part. Thanks so much for the rule! This is going to be great for catching silly mistakes in user code.

noamyogev84:
@JoshuaKGoldberg it's my pleasure man! 🥃

Issue #3755, opened 2018-03-05
`no-unbound-method` in Jasmine spy have been called test


piotrgajow:
### Bug Report

- __TSLint version__: 5.9.1
- __TypeScript version__: 2.5.3
- __Running TSLint via__: Node.js API

#### TypeScript code being linted

```ts
it('on destroy should unsubscribe timer', () => {
    component.timerSubscription = jasmine.createSpyObj('Subscription', ['unsubscribe']);

    component.ngOnDestroy();

    expect(component.timerSubscription.unsubscribe).toHaveBeenCalled();
});
```

with `tslint.json` configuration:

```json
{
    "rulesDirectory": [
        "node_modules/codelyzer"
    ],
    "rules": {
        "no-unbound-method": true
    }
}
```

#### Actual behavior

tslint reports error in the expect line on `component.timerSubscription.unsubscribe` - `Avoid referencing unbound methods which may cause unintentional scoping of 'this'`.

#### Expected behavior

This only references the method to check if it has been called - it does not call it, nor it saves it into variable for later calls - so it should not report any errors.

grimly:
The same happen on type detection.

This code throws a `no-unbound-method` error

```typescript
const property = Object.getOwnPropertyDescriptor(target, propertyName);
if (typeof property === 'object' && typeof property.set === 'function') {
  // ...
}
```

maxi7587:
Have the same issue, how do you use expect() actually with tslint 5.9.1 in jasmine?

Plondrein:
any update on this?

borriej:
I have the same issues since I updated packages.json
```
    "@angular-devkit/build-angular": "^0.7.5",
    "@angular/cli": "^6.2.8",
    "@angular/compiler-cli": "^6.1.10",
    "@angular/language-service": "^6.1.10",
    "@types/jasmine": "^2.8.14",
    "@types/jasminewd2": "^2.0.6",
    "@types/node": "~8.9.4",
    "codelyzer": "^4.5.0",
    "jasmine-core": "~2.99.1",
    "jasmine-spec-reporter": "~4.2.1",
    "karma": "~1.7.1",
    "karma-chrome-launcher": "~2.2.0",
    "karma-coverage-istanbul-reporter": "^2.0.4",
    "karma-jasmine": "~1.1.1",
    "karma-jasmine-html-reporter": "^0.2.2",
    "protractor": "^5.4.1",
    "ts-node": "~5.0.1",
[...]

PR #4472, opened 2019-01-20
Whitelist option for `no-unbound-method`

piotrgajow:
#### PR checklist

- [x] Addresses an existing issue: #3755 
- [x] Enhancement
  - [x] Includes tests
- [ ] Documentation update

#### Overview of change:

Added configuration option to whitelist methods and operators for `no-unbound-method` rule, so that it will not report errors in cases such as:

`expect(x.method).toHaveBeenCalled()`
`typeof x.method === 'function'`

Please verify that options for this rule are backwards compatible.

#### CHANGELOG.md entry:

[new-rule-option] Added `whitelist` option to `no-unbound-method`

### DIFF
--- a/src/rules/noUnboundMethodRule.ts
+++ b/src/rules/noUnboundMethodRule.ts
@@ -15,7 +15,12 @@
  * limitations under the License.
  */
 
-import { hasModifier, isPropertyAccessExpression } from "tsutils";
+import {
+    hasModifier,
+    isCallExpression,
+    isPropertyAccessExpression,
+    isTypeOfExpression,
+} from "tsutils";
 import * as ts from "typescript";
 import * as Lint from "../index";
 
@@ -210,13 +215,12 @@ function isSafeUse(node: ts.Node): boolean {
     }
 }
 
-function isWhitelisted(node: ts.Node, whitelist: string[]): boolean {
-    if (node.parent.kind === ts.SyntaxKind.TypeOfExpression) {
+function isWhitelisted(node: ts.Node, whitelist: Set<string>): boolean {
+    if (isTypeOfExpression(node.parent)) {
         return whitelist.indexOf("typeof") !== -1;
     }
-    if (node.parent.kind === ts.SyntaxKind.CallExpression) {
-        const parent = node.parent as ts.CallExpression;
-        const expression = parent.expression as ts.Identifier;
+    if (isCallExpression(node.parent)) {
+        const expression = node.parent.expression as ts.Identifier;
         const callingMethodName = expression.escapedText as string;
         return whitelist.indexOf(callingMethodName) !== -1;
     }
--- a/src/rules/noUnboundMethodRule.ts
+++ b/src/rules/noUnboundMethodRule.ts
@@ -149,7 +149,7 @@ function parseArguments(args: Array<string | OptionsInput>): Options {
             }
         } else {
             options.ignoreStatic = arg[OPTION_IGNORE_STATIC] || false;
-            (arg[OPTION_WHITELIST] || []).forEach(options.whitelist.add);
+            options.whitelist = new Set(arg[OPTION_WHITELIST] || []);
         }
     }
 
--- a/src/rules/noUnboundMethodRule.ts
+++ b/src/rules/noUnboundMethodRule.ts
@@ -22,6 +22,13 @@ import * as Lint from "../index";
 const OPTION_IGNORE_STATIC = "ignore-static";
 const OPTION_WHITELIST = "whitelist";
 
+const OPTION_WHITELIST_EXAMPLE = [
+    true,
+    {
+        [OPTION_WHITELIST]: ["expect", "typeof"],
+    },
+];
+
 interface Options {
     ignoreStatic: boolean;
     whitelist: string[];
@@ -32,7 +39,16 @@ export class Rule extends Lint.Rules.TypedRule {
     public static metadata: Lint.IRuleMetadata = {
         ruleName: "no-unbound-method",
         description: "Warns when a method is used outside of a method call.",
-        optionsDescription: `You may optionally pass "${OPTION_IGNORE_STATIC}" to ignore static methods.`,
+        optionsDescription: Lint.Utils.dedent`
+            You may additionally pass "${OPTION_IGNORE_STATIC}" to ignore static methods, or an options object.
+            
+            The object might have two properties:
+            
+            * "${OPTION_IGNORE_STATIC}" - to ignore static methods.
+            * "${OPTION_WHITELIST}" - list of method names and keywords where method reference should be ignored, e.g.
+              * when method is referenced as a parameter to whitelisted method.
+              * when method is referenced in a typeof expression.
+            `,
         options: {
             anyOf: [
                 {
@@ -52,7 +68,7 @@ export class Rule extends Lint.Rules.TypedRule {
                 },
             ],
         },
-        optionExamples: [true, [true, OPTION_IGNORE_STATIC]],
+        optionExamples: [true, [true, OPTION_IGNORE_STATIC], OPTION_WHITELIST_EXAMPLE],
         rationale: Lint.Utils.dedent`
             Class functions don't preserve the class scope when passed as standalone variables.
             For example, this code will log the global scope (\`window\`/\`global\`), not the class instance:
--- a/src/rules/noUnboundMethodRule.ts
+++ b/src/rules/noUnboundMethodRule.ts
@@ -232,7 +232,7 @@ function isWhitelisted(node: ts.Node, whitelist: Set<string>, allowTypeof: boole
     }
     if (isCallExpression(node.parent)) {
         const expression = node.parent.expression as ts.Identifier;
-        const callingMethodName = expression.escapedText as string;
+        const callingMethodName = expression.text as string;
         return whitelist.has(callingMethodName);
     }
     return false;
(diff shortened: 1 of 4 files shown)


src/rules/noUnboundMethodRule.ts
```diff
@@ -20,22 +20,55 @@ import * as ts from "typescript";
...
 export class Rule extends Lint.Rules.TypedRule {
     /* tslint:disable:object-literal-sort-keys */
     public static metadata: Lint.IRuleMetadata = {
         ruleName: "no-unbound-method",
         description: "Warns when a method is used outside of a method call.",
-        optionsDescription: `You may optionally pass "${OPTION_IGNORE_STATIC}" to ignore static methods.`,
+        optionsDescription: Lint.Utils.dedent`
+            You may additionally pass "${OPTION_IGNORE_STATIC}" to ignore static methods, or an options object.
+            
+            The object might have two properties:
```
JoshuaKGoldberg: > might

Nit: `may`

src/rules/noUnboundMethodRule.ts
```diff
@@ -20,22 +20,55 @@ import * as ts from "typescript";
...
+            
+            * "${OPTION_IGNORE_STATIC}" - to ignore static methods.
+            * "${OPTION_WHITELIST}" - list of method names and keywords where method reference should be ignored, e.g.
+              * when method is referenced as a parameter to whitelisted method.
+              * when method is referenced in a typeof expression.
```
JoshuaKGoldberg: > `"typeof"`

Love that this works in this PR! But - it seems a little odd to add `"typeof"` in the same array as method names. You can technically have `typeof` as a method name in some cases. Let's switch it to a separate member of the object. How about `allowTypeof`?

If you want to do that in a separate PR (so this one would start on the issue rather than completely fix it), that seems fine too.

src/rules/noUnboundMethodRule.ts
```diff
@@ -87,20 +120,46 @@ export class Rule extends Lint.Rules.TypedRule {
...
+        if (typeof arg === "string") {
+            if (arg === OPTION_IGNORE_STATIC) {
+                options.ignoreStatic = true;
+            }
+        } else {
+            if (arg.hasOwnProperty(OPTION_IGNORE_STATIC)) {
+                options.ignoreStatic = arg[OPTION_IGNORE_STATIC];
+            }
+            if (arg.hasOwnProperty(OPTION_WHITELIST)) {
+                options.whitelist = options.whitelist.concat(arg[OPTION_WHITELIST]);
```
JoshuaKGoldberg: These lines are failing the [lint build on CircleCI](https://circleci.com/gh/palantir/tslint/11243?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link) because `arg` is unsafely being dealt with. See `banRule.ts` for a good example of how this can be done a little more type-safely.

src/rules/noUnboundMethodRule.ts
```diff
@@ -150,3 +209,16 @@ function isSafeUse(node: ts.Node): boolean {
             return false;
     }
 }
+
+function isWhitelisted(node: ts.Node, whitelist: string[]): boolean {
+    if (node.parent.kind === ts.SyntaxKind.TypeOfExpression) {
```
JoshuaKGoldberg: When checking a node kind, use the `is*` methods from `tsutils`. It'll indicate to the type system whether the node is of the proper type, so you won't have to cast it later on.

For example, instead of `kind === ts.SyntaxKind.TypeOfExpression`, use `isTypeOfExpression`. It returns whether the given node `is ts.TypeOfExpression`.

src/rules/noUnboundMethodRule.ts
```diff
@@ -20,22 +20,55 @@ import * as ts from "typescript";
...
+const OPTION_WHITELIST_EXAMPLE = [
+    true,
+    {
+        [OPTION_WHITELIST]: ["expect", "typeof"],
+    },
+];
 
 interface Options {
     ignoreStatic: boolean;
+    whitelist: string[];
```
JoshuaKGoldberg: Looks like this is only ever checked for a name to exist, so it'd be a little more efficient to use a `Set<string>` and `ctx.options.whitelist.has(callingMethodName)`.

src/rules/noUnboundMethodRule.ts
piotrgajow: I have changed the description. What does 'Nit' mean? 🤔

src/rules/noUnboundMethodRule.ts
piotrgajow: Changed to `Set`

src/rules/noUnboundMethodRule.ts
piotrgajow: At first I wanted to make two whitelists for methods and operators, but I couldn't find any other operator that might be used in this context so I merged it to one parameter... But your suggestion feels best, so I added handling of `typeof` as a flag parameter.

src/rules/noUnboundMethodRule.ts
piotrgajow: I have changed the parsing logic - hope it works now. For some reason I cannot run `yarn lint` locally (it just hangs...).

src/rules/noUnboundMethodRule.ts
piotrgajow: Using proper `is*` methods now

piotrgajow:
I see that some tests are failing... Can anyone tell me what is the difference between `test2.1`, `test2.4` and for example `test2.7`?

src/rules/noUnboundMethodRule.ts
JoshuaKGoldberg: Huh, weird. Might be a separate issue to file?

JoshuaKGoldberg:
Hi @piotrgajow! The difference is which TypeScript version the tests are running against. TSLint supports all the versions listed in the tests, down to 2.1. It'd be a breaking change to drop support, even though it's a little annoying how some APIs have changed or added since then.

Suggestion: try installing TypeScript 2.1 locally and compile against it? It could be there's an API change that is missed. Feel free to tag me if that doesn't catch it and I can take a deeper look! This'd be a really great PR to get in.

piotrgajow:
Hi @JoshuaKGoldberg, I have tried installing Typescript 2.1/2.4 locally, but then I didn't manage to run the tests - it fails during compilation phase with:

    $ yarn compile:test && yarn test:rules
    yarn run v1.10.1
    $ tsc -p test
    src/formatterLoader.ts(73,17): error TS1005: '(' expected.
    src/formatterLoader.ts(89,13): error TS1005: '(' expected.
    src/language/utils.ts(128,49): error TS1005: ',' expected.
    src/language/utils.ts(128,53): error TS1005: ',' expected.
    src/ruleLoader.ts(122,13): error TS1005: '(' expected.
    src/rules/fileNameCasingRule.ts(62,13): error TS1005: '(' expected.
    src/rules/noImplicitDependenciesRule.ts(159,17): error TS1005: '(' expected.
    src/runner.ts(149,28): error TS1109: Expression expected.
    src/runner.ts(149,34): error TS1134: Variable declaration expected.
    src/utils.ts(292,13): error TS1005: '(' expected.
    error Command failed with exit code 2.

But these files are not strictly related to changes from this PR... 
Could you maybe direct me somehow? 
Maybe I did something wrong?

I have installed typescipt with version 2.1, and then removed `strictFunctionTypes` and `strictPropertyInitialization` from `tsconfig.json` as compiler complained that these options are unknown.

src/rules/noUnboundMethodRule.ts
```diff
@@ -150,3 +225,15 @@ function isSafeUse(node: ts.Node): boolean {
             return false;
     }
 }
+
+function isWhitelisted(node: ts.Node, whitelist: Set<string>, allowTypeof: boolean): boolean {
+    if (isTypeOfExpression(node.parent)) {
+        return allowTypeof;
+    }
+    if (isCallExpression(node.parent)) {
+        const expression = node.parent.expression as ts.Identifier;
```
JoshuaKGoldberg: > `as ts.Identifier`

It's actually possible for `node.parent.expression` to not be an identifier. You'll want to add test cases for:

```ts
(condition ? expectA : expectB)(c.method);
(await someObject)(c.method);
(await someMethod())(c.method);
```

I don't think we should worry about trying to allow these even if `expectA` and `expectB` are whitelisted. Suggestion: just don't run this check if the parent's expression isn't an identifier.

src/rules/noUnboundMethodRule.ts
```diff
@@ -150,3 +225,15 @@ function isSafeUse(node: ts.Node): boolean {
...
+        const callingMethodName = expression.escapedText as string;
```
JoshuaKGoldberg: > `expression.escapedText`

Looks like `escapedText` is what's giving you trouble in in TS<=2.4. Switch to `.text` and they seem to work fine: https://circleci.com/gh/JoshuaKGoldberg/tslint/1267

JoshuaKGoldberg:
AH, that's fun. I guess you'd have to compile in newer TS so it would understand your syntax, then run using a lower version TS. See comments in PR for resolution! 😊

src/rules/noUnboundMethodRule.ts
piotrgajow: Ok, changed as you suggested. I maybe might include it in some future PR

src/rules/noUnboundMethodRule.ts
piotrgajow: Yes, that was the issue. Thanks!

piotrgajow:
It seems that I finally managed to fix all of the issues 😃 .

Issue #4476, opened 2019-01-22
[Interface-name] Never prefix return error for I18nInterface


VincentLanglet:
### Bug Report

- __TSLint version__: 5.12.1
- __TypeScript version__: 3.2.4
- __Running TSLint via__: CLI / PhpStorm

#### TypeScript code being linted

```ts
interface I18nInterface {
  foo: () => void;
}
```

with `tslint.json` configuration:

```json
{
  "extends": ["tslint:all"],
  "rules": {
    "interface-name": [true, "never-prefix"]
  }
}
```

#### Actual behavior
Got error "Interface must not have an I prefix

#### Expected behavior
No error, I18n should be an allowed word.
Don't know if we have to allow ^I\d+ interfaces or just ^I18n ones.

adidahiya:
looks like a bug. there's already a special case in the rule for `IDB`, might as well add one for `I18`

adidahiya:
Fixed by #4486

PR #4486, opened 2019-01-29
Handle I18n case

VincentLanglet:
#### PR checklist

- [x] Addresses an existing issue: #4476
- [x] Includes tests
- [ ] Documentation update

#### Overview of change:

Interface name (never-prefix) doesn't return error for I18nSomething

#### CHANGELOG.md entry:

[bugfix] `interface-name` now handles interface starting with "I18n" correctly

### DIFF
--- a/src/rules/interfaceNameRule.ts
+++ b/src/rules/interfaceNameRule.ts
@@ -19,7 +19,7 @@ import * as utils from "tsutils";
 import * as ts from "typescript";
 
 import * as Lint from "../index";
-import { isUpperCase } from "../utils";
+import { isLowerCase } from "../utils";
 
 const OPTION_ALWAYS = "always-prefix";
 const OPTION_NEVER = "never-prefix";
@@ -74,10 +74,10 @@ function walk(ctx: Lint.WalkContext<{ never: boolean }>): void {
 }
 
 function hasPrefixI(name: string): boolean {
-    return name.length >= 2 && name[0] === "I" && isUpperCase(name[1]) && !isSpecialCase(name);
+    return name.length >= 2 && name[0] === "I" && !isLowerCase(name[1]) && !isSpecialCase(name);
 }
 
 function isSpecialCase(name: string): boolean {
-    // Allow IndexedDB and I18n interfaces
-    return name.startsWith("IDB") || name.startsWith("I18n");
+    // Allow IndexedDB
+    return name.startsWith("IDB");
 }
--- a/src/rules/interfaceNameRule.ts
+++ b/src/rules/interfaceNameRule.ts
@@ -19,7 +19,7 @@ import * as utils from "tsutils";
 import * as ts from "typescript";
 
 import * as Lint from "../index";
-import { isLowerCase } from "../utils";
+import { isLowerCase, isUpperCase } from "../utils";
 
 const OPTION_ALWAYS = "always-prefix";
 const OPTION_NEVER = "never-prefix";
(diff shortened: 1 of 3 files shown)


reduckted:
Instead of adding a special case just for "I18n", would a better solution be to check if the name without the "I" is a valid identifier? "18nFactory" isn't a valid identifier because it starts with a number, so the "I" prefix must be allowed.

This, of course, wouldn't work for "IDB", though I'm not sure why that was added as a hard-coded special case in the first place.

VincentLanglet:
@reduckted It seems to hard to check what is a valid identifier.

I cant say every number are not valid identifier, because 3rdParty, 1rstItem, 2ndElement can be.

reduckted:
> I cant say every number are not valid identifier, because 3rdParty, 1rstItem, 2ndElement can be.

None of those are valid identifiers. 

> In JavaScript, identifiers can contain only alphanumeric characters (or "$" or "_"), and may not start with a digit.
> \- [Identifier, MDN](https://developer.mozilla.org/en-US/docs/Glossary/Identifier)

> So, what is allowed in an identifier name?
> 
> An identifier must start with $, _, or any character in the Unicode categories “Uppercase letter (Lu)”, “Lowercase letter (Ll)”, “Titlecase letter (Lt)”, “Modifier letter (Lm)”, “Other letter (Lo)”, or “Letter number (Nl)”.
> \- [Valid JavaScript variable names in ES5](https://mathiasbynens.be/notes/javascript-identifiers#valid-identifier-names)

VincentLanglet:
@reduckted Ok, need to do a regex so.

BTW, IDB was added as a short cup for IndexedDataBase, i think.
It's debatable but I'll keep it to not create regression/breaking change.

VincentLanglet:
After some reflexion, I'm not sure it's a good idea to check `name.substr(1)` is a valid identifier to know if `name` has a `I` prefix.

`IDB` case can be leggit. You don't want to forbid acronym. This way : 

- Two letters words may be allowed : `ID` can be acronym for `IndexedData`. 
- Three letters (+) words starting by two uppercases should be allowed if the third letters is also an uppercase. Like `IDB` or `IDKWhatIsThis`. We should also allow `ID42`, `ID_42` or `ID$42`.

A good rule seems to be : 
`hasPrefixI: name[0] = 'I', name[1] Uppercase, name[2] Lowercase`

And best way to check if `name[1]` is uppercase (and avoid special numeric or `_`, `$` cases) is to check that `name[1]` is not lowercase. In the same way, I check `name[2]` is not uppercase.

src/rules/interfaceNameRule.ts
```diff
@@ -74,6 +74,5 @@ function walk(ctx: Lint.WalkContext<{ never: boolean }>): void {
 }
 
 function hasPrefixI(name: string): boolean {
-    // Allow IndexedDB interfaces
-    return name.length >= 2 && name[0] === "I" && isUpperCase(name[1]) && !name.startsWith("IDB");
+    return name.length >= 3 && name[0] === "I" && !isLowerCase(name[1]) && !isUpperCase(name[2]);
```
adidahiya: can you add some test cases in `always-prefix/test.ts.lint` as well? I feel like this will affect that option's behavior.

with `always-prefix`, I think the behavior should be:

- failures: `ID`
- valid: `IDB` (or, any all-caps interface name starting with `I`)

I know it feels a little silly since these are edge cases in naming, but we should try to be consistent in some way

VincentLanglet: I can add test yes.
But I didn't change the behaviour for the àlways-prefix` rule.

Actually the test check only if the first letter is `I`.
Now I'm thinking about it, an interface called `Information` should return an error but it doesn't.

With `always-prefix`, I think `ID` should return no error.
You can create an `A` interface, called `IA` with prefix. The same way a `D` interface would be named `ID` with prefix.

To be consistent the best way is to check
`always-prefix && ! hasPrefixI` or `never-prefix && hasPrefixI`

VincentLanglet:
@adidahiya @JoshuaKGoldberg I add a lot of test and then fix the edge cases.

styu:
I think this change caused a regression, I'm noticing now that correctly prefixed interfaces that have a number in it (e.g. `IS3Foobar`) is now returning errors

VincentLanglet:
@styu I already made a fix here https://github.com/palantir/tslint/pull/4626, I'll check it fix your case

styu:
thanks so much!

PR #4502, opened 2019-02-04
Added unnecessary-else Rule

### LINKED ISSUE #4429: Rule suggestion: unnecessary-else
### Rule Suggestion

General problem

If an if block contains a return, throw, continue or break statement, the else block becomes unnecessary. Its contents can be placed outside of the block.

Instead of:

```typescript
if (foo) {
    throw new Error();
} else {
    // ...
}
```

Prefer:

```typescript
if (foo) {
    throw new Error();
}

// ...
```

See https://github.com/ajafff/tslint-consistent-codestyle/blob/master/docs/no-unnecessary-else.md

**Additional context**

Spinning off of #2040: moving some rules of tslint-consistent-codestyle to tslint core.
Note that as ajafff mentioned, copied code without substantial changes needs to retain the original license.

debsmita1:
#### PR checklist

- [x] Addresses an existing issue: fixes #4429
- [x] New feature
  - [x] Includes tests
- [x] Documentation update

#### Overview of change:
This Rule disallows `else` blocks following `if` blocks ending with a `return` , `throw`, `break` or `continue`statement.

This is my very first open-source contribution, so all kinds of feedback is appreciated.

#### CHANGELOG.md entry:
[new-rule] unnecessary-else﻿


palantirtech:
Thanks for your interest in [palantir/tslint](https://github.com/palantir/tslint), @debsmita1! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

src/rules/noUnnecessaryElseRule.ts
```diff
@@ -0,0 +1,98 @@
...
+    /* tslint:disable:object-literal-sort-keys */
+    public static metadata: Lint.IRuleMetadata = {
+        description: Lint.Utils.dedent`
+            Disallows else block when the If block contains control flow statements, such as \`return\`, \`continue\`,
+            \`break\` or \`throw\`.`,
+        descriptionDetails: "",
+        optionExamples: [true],
+        options: null,
+        optionsDescription: "Not configurable.",
+        rationale: Lint.Utils.dedent`
```
JoshuaKGoldberg: Hmm, this rationale seems to restate the rule description. How about something like:

> ```
> When an `if` block is guaranteed to exit control flow when entered,
> it is unnecessary to add an `else` statement.
> The contents that would be in the `else` block can be placed after the end of the `if` block.
> ```

src/rules/noUnnecessaryElseRule.ts
```diff
@@ -0,0 +1,98 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+import * as ts from "typescript";
+
+import * as Lint from "../index";
+
+export class Rule extends Lint.Rules.AbstractRule {
+    /* tslint:disable:object-literal-sort-keys */
+    public static metadata: Lint.IRuleMetadata = {
```
JoshuaKGoldberg: TSLint's trying to move to more descriptive documentation with code samples for rules. Please add a `codeExamples` section to this rule.

You can see rules such as `arrowReturnShorthandRule.ts` for examples of existing ones.

src/rules/noUnnecessaryElseRule.ts
```diff
@@ -0,0 +1,98 @@
...
+            such as \`return\`, \`continue\`, \`break\` or \`throw\` is written inside \`if\` block,
+            Then \`else\` block becomes unnecessary.`,
+        ruleName: "no-unnecessary-else",
+        type: "functionality",
+        typescriptOnly: false,
+    };
+    /* tslint:disable:object-literal-sort-keys */
+
+    public static FAILURE_STRING(name: string): string {
+        return `If block contains \`${name}\` statement. Consider replacing the contents of else block outside of the block.`;
```
JoshuaKGoldberg: The rule doesn't need to say "consider" because the rule failure is a definite "you must do this":

```suggestion
        return `The preceding \`if\` block ends with a \`${name}\` statement. This \`else\` block is unnecessary.`;
```

src/rules/noUnnecessaryElseRule.ts
```diff
@@ -0,0 +1,98 @@
...
+}
+
+type JumpStatement =
+    | ts.BreakStatement
+    | ts.ContinueStatement
+    | ts.ThrowStatement
+    | ts.ReturnStatement;
+
+function walk(ctx: Lint.WalkContext<void>): void {
+    let inElse = false;
```
JoshuaKGoldberg: It looks like `walk` will consider any jump statement within an `if` block to make that block not need an `else`. What if there's one within a nested `if` statement?

```ts
if (something) {
    if (anotherThing) {
        return;
    } else {
        // Unnecessary else block
    }
} else {
    // Reasonable else block
}
```

Instead of walking down within the `if`, I think you can get away with just checking whether its _last_ statement is a jump statement. In other words, I think you should be able to just use `if (isIfStatement(node)) {` and check the last statement within the `thenStatement` and the node's `elseStatement`.

test/rules/no-unnecessary-else/test.ts.lint
```diff
@@ -0,0 +1,150 @@
...
+function testBreak () {
+   if () {
+       break ;
+       ~~~~~~~  [break]
+   } else {
+   }
+}
+
+const testBreak = () => {
+    switch () {
```
JoshuaKGoldberg: It's invalid to have empty contents in `case`, `if`, and `switch` statements. Rules can behave unpredictably when there are syntax errors, so tests should have valid syntax whenever possible.

test/rules/no-unnecessary-else/test.ts.lint
```diff
@@ -0,0 +1,150 @@
...
+
+const testReturn = () => {
+    if () {
+       return;
+       ~~~~~~~  [return]
+    } else if {
+        if () {
+            return ;
+        } else {
+            return ;
```
JoshuaKGoldberg: Here's an example of the rule not giving a complaint when it should. Unclear if it's from the syntax errors in this test file or from the structural notes above.

debsmita1:
const testReturn = (a) => {
      if (a%2 === 0) {
         return "multiple of 2" ;
      } else if (a%3 === 0) {
          return "multiple of 3" ;
      } else if (a%5 === 0) {
          return "multiple of 5" ;
      } else {
          return "not a multiple of 2 , 3 or 5" ;
      }
  }
For the above scenario what should be the correct behavior ? The linter must throw an error at every `else` block ? @JoshuaKGoldberg

src/configs/recommended.ts
```diff
@@ -91,6 +91,7 @@ export const rules = {
     "no-string-throw": true,
     "no-switch-case-fall-through": false,
     "no-trailing-whitespace": true,
+    "no-unnecessary-else": true,
```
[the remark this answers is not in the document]
debsmita1: Done https://github.com/palantir/tslint/pull/4502/commits/8d4f035a89b74f342aa687724425c8b5406d077d

src/rules/noUnnecessaryElseRule.ts
```diff
@@ -0,0 +1,98 @@
...
+            \`break\` or \`throw\`.`,
+        descriptionDetails: "",
+        optionExamples: [true],
+        options: null,
+        optionsDescription: "Not configurable.",
+        rationale: Lint.Utils.dedent`
+            When control flow statements,
+            such as \`return\`, \`continue\`, \`break\` or \`throw\` is written inside \`if\` block,
+            Then \`else\` block becomes unnecessary.`,
+        ruleName: "no-unnecessary-else",
```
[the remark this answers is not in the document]
debsmita1: Done https://github.com/palantir/tslint/pull/4502/commits/8d4f035a89b74f342aa687724425c8b5406d077d

src/rules/noUnnecessaryElseRule.ts
```diff
@@ -0,0 +1,98 @@
+/**
+ * @license
+ * Copyright 2018 Palantir Technologies, Inc.
```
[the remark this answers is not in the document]
debsmita1: Done https://github.com/palantir/tslint/pull/4502/commits/ee45076c830524d83245b806e4a65bba5a2ca5da

test/rules/no-unnecessary-else/test.ts.lint
```diff
@@ -0,0 +1,150 @@
...
+        if () {
+        return;
+        ~~~~~~~  [return]
+        } else {
+            return;
+        }
+    }
+}
+
+//valid
```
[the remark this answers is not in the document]
debsmita1: Done https://github.com/palantir/tslint/pull/4502/commits/c2db576ebe200c4e74417e56954613d16a3292b9

src/rules/noUnnecessaryElseRule.ts
debsmita1: Done https://github.com/palantir/tslint/pull/4502/commits/b46d5385f6d5848136581f593a1a465db4990ffe

src/rules/noUnnecessaryElseRule.ts
```diff
@@ -0,0 +1,98 @@
...
+ */
+import * as ts from "typescript";
+
+import * as Lint from "../index";
+
+export class Rule extends Lint.Rules.AbstractRule {
+    /* tslint:disable:object-literal-sort-keys */
+    public static metadata: Lint.IRuleMetadata = {
+        description: Lint.Utils.dedent`
+            Disallows else block when the If block contains control flow statements, such as \`return\`, \`continue\`,
```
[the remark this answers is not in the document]
debsmita1: Done https://github.com/palantir/tslint/pull/4502/commits/a1ec00c3360cd8d8b7e36e4ea7b18410a14db156

src/rules/noUnnecessaryElseRule.ts
debsmita1: Done https://github.com/palantir/tslint/pull/4502/commits/a1ec00c3360cd8d8b7e36e4ea7b18410a14db156

src/rules/noUnnecessaryElseRule.ts
debsmita1: Done https://github.com/palantir/tslint/pull/4502/commits/a1ec00c3360cd8d8b7e36e4ea7b18410a14db156

test/rules/no-unnecessary-else/test.ts.lint
debsmita1: Done https://github.com/palantir/tslint/pull/4502/commits/89c814eb7648cbe716af317adb05add694305430

test/rules/no-unnecessary-else/test.ts.lint
debsmita1: Done https://github.com/palantir/tslint/pull/4502/commits/89c814eb7648cbe716af317adb05add694305430

src/rules/noUnnecessaryElseRule.ts
debsmita1: Done https://github.com/palantir/tslint/pull/4502/commits/89c814eb7648cbe716af317adb05add694305430

debsmita1:
I have addressed all your comments.
screenshot from 2019-02-17 18-42-01
Cant figure out why these 2 are failing. Could you please help @JoshuaKGoldberg

src/rules/unnecessaryElseRule.ts
```diff
@@ -0,0 +1,100 @@
...
+    public apply(sourceFile: ts.SourceFile): Lint.RuleFailure[] {
+        return this.applyWithFunction(sourceFile, walk);
+    }
+}
+
+function walk(ctx: Lint.WalkContext<void>): void {
+    ts.forEachChild(ctx.sourceFile, function cb(node: ts.Node): void {
+        if (utils.isIfStatement(node)) {
+            const jumpStatement = getJumpStatement(
+                getLastStatement(node.thenStatement as ts.Block),
```
JoshuaKGoldberg: Fun fact: `node.thenStatement` could also be an `ExpressionStatement`:

```ts
if (condition) someExpression;
```

...because of the lack of `{}`s. You'll need to handle that & have test cases for it as well.

src/rules/unnecessaryElseRule.ts
```diff
@@ -0,0 +1,100 @@
...
+            if (jumpStatement !== undefined && node.elseStatement !== undefined) {
+                ctx.addFailureAtNode(
+                    node.elseStatement.getChildAt(0),
+                    Rule.FAILURE_STRING(jumpStatement),
+                );
+                ts.forEachChild(node.elseStatement, cb);
+            }
+            ts.forEachChild(node.thenStatement, cb);
+        } else {
+            return ts.forEachChild(node, cb);
```
JoshuaKGoldberg: It looks like you have a few different `ts.forEachChild`s: individual ones for the node's `elseStatement` and `thenStatement` if it's an `IfStatement`, but just a regular recursion otherwise? You should be able to simplify this a bit by only having one call to `ts.forEachChild` that's always called.

Also, if `jumpStatement` is undefined but `node.elseStatement` isn't, this code will skip it. That seems like a potential bug?

src/rules/unnecessaryElseRule.ts
debsmita1: Done https://github.com/palantir/tslint/pull/4502/commits/be2f8b4323b7b534130489a8467a91b6962d745e

src/rules/unnecessaryElseRule.ts
debsmita1: Done https://github.com/palantir/tslint/pull/4502/commits/be2f8b4323b7b534130489a8467a91b6962d745e

JoshuaKGoldberg: Sorry @debsmita1 I think I wasn't very clear here. I'm suggesting you don't need to have `ts.forEachChild` statements in the big `if (utils.isIfStatement` block:

```ts
if (utils.isIfStatement(node)) {
    // all your glorious rule logic here
    // no ts.forEachChild in this area
}

ts.forEachChild(node, cb);
```

Note the lack of `else` statement: a single `ts.forEachChild` should recurse on all the node's children no matter what. Could you please switch to that? It's a bit less code and better practice to minimize recursive calls (since they can get tricky in larger rules).

src/rules/code-examples/unnecessaryElse.examples.ts
```diff
@@ -0,0 +1,76 @@
...
+// tslint:disable: object-literal-sort-keys
+export const codeExamples = [
+    {
+        description:
+            'Disallows "else" blocks following "if" blocks ending with "return", "break", "continue" or "throw" statement. ',
+        config: Lint.Utils.dedent`
+            "rules": { "unnecessary-else": true }
+        `,
+        pass: Lint.Utils.dedent`
+            if () {
```
JoshuaKGoldberg: Coming back to these after a bit, the blank `()s` could be a little confusing for newcomers because they're invalid syntax in our code examples. Could you change it to `if (someCondition())` and similar?

src/rules/unnecessaryElseRule.ts
```diff
@@ -0,0 +1,117 @@
...
+        The contents that would be in the \`else\` block can be placed after the end of the \`if\` block.`,
+        ruleName: "unnecessary-else",
+        type: "functionality",
+        typescriptOnly: false,
+        codeExamples,
+    };
+    /* tslint:disable:object-literal-sort-keys */
+
+    public static FAILURE_STRING(name: string): string {
+        return `The preceding \`if\` block ends with a \`${name}\` statement. This \`else\` block is unnecessary.`;
```
JoshuaKGoldberg: Sorry to nitpick on this again: I think it might be a little more clear if the error message said <code>... This \`else\` is unnecessary</code> _(removing the `block`)_. Technically, the block's contents are fine; it's just the `else` that's unnecessary. Would you mind changing it?

src/rules/unnecessaryElseRule.ts
```diff
@@ -0,0 +1,117 @@
...
+    }
+}
+
+function walk(ctx: Lint.WalkContext<void>): void {
+    ts.forEachChild(ctx.sourceFile, function cb(node: ts.Node): void {
+        if (utils.isIfStatement(node)) {
+            let jumpStatement;
+            if (
+                node.thenStatement.kind !== ts.SyntaxKind.Block &&
+                !utils.isSameLine(ctx.sourceFile, node.thenStatement.pos, node.thenStatement.end)
```
JoshuaKGoldberg: I'm a little confused, is it actually necessary to check `utils.isSameLine`? `node` and its `.thenStatement` should behave the same regardless of whitespace position, right?

JoshuaKGoldberg:
> screenshot from 2019-02-17 18-42-01

I also can't figure out why those are failing at first glance. Very spooky. If they're still failing after the introduced lint complaints are fixed, we can take a deeper look...

src/rules/unnecessaryElseRule.ts
debsmita1: I have verified this multiple times,
if() return;
else return;

behaves very differently from 

if()
	return;
else
	return;

which is why I had to check `utils.isSameLine`. Could you kindly suggest a way to optimize this ? @JoshuaKGoldberg

JoshuaKGoldberg: Interesting! In what ways are you finding them to be different?

With this test, I'm getting the same result reported for both:

```ts
function testSameLine(condition) {
    if (condition) return 'a';
    else return 'b';
}

function testNextLine(condition) {
    if (condition)
    return 'a';
    else
    return 'b';
}

for (const condition of [true, false]) {
    console.log(condition);
    console.log("\tSame line:", testSameLine(condition));
    console.log("\tNext line:", testNextLine(condition));
    console.log("");
}
```

```
true
    Same line: a
    Same line: a

false
    Same line: b
    Same line: b
```

debsmita1: If I remove any of the `utils.isSameLine` checks I get the below error and the rule doesnt work
logs

JoshuaKGoldberg: Referring to the `TypeError; cannot read property '0' of undefined`?

It looks like the problem is from the `jumpStatement = getJumpStatement(getLastStatement(node.thenStatement as ts.Block));`. The rule can't assume `node.thenStatement as ts.Block` is valid because `node.thenStatement` might be some other kind of node, such as a `ReturnStatement` in the sample code in this thread.

You should be able to simplify the cases within the `if (utils.isIfStatement` block into just the following:

* `tsutils.isBlock(node.thenStatement)`
* Not ☝

...rather than the three cases you have now.

debsmita1: Done https://github.com/palantir/tslint/pull/4502/commits/f36d4201e2fa8ec3aec60a022db2bf51435aceaf

src/rules/unnecessaryElseRule.ts
debsmita1: Done https://github.com/palantir/tslint/pull/4502/commits/f36d4201e2fa8ec3aec60a022db2bf51435aceaf

src/rules/code-examples/unnecessaryElse.examples.ts
debsmita1: Done https://github.com/palantir/tslint/pull/4502/commits/a2842a9ad5ba17d0bb597296ad99b7bc0bbe5142

src/rules/unnecessaryElseRule.ts
debsmita1: Done https://github.com/palantir/tslint/pull/4502/commits/4c29b4309e698117979d5bd9a8e62b9fbc337b1f

src/rules/unnecessaryElseRule.ts
debsmita1: Done https://github.com/palantir/tslint/pull/4502/commits/f36d4201e2fa8ec3aec60a022db2bf51435aceaf

src/rules/unnecessaryElseRule.ts
```diff
@@ -0,0 +1,100 @@
...
+        descriptionDetails: "",
+        optionExamples: [true],
+        options: null,
+        optionsDescription: "Not configurable.",
+        rationale: Lint.Utils.dedent`
+        When an \`if\` block is guaranteed to exit control flow when entered,
+        it is unnecessary to add an \`else\` statement.
+        The contents that would be in the \`else\` block can be placed after the end of the \`if\` block.`,
+        ruleName: "unnecessary-else",
+        type: "functionality",
```
adidahiya: I think this is a purely stylistic rule; is there any functional difference?

Issue #4513, opened 2019-02-08
jsRules:true does not copy over explicitly disabled rules


UselessPickles:
### Bug Report

- __TSLint version__: 5.12.1
- __TypeScript version__: 3.2.4
- __Running TSLint via__: CLI

#### TypeScript code being linted
(any JS file)

with `tslint.json` configuration:

```json
{
    "jsRules": true,
    "rules": {
        // any JS-compatible rule that is enabled by default?, disabled
        "object-literal-sort-keys": false    
    }
}
```

#### Actual behavior
Default enabled rules are still applied to JS files, even when explicitly disabled in `"rules"` and `"jsRules": true` is used.

In this PR: https://github.com/palantir/tslint/pull/3641/files

The implementation of `filterValidJsRules` only copies over JS-compatible rules if `ruleOptions.ruleSeverity !== "off"`.

This seems like an odd intentional behavior. I think simply removing that `if` check will produce expected behavior, but I'm hesitant because it appears that this was intentionally done for some reason. Is there a good reason that disabled rules are ignored?

#### Expected behavior
Configuration of all JS-compatible rules should be copied over to and applied to JS files, including those that are explicitly disabled.

UselessPickles:
@mlavina - As the original PR author, can you provide any insight into why explicitly disabled rules are intentionally ignored when copying over to `jsRules`?

mlavina:
@UselessPickles  my best guess was that it was because that was my best guess for the issue here https://github.com/palantir/tslint/issues/2198

Which specifically states `Active` rules.

mlavina:
Though I don't have a strong opinion either way haha

UselessPickles:
Hmmm... is there a distinction between a "non-active" rule and a rule that is explicitly disabled via the config.

I wonder if the original intent/desire was to copy over all "specified" rules, rather than all "active" rules?

UselessPickles:
Either way, I feel like my expectations are quite reasonable: however I configured the `"rules"`, I expect it to be also be applied to JS files when `"jsRules": true`. I can't think of a good reason for a rule that I explicitly configured as disabled should be exempt from this behavior.

If someone with authority can make the decision, I will gladly create a PR for this.

adidahiya:
@UselessPickles I think your expectations are reasonable, it might end up being a breaking change but I'm open to PRs for this

retwedt:
[...]

PR #4517, opened 2019-02-14
Fix behavior of jsRules:true to include explicitly disabled rules.

UselessPickles:
#### PR checklist

- [x] Addresses an existing issue: #4513
- [x] bugfix
  - [x] Includes tests
- [x] Documentation update

#### Overview of change:

Changed behavior of `jsRules:true` to copy over ALL rule configurations that are compatible with JS, including any that are explicitly disabled.

The original PR (https://github.com/palantir/tslint/pull/3641) that added `jsRules:true` functionality seemed to make a point to describe its desired/implemented behavior to copy over all "active" rules that are valid for JS. Does the term "active" mean something special that is not simply "those that are not disabled"? Is there still some kind of rule configurations that I should be skipping over?

#### CHANGELOG.md entry:
[bugfix] Explicit disabling of rules is now copied over to jsRules when using jsRules:true

### DIFF
--- a/docs/usage/configuration/index.md
+++ b/docs/usage/configuration/index.md
@@ -31,7 +31,7 @@ configure which rules get run and each of their options.
     -   Legacy: A boolean value may be specified instead of the above object, and is equivalent to setting no options with default severity.
     -   Any rules specified in this block will override those configured in any base configuration being extended.
     -   [Check out the full rules list here][3].
--   `jsRules?: any | boolean`: Same format as `rules` or explicit `true` to copy all valid active rules from rules. These rules are applied to `.js` and `.jsx` files.
+-   `jsRules?: any | boolean`: Same format as `rules` or explicit `true` to copy all rule configurations for JS-compatible rules from `rules`. These rules are applied to `.js` and `.jsx` files.
 -   `defaultSeverity?: "error" | "warning" | "off"`: The severity level that is applied to rules in this config file as well as rules in any inherited config files which have their severity set to "default". If undefined, "error" is used as the defaultSeverity.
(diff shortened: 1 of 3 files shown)


test/configurationTests.ts
```diff
@@ -151,7 +152,9 @@ describe("Configuration", () => {
             rawConfig = {
                 jsRules: true,
                 rules: {
-                    eofline: true,
+                    // valid rule for JS, disabled (should be copied over)
+                    eofline: false,
```
UselessPickles: I tweaked this existing test case to cover the new behavior of keeping disabled rules

docs/usage/configuration/index.md
```diff
@@ -31,7 +31,7 @@ configure which rules get run and each of their options.
     -   Legacy: A boolean value may be specified instead of the above object, and is equivalent to setting no options with default severity.
     -   Any rules specified in this block will override those configured in any base configuration being extended.
     -   [Check out the full rules list here][3].
--   `jsRules?: any | boolean`: Same format as `rules` or explicit `true` to copy all valid active rules from rules. These rules are applied to `.js` and `.jsx` files.
+-   `jsRules?: any | boolean`: Same format as `rules` or explicit `true` to copy all rule configurations for JS-compatible rules from `rules`. These rules are applied to `.js` and `.jsx` files.
```
UselessPickles: Documentation is clarified

UselessPickles:
Am I supposed to add an entry to CHANGELOG.md as part of this PR?

adidahiya:
@UselessPickles nope, those entries get autogenerated by the release script

PR #4519, opened 2019-02-15
Add strict-comparisons Rule

### LINKED ISSUE #3574: Rule suggestion: Ban comparisons of non-primitives
The comparison operators `===`, `>`, `<`, `>=`, `<=` can be used to compare primitive values, and this is usually good and fine.

But when they are used to compare between non-primitives (objects, arrays, functions, etc..) they compare by reference, instead of by value. This is often done accidentally, when the programmer really meant to compare by value (using a ".equals" method, or a deep-equal library).

I propose a rule that bans all comparisons where the operands are a non-primitive type. This will catch accidents where the programmer made a mistake and meant to use ".equals" or a deep-equal library. Many code bases never have the need to compare for reference-equality, and for those that do, it is usually pretty rare, and can be done with a user-defined function such as "isSameObject(a, b)" that makes it much more explicit what check is being performed.

Questions:
How should `any` be treated?
How should generic parameters be treated?

AndreasGassmann:
#### PR checklist

- [x] Addresses an existing issue: fixes #3574
- [x] New feature
  - [x] Includes tests
- [x] Documentation update

#### Overview of change:

This rule disallows the comparison of objects.

`>`, `>=`, `<`, `<=` are only allowed if the operands are of type number.
`===` and `!==` are only allowed for operands of type `number`, `string`, `boolean` and `enum` or if either of the operands is `null` or `undefined`.

This is my first tslint rule and in general the first time I took a closer look at the typescript internals.

I'm not sure if all the type checks are bullet proof. I'm pretty sure there's a better way to do those which I haven't figured out yet.

#### CHANGELOG.md entry:

[new-rule] `strict-comparison`


AndreasGassmann:
Looks like there are some cases that I didn't think about. Some of the things I saw:

* Numeric enums should be allowed with `>`, `<`, `<=` and `>=`. (How can I find out if it's a numeric enum or not?)
* Some people do use `>`, `<`, `<=` and `>=` with strings. Should I add an option to allow that?

src/rules/noObjectComparisonRule.ts
```diff
@@ -0,0 +1,181 @@
...
+            !(
+                isEqualityOperator(node) &&
+                (isEnumType(node.left, checker) || isEnumType(node.right, checker))
+            ) &&
+            !(
+                isEqualityOperator(node) &&
+                (isNullOrUndefinedType(node.left, checker) ||
+                    isNullOrUndefinedType(node.right, checker))
+            ) &&
+            !(ctx.options.allowEquals && isEqualityOperator(node))
```
JoshuaKGoldberg: This >20 line comparison, although impressively long (would this be a record in the project?), is very difficult to read and seems to be doing a lot of checks multiple time. `isEqualityOperator(node)` is called a lot and many of the utility functions call `checker.getTypeAtLocation`.

Please split this logic out into a new function with separate conditionals so you can save the results of those checks & avoid some of the duplicate work.

src/rules/noObjectComparisonRule.ts
```diff
@@ -0,0 +1,181 @@
...
+}
+
+export class Rule extends Lint.Rules.TypedRule {
+    /* tslint:disable:object-literal-sort-keys */
+    public static metadata: Lint.IRuleMetadata = {
+        ruleName: "no-object-comparison",
+        description: "Only allow comparisons between primitives.",
+        optionsDescription: Lint.Utils.dedent`
+                One argument may be optionally provided:
+                * \`${OPTION_ALLOW_EQUAL}\` allows \`!=\` \`==\` \`!==\` \`===\` comparison between any types.`,
```
JoshuaKGoldberg: Although TSLint has traditionally had rule arguments provided as strings, that's starting to be a bit of a pain when they become more complex. This rule seems powerful enough that folks are going to want to add more options later on. Let's have the rule take in an object:

```ts
[true, {
    "allow-equal": true
}]
```

src/rules/noObjectComparisonRule.ts
```diff
@@ -0,0 +1,181 @@
...
+const OPTION_ALLOW_EQUAL = "allow-equal";
+
+interface Options {
+    allowEquals: boolean;
+}
+
+export class Rule extends Lint.Rules.TypedRule {
+    /* tslint:disable:object-literal-sort-keys */
+    public static metadata: Lint.IRuleMetadata = {
+        ruleName: "no-object-comparison",
```
JoshuaKGoldberg: > `"no-object-comparison"`

That name seems a little inaccurate, since this rule can allow some kinds of object comparisons (and sometimes that is preferable). I'm also not a fan of the `no-` prefix for rule names, as that makes it less convenient to add the opposite behavior as an option and plays with alphabetical ordering of rules. 

How about `object-comparison` as a rule name?

AndreasGassmann: I thought in this case the `no-` was ok because all options will only weaken the constraints, but they will never reverse it, so it will never "force" you to do object comparison.

But removing the `no` also works for me.

src/rules/noObjectComparisonRule.ts
AndreasGassmann: Yeah, you are right, not the most readable statement. 😅

AndreasGassmann:
I see a couple of things that are still open:

- [ ] Sometimes, enum comparisons seem incorrect (eg. looking at the failing circleci: lint step here). I was not able to reproduce this, so I'm not sure what's going on.

- [ ] `<` and `>` should be allowed between Numeric Enums, but not between String Enums (without the `allow-string-comparison` option). However, I'm not sure how to do that because it doesn't seem like those are different types in typescript.

- [ ] Rename `allow-equal` and `allow-string-comparison` because it's not clear what they do right now.
`allow-equal` could be named `allow-equal-object-comparison`
`allow-string-comparison` could be named `allow-string-ordering-comparison` (I'm not sure how to call the "ordering" operators like `>`, `<`, `>=`, `<=`).

- [ ] Corner cases including `any` `undefined` and `null` have to be handled.

AndreasGassmann:
@JoshuaKGoldberg I think this is done now. Can you take another look?

I'm not sure what to do about the failing tests in the `lint` step, because the rule works as it should.

AndreasGassmann:
Also, maybe this rule should be renamed. It does a lot more than check object comparisons. How about `strict-comparisons`?

src/rules/code-examples/objectComparison.examples.ts
```diff
@@ -0,0 +1,55 @@
...
+            const object2 = {};
+            if (isEqual(object1, object2)) {}
+        `,
+        fail: Lint.Utils.dedent`
+            const object1 = {};
+            const object2 = {};
+            if (object1 === object2) {}
+        `,
+    },
+    {
```
JoshuaKGoldberg: Missing example for `allow-string-order-comparison`?

src/rules/objectComparisonRule.ts
```diff
@@ -0,0 +1,278 @@
...
+    Object = 6,
+}
+
+const typeNames = {
+    [TypeKind.Any]: "any",
+    [TypeKind.Number]: "number",
+    [TypeKind.Enum]: "enum",
+    [TypeKind.String]: "string",
+    [TypeKind.Boolean]: "boolean",
+    [TypeKind.NullOrUndefined]: "null | undefined",
```
JoshuaKGoldberg: Error messages mentioning `null | undefined` for types that are just `undefined` are a little misleading. `--strictNullCheck` projects make a distinction between the `null` and `undefined` types. `null` and `undefined` should be separate messages & test cases.

JoshuaKGoldberg:
@AndreasGassmann `strict-comparisons` SGTM.

/cc @adidahiya for any other thoughts?

JoshuaKGoldberg:
Re the lint failures in CI: you'll need to fix and/or disable the existing failures in the codebase. Are they finding any issues with this rule, or are they just existing failures that can be fixed?

Re 2.1: it looks like TypeScript is treating enum values as objects. Is it possible to have the rule check the definition of the object to see if it's actually an enum?

test/rules/object-comparison/allow-object-equal-comparison/test.ts.lint
```diff
@@ -0,0 +1,147 @@
...
+if (2 > 1) {}
+if (2 < 1) {}
+if (2 >= 1) {}
+if (2 <= 1) {}
+if (2 == 1) {}
+if (2 === 1) {}
+if (2 != 1) {}
+if (2 !== 1) {}
+if (2 > undefined) {}
+    ~~~~~~~~~~~~~     [cannot compare type 'number' to type 'null | undefined']
```
adidahiya: Can we use sentence case for these failure strings? let's capitalize the first word

test/rules/object-comparison/allow-object-equal-comparison/test.ts.lint
```diff
@@ -0,0 +1,147 @@
...
+const h1: string | number = Math.random() > 0.5 ? 'text' : 5;
+const h2: string | number = Math.random() > 0.5 ? 'test' : 2;
+if (h1 > h2) {}
+    ~~~~~~~     [cannot use '>' comparator for type 'string']
+if (h2 > h1) {}
+    ~~~~~~~     [cannot use '>' comparator for type 'string']
+if (h1 === h2) {}
+if (h2 === h1) {}
+
+[0]: Invalid comparison
```
adidahiya: this looks unused

test/rules/object-comparison/allow-string-order-comparison/test.ts.lint
```diff
@@ -0,0 +1,149 @@
...
+if (TestStringEnum.Two > g1) {}
+
+const h1: string | number = Math.random() > 0.5 ? 'text' : 5;
+const h2: string | number = Math.random() > 0.5 ? 'test' : 2;
+if (h1 > h2) {}
+if (h2 > h1) {}
+if (h1 === h2) {}
+if (h2 === h1) {}
+
+[0]: Invalid comparison
```
adidahiya: this looks unused

AndreasGassmann:
I included the suggested changes from here and the eslint PR. I'll still have to fix the ts2.1 bug and review the errors in the tslint codebase.

adidahiya:
@AndreasGassmann it would be great to address the errors in the tslint codebase itself, but it's not strictly necessary for this PR. You could disable this new rule in the root `tslint.json` near the other rules we have listed as `false` but "we want to enable these". Enabling the rule for tslint itself could happen in another PR.

JoshuaKGoldberg:
All right, this is in! Thanks so much @AndreasGassmann for getting started on the rule implementation and working earlier this year on the initial rounds of feedback. `strict-comparisons` will be available in the next TSLint version.

rolftimmermans:
Feedback from a user who likes to use `"extends": ["tslint:all"]` to test new rules: I don't think this rule is correct in its current form.

Example code :

```function foo(s: string | undefined) {
  if (s === undefined) return "none"
  return s
}

foo(undefined)
```

Lint error with this rule enabled:
`foo.ts[2, 7]: Cannot use '===' comparator for type 'undefined'.`

To me there seems absolutely nothing wrong with testing a variable of type `string | undefined` for equality with `undefined`.

JoshuaKGoldberg:
> Feedback from a user who likes to use `"extends": ["tslint:all"]`

Love it! 💖

Could you please file a new issue for feedback? Separate issues and PRs for different topics are easier to track and search for.

rolftimmermans:
> Could you please file a new issue for feedback? Separate issues and PRs for different topics are easier to track and search for.
Absolutely! Searching through the issues my experience seems similar or identical to #4774, so I'll leave it at that. Thanks!

Issue #4333, opened 2018-12-02
no-object-literal-type-assertion rule in Reduce


zheeeng:
### Rule Suggestion

**Is your rule for a general problem or is it specific to your development style?**
It is general that constructing an array/object from empty array and object.

**What does your suggested rule do?**
Allow use assertion in reduce function by default

**List several examples where your rule could be used**
```typescript
    const groupMap = activities.reduce(
        (m, act) => {
          m[act.date] = m[act.date] ? [...m[act.date], act] : [act]

          return m
        },
        // tslint:disable-next-line:no-object-literal-type-assertion
        {} as { [key: string]: Activity[] },
      )
```

**Additional context**

JoshuaKGoldberg:
Hey @zheeeng, thanks for posting this! Could you please post a code snippet that includes the definitions of the `Activity` type/interface and the type of the `activities` variable? I'm finding it hard to get your sample to compile.

zheeeng:
@JoshuaKGoldberg 

There is a sample:
```
const arr = ['foo', 'bar']

const map = arr.reduce(
  (m, key) => {
    m[key] = 42

    return m
  },
  {} as { [key: string]: number },
)
```
tslint reports: `[tslint] Type assertion on object literals is forbidden, use a type annotation instead. [no-object-literal-type-assertion]`

It is a common way in `reduce` but now we have to suppress the lint reporting every time.

zheeeng:
@JoshuaKGoldberg We lost the accurate type of `map`, it is `{}`.

Thought we a workaround that:
```ts

const arr = ['foo', 'bar']

const map = arr.reduce<{ [key: string]: number }>(
  (m, key) => {
    m[key] = 42

    return m
  },
  {},
)
```

But when we encount Generic, we have to use assertion:

```ts
function Test<O extends { value: number }> (o: O) {

  const arr = [42, 43]

  // { value: 5 } is not assignable to generic O
  const ret =  arr.reduce<O>(
    (sum, num) => {
      sum.value = o.value + num + o.value

      return sum
    },
    // have to use assertion there
    { value: 5 } as O,
  )

  return ret
}
```

ford04:
Hey,
If I take your example mentioned above,
```
const arr = ['foo', 'bar']

const map = arr.reduce(
  (m, key) => {
    m[key] = 42 // error, no index signature.ts(7017)

    return m
  },
  {} /* as {[key:string]: any} */,
)
```
Without the cast, I get 
[...]

PR #4521, opened 2019-02-17
[no-object-literal-type-assertion] New option: "allow-arguments"

karol-majewski:
#### PR checklist

- [x] Addresses an existing issue: #4333
- [x] Enhancement
  - [x] Includes tests
- [x] Documentation update

#### Overview of change:

This option allows using a type assertion when an object literal is used as an argument. A common use case is the second parameter for `Array.prototype.reduce`, where an empty object is often used as the initial value.

#### CHANGELOG.md entry:

[new-rule-option] [`no-object-literal-type-assertion`](https://palantir.github.io/tslint/rules/no-object-literal-type-assertion/) rule accepts an option called `allow-arguments`. It allows using type assertions on object literals used inside call expressions.


palantirtech:
Thanks for your interest in [palantir/tslint](https://github.com/palantir/tslint), @karol-majewski! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

src/rules/noObjectLiteralTypeAssertionRule.ts
```diff
@@ -37,23 +45,35 @@ export class Rule extends Lint.Rules.AbstractRule {
             The compiler will warn for excess properties with this syntax, but not missing required fields.
             For example: \`const x: { foo: number } = {}\` will fail to compile, but
             \`const x = {} as { foo: number }\` will succeed.`,
-        optionsDescription: "Not configurable.",
-        options: null,
-        optionExamples: [true],
+        optionsDescription: Lint.Utils.dedent`
+            One argument may be optionally provided:
+
+            * \`${OPTION_ALLOW_ARGUMENTS}\` allows type assertions to be used on object literals inside call expressions.`,
```
JoshuaKGoldberg: Although TSLint has traditionally had rule arguments provided as strings, that's starting to be a bit of a pain when they become more complex. Let's have the rule take in an object:

```ts
[true, {
    "allow-arguments": true
}]
```

test/rules/no-object-literal-type-assertion/allow-arguments/test.ts.lint
```diff
@@ -0,0 +1,12 @@
+// Allow object literals to be asserted when used as arguments.
```
JoshuaKGoldberg: Just to make sure the `allow-arguments` option doesn't completely break the rule in the future, could you copy & paste the existing test cases into this file? They should still cause complaints for non-argument assertions.

src/rules/noObjectLiteralTypeAssertionRule.ts
karol-majewski: Done 👍 I borrowed the new schema from the `noImportSideEffectRule.ts` rule.

https://github.com/palantir/tslint/blob/9785477f26c1e0f926a5a34ba9170f1784a728c7/src/rules/noImportSideEffectRule.ts#L29-L41

test/rules/no-object-literal-type-assertion/allow-arguments/test.ts.lint
karol-majewski: Of course. Done 👍

Luixo:
@adidahiya 
Hey! I couldn't find any info about the review routine in palantir and I hope it's not a very rude question to ask, but is there any evaluated timing for the review? :)

src/rules/noObjectLiteralTypeAssertionRule.ts
```diff
@@ -37,23 +49,44 @@ export class Rule extends Lint.Rules.AbstractRule {
...
             \`const x = {} as { foo: number }\` will succeed.`,
-        optionsDescription: "Not configurable.",
-        options: null,
-        optionExamples: [true],
+        optionsDescription: Lint.Utils.dedent`
+            One option may be configured:
+
+            * \`${OPTION_ALLOW_ARGUMENTS}\` allows type assertions to be used on object literals inside call expressions.`,
+        options: {
+            type: "array",
```
adidahiya: you don't need to wrap the object in an array, you can just bring the object out as the top level `options` value

karol-majewski: Done! Thank you.

Issue #3418, opened 2017-10-30
no-duplicate-imports false positive with namespace imports


Jessidhia:
### Bug Report

- __TSLint version__: 5.8.0
- __TypeScript version__: 2.6.0
- __Running TSLint via__: CLI

#### TypeScript code being linted

```ts
import * as fooApi from 'api/foo'
import { FooContext } from 'api/foo'
```

with `tslint.json` configuration:

```json
{
  "rules": {
    "no-duplicate-imports": true
  }
}
```

#### Actual behavior

```
Multiple imports from 'api/foo' can be combined into one. (no-duplicate-imports)
```

#### Expected behavior

For the rule to ignore, at least through an option, this specific case of combining namespace imports with named imports.

While default imports can be combined with named imports (`import Foo, { Bar } from 'foobar'`), it's not possible to combine them with namespace imports other than changing all references to refer to the namespace instead.

It is sometimes useful to import certain names of a namespace for commonly used things like types, while still accessing the less commonly used exports through the namespace object.

adidahiya:
Another use case might be importing symbols of the same name from different packages:

```ts
import * as fooApi from "api/foo";
import { FooContext } from "api/foo";
import * as barApi from "api/bar";

const c: FooContext = {};

fooApi.doStuff(c);
barApi.doStuff();
```

mnn:
Yeah, [hit this error few months back](https://stackoverflow.com/questions/45669863/how-to-combine-import-all-with-others) and now again. The error message `can be combined into one` is misleading and incorrect, as demonstrated by OP (assuming we don't want to change rest of the code, especially when it's deliberate design). In my opinion there are cases this has its legitimate uses (e.g. a lot of action classes with generic names [e.g. `Set`, `Response`] to be imported with `*` and one utility "static" class very closely related to those actions).

robpaveza:
Another use case is when importing modules for side effects:

    import "./foo";
    import { IFoo, IFooType } from "./foo";

The consideration here is that when the TypeScript compiler imports disappearing types only from a module (e.g., interfaces, type aliases, enum declarations, etc.), then the import of that module disappears.  As a result, side-effects-only imports sometimes have to be emitted as well.

DavidSouther:
@robpaveza If I were coming to your codebase fresh, I'd love to see that called out explicitly:

```typescript
import {IFoo, IFooType} from './foo';
// tslint:disable-next-line:no-duplicate-any: Also import the side-effects
import "./foo";
```

[...]

PR #4524, opened 2019-02-19
Add option for "allow-namespace-imports" to noDuplicateImports rule

EdwardDrapkin:
#### PR checklist

- [X] Addresses an existing issue: #3418 

#### Overview of change:
Adds an option "allow-namespace-imports" that can be passed to the no duplicate imports rule that will make it skip over lines containing namespace imports.

This option will fix the following fairly common pattern that currently errors the rule:

```
import * as React from 'react';
import { NamedExoticComponent } from 'react';
```

With the option set, the following error is still caught:

```
import * as React from 'react';
import { NamedExoticComponent } from 'react';
import { ExoticComponent } from 'react';
```

When the new option is enabled, wildcard imports will be allowed on a separate line from named imports, but the rule will still fail for multiple lines of named imports as well as multiple wildcard imports from the same module.  To enable the new behavior pass the following object as an option to the rule in your tslint.json: `{ "allow-namespace-imports": true }`

#### CHANGELOG.md entry:

[new-rule-option] added `allow-namespace-imports ` option to `no-duplicate-imports` rule


palantirtech:
Thanks for your interest in [palantir/tslint](https://github.com/palantir/tslint), @EdwardDrapkin! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

src/rules/noDuplicateImportsRule.ts
```diff
@@ -41,21 +47,43 @@ export class Rule extends Lint.Rules.AbstractRule {
     }
 
     public apply(sourceFile: ts.SourceFile): Lint.RuleFailure[] {
-        return this.applyWithFunction(sourceFile, walk);
+        return this.applyWithFunction(sourceFile, (ctx: Lint.WalkContext<void>) =>
```
JoshuaKGoldberg: This looks like it works, but to simplify a bit, you can pass an options object as a third argument to `applyWithFunction`. That'll allow you to access it on `ctx.options` instead of passing a new parameter.

src/rules/noDuplicateImportsRule.ts
```diff
@@ -29,9 +31,13 @@ export class Rule extends Lint.Rules.AbstractRule {
         rationale: Lint.Utils.dedent`
             Using a single import statement per module will make the code clearer because you can see everything being imported
             from that module on one line.`,
-        optionsDescription: "Not configurable",
-        options: null,
-        optionExamples: [true],
+        optionsDescription: Lint.Utils.dedent`
+            "${OPTION_ALLOW_SEPARATE_NAMESPACE_IMPORTS}" allows you to import namespaces on separate lines.`,
+        options: {
+            type: "string",
```
JoshuaKGoldberg: Although TSLint has traditionally had rule arguments provided as strings, that's starting to be a bit of a pain when they become more complex. Let's have the rule take in an object:

```ts
[true, {
    "allow-namespace-imports": true
}]
```

`memberOrderingRule.ts` is a good example of this.

EdwardDrapkin:
Thanks for the feedback.

I believe I have addressed all your issues, but for brevity, here are my comments on your comments.

1. I updated the call to pass options per your suggestion.  Didn't know this was possible, thanks for the help.

2. I figured out how to write tests and added a test case for this option because the option is a bit more complex.  I believe I covered all of the possible permutations of error/not error when the rule is enabled.

3. I updated the option to be an object with the format `{ "allow-namespace-imports": boolean }`.

4.  The rule now tracks named imports separately from namespace imports.  The benefit is now that multiple wildcard imports will error out alongside multiple named imports.  You also get a separate, special error message specifically for multiple wildcard imports.

Thanks for your help!!

EdwardDrapkin:
Looks like all the test passed this time 🎉

JoshuaKGoldberg:
Yes, the changelog entry is a bit too long haha. I'll go ahead and edit it.

huyz:
Is there a reason why this option isn't the default?

JoshuaKGoldberg:
@huyz it happened not to be in this PR, but that's a good idea. Perhaps it should - would you mind filing a new issue?

EdwardDrapkin:
> Is there a reason why this option isn't the default?

I didn't want to presume to change default behavior is the only reasoning I had to make it opt-in instead of opt-out.

Issue #2000, opened 2017-01-06
arrow-call-signature Not working with arrow function missing type definition


cjbarth:
### Bug Report

- __TSLint version__: `4.3.1`
- __TypeScript version__: `2.1.4`
- __Running TSLint via__: CLI

#### TypeScript code being linted

```ts
let t: string[] = [
 "hi",
 "there"
];
let u: string = t.find((item: string) => { return item === "hi"; });
```

with `tslint.json` configuration:

```json
{
 "rules": {
  "typedef": [
   true,
   "call-signature",
   "arrow-call-signature",
   "parameter",
   "arrow-parameter",
   "property-declaration",
   "variable-declaration",
   "member-variable-declaration"
  ]
 }
}
```

#### Actual behavior

I don't get a warning about the arrow function call signature not having a type defined.

#### Expected behavior

I should get a warning about the arrow function call signature not having a type defined.

andy-hanson:
Looks like this behavior is intentional: it doesn't lint arrow functions inside CallExpressions. (https://github.com/palantir/tslint/blob/master/src/rules/typedefRule.ts#L83)
Did you really want to have to define the return type in every callback?

nchen63:
@YuichiNukiyama: was it intentional to not warn on callbacks? I see one of the revisions in #1284 which disables the warning for the line @andy-hanson mentioned. If so, what was the thought behind that?

YuichiNukiyama:
@nchen63 
@andy-hanson is right. When defining arrow function in some method, outer method checks type of retern value. So, there is no point in checking the type.

nchen63:
@YuichiNukiyama I thought that the point of the rule wasn't to declare types because they were necessary, but to do it to be explicit, as a matter of style

cjbarth:
He is an example of where linting would be valuable. Currently, no linting error is generated:

```
let t: string[] = [
 "hi",
 "there"
];
let u: number[] = t.map((item: string) => { return item === "hi"; });
```

Having said that, it does appear that such a problem is picked up by the TypeScript compiler.

JoshuaKGoldberg:
Sounds like the general thought is that the rule should be given an opt-in option to allow this behavior. Perhaps `"child-arrow-call-signatures"`? Any PR that adds the option should document why it isn't enabled by default (as per the above discussion).

VincentLanglet:
I agree with @nchen63 ; this rule is not to declare types because they are necessary, but to document the code.

[...]

Issue #813, opened 2015-11-21
typedef: split "callback-parameter" from "parameter"


leonyu:
There should be an option to split callback parameter from `parameter`  `typedef`. One of the reasons for this is that type inference can flow the other way in Typescript, for example in the following code: 

``` ts
let a = document.createElement('a');
a.addEventListener('click', (event) => {
  // something
});
```

the type of `event` can be inferred as `MouseEvent` by the compiler, however given the current `typedef: parameter` option, user would be forced to put something there, which usually is `Event`. `Event` is worse than what could be inferred. 

This also occurs for the following: 

``` ts
let p = new Promise(function(resolve, reject){
  // do something
});
```

Both `resolve` and `reject` would be inferences by the compiler. However with `typedef: parameter`, user could be easy overwriting with something that's worse, such as `Function`

adidahiya:
In this proposal, do you want the "parameter" option to only apply to named functions then? And prefer type inference for anonymous functions?

This sounds similar to #669 and my general advice would be to try the `--noImplicitAny` compiler option rather than the `typedef` rule if you want to leverage type inference. But i'll leave this issue open for further discussion.

leonyu:
That would only work for new code, not copy and pasted code.

Not sure if it is possible to look at all callbacks with dual declarations and see if one side is better than the other. If it is heuristically display it. 

* Display if the better definition is from `.d.ts`
* Display if the better definition is in user's code somewhere else
* Don't display if better code in user's code, and user is calling `.d.td`

erictsangx:
+1 typescript is smart enough to infer the types but tslint keep warning me missing ``typedef``, especially callbacks.
It would be great to split them.

adidahiya:
@erictsangx is your use of the typedef rule also for legacy code that has been copied over to TS as @leonyu mentioned above? I'm trying to understand the motivation for using the rule. I generally don't think it's useful since the compiler does a lot of inference work for you and noImplicitAny forces typedefs everywhere else.

JKillian:
@adidahiya Do you mean the motivation for `typedef` in general or for a `callback-parameter` option? There are a few reasons I see for the rule:

```ts
// what in the world is myVar???
let myVar = mysteriousObj.mysteriousFunc();
// ah, it's a spaceship!
let myVar: Spaceship = mysteriousObj.mysteriousFunc();
```

```ts
// all good, right?
function doubleExp(a: number, b: number, c: number): number {
  let step1 = a ** b;
[...]

Issue #227, opened 2014-10-14
Inconsistency of typedef:parameter with lambdas


mihhail-lapushkin:
`array.forEach(element => {})` does not complain
`array.forEach((element) => {})` complains `expected to have a typedef.`
`array.sort((a,b) => {})` complains `expected to have a typedef.`

rokerkony:
hi guys, is it possible to look into it again?

I would like to enforce typedef for lambdas...
`.filter((e: Error): boolean => this.belongsToChannel(e))`

there is no way to enforce *: boolean* in this example ... or is there such a rule?

JoshuaKGoldberg:
☠️ *TSLint's time has come!* ☠️ 

TSLint is no longer accepting most feature requests per #4534. See **[typescript-eslint.io](https://typescript-eslint.io)** for the new, shiny way to lint your TypeScript code with **ESLint**. ✨ 

It was a pleasure open sourcing with you all!

JoshuaKGoldberg:
🤖 Beep boop! 👉  **TSLint is deprecated** 👈 _(#4534)_  and you should **switch to [typescript-eslint](https://typescript-eslint.io)**! 🤖 

🔒 This issue is being locked to prevent further unnecessary discussions. Thank you! 👋

PR #4533, opened 2019-02-23
Improve typedef for callback used as parameter

VincentLanglet:
#### PR checklist

- [x] Addresses an existing issue: #227 ; #2000
- [x] Includes tests

#### Overview of change:
Typedef `arrow-call-signature` now return error for
```
someFunc(() => {});
```

Actually, with all typedef rules actives.
```
const A = () => {} // error (return type)
someFunc(() => {}); // no error
someFunc((n) => {}); // error (param type)
setTimeout(function() {}, 1000); // error (return type)
```
After the PR, error everywhere ; for consistency.

Here https://github.com/palantir/tslint/issues/2000#issuecomment-435629910, a suggestion is a `child-arrow-call-signature`. That's a possibility, but we have to (for consistency) add a `child-arrow-parameter` too, which is also a suggestion here https://github.com/palantir/tslint/issues/813.
We, then, have to decide before if we want also a `child-parameter` and a `child-call-signature`.

I can do it if you prefer. 
Personally I don't think all these options are necessary. The typedef is a _style rule_ for people who want to type the code (more like a documentation) even if it's not necessary thanks to inference. If you don't want to be redundant and infer as much type you can, just use the `noImplicitAny` option in `tsconfig`.

#### CHANGELOG.md entry:

[enhancement] `typedef` rule `arrow-call-signature` option is more consistent in reporting errors on lambdas and will flag more violations that were missed in the previous rule implementation.

### DIFF
--- a/test/rules/typedef/arrow-call-signature/test.ts.lint
+++ b/test/rules/typedef/arrow-call-signature/test.ts.lint
@@ -18,7 +18,13 @@ var obj = {
 }
 
 setTimeout(() => {}, 1000);
+           ~~       [expected arrow-call-signature to have a typedef]
 setTimeout(function() {}, 1000);
 
+const foo = n => n+1
+          ~~~~       [expected arrow-call-signature to have a typedef]
+
 someFunc(n => n+1);
+        ~~~       [expected arrow-call-signature to have a typedef]
 someFunc(n => {});
+        ~~~       [expected arrow-call-signature to have a typedef]
--- a/src/rules/typedefRule.ts
+++ b/src/rules/typedefRule.ts
@@ -148,7 +148,7 @@ class TypedefWalker extends Lint.AbstractWalker<Options> {
     }
 
     private checkArrowFunction({ parent, parameters, type }: ts.ArrowFunction): void {
-        if (parent.kind !== ts.SyntaxKind.CallExpression && !isTypedPropertyDeclaration(parent)) {
+        if (!isTypedPropertyDeclaration(parent)) {
             this.checkTypeAnnotation("arrow-call-signature", parameters, type);
         }
     }
--- a/test/rules/typedef/all/test.ts.lint
+++ b/test/rules/typedef/all/test.ts.lint
@@ -76,12 +76,15 @@ class TsLint {
 }
 
 setTimeout(() => {}, 1000);
+           ~~            [expected arrow-call-signature to have a typedef]
 setTimeout(function() {}, 1000);
                    ~~            [expected call-signature to have a typedef]
 
 someFunc(n => n+1);
+        ~~~            [expected arrow-call-signature to have a typedef]
          ~        [expected arrow-parameter: 'n' to have a typedef]
 someFunc(n => {});
+        ~~~            [expected arrow-call-signature to have a typedef]
          ~        [expected arrow-parameter: 'n' to have a typedef]
 
 class A {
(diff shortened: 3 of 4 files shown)


VincentLanglet:
@JoshuaKGoldberg
The question "Is it breaking change or is it bug fix" is in a way kinda debatable.

There was no consistency in this example
```
const A = () => {} // error (return type)
someFunc(() => {}); // no error
someFunc((n) => {}); // error (param type)
setTimeout(function() {}, 1000); // error (return type)
```
So you can think it was a bug to not detect the error for callback return type.

But I can understand the "breaking change" point of view.

VincentLanglet:
@JoshuaKGoldberg @adidahiya Is it a breaking change or a bug fix ? I think it's debatable.

JoshuaKGoldberg:
> breaking change or a bug fix

Both? 😄 

Even if this is considered a bug, it's a pretty big change. If a project already enables the rule under the expectation that scenario of arrow lambdas passed to methods shouldn't be flagged, then it starts flagging them, it'd become more difficult to migrate to the next TSLint version (even if this is desirable).

test/rules/typedef/arrow-call-signature/test.ts.lint
```diff
@@ -18,7 +18,13 @@ var obj = {
 }
 
 setTimeout(() => {}, 1000);
+           ~~       [expected arrow-call-signature to have a typedef]
 setTimeout(function() {}, 1000);
 
+const foo = n => n+1
+          ~~~~       [expected arrow-call-signature to have a typedef]
```
adidahiya: why does the failure range extend to the `=`? that seems a little off

test/rules/typedef/arrow-call-signature/test.ts.lint
```diff
@@ -18,7 +18,13 @@ var obj = {
...
+
 someFunc(n => n+1);
+        ~~~       [expected arrow-call-signature to have a typedef]
```
adidahiya: same here, feels strange to include the `(` in the failure range

VincentLanglet: It's made for (n) => n+1
.....................~~~       [expected arrow-call-signature to have a typedef]

The Node is only `n`.
But the code use `this.addFailure(location.pos - 1, location.end + 1, failure);`
I didn't touch this part

test/rules/typedef/arrow-call-signature/test.ts.lint
VincentLanglet: In `const foo = n => n+1` location.pos = 11
in `const foo = (n => n+1)` location.pos = 13

I didn't touch this part

VincentLanglet:
@JoshuaKGoldberg @adidahiya I agree it will be easier with something like `arrow-callback-call-signature` option. But it does not make any sense having an option for callback in arrow function, and not in standard function, or an option for callback return type in arrow function, and not parameter type in callback function.

A bug fix is always a breaking change. Before the rule had false positive, now it correctly find the errors.

But if you prefer, you can wait for the next major version. But in this case I have a question. When is it planned ? There is already a lot of Breaking changes waiting for it.
https://github.com/palantir/tslint/pulls?q=is%3Aopen+is%3Apr+label%3A%22Type%3A+Breaking+Change%22

JoshuaKGoldberg:
@VincentLanglet great question! 😄 just filed #4811 to discuss.

VincentLanglet:
@JoshuaKGoldberg Great ! Is this PR ready for 6.0.0 version ?

JoshuaKGoldberg:
Deferring to @adidahiya for review now that we're ready for breaking changes (#4811).

adidahiya:
thanks for the PR @VincentLanglet!

PR #4540, opened 2019-02-24
Fix #4530: no-string-throw fixer creates syntax errors

### LINKED ISSUE #4530: Bug: no-string-throw fixer creates syntax errors
### Bug Report

- __TSLint version__: 5.12.1
- __TypeScript version__: 2.9.2
- __Running TSLint via__: Webpack

```ts
throw('component requires CSS height property');
```

```json
{
}
```

#### Actual behavior
Should fix or ignore the code

#### Expected behavior
Generates this invalid code:
thrownew; Error(('component requires CSS height property'));

rrogowski:
#### PR checklist

- [x] Addresses an existing issue: #4530 
  - [x] Includes tests
- [ ] Documentation update

#### Overview of change:

Insert whitespace after throw keyword (if not present) when fixing `no-string-throw`.

**Input**

throw('component requires CSS height property');

**Fix (Before)**

thrownew Error(('component requires CSS height property'));

**Fix (After)**

throw new Error(('component requires CSS height property'));

- The double parentheses behavior is unrelated to this change, and existed before this bugfix.
    - I am happy to create an issue for this if so desired.

#### CHANGELOG.md entry:

[bugfix] `no-string-throw` fix inserts whitespace if not present after `throw` keyword


test/rules/no-string-throw/test.ts.lint
```diff
@@ -28,3 +28,6 @@ let i: never = () => throw `some template string`;
 const someVariable = 123;
 let j: never = () => throw `template with ${someVariable} string`;
                      ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~ [Throwing plain strings (not instances of Error) gives no stack traces]
+
+throw('component requires CSS height property');
```
JoshuaKGoldberg: A few more test cases:

```ts
throw'component...'
throw(('component...'))
throw/**/'component'
```

src/rules/noStringThrowRule.ts
```diff
@@ -73,10 +73,20 @@ function walk(ctx: Lint.WalkContext<void>): void {
...
             const { expression } = node;
             if (expression !== undefined && isString(expression)) {
-                ctx.addFailureAtNode(node, Rule.FAILURE_STRING, [
+                const fix = [
                     Lint.Replacement.appendText(expression.getStart(sourceFile), "new Error("),
                     Lint.Replacement.appendText(expression.getEnd(), ")"),
-                ]);
+                ];
+
+                const hasWhitespaceAfterThrow = /^throw\s+/.test(node.getText());
```
JoshuaKGoldberg: We normally try not to use regular expressions on `getText` to check for positioning. It's a bit cleaner and faster to use `.pos` when possible.

How about checking `node.expression.pos`? It should be `"throw.length"` greater than `node.pos`.

test/rules/no-string-throw/test.ts.lint
rrogowski: Done.

src/rules/noStringThrowRule.ts
rrogowski: I ended up using the `getStart()` method and doing a quick calculation to check if there are any characters preceding the string literal directly after the throw.

rrogowski:
@JoshuaKGoldberg finished those changes you requested.

Oddly enough, I'm seeing that my implementation is failing on CircleCI, but all tests are passing locally on my end. I suspect it has to do with the fact that a Replacement fix is applied to the same position (`expression.getStart`). Any thoughts as to why there might be this disparity?

JoshuaKGoldberg:
Very spooky 👻... Have you tried installing a version of TypeScript that's failing in CI, such as 2.7 or 3.0? Perhaps `.pos` instead of `.getStart()` might help?

Alternately, you could try reordering the lines? The TSLint application of fixes is known to be not great at overlapping fixes, but that doesn't seem like it should be affecting these?

rrogowski:
I have tried all versions of TypeScript that are failing on CI: `2.1`, `2.4`, `2.7`, `2.8`, `2.9`, `3.0`, `rc`, and `next`. All tests are all passing locally with each of these TypeScript versions installed, yet the only version passing on CI is `next`. Using `.pos` instead of `.getStart` did not remedy the situation either. I will continue to investigate this, but I am a bit stumped...

rrogowski:
To clarify, my workflow is this.
1. Checkout to fresh version of this branch
2. `yarn`
3. `yarn compile`
4. `yarn test` (all tests pass)
5. `yarn add typescript@2.1`
6. `yarn test` (again, all tests pass)
7. Repeat steps 5 & 6 for all failing TypeScript versions

Am I missing something?

rrogowski:
@JoshuaKGoldberg ready for your pass.

Issue #4518, opened 2019-02-14
(no-trailing-whitespace) flags leading empty line as error


itanex:
### Bug Report

- __TSLint version__: 5.11.0
- __TypeScript version__: 2.9.2
- __Running TSLint via__:  CLI / (Visual Studio - does not show this error)

#### TypeScript code being linted
Multiple classes with a leading line that is empty. This line only contains line endings when analyzed (e.g. \r\n)
```ts

export class User {
    userName: string;
}
```

with `tslint.json` configuration:

```json
{
  "defaultSeverity": "error",
  "extends": "tslint:recommended",
  "jsRules": {},
  "rules": {
    "quotemark": [
      true,
      "single",
      "avoid-escape"
    ],
    "member-access": [
      true,
      "no-public"
    ],
    "object-literal-sort-keys": [
      false
    ],
    "object-literal-shorthand": [
      true,
      "never"
    ],
    "ordered-imports": [
      false
    ],
    "trailing-comma": [
      true,
      {
        "esSpecCompliant": true
      }
    ]
  },
  "rulesDirectory": [],
  "linterOptions": {
    "exclude": [
      "src/test.ts",
      "src/main.ts",
      "src/polyfills.ts"
    ]
  }
}
```

#### Actual behavior
Multiple results identical to the following

`ERROR: ClientApp/src/app/models/ActionItem.ts:1:1 - trailing whitespace`

#### Expected behavior
No results generated

JoshuaKGoldberg:
@itanex thanks for the bug report! I tried reproducing this locally but couldn't. Is the project you're experiencing this with open source, or failing that, could you upload `ActionItem.ts` or another file it's happening on?

One guess would be that the file has some `\r\n` lines and some `\n`.

itanex:
@JoshuaKGoldberg Yes the file line endings are windows line endings (\r\n).

I have uploaded the `ActionItem.ts` file along with the `tslint.json`. 

Please note that the `ActionItems.ts` file is generated through TypeWriter, so simply deleting the line is not a viable option, as this would need to be done anytime the file is regenerated.

JoshuaKGoldberg:
This was a fun one to debug! The root issue is that `getLineRanges` from the `tsutils` library is reporting that the first line in the `ActionItem.ts` file has a `contentLength` of 1 instead of 0. Filed https://github.com/ajafff/tsutils/issues/94 to track.

[...]

PR #4543, opened 2019-02-25
fix for #4518: (no-trailing-whitespace) flags leading empty line as error

rrogowski:
#### PR checklist

- [x] Addresses an existing issue: #4518 
  - [x] Includes tests
- [ ] Documentation update

#### Overview of change:

Temporary fix to handle files starting with a byte order mark (see https://github.com/ajafff/tsutils/issues/94).

#### CHANGELOG.md entry:

[bugfix] `no-trailing-whitespace` no longer flags files starting with a byte order mark

### DIFF
--- a/src/rules/noTrailingWhitespaceRule.ts
+++ b/src/rules/noTrailingWhitespaceRule.ts
@@ -92,10 +92,7 @@ function walk(ctx: Lint.WalkContext<Options>) {
         if (
             match !== null &&
             !(ctx.options.ignoreBlankLines && match.index === 0) &&
-            // See https://github.com/palantir/tslint/issues/4518
-            // and https://github.com/ajafff/tsutils/issues/94
-            // TODO: Remove this check once tsutils handles Byte Order Markers correctly.
-            match[0].charCodeAt(0) !== ZERO_WIDTH_NO_BREAK_SPACE
+            (match[0] !== String.fromCharCode(ZERO_WIDTH_NO_BREAK_SPACE))
         ) {
             possibleFailures.push({
                 end: line.pos + line.contentLength,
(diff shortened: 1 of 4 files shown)


src/rules/noTrailingWhitespaceRule.ts
```diff
@@ -87,7 +89,14 @@ function walk(ctx: Lint.WalkContext<Options>) {
     for (const line of getLineRanges(sourceFile)) {
         // \s matches any whitespace character (equal to [\r\n\t\f\v ])
         const match = text.substr(line.pos, line.contentLength).match(/\s+$/);
-        if (match !== null && !(ctx.options.ignoreBlankLines && match.index === 0)) {
+        if (
+            match !== null &&
+            !(ctx.options.ignoreBlankLines && match.index === 0) &&
+            // See https://github.com/palantir/tslint/issues/4518
+            // and https://github.com/ajafff/tsutils/issues/94
+            // TODO: Remove this check once tsutils handles Byte Order Markers correctly.
```
JoshuaKGoldberg: You can remove this TODO; see discussion in the tsutils issue.

src/rules/noTrailingWhitespaceRule.ts
```diff
@@ -87,7 +89,14 @@ function walk(ctx: Lint.WalkContext<Options>) {
...
+            match[0].charCodeAt(0) !== ZERO_WIDTH_NO_BREAK_SPACE
```
JoshuaKGoldberg: Missing test case: what if the line starts with a BOM but also has trailing whitespace?

src/rules/noTrailingWhitespaceRule.ts
rrogowski: Got it!

src/rules/noTrailingWhitespaceRule.ts
rrogowski: Thanks, I've handled this case now!

rrogowski:
@JoshuaKGoldberg suggestions have been addressed. Ready for your review.

PR #4546, opened 2019-02-26
Implement 4201: add option to ignore shadowed underscore

### LINKED ISSUE #4201: no-shadowed-variable should ignore shadowed `_`-params
Having a few nested arrow functions that throw away the argument shouldn't be considered shadowing. Bad taste perhaps. But not shadowing.

rrogowski:
#### PR checklist

- [x] Addresses an existing issue: fixes #4201
  - [x] Includes tests
- [x] Documentation update

#### Overview of change:

Add option to ignore shadowed underscore.

**Usage**

```json
{
  "rules": {
    "no-shadowed-variable": [true, {"underscore": false}]
  }
}
```

**Before**

```javascript
import _ = require('lodash');
function foo(_) {
             ~ [Shadowed name: '_']
    console.log(_);
}
```

**After**

```javascript
import _ = require('lodash');
function foo(_) {
    // No lint error
    console.log(_);
}
```

Have I made all the necessary changes to documentation?

#### CHANGELOG.md entry:

[enhancement]: added option to `no-shadowed-variable` to ignore shadowed underscores

### DIFF
--- a/src/rules/noShadowedVariableRule.ts
+++ b/src/rules/noShadowedVariableRule.ts
@@ -63,8 +63,8 @@ export class Rule extends Lint.Rules.AbstractRule {
         `,
         optionsDescription: Lint.Utils.dedent`
             You can optionally pass an object to disable checking for certain kinds of declarations.
-            Possible keys are \`"class"\`, \`"enum"\`, \`"function"\`, \`"import"\`, \`"interface"\`, \`"namespace"\`, \`"typeAlias"\`
-            and \`"typeParameter"\`. Just set the value to \`false\` for the check you want to disable.
+            Possible keys are \`"class"\`, \`"enum"\`, \`"function"\`, \`"import"\`, \`"interface"\`, \`"namespace"\`, \`"typeAlias"\`,
+            \`"typeParameter"\` and \`"underscore"\`. Just set the value to \`false\` for the check you want to disable.
             All checks default to \`true\`, i.e. are enabled by default.
             Note that you cannot disable variables and parameters.
 
@@ -101,6 +101,7 @@ export class Rule extends Lint.Rules.AbstractRule {
                 typeAlias: { type: "boolean" },
                 typeParameter: { type: "boolean" },
                 temporalDeadZone: { type: "boolean" },
+                underscore: { type: "boolean" },
             },
         },
         optionExamples: [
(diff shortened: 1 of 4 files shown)


src/rules/noShadowedVariableRule.ts
```diff
@@ -63,8 +63,8 @@ export class Rule extends Lint.Rules.AbstractRule {
         `,
         optionsDescription: Lint.Utils.dedent`
             You can optionally pass an object to disable checking for certain kinds of declarations.
-            Possible keys are \`"class"\`, \`"enum"\`, \`"function"\`, \`"import"\`, \`"interface"\`, \`"namespace"\`, \`"typeAlias"\`
-            and \`"typeParameter"\`. Just set the value to \`false\` for the check you want to disable.
+            Possible keys are \`"class"\`, \`"enum"\`, \`"function"\`, \`"import"\`, \`"interface"\`, \`"namespace"\`, \`"typeAlias"\`,
+            \`"typeParameter"\` and \`"underscore"\`. Just set the value to \`false\` for the check you want to disable.
```
JoshuaKGoldberg: The rest of the types in this list are node types, whereas `"underscore"` refers to the text of the variable's name.

```suggestion
            , and \`"typeParameter"\`. You can also pass \`"underscore\`" to ignore variable names that begin with `_`. Just set the value to \`false\` for the check you want to disable.
```

JoshuaKGoldberg: _(I think that suggested change will end up with a duplicate `,` by the way...)_

test/rules/no-shadowed-variable/ignore-underscore/test.ts.lint
```diff
@@ -0,0 +1,5 @@
+import _ = require('lodash');
+
+function foo(_) {
```
JoshuaKGoldberg: What about multiple unused parameters? The rule should also apply to names like `_a`, `_match`, etc., right?

```ts
function foo(_a, _b) { }
```

TypeScript doesn't allow duplicate identifiers of the same name unfortunately.

rrogowski: Thanks, forgot about this case!

src/rules/noShadowedVariableRule.ts
rrogowski: Good suggestion.

rrogowski:
@JoshuaKGoldberg suggestions have been addressed. Ready for your review.

Issue #3801, opened 2018-03-29
no-unbound-method warns when deleting property


tvuotila:
### Bug Report

- __TSLint version__: 5.9.1
- __TypeScript version__: 2.4.2
- __Running TSLint via__: CLI

#### TypeScript code being linted

```ts
const a = {
  getText() {
    return 'text';
  }
};

delete a.getText;
```

with `tslint.json` configuration:

```json
{
  "rules": {
    "no-unbound-method": true
  }
}
```

#### Actual behavior

Error `Avoid referencing unbound methods which may cause unintentional scoping of 'this'.`

#### Expected behavior
No error. I am deleting a property. There is no risk on deleting unbound method compared to other kind of properties.

piotrgajow:
Hi, I think I could fix this issue by adding `allow-delete` parameter similar to what I have done previously for `typeof` operator. What do you think about such solution?

JoshuaKGoldberg:
@piotrgajow that sounds great, yes please! 😄

piotrgajow:
Ok, I have made PR for this change.

JoshuaKGoldberg:
Fxed by #4548! ✨

PR #4548, opened 2019-02-26
Add `allow-delete` option to `no-unbound-method` rule

piotrgajow:
#### PR checklist

- [x] Addresses an existing issue: #3801
- [x] enhancement
  - [x] Includes tests
- [x] Documentation update

#### Overview of change:

Added configuration option to `no-unbound-method` that does allows not reporting errors in `delete` statements `delete obj.method;`

#### CHANGELOG.md entry:

[new-rule-option] Added `allow-delete` option to `no-unbound-method` rule

### DIFF
--- a/src/rules/noUnboundMethodRule.ts
+++ b/src/rules/noUnboundMethodRule.ts
@@ -66,6 +66,7 @@ export class Rule extends Lint.Rules.TypedRule {
             The object may have three properties:
             
             * "${OPTION_IGNORE_STATIC}" - to ignore static methods.
+            * "${OPTION_ALLOW_DELETE}" - ignore methods referenced in a delete expression.
             * "${OPTION_ALLOW_TYPEOF}" - ignore methods referenced in a typeof expression.
             * "${OPTION_WHITELIST}" - ignore method references in parameters of specifed function calls.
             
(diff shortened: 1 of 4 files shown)


src/rules/noUnboundMethodRule.ts
```diff
@@ -62,6 +66,7 @@ export class Rule extends Lint.Rules.TypedRule {
             The object may have three properties:
```
JoshuaKGoldberg: Since it's no longer three, and keeping the number up-to-date is a bit of a pain...

```suggestion
             The object may have the following properties:
```

test/rules/no-unbound-method/whitelist/test.tsx.lint
```diff
@@ -69,4 +69,12 @@ Validators.compose(Validators.required);
 (await someMethod())(c.method);
                      ~~~~~~~~ [0]
 
+const a = {
+  getText() {
+    return 'text';
+  }
+};
+
+delete a.getText;
```
JoshuaKGoldberg: This test should also be in `test/rules/no-unbound-method/default/test.ts.lint` to make sure the option isn't enabled by default.

piotrgajow: I have added the test to the default test suite

JoshuaKGoldberg:
Super, thanks @piotrgajow!

PR #4560, opened 2019-02-28
Address 3180: Rule file-header should have an option to allow // comments

### LINKED ISSUE #3180: Rule file-header should have an option to allow // comments
Rule file-header is not working. Even if particular header in a file like  Copyright \\d{4} didn't add in the file or any wrong text adds, No Lint errors are getting.

- __TSLint version__: ~5.3.2
- __TypeScript version__: ~2.3.3
- __Running TSLint via__: angular CLI 

```ts
"file-header": [
    true,
    "Copyright \\d{4}"
  ]
```

with `tslint.json` configuration: tried in 2 different ways

``` json
{
  "rulesDirectory": [
    "node_modules/codelyzer"
  ],
  "rules": {
   // lint
  },
  "file-header": [
    true,
    "Copyright \\d{4}"
  ]
}

```

```json
{
  "rulesDirectory": [
    "node_modules/codelyzer"
  ],
  "rules": {
   // lint,
   "file-header": [
    true,
    "Copyright \\d{4}"
  ]
  },
  
}

```

#### Actual behavior
TS lint is not validating wheather the header text is present or not present
#### Expected behavior
TSLint is failing to validate the file header text

rrogowski:
#### PR checklist

- [x] Addresses an existing issue: fixes #3180 
  - [x] Includes tests
- [x] Documentation update

#### Overview of change:

**Config**
```json
{
  "rules": {
    "file-header": [true, "Good header", "", "allow-multiline"]
  }
}
```

**Before**

```typescript
// **********************************
~nil [missing file header]
// Good header
// **********************************
```

**After**
```typescript
// **********************************
// Good header
// **********************************

// No lint error for missing file header
```

#### CHANGELOG.md entry:

[enhancement] added `allow-multiline` option to `file-header` rule to allow header to consist of consecutive single-line comments


src/rules/fileHeaderRule.ts
```diff
@@ -32,7 +33,9 @@ export class Rule extends Lint.Rules.AbstractRule {
             The second argument, which is optional, is a string that should be inserted as a header comment
             if fixing is enabled and no header that matches the first argument is found.
             The third argument, which is optional, is a string that denotes whether or not a newline should
-            exist on the header.`,
+            exist on the header.
+            The fourth argument, which is optional, is a string that denotes whether or not a header is
+            allowed to consist of consecutive single-line comments.`,
         options: {
```
JoshuaKGoldberg: I know this is a bit of a pain, but for the same reason as a few other PRs, we should really switch this rule to allowing an options object. Strings directly in the options list can sometimes intersect with package names that happen to be the same.

`importBlacklistRule.ts` is a good example of one that allows either an array of items or a options object.

JoshuaKGoldberg: Yeah, this area is a little tricky... The suggestion is to allow two forms of configuration:

1. [string, string, string] // the current one without `enforce-trailing-newline`
2. [object] // a new form that has existing options as well as `enforce-trailing-newline`

Starting text suggestion for the documentation:

```text
A single object may be passed in for configuration that must contain:

* `match`: a regular expression that all headers should match

Any of the following optional fields may also be provided:

* `allow-single-line-comments`: whether `//` should be considered file headers in additional to `/*` comments
* `default`: text to add for file headers when running in `--fix` mode
* `enforce-trailing-newline`: a boolean for whether a newline must follow the header

The rule will also accept array of strings as a legacy form of options, though the object form is recommended. <insert existing description here>
```

src/rules/fileHeaderRule.ts
```diff
@@ -20,6 +20,7 @@ import * as ts from "typescript";
 import * as Lint from "../index";
 
 const ENFORCE_TRAILING_NEWLINE = "enforce-trailing-newline";
+const ALLOW_MULTILINE = "allow-multiline";
```
adidahiya: I don't think this is a good option name. The rule _currently_ supports block comments, which are typically "multiline". So I think this should be called `allow-singleline`

src/rules/fileHeaderRule.ts
rrogowski: Got it.

src/rules/fileHeaderRule.ts
rrogowski: Option is called `allowSingleLineComments`.

rrogowski:
@adidahiya @JoshuaKGoldberg ready for second review.

test/rules/file-header/good-allow-single-line-comments/tslint.json
```diff
@@ -0,0 +1,8 @@
+{
+    "rules": {
+      "file-header": [true, {
+        "match": "Good header",
+        "allowSingleLineComments": true
```
adidahiya: none of our options use camelCase, we only use kebab-case

src/rules/fileHeaderRule.ts
```diff
@@ -21,36 +21,90 @@ import * as Lint from "../index";
...
         optionsDescription: Lint.Utils.dedent`
+            A single object may be passed in for configuration that must contain:
+
+            * \`match\`: a regular expression that all headers should match
+
+            Any of the following optional fields may also be provided:
+
+            * \`allowSingleLineComments\`: a boolean for whether \`//\` should be considered file headers in addition to \`/*\` comments
+            * \`default\`: text to add for file headers when running in \`--fix\` mode
+            * \`enforceTrailingNewline\`: a boolean for whether a newline must follow the header
```
adidahiya: this is not the right option name. see line 22

test/rules/file-header/good-allow-single-line-comments/tslint.json
rrogowski: Woops, I thought I saw an example of options using camelCase. Fixed.

src/rules/fileHeaderRule.ts
rrogowski: Fixed.

rrogowski:
@adidahiya Thanks for your comments. They have been addressed. Ready for another pass.

PR #4563, opened 2019-03-02
Added an overloads option in completed-docs

### LINKED ISSUE #4416: completed-docs should have an option to check every method/function overload
### Feature request for `completed-docs` rule
Is it possible to enforce (add option if does not exist) jsDoc comment for function/method overloading?

JoshuaKGoldberg:
#### PR checklist

- [x] Addresses an existing issue: fixes #4416
  - [x] Includes tests
- [x] Documentation update

#### Overview of change:

Opt-in (false by default) boolean flag that can be applied to functions and methods to require their overloads each receive their own documentation comment.

Also deleted a bunch of unnecessary `tsconfig.json` files that were still in the tests directory for completed-docs. I should have removed them when converting it to TypeScript.

#### CHANGELOG.md entry:

[new-rule-option] `completed-docs`: `overloads` option to enforce documentation of each overload signature

### DIFF
--- a/src/rules/completedDocsRule.ts
+++ b/src/rules/completedDocsRule.ts
@@ -592,7 +592,7 @@ function getSiblings(node: ts.Node) {
         return [...parent.statements];
     }
 
-    return parent.getChildren()!;
+    return parent.getChildren();
 }
 
 function isGetAccessor(node: ts.Node): node is ts.GetAccessorDeclaration {
--- a/src/rules/completedDocsRule.ts
+++ b/src/rules/completedDocsRule.ts
@@ -21,7 +21,7 @@ import * as ts from "typescript";
 import * as Lint from "../index";
 
 import { IInputExclusionDescriptors } from "./completed-docs/exclusionDescriptors";
-import { constructExclusionsMap, ExclusionsMap } from "./completed-docs/exclusionFactory";
+import { constructExclusionsMap, ExclusionsMap } from "./completed-docs/exclusions";
 
 export const ALL = "all";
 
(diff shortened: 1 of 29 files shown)


JoshuaKGoldberg:
> ```
> This assertion is unnecessary since it does not change the type of the expression. (no-unnecessary-type-assertion)
>   593 |     }
>   594 | 
> > 595 |     return parent.getChildren()!;
>       |           ^
>   596 | }
>   597 | 
>   598 | function isGetAccessor(node: ts.Node): node is ts.GetAccessorDeclaration {
> ```

🤔

src/rules/completedDocsRule.ts
```diff
@@ -21,7 +21,7 @@ import * as ts from "typescript";
 import * as Lint from "../index";
 
 import { IInputExclusionDescriptors } from "./completed-docs/exclusionDescriptors";
-import { ExclusionFactory, ExclusionsMap } from "./completed-docs/exclusionFactory";
+import { constructExclusionsMap, ExclusionsMap } from "./completed-docs/exclusionFactory";
```
adidahiya: let's rename this file to `exclusions.ts` now that there's no "factory"

src/rules/completedDocsRule.ts
```diff
@@ -38,6 +38,7 @@ export const ARGUMENT_VARIABLES = "variables";
 
 export const DESCRIPTOR_TAGS = "tags";
 export const DESCRIPTOR_LOCATIONS = "locations";
+export const DESCRIPTOR_OVERLOADS_SEPARATE_DOCS = "overloads-separate-docs";
```
adidahiya: can we just name it "overloads"? I think seeing that in the config in context will be clear enough

PR #4566, opened 2019-03-05
Add junit output for files which are successfully linted.

### LINKED ISSUE #3744: Junit formatter prevents Jenkins picking up results on success
### Bug Report

- __TSLint version__: 5.9.1
- __TypeScript version__: 2.7.1
- __Running TSLint via__: CLI

#### Actual behavior

Running `tslint -o lint-results.xml -t junit -p tsconfig.json` on a lint-free project results in an xml file with an empty `<testsuites>` element. While this is acceptable by some consuming tools, others reject it. Unfortunately for me, the latter includes Jenkins. Jenkins hits a non-breaking error trying to consume `lint-results.xml`, and skips it. Which makes generating the file in the first place pointless.

Before you added the core junit formatter, we could use tslint-junit-formatter, whose [last commit] before then seems to have been a fix for this very issue. But because tslint prioritises its own core formatters over external ones with the same name, this formatter can no longer be used (it's mostly redundant except for that one fix, anyway).

#### Expected behavior

A single "success" `<testsuite>` element is generated inside the `<testuites>`, and Jenkins is able to consume it (as is Bamboo, which has the same problem, according to the referenced tslint-junit-formatter [issue]).

I will say I'm happy to submit a PR for this one, if you'll have me!

[last commit]: https://github.com/GenoLogics-OpenSource/tslint-junit-formatter/commit/fc04c216465a16e3ceb98e512696e432539babce
[issue]: https://github.com/GenoLogics-OpenSource/tslint-junit-formatter/issues/8

darrena092:
#### PR checklist

- [x] Addresses an existing issue: fixes #3744 
  - [x] Includes tests
- [ ] Documentation update

#### Overview of change:
This fixes an issue with the junit formatter. Successful test cases are now output when files are linted without error.

### DIFF
--- a/src/formatters/junitFormatter.ts
+++ b/src/formatters/junitFormatter.ts
@@ -76,20 +76,20 @@ export class Formatter extends AbstractFormatter {
             }
         }
 
-        if(fileNames && fileNames.length !== 0) {
+        if (fileNames !== undefined && fileNames.length !== 0) {
             // Filter out files which have had a failure associated with them.
-            const filteredFileNames = fileNames.filter((fileName) => {
-                for(const failure of failures) {
-                    if(fileName === failure.getFileName()) {
+            const filteredFileNames = fileNames.filter(fileName => {
+                for (const failure of failures) {
+                    if (fileName === failure.getFileName()) {
                         return false;
                     }
                 }
                 return true;
             });
 
-            for(const fileName of filteredFileNames) {
+            for (const fileName of filteredFileNames) {
                 output += `<testsuite name="${this.escapeXml(fileName)}" errors="0">`;
-                output += `<testcase name="${this.escapeXml(fileName)}" />`
+                output += `<testcase name="${this.escapeXml(fileName)}" />`;
                 output += `</testsuite>`;
             }
         }
--- a/src/formatters/junitFormatter.ts
+++ b/src/formatters/junitFormatter.ts
@@ -41,7 +41,7 @@ export class Formatter extends AbstractFormatter {
     };
     /* tslint:enable:object-literal-sort-keys */
 
-    public format(failures: RuleFailure[]): string {
+    public format(failures: RuleFailure[], _fixes?: RuleFailure[], fileNames?: string[]): string {
         let output = '<?xml version="1.0" encoding="utf-8"?><testsuites package="tslint">';
 
         if (failures.length !== 0) {
@@ -73,6 +73,24 @@ export class Formatter extends AbstractFormatter {
             }
             if (previousFilename !== null) {
                 output += "</testsuite>";
+            }
+        }
+
+        if(fileNames && fileNames.length !== 0) {
+            // Filter out files which have had a failure associated with them.
+            const filteredFileNames = fileNames.filter((fileName) => {
+                for(const failure of failures) {
+                    if(fileName === failure.getFileName()) {
+                        return false;
+                    }
+                }
+                return true;
+            });
+
+            for(const fileName of filteredFileNames) {
+                output += `<testsuite name="${this.escapeXml(fileName)}" errors="0">`;
+                output += `<testcase name="${this.escapeXml(fileName)}" />`
+                output += `</testsuite>`;
             }
         }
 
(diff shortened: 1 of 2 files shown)


palantirtech:
Thanks for your interest in [palantir/tslint](https://github.com/palantir/tslint), @darrena092! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

darrena092:
Just another note on this. It has been successfully tested with Jenkins, and it works great.

src/formatters/junitFormatter.ts
```diff
@@ -76,6 +76,24 @@ export class Formatter extends AbstractFormatter {
             }
         }
 
+        if (fileNames !== undefined && fileNames.length !== 0) {
+            // Filter out files which have had a failure associated with them.
+            const filteredFileNames = fileNames.filter(fileName => {
+                for (const failure of failures) {
+                    if (fileName === failure.getFileName()) {
```
JoshuaKGoldberg: This loops over all of `failures` for each `fileName`, so its runtime complexity is slightly better than `O(fileNames.length * failures.length)`, but still at worst a larger-than-necessary loop.

It'd be more efficient to create some kind of mapping between file names and their failures. How about something like a `Set<string>` of file names that _do_ have failures? That'll change your runtime complexity to roughly `O(fileNames.length + failures.length)`.

Normally the performance & runtime complexities for a method that runs once to format something wouldn't be too important, but in this case it might be called with thousands of file names and hundreds of failures.

test/formatters/junitFormatterTests.ts
```diff
@@ -106,14 +106,24 @@ describe("JUnit Formatter", () => {
                 </testsuite>
             </testsuites>`.replace(/>\s+/g, ">"); // Remove whitespace between tags;
 
-        assert.equal(formatter.format(failures), expectedResult);
+        assert.equal(formatter.format(failures, [], [TEST_FILE_1, TEST_FILE_2]), expectedResult);
     });
```
JoshuaKGoldberg: Could you please add a unit test for a result where 1 file has a failure and 2 file does not? Just to make sure both get into the output.

src/formatters/junitFormatter.ts
darrena092: No worries! Added something that's a bit more optimised using a set as you suggest. Will add the new unit test ASAP.

test/formatters/junitFormatterTests.ts
darrena092: Done :)

src/formatters/junitFormatter.ts
```diff
@@ -41,15 +41,20 @@ export class Formatter extends AbstractFormatter {
     };
     /* tslint:enable:object-literal-sort-keys */
 
-    public format(failures: RuleFailure[]): string {
+    public format(failures: RuleFailure[], _fixes?: RuleFailure[], fileNames?: string[]): string {
         let output = '<?xml version="1.0" encoding="utf-8"?><testsuites package="tslint">';
 
+        const failureFileNames: Set<string> = new Set<string>();
```
mastermatt: This line could be written as such, then the `if` stmt below could be removed.
```ts
const failureFileNames: Set<string> = new Set([...failures.map(f => f.getFileName())]);
```

darrena092: Good call. I've added that.

src/formatters/junitFormatter.ts
```diff
@@ -41,9 +41,11 @@ export class Formatter extends AbstractFormatter {
     };
     /* tslint:enable:object-literal-sort-keys */
 
-    public format(failures: RuleFailure[]): string {
+    public format(failures: RuleFailure[], _fixes?: RuleFailure[], fileNames?: string[]): string {
         let output = '<?xml version="1.0" encoding="utf-8"?><testsuites package="tslint">';
 
+        const failureFileNames: Set<string> = new Set([...failures.map(f => f.getFileName())]);
```
JoshuaKGoldberg: Even simpler 😉 

```suggestion
        const failureFileNames: Set<string> = new Set(failures.map(f => f.getFileName());
```

JoshuaKGoldberg: Hmm, I don't have permission to apply this in the fork. Will just merge anyway.

PR #4580, opened 2019-03-14
Fix #4579: false positive static-this in HOC

### LINKED ISSUE #4579: static-this - Improperly flags returned anonymous class
### Bug Report

- 5.14:
- 3..3.3333:
- CLI

```ts
interface Bar {}

class Foo {
  public static higherOrderComponent(): any {
    return class implements Bar {
      public constructor() {
        console.log(this);   // "this" is valid here but tslint is complaining
      }
    };
  }
}
```

```json
{
  "defaultSeverity": "error",
  "extends": [
    "tslint:all"
  ],
  "rules": {
    "static-this": true
  }
}
```

#### Actual behavior
This improperly flags this in an anonymous class returned by a static function, something that is typically used to construct higher-order components.

#### Expected behavior
`this` used in a class scoped inside a static function is valid.

mbelsky:
#### PR checklist

- [x] Addresses an existing issue: #4579 
- [x] bugfix
  - [x] Includes tests

[bugfix]

### DIFF
--- a/src/rules/staticThisRule.ts
+++ b/src/rules/staticThisRule.ts
@@ -58,13 +58,17 @@ function walk(ctx: Lint.WalkContext<void>) {
     const cb = (node: ts.Node): void => {
         const originalParentClass = currentParentClass;
 
-        if (
-            utils.isClassLikeDeclaration(node.parent) &&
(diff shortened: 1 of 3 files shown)


JoshuaKGoldberg:
Thanks @mbelsky!

Issue #4569, opened 2019-03-06
fix on grouped-import is deleting export statement


jonyeezs:
### Bug Report

- __TSLint version__: 5.13.1
- __TypeScript version__: 3.1.6
- __Running TSLint via__: (pick one) CLI - `npx tslint -p tsconfig.json`

#### TypeScript code being linted

```ts
import { version } from '../../package.json';

// This file can be replaced during build by using the `fileReplacements` array.
// `ng build --prod` replaces `environment.ts` with `environment.prod.ts`.
// The list of file replacements can be found in `angular.json`.

export const environment = {
  // Default Angular configuration.
  production: false,

  // Custom app configuration.
  Version: version,
  ApiUrl: 'https://url',
  IdentityUrl: 'https://url',
  MarketingUrl: 'https://url',
  ClientId: 'webapp',
  LogentriesToken: '',
  MixpanelToken: '',
  IntercomAppId: '',
  LogToConsole: true,
};

import 'zone.js/dist/zone-error'; // Included with Angular CLI.
```

with `tslint.json` configuration:

```json
{
  "rulesDirectory": ["codelyzer"],
  "rules": {
    "max-length": false /* Let prettier handle it */,
    "arrow-return-shorthand": true,
    "callable-types": true,
    "class-name": true,
    "comment-format": [true, "check-space"],
    "curly": [true, "ignore-same-line"],
    "deprecation": {
      "severity": "warn"
    },
    "eofline": true,
    "forin": true,
    "import-blacklist": [true, "rxjs/Rx"],
    "import-spacing": true,
    "indent": [true, "spaces"],
    "interface-over-type-literal": true,
    "label-position": true,
    "member-access": false,
    "member-ordering": [
      true,
      {
        "order": ["static-field", "instance-field", "static-method", "instance-method"]
      }
    ],
    "no-arg": true,
    "no-bitwise": true,
    "no-console": [true, "debug", "info", "time", "timeEnd", "trace"],
    "no-construct": true,
    "no-debugger": true,
    "no-duplicate-super": true,
    "no-empty": false,
    "no-empty-interface": true,
    "no-eval": true,
    "no-inferrable-types": [true, "ignore-params"],
    "no-misused-new": true,
    "no-non-null-assertion": true,
    "no-redundant-jsdoc": true,
    "no-shadowed-variable": true,
    "no-string-literal": false,
    "no-string-throw": true,
    "no-switch-case-fall-through": true,
    "no-trailing-whitespace": true,
[...]

PR #4583, opened 2019-03-18
ordered-imports: don't remove linebreaks if text contains non-whitespace characters

lukeautry:
#### PR checklist

- [X] Addresses an existing issue: #4569

#### Overview of change:

Existing replacement code deletes lines around import declarations if they contain any line break. This means it deletes actual code in the case where a fragment of text has both line breaks and characters.

An ideal fix would be to keep the (previously deleted) code as close to the import it depends on...assuming it depends on an import at all. Of course, this ramps up the complexity, because any given line of code could use a combination of imports.

This fix results in the code just appearing below the import blocks in the order they were processed.

#### CHANGELOG.md entry:

[bugfix] fix linebreak handling in `ordered-imports`

### DIFF
--- a/src/rules/orderedImportsRule.ts
+++ b/src/rules/orderedImportsRule.ts
@@ -538,7 +538,10 @@ class Walker extends Lint.AbstractWalker<Options> {
             if (index > 0) {
                 const prevItems = groupedDeclarations[index - 1];
                 const last = prevItems[prevItems.length - 1];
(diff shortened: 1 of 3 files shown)


palantirtech:
Thanks for your interest in [palantir/tslint](https://github.com/palantir/tslint), @lukeautry! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

PR #4589, opened 2019-03-20
Add "no-null-undefined-union" rule.

nrathi:
#### PR checklist

  - [x] Includes tests
- [ ] Documentation update

#### CHANGELOG.md entry:
[new-rule] `no-null-undefined-union`

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/tslint/4589)

### DIFF
--- a/src/rules/noNullUndefinedUnionRule.ts
+++ b/src/rules/noNullUndefinedUnionRule.ts
@@ -1,0 +1,100 @@
+/**
+ * @license Copyright 2018 Palantir Technologies, Inc. All rights reserved.
+ */
+
+import * as Lint from "tslint";
+import { isTypeReference, isUnionType } from "tsutils";
+import * as ts from "typescript";
+
+export class Rule extends Lint.Rules.TypedRule {
+    /* tslint:disable:object-literal-sort-keys */
+    public static metadata: Lint.IRuleMetadata = {
+        ruleName: "no-null-undefined-union",
+        description: "Disallows union types with both `null` and `undefined` as members.",
+        rationale: Lint.Utils.dedent`
+            A union type that includes both \`null\` and \`undefined\` is either redundant or fragile.
+            Enforcing the choice between the two allows the \`triple-equals\` rule to exist without
+            exceptions, and is essentially a more flexible version of the \`no-null-keyword\` rule.
+        `,
+        optionsDescription: "True if the rule should be enabled.",
+        options: null,
+        optionExamples: [true],
+        type: "functionality",
+        typescriptOnly: true,
+        requiresTypeInfo: true,
+    };
+    /* tslint:enable:object-literal-sort-keys */
+
+    public static FAILURE_STRING = "Union type cannot include both 'null' and 'undefined'.";
+
+    public applyWithProgram(sourceFile: ts.SourceFile, program: ts.Program): Lint.RuleFailure[] {
+        return this.applyWithFunction(sourceFile, walk, undefined, program.getTypeChecker());
+    }
+}
+
+function walk(ctx: Lint.WalkContext<void>, tc: ts.TypeChecker): void {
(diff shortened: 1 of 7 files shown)


palantirtech:
Thanks for your interest in [palantir/tslint](https://github.com/palantir/tslint), @nrathi! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

src/rules/noNullUndefinedUnionRule.ts
```diff
@@ -0,0 +1,105 @@
...
+    /* tslint:disable:object-literal-sort-keys */
+    public static metadata: Lint.IRuleMetadata = {
+        ruleName: "no-null-undefined-union",
+        description: "Disallows union types with both `null` and `undefined` as members.",
+        rationale: Lint.Utils.dedent`
+            A union type that includes both \`null\` and \`undefined\` is either redundant or fragile.
+            Enforcing the choice between the two allows the \`triple-equals\` rule to exist without
+            exceptions, and is essentially a more flexible version of the \`no-null-keyword\` rule.
+        `,
+        optionsDescription: "True if the rule should be enabled.",
```
adidahiya: There are no options. You should use:

```suggestion
        optionsDescription: "Not configurable.",
```

src/rules/noNullUndefinedUnionRule.ts
```diff
@@ -0,0 +1,105 @@
...
+        const type = getType(node, tc);
+        if (type !== undefined && isNullUndefinedUnion(type)) {
+            ctx.addFailureAtNode(node, Rule.FAILURE_STRING);
+        }
+        return ts.forEachChild(node, cb);
+    });
+}
+
+function getType(node: ts.Node, tc: ts.TypeChecker): ts.Type | undefined {
+    // NOTE: This is a comprehensive intersection between `HasType` and has property `name`.
```
adidahiya: nit: it's a comment... no need to say "NOTE"

```suggestion
    // This is a comprehensive intersection between `HasType` and has property `name`.
```

test/rules/no-null-undefined-union/test.ts.lint
```diff
@@ -0,0 +1,55 @@
...
+    if (somePredicate()) {
+~~~~~~~~~~~~~~~~~~~~~~~~~~
+        foo = someFunc();
+~~~~~~~~~~~~~~~~~~~~~~~~~
+    }
+~~~~~
+    return foo;
+~~~~~~~~~~~~~~~
+}
+~ [0]
```
adidahiya: this is a really large error range... I don't think you should report the failure on the whole function body. Instead, report it on the function name or near the return value typedef (or where the typedef would go, if it's inferred). consider how tsc reports errors on return values:

image

nrathi:
```
    if (isSignatureDeclaration(node)) {
        const signature = tc.getSignatureFromDeclaration(node);
        return signature === undefined ? undefined : signature.getReturnType();
    }
```

In this bit, I'm not sure what I can extract off of either `node` or `signature` to get a smaller range.

adidahiya: alright, this is fine for now, maybe we can improve this in a future PR

DarryQueen:
@nrathi as per this, should we stop using `React.ReactNode`? At least in 15.6.1, it has a type union of `null | undefined`.

styu:
+1 to @rafeememon ’s point, it feels like the rule should be enforced on type definitions rather than on consumed types, so you wouldn’t be able to define types with the union, but could still consume types that have defined the union already. That way you can trend towards the right direction without encouraging the disabling of this rule due to external types

adidahiya:
@styu that makes sense. I wouldn't recommend enabling this rule right away across the board. I filed https://github.com/palantir/tslint/issues/4617

PR #4619, opened 2019-03-29
Check for non-public modifiers in constructors (#4511)

### LINKED ISSUE #4511: [unnecessary-constructor] Incorrectly flagging accessibility-restricting constructors
### Bug Report

- __TSLint version__: 5.12.1

```ts
class SomethingNotConstructible {
	private constructor() {}
}
```

```json
{
	"rules": {
		"unnecessary-constructor": true
	}
}
```

#### Actual behavior

The private constructor is flagged as unnecessary.

#### Expected behavior

The private constructor is not flagged.

There are cases where a class shouldn't be constructible (factory classes or something) where it would make sense to have an empty constructor with the private/etc modifier to prevent it from being constructed externally. Right now, this private empty constructor is flagged by the `unnecessary-constructor` rule. In my estimation, this constructor _is_ necessary, because it performs a task that would not performed were it not included (makes the constructor inaccessible).

ericbf:
This commit checks the modifiers of each constructor declaration. If it has any modifier that isn't `public`, then it is actually necessary and shouldn't be flagged.

#### PR checklist

- [x] Addresses an existing issue: fixes #4511 
  - [x] Includes tests
- [ ] Documentation update

#### Overview of change:

Iterate through the modifiers of each constructor declaration. If it has any modifiers that isn't `public`, keep it.

#### CHANGELOG.md entry:

[bugfix] unnecessary-constructor: don't flag constructors that are actually necessary

### DIFF
--- a/src/rules/unnecessaryConstructorRule.ts
+++ b/src/rules/unnecessaryConstructorRule.ts
@@ -45,22 +45,23 @@ export class Rule extends Lint.Rules.AbstractRule {
 const isEmptyConstructor = (node: ts.ConstructorDeclaration): boolean =>
     node.body !== undefined && node.body.statements.length === 0;
 
-const containsConstructorParameter = (node: ts.ConstructorDeclaration): boolean => {
-    for (const parameter of node.parameters) {
-        if (isParameterProperty(parameter)) {
-            return true;
-        }
-    }
+const containsConstructorParameter = (node: ts.ConstructorDeclaration): boolean =>
+    // If this has any parameter properties
+    node.parameters.some(isParameterProperty);
 
-    return false;
-};
+const isAccessRestrictingConstructor = (node: ts.ConstructorDeclaration): boolean =>
+    // No modifiers implies public
+    node.modifiers != undefined &&
+    // If this has any modifier that isn't public, it's doing something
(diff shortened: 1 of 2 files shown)


src/rules/unnecessaryConstructorRule.ts
```diff
@@ -45,6 +45,15 @@ export class Rule extends Lint.Rules.AbstractRule {
...
     node.body !== undefined && node.body.statements.length === 0;
 
+const isAccessRestrictingConstructor = (
+    node: ts.ConstructorDeclaration,
+    context: Lint.WalkContext,
+): boolean =>
+    // No modifiers implies public
+    node.modifiers != undefined &&
+    // If this has any modifier that isn't public, it's doing something
+    node.modifiers.findIndex(modifier => modifier.getText(context.sourceFile) !== "public") >= 0;
```
JoshuaKGoldberg: It's a bit speedier to check the modifier's `.kind`, which might be of type `ts.SyntaxKind.PublicKeyword`. 

```suggestion
    node.modifiers.some(modifier => modifier.kind !== ts.SyntaxKind.PublicKeyword);
```

ericbf: Nice – will make this update! Thanks :)

ericbf:
Didn't know if GitHub would automatically identify the change as made 😆. Updated now.

Issue #4617, opened 2019-03-28
Relax no-null-undefined-union rule


adidahiya:
especially as it pertains to React render functions, `null | undefined` is sometimes unavoidable when dealing with external type definitions.

from @styu:

>  the rule should be enforced on type definitions rather than on consumed types, so you wouldn’t be able to define types with the union, but could still consume types that have defined the union already. That way you can trend towards the right direction without encouraging the disabling of this rule due to external types

VincentLanglet:
Something like this
```
interface Props {
  title?: string | null
}
```
Or
```
type Text = string | null

interface Props {
  title?: Text
}
```
Should be allowed, or at least having an options to be allowed.

It's more convenient to use `?` to inform it's an optional props, even if the type already have `null` or `undefined` somewhere lost in his definition.

adidahiya:
@VincentLanglet if you want to allow that then it seems like you should just disable the rule altogether. It is a relatively strict style choice, definitely not for everyone

VincentLanglet:
See https://github.com/palantir/tslint/pull/4625#discussion_r271440539

`?` is more than `undefined`

```
interface A {
 a: number | undefined
}
interface B {
 b?: number
}

const a: A = {} // $error
const b: B = {} // no error
```

Plus the actual `tslint:all` has conflict between this rule and `no-null-keyword`.

glen-84:
@adidahiya This was fixed in #4625.

PR #4625, opened 2019-04-02
Relax "no-null-undefined-union" rule.

nrathi:
#### PR checklist

- [x] Addresses an existing issue: fixes https://github.com/palantir/tslint/issues/4617
  - [x] Includes tests
- [x] Documentation update

#### CHANGELOG.md entry:
[rule-change] Relax `no-null-undefined-union` rule.

### DIFF
--- a/test/rules/no-null-undefined-union/test.ts.lint
+++ b/test/rules/no-null-undefined-union/test.ts.lint
@@ -13,16 +13,16 @@ const x: SomeType;
 
 const someFunc = (): SomeType => {}
 
-function(foo: SomeType) {}
-
-const y: string | null | undefined;
-         ~~~~~~~~~~~~~~~~~~~~~~~~~ [0]
-
-interface someInterface {
+interface SomeInterface {
     foo: number | undefined | null;
          ~~~~~~~~~~~~~~~~~~~~~~~~~ [0]
     bar: boolean;
 }
+
+function(foo: SomeInterface) {}
+
+const y: string | null | undefined;
+         ~~~~~~~~~~~~~~~~~~~~~~~~~ [0]
 
 const someFunc = (): string | undefined | null => {}
                      ~~~~~~~~~~~~~~~~~~~~~~~~~ [0]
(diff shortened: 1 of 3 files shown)


VincentLanglet:
Hi @nrathi 

Can you add a test for
```
type Text = string | null

interface Props {
  title?: Text
}
interface SuperProps {
  title?: Text | number
}

function someFunc(bar?: Text, foo?: Text | number) {}
```

I don't expect error from the rule, if you do, there should be at least an option to disable these cases.

nrathi:
@VincentLanglet added, no errors!

test/rules/no-null-undefined-union/test.ts.lint
```diff
@@ -52,4 +58,18 @@ function someFunc() {
 }
 ~ [0]
 
+const z = testFunc();
+
+type Text = string | null
+
+interface TextInterface {
+  foo?: Text
```
adidahiya: 1. indentation is inconsistent on this line (use 4 spaces)
2. I don't agree with this change... this is just syntactic sugar for `Text | undefined`, which is in turn sugar for `string | null | undefined`. Why should we allow this? What's the point of the rule? How do you explain to devs which style is allowed and which isn't?

VincentLanglet: Plus `?` is just not syntactic sugar for Text | undefined.
```
interface A {
 a: number | undefined
}
interface B {
 b?: number
}

const a: A = {} // $error
const b: B = {} // no error
```

When you declare props in react, all I saw
```
interface Props {
  requiredValue: SomeType
  optionalValue?: SomeType
}

<Component requiredValue={...}  />
```

I tried
```
interface Props {
  value: SomeType | undefined
}

<Component /> // $error property value is missing
```

Plus actually the rule create conflict with the `non-null-keyword` in the `tslint:all` config. With the precedent props, you have to use the component this way
```
interface Props {
  value: SomeTypeCouldBeNull
}

<Component value={null} /> // $error prefer use undefined
```

Every user can disable the rules he does not want, but at least the proposal `tslint:all` should not have conflict.

adidahiya: ok, that's fair, thanks for the examples. after looking at this again, the rule makes more sense now. I think we should update the documentation to also explain this, though. we should mention that interface fields and function params with `?` are expected to be exempt from the rule.

PR #4635, opened 2019-04-03
Include actual types for restrict-plus-operands.

mprobst:
It can be difficult for users to find out why exactly they received the
"restrict-plus-operands" lint error. In particular, users often end up
with `any` types leaking into their code (e.g. in tests), and are then
puzzled why this triggers.

To fix, this change includes a textual representation of the type in the
error message, e.g.:

    [Operands of '+' operation must either be both strings or both numbers, but found 5 + undefined[]]

The type representation isn't always super pretty (e.g. the
`undefined[]` bit above), but should still be helpful, and also matches
TS compiler's representation of these.

#### PR checklist

- [x] bugfix
  - [x] Includes tests

#### CHANGELOG.md entry:

[enhancement] Improve error message for `restrict-plus-operands` rule.

### DIFF
--- a/src/rules/restrictPlusOperandsRule.ts
+++ b/src/rules/restrictPlusOperandsRule.ts
@@ -37,7 +37,7 @@ export class Rule extends Lint.Rules.TypedRule {
 
     public static INVALID_TYPES_ERROR =
         "Operands of '+' operation must either be both strings or both numbers";
-    public static SUGGEST_TEMPLATE_LITERALS = ". Consider using template literals.";
+    public static SUGGEST_TEMPLATE_LITERALS = "; consider using template literals.";
 
     public applyWithProgram(sourceFile: ts.SourceFile, program: ts.Program): Lint.RuleFailure[] {
         return this.applyWithFunction(sourceFile, walk, undefined, program.getTypeChecker());
--- a/src/rules/restrictPlusOperandsRule.ts
+++ b/src/rules/restrictPlusOperandsRule.ts
@@ -52,7 +52,7 @@ function walk(ctx: Lint.WalkContext, tc: ts.TypeChecker) {
             const rightType = tc.getTypeAtLocation(node.right);
             const rightTypeStr = getBaseTypeOfLiteralType(rightType);
             if (leftTypeStr === "invalid" || rightTypeStr === "invalid" || leftTypeStr !== rightTypeStr) {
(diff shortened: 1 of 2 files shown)


mprobst:
@bowenni can you take a look at this?

mprobst:
I think the test failure is an infrastructure failure, but I cannot retry because I don't have permissions.

JoshuaKGoldberg:
I love this idea! Marked as accepting PRs.

In the future @mprobst, please file an issue describing your requested changes before sending a PR that non-trivially changes TSLint. This one happens to be immediately acceptable but most aren't.

src/rules/restrictPlusOperandsRule.ts
```diff
@@ -37,7 +37,7 @@ export class Rule extends Lint.Rules.TypedRule {
 
     public static INVALID_TYPES_ERROR =
         "Operands of '+' operation must either be both strings or both numbers";
-    public static SUGGEST_TEMPLATE_LITERALS = ", consider using template literals";
+    public static SUGGEST_TEMPLATE_LITERALS = ". Consider using template literals.";
```
JoshuaKGoldberg: Additional +1 to this grammar fix :wink:

test/rules/restrict-plus-operands/test.ts.lint
```diff
@@ -17,35 +17,38 @@ var pair: NumberStringPair = {
 
 // bad
 var bad1  = 5 + "10";
-            ~~~~~~~~  [Operands of '+' operation must either be both strings or both numbers, consider using template literals]
+            ~~~~~~~~  [Operands of '+' operation must either be both strings or both numbers, but found 5 + "10". Consider using template literals.]
 var bad2  = [] + 5;
-            ~~~~~~  [Operands of '+' operation must either be both strings or both numbers]
+            ~~~~~~  [Operands of '+' operation must either be both strings or both numbers, but found undefined[] + 5]
```
JoshuaKGoldberg: > `undefined[]`

Agreed, this is a little confusing (and `never[]` would be more accurate in some situations). For array literals of size 0, let's just have the rule print `[]`.

> `Operands of '+' operation must either be both strings or numbers, but found [] + {}`

mprobst: Done.

mprobst:
> In the future @mprobst, please file an issue describing your requested changes before sending a PR that non-trivially changes TSLint. This one happens to be immediately acceptable but most aren't.

Will do.

PR #4637, opened 2019-04-04
Add prettierrc

VincentLanglet:
#### Overview of change:

Use prettierrc instead of a config in tslint.json
User with prettier have now a correct auto-formatting of the project


adidahiya:
Does `tslint --fix` still apply prettier in this repo?

Issue #4588, opened 2019-03-20
[quotemark: backtick] Incorrectly flags string literals in export line


ericbf:
### Bug Report

- __TSLint version__: 5.14.0
- __TypeScript version__: Version 3.3.3333
- __Running TSLint via__: Atom
- __Related issues__: #4514 

#### TypeScript code being linted

```ts
export * from "./some-module"
```

with `tslint.json` configuration:

```json
{
    "rules": {
        "quotemark": "backtick"
    }
}
```

#### Actual behavior

The quotes get flagged (and changed to backticks if fix is enabled).

#### Expected behavior

The quotes should remain as double/single quotes, as typescript/ecma doesn't allow backticks here.

ericbf:
Here's another example:

**Code being linted**
```typescript
const array: SomeType<string, "hello"> = []
const obj: {
	"kebab-case": string
} = {
	"kebab-case": "value"
}
```

**Expected result of fix**
```typescript
const array: SomeType<string, "hello"> = []
const obj: {
	"kebab-case": string
} = {
	"kebab-case": `value`
//                    ~~~~~~~ [" should be `]
}
```

**Actual result of fix**
```typescript
const array: SomeType<string, `hello`> = []
//                            ~~~~~~~ [" should be `]
const obj: {
	`kebab-case`: string
//      ~~~~~~~~~~~~ [" should be `]
} = {
	"kebab-case": "value"
//                    ~~~~~~~ This one is ignored, but should be flagged
}
```

**Problem summary**

1. The fixing/flagging here flags some things that when changed breaks typescript syntax and doesn't flag other instances where it should be flagged.
1. Typescript required that the string literal type passed to the generic type be single/double, but the implementation of the backtick rule does not make that distinction. Seems that node doesn't have a clear-cut type though to exclude it.
1. For the type definition, the kebab case property name has to remain single/double quotes, or typescript breaks as well. What is the associated node type there? Mucking in the code trying to PR for this, seems that the type definition property name here doesn't have a clean-cut node type either.
1. The string literal in the property value assignment is not flagged, but it should. Seems that looking at the parent node (`"kebab-case": "value"`) gives you "PropertyAssignment" for both the property key and the property value, while the key should be ignored in this case, the value should not. How would one implement this part when they don't have their own node types?

ericbf:
Actually, I think I got something. PR soon.

PR #4642, opened 2019-04-05
[quotemark] Excuse more backtick edge cases

### LINKED ISSUE #4645: [quotemark] backtick breaks old typescript
### Bug Report

- __TSLint version__: 5.15.0
- __TypeScript version__: 2.4.x

```ts
declare let v: string | number

if (typeof v === "string") {
    v.substring(1)
}

// and

let obj: {
    key: "literal"
}
```

```json
{
    "rules": {
        "quotemark": [true, "backtick"]
    }
}
```

#### Actual behavior

Quotes are flagged (and changed to backticks with --fix)

#### Expected behavior

The quotes remain as they are, because this version of TS does not narrow the type of v when using `typeof` with a backtick string, and does not compiler at all when using a backtick string as a literal type

ericbf:
This edge cases were previously flagged when they should be ignored, as changing them breaks typescript. This commit makes it so they are ignored. It also organizes a little better, using functions instead of multi-layered conditionals (it was getting confusing).

#### PR checklist

- [x] Addresses an existing issue: #4588, fixes #4645
  - [x] Includes tests
- [ ] Documentation update

#### Overview of change:

I keep finding other edge cases that I previously missed. I *think* this is all of them now, but we shall see. Related to #4588.

See if y'all find any other edge cases.

#### CHANGELOG.md entry:

[bugfix] Excuse more quotemark backtick edge cases

### DIFF
--- a/src/rules/quotemarkRule.ts
+++ b/src/rules/quotemarkRule.ts
@@ -324,7 +324,7 @@ function isNameInAssignment(node: ts.StringLiteral | ts.NoSubstitutionTemplateLi
 }
 
 function isTypeCheckWithOldTsc(node: ts.StringLiteral | ts.NoSubstitutionTemplateLiteral) {
-    if (hasOldTscBacktickBehavior()) {
+    if (!hasOldTscBacktickBehavior()) {
         // This one only affects older typescript versions
         return false;
     }
--- a/src/rules/quotemarkRule.ts
+++ b/src/rules/quotemarkRule.ts
@@ -134,13 +134,14 @@ function walk(ctx: Lint.WalkContext<Options>) {
                 (isExportDeclaration(node.parent) ||
                     // This captures `import blah from "package"`
                     isImportDeclaration(node.parent) ||
-                    // This captures kebab-case property names in object literals (only when the node is not at the end of the parent node)
-                    (node.parent.kind === ts.SyntaxKind.PropertyAssignment &&
-                        node.end !== node.parent.end) ||
-                    // This captures the kebab-case property names in type definitions
-                    node.parent.kind === ts.SyntaxKind.PropertySignature ||
+                    // This captures quoted names in object literal keys
+                    isNameInAssignment(node) ||
+                    // This captures quoted signatures (property or method)
+                    isSignature(node) ||
                     // This captures literal types in generic type constraints
-                    node.parent.parent.kind === ts.SyntaxKind.TypeReference)
+                    isTypeConstraint(node) ||
+                    // Whether this is the type in a typeof check with older tsc
+                    isTypeCheckWithOldTsc(node))
             ) {
                 return;
             }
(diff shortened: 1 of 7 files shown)


src/rules/noInferredEmptyObjectTypeRule.ts
```diff
@@ -68,6 +68,7 @@ class NoInferredEmptyObjectTypeRule extends Lint.AbstractWalker {
             } else if (node.kind === ts.SyntaxKind.NewExpression) {
                 this.checkNewExpression(node as ts.NewExpression);
             }
+
```
JoshuaKGoldberg: Unrelated changes; please revert.

src/rules/quotemarkRule.ts
```diff
@@ -245,3 +246,91 @@ function getJSXQuotemarkPreference(
...
+        node.parent.kind !== ts.SyntaxKind.PropertyAssignment &&
+        node.parent.kind !== ts.SyntaxKind.MethodDeclaration
+    ) {
+        // If the node is neither a property assignment or method declaration, it's not a name in an assignment
+        return false;
+    }
+
+    return (
+        // In typescript below 2.7.1, don't change values either
+        ts.version < "2.7.1" ||
```
JoshuaKGoldberg: Directly checking `ts.version` is a little unusual... maybe there's a cleaner way? Thoughts @adidahiya? 

Also, @ericbf, could you elaborate on why this is necessary?

src/rules/noInferredEmptyObjectTypeRule.ts
ericbf: Gladly. Just making all merge checks pass.

src/rules/quotemarkRule.ts
ericbf: Typescript below the listed version fails compilation (or otherwise misbehaves) when backticks are used in some places where it succeeds in future versions. For instance, ``typeof str === `string``` does not narrow the type of `str` in TS versions below 2.7.1, causing compilation errors. In the other cases, older TS throws errors where NoSubstitutionLiterals are used instead of single/double quoted strings, but works fine in above 2.7.1; rather than cripple the rule to support order versions, I split it up.

ericbf: See the `typeof` issue [here](https://github.com/Microsoft/TypeScript/issues/20949). Its milestone was 2.7 (hence the split in this PR).
And [here's](https://github.com/Microsoft/TypeScript/issues/16592) the issue for backticks strings as types. This one was 2.6, but rather than doing multiple splits, I just did one at 2.7.

ericbf: Created corresponding issue, #4645

JoshuaKGoldberg: Thanks! This LGTM but before merging I'd like to hear from adidahiya on whether there's a cleaner way to check on these things.

ericbf:
Any update here? @adidahiya? @JoshuaKGoldberg?

ericbf:
Any update here?

src/rules/quotemarkRule.ts
adidahiya: I agree it's unusual and should be avoided if possible, but this seems like a reasonable use case for checking ts.version.

However, can you please use the semver package and the helper function from `parse.ts` to check version ranges instead?

```ts
import { lt } from "semver";
import { getNormalizedTypescriptVersion } from "../../verify/parse";

function hasOldTscBacktickBehavior() {
    return lt(getNormalizedTypescriptVersion(), "2.7.1");
}
```

adidahiya:
Also if you merge master you'll get the new required CI check, "clean-lockfile"

PR #4652, opened 2019-04-09
Improve "jsdoc-format" rule.

bowenni:
#### PR checklist

- [ ] Addresses an existing issue: fixes #0000
  - [x] Includes tests
- [ ] Documentation update

#### Overview of change:
When an asterisk is missing on the current line the rule emits two errors. The second error message confuses the users because there are no asterisks on that line.
This PR makes the rule only emit `asterisks in jsdoc must be aligned` if there is an asterisks.


bowenni:
I think the failed test is unrelated to this change.

JoshuaKGoldberg:
Hi @bowenni - this change seems a little arbitrary. It's unclear if it fixes the issue as described: one of the changed test lines also has two errors on it.

Could you please file an issue in the standard template to discuss more? There might be an even cleaner solution to go for.

PR #4681, opened 2019-04-24
Allow const assertions in no-object-literal-type-assertion

### LINKED ISSUE #4628: no-object-literal-type-assertion should have an option to allow const assertions
### Option to rule Suggestion

Since TypeScript 3.4 there are "const contexts for literal expressions" (Microsoft/TypeScript#29510) that allow an `as const` statement in certain scenarios.

Right now the rule report an error:

```ts
// Type assertion on object literals is forbidden, use a type annotation instead.
let bar = { type: 'foo' } as const;
```
The new const assertions are really useful for complex type constructs, specially if they need to be used in discriminated unions.

So, add an option like "`allow-const-assertions`" to support this case would be very nice.

asmundg:
These are not normal type assertions and asserting that literal objects
are immutable is a normal usage pattern.

#### PR checklist

- [x] Addresses an existing issue: fixes #4628
- [x] enhancement
  - [x] Includes tests
- [x] Documentation update

#### Overview of change:

Allow const assertions in no-object-literal-type-assertion. These are not equivalent to normal type assertions, and using them on literal objects is expected behavior.

#### CHANGELOG.md entry:

[enhancement] Allow const assertions in no-object-literal-type-assertion

### DIFF
--- a/src/rules/noObjectLiteralTypeAssertionRule.ts
+++ b/src/rules/noObjectLiteralTypeAssertionRule.ts
@@ -86,6 +86,12 @@ function walk(ctx: Lint.WalkContext<Options>): void {
         if (
             isAssertionExpression(node) &&
             node.type.kind !== ts.SyntaxKind.AnyKeyword &&
+            // If using TS >=3.4, allow const assertions
+            !(
+                (ts as any).isConstTypeReference &&
+                node.type.kind === ts.SyntaxKind.TypeReference &&
+                (ts as any).isConstTypeReference(node.type)
+            ) &&
             // Compare with UnknownKeyword if using TS 3.0 or above
             (!!(ts.SyntaxKind as any).UnknownKeyword
                 ? node.type.kind !== (ts.SyntaxKind as any).UnknownKeyword
--- a/src/rules/noObjectLiteralTypeAssertionRule.ts
+++ b/src/rules/noObjectLiteralTypeAssertionRule.ts
@@ -88,9 +88,9 @@ function walk(ctx: Lint.WalkContext<Options>): void {
             node.type.kind !== ts.SyntaxKind.AnyKeyword &&
             // If using TS >=3.4, allow const assertions
             !(
-                (ts as any).isConstTypeReference &&
                 node.type.kind === ts.SyntaxKind.TypeReference &&
-                (ts as any).isConstTypeReference(node.type)
+                ts.isConstTypeReference !== undefined &&
+                ts.isConstTypeReference(node.type)
             ) &&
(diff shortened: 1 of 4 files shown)


palantirtech:
Thanks for your interest in [palantir/tslint](https://github.com/palantir/tslint), @asmundg! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

asmundg:
@JoshuaKGoldberg this relies on a new utility function added in typescript 3.4. Is it reasonable to bump tslint's typescript dev dependency?

src/rules/noObjectLiteralTypeAssertionRule.ts
```diff
@@ -86,6 +86,13 @@ function walk(ctx: Lint.WalkContext<Options>): void {
         if (
             isAssertionExpression(node) &&
             node.type.kind !== ts.SyntaxKind.AnyKeyword &&
+            // Allow const assertions, introduced in TS 3.4
+            !(
+                node.type.kind === ts.SyntaxKind.TypeReference &&
+                (ts.isConstTypeReference !== undefined
+                    ? ts.isConstTypeReference(node.type)
+                    : node.type.getText(ctx.sourceFile) === "const")
```
JoshuaKGoldberg: > `node.type.getText(ctx.sourceFile) === "const"`

If the user isn't running TSLint with a TypeScript version that understands `as const` then TypeScript will already be giving them errors. You can ignore this case, I think.

test/rules/no-object-literal-type-assertion/default/test.ts.lint
```diff
@@ -16,6 +16,10 @@ x as T;
 {} as unknown;
 <unknown> {};
 
+// Allow const assertion
+({}) as const;
+<const> ({});
```
JoshuaKGoldberg: See https://github.com/palantir/tslint/blob/c01bb2cdc258da4f9727781151d91773fba3e283/test/rules/no-unused-variable/default/var.ts.lint for an example of how to only include some lines of tests in particular TypeScript versions. That should help with your CI failures!

asmundg: Thanks!

src/rules/noObjectLiteralTypeAssertionRule.ts
asmundg: Agreed.

asmundg:
@JoshuaKGoldberg thanks for the input, I've addressed your change requests.

PR #4693, opened 2019-05-01
[quotemark] Exclude some cases from backtick rule

### LINKED ISSUE #4692: [quotemark] backticks break lookup types / use strict / octal escape sequences / enum members
### Bug Report

- __TSLint version__: 5.16.0
- __TypeScript version__: 3.4.2

```ts
"use strict"

enum Hello {
    "world",
    "foo"
}

function hi() {
    "use strict"
}

type hello = typeof obj["key"]
type hello2 = InterfaceName["key"]

const octal = "\110\151"
```

```json
{
    "rules": {
        "quotemark": [true, "backtick"]
    }
}
```

#### Actual behavior

All the strings in the example are flagged, even though they must remain as single/double quotes in order to compile or otherwise behave correctly.

#### Expected behavior

None of the strings in the example are flagged.

ericbf:
This commit makes quotemark backtick ignore use strict declarations, enum members, lookup types, and strings containing octal escape sequences.

#### PR checklist

- [x] Addresses an existing issue: fixes #4692 
  - [x] Includes tests
- [x] Documentation update

#### Overview of change:

See description. Just ignore the cases mentioned in the bug.

Make sure I captured all the variations of the said cases in the tests.

#### CHANGELOG.md entry:

[bugfix] Quotemark backtrack now ignores enum members, use strict declarations, lookup types, and strings containing octal escape sequences.

### DIFF
--- a/src/rules/noInferredEmptyObjectTypeRule.ts
+++ b/src/rules/noInferredEmptyObjectTypeRule.ts
@@ -30,7 +30,7 @@ export class Rule extends Lint.Rules.TypedRule {
         options: null,
         optionExamples: [true],
         rationale: Lint.Utils.dedent`
-            Prior to TypeScript 3.4, generic type parameters for functions and constructors are inferred as
+            Prior to TypeScript 3.5, generic type parameters for functions and constructors are inferred as
             \`{}\` (the empty object type) by default when no type parameter is explicitly supplied or when
             the compiler cannot infer a more specific type.
             This is often undesirable as the call is meant to be of a more specific type.
--- a/src/rules/noInferredEmptyObjectTypeRule.ts
+++ b/src/rules/noInferredEmptyObjectTypeRule.ts
@@ -30,7 +30,7 @@ export class Rule extends Lint.Rules.TypedRule {
         options: null,
         optionExamples: [true],
         rationale: Lint.Utils.dedent`
-            Prior to TypeScript 3.5, generic type parameters for functions and constructors are inferred as
+            Prior to TypeScript 3.4, generic type parameters for functions and constructors are inferred as
             \`{}\` (the empty object type) by default when no type parameter is explicitly supplied or when
             the compiler cannot infer a more specific type.
             This is often undesirable as the call is meant to be of a more specific type.
--- a/src/rules/quotemarkRule.ts
+++ b/src/rules/quotemarkRule.ts
@@ -17,8 +17,12 @@
 
 import { lt } from "semver";
 import {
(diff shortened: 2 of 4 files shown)


src/rules/quotemarkRule.ts
```diff
@@ -338,6 +354,20 @@ function isTypeCheckWithOldTsc(node: ts.StringLiteral | ts.NoSubstitutionTemplat
     return node.parent.getChildren().some(n => n.kind === ts.SyntaxKind.TypeOfExpression);
 }
 
+function containsOctalEscapeSequence(node: StringLiteralLike, sourceFile: ts.SourceFile) {
+    // Octal sequences can go from 1-377 (255 in octal), but let's match the prefix, which will at least be \1-\77
+    return /\\[1-7][0-7]?/.test(node.getText(sourceFile));
```
JoshuaKGoldberg: If the node is the actual `ts.StringLiteral` type, this should go off `node.text`, right? IIRC that's a bit more performant than getting text from the source file.

It's unlikely this would be an actual performance concern given the context this is used in, but I'd really like to discourage folks from using `.getText` whenever possible.

src/rules/noInferredEmptyObjectTypeRule.ts
```diff
@@ -30,7 +30,7 @@ export class Rule extends Lint.Rules.TypedRule {
         options: null,
         optionExamples: [true],
         rationale: Lint.Utils.dedent`
-            Prior to TypeScript 3.4, generic type parameters for functions and constructors are inferred as
+            Prior to TypeScript 3.5, generic type parameters for functions and constructors are inferred as
```
JoshuaKGoldberg: > This PR also fixes a bit of documentation in the no-inferred-empty-object rule (it previously said "prior to 3.4", but it really is supposed to be "prior to 3.5". I didn't want to create another PR just for that, but can if that is y'all's call).

Yes please! Separate PRs are preferable to make it easier to work on and review things in parallel, especially so that requested changes to one area don't block accepting others.

src/rules/quotemarkRule.ts
ericbf: `getText` actually did not include the literal backslashes here, for some reason. Since I needed to match the backslash, I had to get the text from the file directly.

src/rules/noInferredEmptyObjectTypeRule.ts
ericbf: Got it! Will revert this.

Issue #4534, opened 2019-02-23
Roadmap: TSLint -> ESLint


adidahiya:
As you may have read in [this blog post](https://medium.com/palantir/tslint-in-2019-1a144c2317a9), we plan to deprecate TSLint in 2019 and support the migration to ESLint as the standard linter for both TypeScript & JavaScript. This will not be an immediate deprecation; on the contrary, there is a lot of work to do to ensure a smooth transition to the new tooling without any regressions. There are features, test suites, and conveniences in TSLint which we hope to retain in the migration. There may be a period of time when there is overlap between the two tools and TSLint early adopters are recommended to run _both_ linters to ensure full code check coverage (to a reasonable degree such that performance doesn't suffer drastically).

I will be closing some feature requests in this repo which now feel out of scope because we expect them to be handled in the ESLint / [typescript-eslint](https://github.com/typescript-eslint/typescript-eslint) roadmap. One example of a category of rules for which __new feature requests__ are most likely to be closed / rejected is [formatting rules](https://github.com/palantir/tslint/labels/Formatting%20rule). I have suggested [splitting out these rules for quite some time](https://github.com/palantir/tslint/issues/3592) because we use Prettier at Palantir and consider it a better tool for the job of formatting code.

TSLint will continue to release with important bug fixes and updates which keep it up-to-date with the latest compiler / language features.

felixfbecker:
I would be great to have a CLI command that migrates a tslint.json to an eslint.json, mapping equivalent rules and options. Ideally it would remove rules that can be migrated from tslint.json and keep rules that don't have an equivalent yet (or not support the used options), so that the rule can be run idempotently over repeatedly over time until tslint.json becomes empty at some point and we can fully rely on ESLint.

hinell:
Good news.
ESLint guys already started [typescript-eslint](https://github.com/typescript-eslint/typescript-eslint) project to support TS.
They are looking for help of united community effort to bring full scale support of TS into life. [Here is the statement.](https://eslint.org/blog/2019/01/future-typescript-eslint#the-future-of-typescript-on-eslint).

cyberhck:
what will writing tslint rules mean after the migration? On tslint I'm happy to write new rules, because it's easy as tslint itself uses typescript and I know type of data while writing rule.

Is there plan to migrate eslint from JS to TS? If not then writing rule will be a bit harder. I hate to say it but if eslint isn't migrated to ts, it'd not be as pleasant to write rules.

[...]

Issue #2551, opened 2017-04-11
use-before-declare: false positive in destructuring assignment


tamird:
### Bug Report

- __TSLint version__: 
```
yarn run tslint -- --version
yarn run v0.22.0
$ "/Users/tamird/src/go/src/github.com/cockroachdb/cockroach/pkg/ui/node_modules/.bin/tslint" --version
5.1.0
✨  Done in 0.44s.
```
- __TypeScript version__:
```
yarn run tsc -- --version
yarn run v0.22.0
$ "/Users/tamird/src/go/src/github.com/cockroachdb/cockroach/pkg/ui/node_modules/.bin/tsc" --version
Version 2.3.0
✨  Done in 0.29s.
```
- __Running TSLint via__: (pick one) CLI / Node.js API / VSCode / grunt-tslint / Atom / Visual Studio / etc
```
$ "/Users/tamird/src/go/src/github.com/cockroachdb/cockroach/pkg/ui/node_modules/.bin/tslint" -c tslint.json -p tsconfig.json --type-check
```
#### TypeScript code being linted

```ts
const { key: saveErrorKey, error: saveError } = (action as PayloadAction<KeyedError>).payload;
```

with `tslint.json` configuration:

```json
{
  "rules": {
    "align": [
      true,
      "parameters",
      "arguments",
      "statements"
    ],
    "ban": false,
    "class-name": true,
    "comment-format": [
      true,
      "check-space"
    ],
    "curly": true,
    "eofline": true,
    "forin": true,
    "indent": [
      true,
      "spaces"
    ],
    "interface-name": false,
    "jsdoc-format": true,
    "label-position": true,
    "max-line-length": [
      false,
      140
    ],
    "member-ordering": [
      true,
      "public-before-private",
      "static-before-instance",
      "variables-before-functions"
    ],
    "no-any": false,
    "no-arg": true,
    "no-bitwise": true,
    "no-console": [
      true,
      "debug",
      "info",
      "time",
      "timeEnd",
      "trace"
    ],
    "no-consecutive-blank-lines": true,
    "no-construct": true,
    "no-parameter-properties": false,
    "no-debugger": true,
    "no-duplicate-variable": true,
    "no-shadowed-variable": true,
    "no-empty": false,
    "no-eval": true,
    "no-require-imports": false,
    "no-string-literal": true,
    "no-switch-case-fall-through": true,
    "trailing-comma": [true, {
        "multiline": "always",
        "singleline": "never"
    }],
[...]

PR #4695, opened 2019-05-05
Deprecate no-use-before-declare rule for typescript 2.9+

### LINKED ISSUE #4666: Task: document the deprecation of no-use-before-declare
the [documentation](https://palantir.github.io/tslint/rules/no-use-before-declare/) already has some warnings, but it would be nice to spell it out even more clearly in the rule metadata (like no-unused-variable)

Related: https://github.com/palantir/tslint/issues/2551#issuecomment-483787682

waseemahmad31:
#### PR checklist

- [x] Addresses an existing issue: fixes #4666
  - [ ] Includes tests
- [ ] Documentation update

#### Overview of change:
deprecate no-use-before-declare for TS 2.9+

### DIFF
--- a/src/rules/noUseBeforeDeclareRule.ts
+++ b/src/rules/noUseBeforeDeclareRule.ts
@@ -15,6 +15,7 @@
  * limitations under the License.
  */
 
+import * as semver from "semver";
 import { isBindingElement } from "tsutils";
 import * as ts from "typescript";
 
@@ -42,6 +43,9 @@ export class Rule extends Lint.Rules.TypedRule {
         typescriptOnly: false,
         requiresTypeInfo: true,
         codeExamples,
+        deprecationMessage: semver.gte(ts.version, "2.9.0-dev.0")
+            ? "Since TypeScript 2.9. Please use the built-in compiler checks instead."
+            : undefined,
     };
     /* tslint:enable:object-literal-sort-keys */
 


palantirtech:
Thanks for your interest in [palantir/tslint](https://github.com/palantir/tslint), @waseemahmad31! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

karol-majewski:
It seems like discouraging TSLint users from using `no-use-before-declare` might have been premature. There is a class of errors this rule helps catch and these errors are not detected by TypeScript itself.

Consider this example:

```ts
const consumer = () => {
  console.log(variable); // no-use-before-declare: "variable 'variable' used before declaration"
};

consumer();

const variable = 'foo';
```

This snippet represents perfectly valid TypeScript — no compile-time error will be issued. Yet, an attempt to run it will cause a runtime exception.

See [TypeScript playground](https://www.typescriptlang.org/play/#src=const%20consumer%20%3D%20()%20%3D%3E%20%7B%0D%0A%20%20console.log(variable)%3B%20%2F%2F%20no-use-before-declare%3A%20%22variable%20'variable'%20used%20before%20declaration%22%0D%0A%7D%3B%0D%0A%0D%0Aconsumer()%3B%0D%0A%0D%0Aconst%20variable%20%3D%20'foo'%3B%0D%0A).

Thanks to @OliverJAsh for providing an example.

JoshuaKGoldberg:
@karol-majewski one difficulty of rules like `no-use-before-declare` is figuring out which nodes are truly used before declaration vs. ones that just _seem_ to be used before declaration.

```ts
const consumer = () => {
    setTimeout(() => {
        console.log(variable);
    }, 1);
};
  
consumer();
  
const variable = 'foo';
```

The rule has been somewhat marked as deprecation for a while (#4666); if you'd like to see it un-deprecated or generally re-implemented please file a new issue.

OliverJAsh:
The docs seem to suggest this rule was deprecated because it no longer has any use (since `var` is no longer used. Is that correct?

In addition to @karol-majewski's point about catching runtime errors: I would still like this rule for a code style preference of mine: enforcing single direction of code definitions—details at top, composition/usages below. E.g.

#### Without

``` ts
const y = 1;

const foo = () => {
    x;
    y;
}

const x = 1;
```

#### With


``` ts
const x = 1;
const y = 1;

const foo = () => {
    x;
    y;
}
```

Burgestrand:
I'm a new user of typescript, and received this warning. However, I believe it's unclear:

> Please use the built-in compiler checks instead.

Does this mean I need to enable a compiler option in order to have the same guarantees as I would have had before this rule was deprecated?

I propose changing the message to either "This rule is no longer necessary and can be safely removed" or pointing the user (in this case me) to documentation for which compiler options are relevant.

bolatovumar:
@Burgestrand I can't find an equivalent compiler option for this rule but it looks like you don't really need this rule if you don't use the "var" keyword. So, if you have the "no-var-keyword" rule enabled then this rule is not needed at all.

Burgestrand:
@bolatovumar Great, thank you!

karol-majewski:
@bolatovumar None of the examples in this thread is using the `var` keyword. 🤔 Yet using `no-use-before-declare` would have helped catch errors for them.

bolatovumar:
@karol-majewski you get compiler warnings if you use a `const` or `let` variable before it's declared so you don't need this rule in that case.

karol-majewski:
@bolatovumar TypeScript doesn't issue any warnings for this piece of code and yet it breaks in runtime:

```ts
const consumer = () => {
  console.log(variable);
};

consumer();

const variable = 'foo';
```

See for yourself in [TypeScript playground](https://typescript-play.js.org/#code/MYewdgzgLgBKkFcC2BTATjAvDAFASiwD4YBvAKBjnAhABsUA6WkAcxwDcBDNAS04CN6eANxkAvqLLwIydPknTYXXgPpYYAcgBmIEBuFA).

bolatovumar:
@karol-majewski right, I see. You might wanna file a separate issue as @JoshuaKGoldberg suggested above . I'm not actually sure why this rule is deprecated specifically for TS 2.9+ so I can't really add much more here.

JoshuaKGoldberg:
Posting again here: the rule seems to have some buggy behavior, such as the false positive mentioned in #2551. If you'd like the rule to be un-deprecated despite that _(maybe the bug doesn't exist? maybe it's worth it to keep the rule?)_ please file a new issue so we can discuss there. PR responses aren't very search-discoverable.

Also, keep #4534 in mind: TSLint is in the process being deprecated.

Same goes for suggestions to improve the rule's docs or error messaging. I'm all for improving docs, but we should discuss them in a dedicated thread. It's hard to keep track of this one... 😞

ikokostya:
> please file a new issue so we can discuss there

Come on. I created new *issue* https://github.com/palantir/tslint/issues/4789, but you closed it with `blocked` status without any discussion.

> Also, keep #4534 in mind: TSLint is in the process being deprecated.

Yes, maybe TSLint team will stop deprecate existing rules? False positive results is not a reason for removing rule. By this logic any buggy rule can be removed.

> Posting again here: the rule seems to have some buggy behavior, such as the false positive mentioned in #2551. If you'd like the rule to be un-deprecated despite that (maybe the bug doesn't exist? maybe it's worth it to keep the rule?

So, example from issue #2551 is not isolated and contains some project specific entities.

I tried to test code from replies:
https://github.com/palantir/tslint/issues/2551#issuecomment-300904638
https://github.com/palantir/tslint/issues/2551#issuecomment-424837231

There are no errors with `no-use-before-declare` rule.

JoshuaKGoldberg:
> Come on. I created new issue #4789, but you closed it with blocked status without any discussion.

Oh oops, sorry! I did a dumb and thought it was a PR. That is entirely my fault; will re-open.

PR #4734, opened 2019-05-20
[new-rule] Add no-unnecessary-type-check rule

### LINKED ISSUE #4733: Add no-unnecessary-type-check rule
### Rule Suggestion

General problem

Warns you if you compare against `null` or `undefined` unnecessarily.

```typescript
const x = "my string"; // definitely not undefined or null

if (x !== undefined) { // should warn
    console.log(x);
}
if (x !== null) { // should warn
    console.log(x);
}

function getStringOrUndefined() {
    return Math.random() > 0.5 ? "string" : undefined;
}

const y: string | undefined = getStringOrUndefined();
if (y !== undefined) { // should not warn
    console.log(y);
}
```

**Additional context**

Checking if a variable is `undefined` unnecessarily is a problem for two reasons. Either the variable can actually never be `undefined` - in which case the check is unnecessary and probably confusing to a reader. Or, the variable _can_ be undefined (i.e. the code-writer knows something that the compiler doesn't) - in which case the typings should be updated to reflect that the variable _might_ be undefined.

In the former case, this rule more more for style. But in the latter (and imo more important) case, this rule can prevent bugs - i.e. since the type is wrong, there could be a variable of that type somewhere that could be `undefined` but neither the person using the variable nor the compiler know this.

Submitting a PR with this rule shortly

zachkirsch:
#### PR checklist

- [x] Addresses an existing issue: fixes #4733
  - [x] Includes tests
- [ ] Documentation update

#### Overview of change:

Adds the rule `no-unnecessary-type-check` (see https://github.com/palantir/tslint/issues/4733 for more details)

- Not sure how to add documentation - is there a command or is it manual?
- `yarn verify` is failing because there are some places in this repo where this rule fails (see below for example). Should I try to fix all these cases or just add a `disable-line` for this rule where the errors occur?
```
/Users/zkirsch/Documents/tslint/src/configuration.ts
This comparison is unnecessary because the left side is never undefined (no-unnecessary-type-check)
  151 |     inputFilePath?: string,
  152 | ): string | undefined {
> 153 |     if (suppliedConfigFilePath != undefined) {
      |        ^
  154 |         if (!fs.existsSync(suppliedConfigFilePath)) {
  155 |             throw new FatalError(
  156 |                 `Could not find config file at: ${path.resolve(suppliedConfigFilePath)}`,
```
- This is my first time working the TSLint or anything to do with the TS compiler so please let me know if I did anything incorrectly or there's a corner case I didn't consider.

#### CHANGELOG.md entry:

[new-rule] `no-unnecessary-type-check`

### DIFF
--- a/README.md
+++ b/README.md
@@ -63,6 +63,7 @@ Prerequisites:
 
 ```bash
 git clone git@github.com:palantir/tslint.git --config core.autocrlf=input --config core.eol=lf
+cd tslint
 yarn
 yarn compile
 yarn test
--- a/src/configs/all.ts
+++ b/src/configs/all.ts
@@ -221,6 +221,7 @@ export const rules = {
     "no-unnecessary-initializer": true,
     "no-unnecessary-qualifier": true,
     "no-unnecessary-type-assertion": true,
+    "no-unnecessary-type-check": true,
     "number-literal-format": true,
     "object-literal-key-quotes": { options: "consistent-as-needed" },
     "object-literal-shorthand": true,
--- a/src/rules/noUnnecessaryTypeCheckRule.ts
+++ b/src/rules/noUnnecessaryTypeCheckRule.ts
@@ -1,0 +1,171 @@
+/**
+ * @license
+ * Copyright 2017 Palantir Technologies, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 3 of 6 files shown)


palantirtech:
Thanks for your interest in [palantir/tslint](https://github.com/palantir/tslint), @zachkirsch! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

README.md
```diff
@@ -63,6 +63,7 @@ Prerequisites:
 
 ```bash
 git clone git@github.com:palantir/tslint.git --config core.autocrlf=input --config core.eol=lf
+cd tslint
```
JoshuaKGoldberg: Haha good catch! Would you mind sending this in as a separate PR? That way it's not blocked on requested changes in this PR.

PR #4737, opened 2019-05-21
[new-rule] Added new no-property-getter-setter rule

### LINKED ISSUE #3340: [Rule suggestion] Prefer or disallow get/set accessors
A rule which either disallows get/set accessors, or requires them when possible.

#### Enforces this
```javascript
getSomething() {
    return this.aProp * this.cProp;
}

setSomething(a: string) {
    this.bProp = a + '!!!';
}
```
#### or this
```javascript
get something() {
    return this.aProp * this.cProp;
}

set aDiffThing(a: string) {
    this.bProp = a + '!!!';
}
```

Accessors, no accessors--I don't have a preference, really. Just looking for a rule to help w/ consistency on this matter.

timocov:
#### PR checklist

- [x] Addresses an existing issue: #3340
  - [x] Includes tests
- [x] Documentation update

#### Overview of change:

Added new rule `no-property-getter-setter` to disallow usage of properties.

#### CHANGELOG.md entry:

[new-rule] `no-property-getter-setter`

### DIFF
--- a/src/rules/noPropertyGetterSetterRule.ts
+++ b/src/rules/noPropertyGetterSetterRule.ts
@@ -1,0 +1,49 @@
+/**
+ * @license
+ * Copyright 2019 Palantir Technologies, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import * as ts from "typescript";
+
+import * as Lint from "../index";
+
+export class Rule extends Lint.Rules.AbstractRule {
+    /* tslint:disable:object-literal-sort-keys */
+    public static metadata: Lint.IRuleMetadata = {
+        ruleName: "no-property-getter-setter",
+        description: "Forbids usage properties (get/set).",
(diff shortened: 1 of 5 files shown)


src/rules/noPropertyGetterSetterRule.ts
```diff
@@ -0,0 +1,49 @@
...
+ */
+
+import * as ts from "typescript";
+
+import * as Lint from "../index";
+
+export class Rule extends Lint.Rules.AbstractRule {
+    /* tslint:disable:object-literal-sort-keys */
+    public static metadata: Lint.IRuleMetadata = {
+        ruleName: "no-property-getter-setter",
```
JoshuaKGoldberg: From https://github.com/palantir/tslint/issues/3340#issuecomment-494076011:

> But: we should note that any rule to fix this issue should be configured to prefer either way, not hardcoded to just one.

This rule should be named `property-getter-setter`, then, since it's about configuring _how_ to set these up. It should also allow either style preference.

timocov:
> who prefer either get prop() or getProp() can use it.

I'm not sure that it's even possible to check: every project can have it's own code style and naming convention (get_prop, getProp, getprop, etc). What conventions we should support for that? Do you have any thoughts about it?

src/rules/noPropertyGetterSetterRule.ts
timocov: > any rule to fix this issue should be configured 

I'm not sure that this rule fixes that issue. We treat this rule as rule about preventing usage of JS properties, not about forcing props style.

JoshuaKGoldberg:
@timocov great question. Perhaps some configurable regular expression that defaults to something like:

```ts
["get(.*)`", "set(.*)"]
```

... which would match any two method declarations where:
* the `(.*)` matched in two method declarations are the same,
* the getter takes no parameters, 
* the setter takes in a single parameter the same type as the getter's return

adidahiya:
Declining as per https://github.com/palantir/tslint/issues/3340

PR #4744, opened 2019-05-25
--print-config flag

### LINKED ISSUE #3720: [Feature Request]: --print-config flag like eslint
I'm not using the issue template because I'm not reporting a bug and all the fields appear to relate to bug reporting. Please let me know if I can provide any other information.

### Description

It would be great to have a way to audit the rules being applied to a given file or project, as [eslint does with `--print-config [FILENAME]`](https://eslint.org/docs/user-guide/command-line-interface). I've currently got a number of `extends` in my tslint.json and I'm getting unexpected results; being able to print the config would help. Ideally something like:

```sh
$ tslint --print-config path/to/file.ts
```

With the result being a simple JSON configuration of the evaluated set of rules.

### Implementation

It looks like it _might_ be pretty easy to accomplish, if I understand the [library usage](https://palantir.github.io/tslint/usage/library/) example properly—it seems like you'd just have to pretty-format the object returned by this command (assuming `commandLineConfigArg?: string` and `commandLineFileArg: string` were supplied through the command line):

```js
tslint.Configuration.findConfiguration(commandLineConfigArg, commandLineFileArg).results
```

Is that a reasonable approach? In experimenting, I'm getting a pretty short set of rules compared to my eslint output on another project, so I might be missing something—ah it looks like the eslint output includes a lot of "off" rules, so I don't know how this flag might account for that. But if I understand the above line correctly, then in the short term I think I can just `npx node` and run these commands myself in a node console, but I think this would be a useful flag and help to align this tool with its ES analog.

JoshuaKGoldberg:
#### PR checklist

- [x] Addresses an existing issue: fixes #3720
  - [x] Includes tests
- [x] Documentation update

#### Overview of change:

Adds a `--print-config` flag as suggested in #3720. It's roughly equivalent to ESLint's: it takes in a file name and prints the JSON result of parsing the configuration that would be used for that file.

Where should I add tests for the output of this? I could only find precedent for testing the exit code. 😕 

Cleaned up `src/runner.ts` a bit to help keep things organized. See first 3 commits in order for a more logical flow of the changes.

~Edit: managed to bust my local install of TSLint; fixing lint rules now... 😂~ RC and Next lint failures are unrelated: #4745

#### CHANGELOG.md entry:

[enhancement]: Added --print-config CLI flag


src/configuration.ts
```diff
@@ -660,3 +660,24 @@ export function isFileExcluded(filepath: string, configFile?: IConfigurationFile
     const fullPath = path.resolve(filepath);
     return configFile.linterOptions.exclude.some(pattern => new Minimatch(pattern).match(fullPath));
 }
+
+export function stringifyConfiguration(configFile: IConfigurationFile) {
```
adidahiya: re: testing. I think a couple unit tests for this function should be sufficient; we don't really need to verify this feature through the CLI (but you can keep the tests you have added thus far)

src/runner.ts
```diff
@@ -165,6 +176,29 @@ async function runWorker(options: Options, logger: Logger): Promise<Status> {
     return options.force || errorCount === 0 ? Status.Ok : Status.LintError;
 }
 
+async function runConfigPrinting(options: Options, logger: Logger): Promise<Status> {
```
adidahiya:
```suggestion
async function printConfiguration(options: Options, logger: Logger): Promise<Status> {
```

src/runner.ts
```diff
@@ -145,6 +152,10 @@ async function runWorker(options: Options, logger: Logger): Promise<Status> {
         return Status.Ok;
     }
 
+    if (options.printConfig) {
+        return runConfigPrinting(options, logger);
```
adidahiya:
```suggestion
        return printConfiguration(options, logger);
```

src/files/finding.ts
```diff
@@ -0,0 +1,67 @@
+/**
```
adidahiya: can we combine this file with `./program.ts` and call the new file `resolution.ts`?

PR #4778, opened 2019-06-26
Change interface-name rule to check suffix and prefix

omarkdev:
Thinking about the interface-name rule, it's very common discussions in the development world if developers prefer to use suffixes or prefixes specific to their interfaces, eg TypeSomething, SomethingContract or ISomething, I Realized that it would make more sense for developers to use the rule by defining whether they want a specific suffix or prefix, here are some examples:

Always interfaces prefix `Type`:
```json
{
  "rules": {
    "interface-name": [
        true,
        {
            "prefix-needs": ["Type"]
        }
    ]
  }
}
```

Always interfaces prefix `I` (keep with same older implementation):
```json
{
    "rules": {
        "interface-name": [true,
            {
                "prefix-needs": ["I"]
            }
        ]
    }
}
```

Never interfaces prefix `Type`:

```json
{
  "rules": {
    "interface-name": [true,
        {
            "prefix-never": ["Type"]
        }
    ]
  }
}
```

Never interfaces prefix `I` (keep with same older implementation):

```json
{
    "rules": {
        "interface-name": [true,
            {
                "prefix-never": ["I"]
            }
        ]
    }
}
```

Always interface suffix `Contract`:

```json
{
  "rules": {
    "interface-name": [true,
        {
            "suffix-needs": ["Contract"]
        }
    ]
  }
}
```

Never interface suffix `Contract`:

```json
{
  "rules": {
    "interface-name": [true,
        {
            "suffix-never": ["Contract"]
        }
    ]
  }
}
```

#### PR checklist

- [ ] Addresses an existing issue: fixes #0000
  - [X] Includes tests
- [ ] Documentation update

### DIFF
--- a/src/configs/recommended.ts
+++ b/src/configs/recommended.ts
@@ -52,7 +52,9 @@ export const rules = {
         options: ["spaces"],
     },
     "interface-name": {
-        options: ["always-prefix"],
+        options: {
+            "prefix-never": ["I"],
+        },
     },
     "interface-over-type-literal": true,
     "jsdoc-format": true,
--- a/src/rules/interfaceNameRule.ts
+++ b/src/rules/interfaceNameRule.ts
@@ -20,62 +20,153 @@ import * as ts from "typescript";
 
 import * as Lint from "../index";
 
-const OPTION_ALWAYS = "always-prefix";
-const OPTION_NEVER = "never-prefix";
+const OPTION_SUFFIX_NEEDS = "suffix-needs";
+const OPTION_SUFFIX_NEVER = "suffix-never";
+const OPTION_PREFIX_NEEDS = "prefix-needs";
+const OPTION_PREFIX_NEVER = "prefix-never";
+
+interface WalkOptions {
+    prefixNeeds: string[];
+    prefixNever: string[];
+    suffixNeeds: string[];
+    suffixNever: string[];
+}
 
 export class Rule extends Lint.Rules.AbstractRule {
     /* tslint:disable:object-literal-sort-keys */
     public static metadata: Lint.IRuleMetadata = {
         ruleName: "interface-name",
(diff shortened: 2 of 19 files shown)


palantirtech:
Thanks for your interest in [palantir/tslint](https://github.com/palantir/tslint), @omarkdev! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

JoshuaKGoldberg:
Hi @omarkdev, thanks for sending a PR! 

Just a heads up, we (like most projects) typically discuss larger changes such as this one in an issue first. This rule change seems reasonable so I'll go ahead and review it in-place.

JoshuaKGoldberg:
Ah, actually, @omarkdev - the changes as you've implemented are a breaking change. This rule needs to still work under the existing configuration format for the folks who have it configured that way.

Could you please file an issue to discuss? My suspicion is we can allow both formats to work in a relatively straightforward manner, but it'd be good to discuss in depth.

adidahiya:
Hey, actually, I'm not sure a rule like this belongs in TSLint core at this time. I am reluctant to add major opinionated code style features like this which will require tweaking and iteration as the inevitable bikeshedding ensues over the coming months. I would rather this rule live in a custom rules repo outside of TSLint core. As @JoshuaKGoldberg suggested, it's best to file a separate issue to track this request, and if you end up publishing your "interface naming convention" rule implementation for others to share, you can link to that in the issue you file.

Also, have you looked into the [naming-convention](https://github.com/ajafff/tslint-consistent-codestyle/blob/master/docs/naming-convention.md) rule from tslint-consistent-codestyle? It may do what you want here.

PR #4790, opened 2019-07-10
Add rule 'no-promise-as-boolean'

### LINKED ISSUE #3983: [Rule Request] no-misused-promises
This is related to #3804 and of course [`no-floating-promises`](https://palantir.github.io/tslint/rules/no-floating-promises/). There are a number of pitfalls when working with promises and async code.

### First off:

```
async function printUrl(url: string) {
  console.log(await (await fetch(url)).json());
}

const Urls = ['palantir.com', 'google.com'];

async const myFunction() {
  Urls.forEach(async url => await printUrl(url));
  console.log('Done!');
}
```

What happens here is that `Array#forEach` doesn't support promises natively, and will simply fire them off and run onto the `Done` line before the URLs are fetched. `printUrl()` has a return type of `Promise<void>`, but [non-void functions in TS are assignable to void function parameters](https://github.com/Microsoft/TypeScript/wiki/FAQ#why-are-functions-returning-non-void-assignable-to-function-returning-void). Generally that rule is fine, but it'd be nice to handle promises specially to avoid this issue. One place where this can be a problem is in unit tests, where using `forEach` with async handlers can end up not actually running the test code.

### Here's another scenario:

```
async function checkIsUserAllowed(id: number) {
  const user = await postgres.fetchUser(id)
  return Boolean(user && user.hasAccess);
}

async const myFunction() {
  const user = 42;
  if (checkIsUserAllowed(user)) { // should be `await checkIsUserAllowed(user)`!
    console.log('Access granted');
  }
}

```

It would be great to be able to catch this scenario (see #3804), though one thing to be cautious about is to still allow handling working with types like `Promise<any> | undefined` when explicitly needed to check whether a value is there or not. This could be allowed using `Boolean(promiseValue)`, for instance.

Thoughts on whether handling these issues can be baked into an existing rule or be worthwhile to stand on its own as part of core rules?

guidsdo:
This commit fixes #3983

#### PR checklist

- [x] Addresses an existing issue: fixes #3983
  - [x] Includes tests
- [x] Documentation update

#### Overview of change:
Added rule to prevent people from using a Promise in a boolean expression. 

No.

#### CHANGELOG.md entry:
 "[new-rule] `no-promise-as-boolean`"


palantirtech:
Thanks for your interest in [palantir/tslint](https://github.com/palantir/tslint), @guidsdo! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

guidsdo:
Build is failing due to a bug in `typescript@next`, please ignore this.

test/rules/no-promise-as-boolean/test.ts.lint
```diff
@@ -0,0 +1,54 @@
+[typescript]: >= 2.4.0
+
+async function promiseFunc() {
```
JoshuaKGoldberg: We'll need some more test cases to really exercise this rule's edge cases. The following should be included:

* `Promise | undefined`
* `Promise | false`
* `Promise | null`

Additionally, the `Array.forEach` case in the linked issue is missing.

src/rules/noPromiseAsBooleanRule.ts
```diff
@@ -0,0 +1,128 @@
...
+        options: {
+            type: "list",
+            listType: {
+                type: "array",
+                items: { type: "string" },
+            },
+        },
+        optionExamples: [true, [true, "Thenable"]],
+        rationale: Lint.Utils.dedent`
+            There are no situations where one would like to check if a variable's value is truthy if it's type
```
JoshuaKGoldberg: Grammar nitpicking:

```suggestion
            There are no situations where one would like to check whether a variable's value is truthy if its type
```

src/rules/noPromiseAsBooleanRule.ts
```diff
@@ -0,0 +1,128 @@
...
+            only is Promise.
+            This may only occur when the typings are incorrect or the variable has a union type
+            (like Promise | undefined), of which the latter is allowed.
+
+            This rule prevents common bugs, where the user most of the times wants to 'await' a Promise first.
```
JoshuaKGoldberg:
```suggestion
            This rule prevents common bugs from forgetting to 'await' a Promise.
```

src/rules/noPromiseAsBooleanRule.ts
```diff
@@ -0,0 +1,128 @@
...
+        `,
+        options: {
+            type: "list",
+            listType: {
+                type: "array",
+                items: { type: "string" },
+            },
+        },
+        optionExamples: [true, [true, "Thenable"]],
+        rationale: Lint.Utils.dedent`
```
JoshuaKGoldberg: Per #3983 it's rather confusing how these rules all work together. In order to get this new rule in without confusing users, you'll want to:

* Export examples from a `noPromiseAsBoolean.examples.ts` file in `src/rules/code-examples/`
* Link to the relevant docs pages for other Promise rules in the description here: see https://github.com/palantir/tslint/issues/3983#issuecomment-433939470

src/rules/noPromiseAsBooleanRule.ts
```diff
@@ -0,0 +1,128 @@
...
+            type reference names.
+        `,
+        options: {
+            type: "list",
+            listType: {
+                type: "array",
+                items: { type: "string" },
+            },
+        },
+        optionExamples: [true, [true, "Thenable"]],
```
JoshuaKGoldberg: Although some older rules don't adhere to this yet, it's preferable for configuration options to be taken in as an object with a shape like `{ promiseClasses: ["Thenable"] }`. That makes it easier to add more options in the future. See `src/rules/adjacentOverloadSignaturesRule.ts` for an example of how to set this up.

JoshuaKGoldberg: Missing test cases to ensure these extra Promise class names are respected. See the `no-floating-promises` rule tests for an example of a rule that has multiple test directories.

test/rules/no-promise-as-boolean/test.ts.lint
guidsdo: The extra edge cases are added, but I won't add the `Array.forEach` one, since that's more specific about not passing Promises in normal callbacks. I can try to add another rule for this, since this can be a bug (but not always is). 

The `no-promise-as-boolean` rule is always a bug, unless someones typings are wrong. :-)

src/rules/noPromiseAsBooleanRule.ts
guidsdo: Fixed

guidsdo: Fixed. Also supported union promises.

src/rules/noPromiseAsBooleanRule.ts
guidsdo: Fixed

guidsdo: Fixed

guidsdo:
@JoshuaKGoldberg fixed all the issues, please re-review

guidsdo:
@JoshuaKGoldberg can I expect you to look at this before the 1st of August?

nomcopter:
Would love to use this in tslint before it is too late!

guidsdo:
> Would love to use this in tslint before it is too late!

The end is near

JoshuaKGoldberg:
Hi folks! The 1st of August deadline is for _sending_ PRs. I'm a bit swamped now but promise to review this, even if it's after then! 💖

src/rules/noPromiseAsBooleanRule.ts
```diff
@@ -0,0 +1,163 @@
...
+
+export class Rule extends Lint.Rules.TypedRule {
+    /* tslint:disable:object-literal-sort-keys */
+    public static metadata: Lint.IRuleMetadata = {
+        ruleName: "no-promise-as-boolean",
+        description: "Warns for Promises that are used for boolean conditions.",
+        descriptionDetails: Lint.Utils.dedent`
+            For the most accurate findings, set \`"strict": true\` in your \`tsconfig.json\`.
+
+            It's recommended to enable to following tslint rules as well:
```
JoshuaKGoldberg:
```suggestion
            It's recommended to enable the following rules as well:
```

JoshuaKGoldberg:
Ah, I don't have permissions to make the one suggested change @guidsdo. Could you please commit that suggestion and merge the latest master into this PR? Once you do, we'll be good to merge! 🚀

guidsdo:
> Ah, I don't have permissions to make the one suggested change @guidsdo. Could you please commit that suggestion and merge the latest master into this PR? Once you do, we'll be good to merge! 🚀

Thought I put that on, but thanks I applied the change and merged master into my branch :-)

JoshuaKGoldberg:
giphy

Thanks so much @guidsdo! Rule is looking great; let's get it in.

Issue #4784, opened 2019-07-04
testNext failures in CI: for-await-of


JoshuaKGoldberg:
Example build failure: https://circleci.com/gh/palantir/tslint/16351?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

```
test/rules/await-promise/for-await-of/test.ts.lint: Failed!
Expected (from .lint file)
Actual (from TSLint)
  async function correct(foo: AsyncIterableIterator<string>) {
      for await (const element of foo) {}
  }
  
  async function correct2() {
      for await (const element of asyncGenerator()) {}
-                                 ~~~~~~~~~~~~~~~~     [Invalid 'for-await-of' of a non-AsyncIterable value.]
  }
```

PR #4798, opened 2019-07-13
[max-line-length] ignore strings and regex in max line length

### LINKED ISSUE #3905: max-line-length with template literals
### Bug Report

- __TSLint version__: 5.9.1
- __TypeScript version__: 2.8.3
- __Running TSLint via__: (pick one) CLI / VS Code

```ts
const str = `134567890 1234567890 134567890 1234567890 134567890 1234567890 134567890 1234567890 1234567890 1234567890
1234567890 1234567890 1234567890 1234567890 1234567890`;
```

```json
{
  "rules": {
    "max-line-length": [true, {"limit": 80}]
  }
}
```

#### Actual behavior

```sh
length: 118 const str = `134567890 1234567890 134567890 1234567890 134567890 1234567890 134567890 1234567890 1234567890 1234567890
1234567890 1234567890 1234567890 1234567890 1234567890`;

const str = `134567890 1234567890 134567890 1234567890 134567890 1234567890 134567890 1234567890 1234567890 1234567890

ERROR: /file.ts[1, 1]: Exceeds maximum line length of 80
```
#### Expected behavior

This rule should skip verification for "some" strings, including multiline template literal.

vmk1vmk:
#### PR checklist

- [X] Addresses an existing issue: fixes #3905
  - [X] Includes tests
- [X] Documentation update

#### Overview of change:

Added functionality to ignore strings exceeding the max line length, it is by default enabled.
Added option `check-strings` to control if exceeding strings are skipped or treated as errors.
Added the same option for regex called `check-regex`.

* Are all "string"-cases covered?
* Does this new functionality behave as expected?
* Should it be split into different options like `check-plain-strings` and `check-template-strings`? 

#### CHANGELOG.md entry:

[enhancement] Added `check-strings` and `check-regex` options to `max-line-lenght` rule.


vmk1vmk:
Failing in `testNext` because of #4784

src/rules/maxLineLengthRule.ts
```diff
@@ -57,10 +66,11 @@ export class Rule extends Lint.Rules.AbstractRule {
...
                         type: "object",
                         properties: {
-                            limit: { type: "number" },
-                            "ignore-pattern": { type: "string" },
+                            [OPTION_LIMIT]: { type: "number" },
+                            [OPTION_IGNORE_PATTERN]: { type: "string" },
+                            [OPTION_CHECK_STRINGS]: { type: "boolean" },
+                            [OPTION_CHECK_REGEX]: { type: "boolean" },
                         },
-                        additionalProperties: false,
```
JoshuaKGoldberg: Could you please put the `additionalProperties: false` back in? It's an intentional part of the schema.

src/rules/maxLineLengthRule.ts
```diff
@@ -96,32 +108,98 @@ export class Rule extends Lint.Rules.AbstractRule {
...
-    const limit = ctx.options.limit;
-    const ignorePattern = ctx.options.ignorePattern;
-    for (const line of getLineRanges(ctx.sourceFile)) {
-        if (line.contentLength <= limit) {
-            continue;
+    const { [OPTION_LIMIT]: limit } = ctx.options;
+
+    getLineRanges(ctx.sourceFile)
+        .filter(({ contentLength }: LineRange): boolean => contentLength > limit)
+        .filter((line: LineRange): boolean => !shouldIgnoreLine(line, ctx))
```
JoshuaKGoldberg: Somewhat nitpicking here: could you please combine these two `.filter` statements? Seems odd to have two `.filter`s in a row when you could use a `&&` instead.

src/rules/maxLineLengthRule.ts
```diff
@@ -96,32 +108,98 @@ export class Rule extends Lint.Rules.AbstractRule {
...
-            options = argument as MaxLineLengthRuleOptions;
-            const ignorePattern = (argument as { [key: string]: string })["ignore-pattern"];
-            options.ignorePattern =
-                typeof ignorePattern === "string" ? new RegExp(ignorePattern) : undefined;
+            const {
+                [OPTION_CHECK_REGEX]: checkRegex,
+                [OPTION_CHECK_STRINGS]: checkStrings,
+                [OPTION_IGNORE_PATTERN]: ignorePattern,
+                [OPTION_LIMIT]: limit,
+            } = argument;
```
JoshuaKGoldberg: It's rather difficult to parse through `[OPTON_CHECK_REGEX]` destructurings and co. later on in this file. You're using the same verbose rule format in both the _external_ options for the rule (e.g. this line) and the _internal_ options (the type in the `WalkContext`).

Requested change: please switch back the `MaxLineLengthRuleOptions` type to have camelCase keys, so it can be used directly like `options.limit`.

vmk1vmk:
@JoshuaKGoldberg Thanks for reviewing so fast. I made all requested changes.
How about the functionality and the options? Do you think they should be more granular like `check-strings` and `check-template-strings`?

JoshuaKGoldberg:
> How about the functionality and the options? Do you think they should be more granular like `check-strings` and `check-template-strings`?

No, this looks fine. TSLint's being deprecated soon (#4534) and has been recommending using Prettier instead of formatting rules for quite some time. I don't think that extra logic will be worth the added complexity.

> Are all "string"-cases covered?
> Does this new functionality behave as expected?

Looks good! 😄 I played around with this locally a bit and it worked great.

PR #4807, opened 2019-07-22
Add rule: strict-string-expressions

### LINKED ISSUE #4512: Rule suggestion: strict-string-expressions (no implicit function to string)
### Rule Suggestion

There is some common pattern

```js
const barFunc = () => 'World';
const fooStr = `Hello, ${barFunc()}`; // Hello, World
```

But it's easy to miss brackets around function and this would be not an error

```js
const barFunc = () => 'World';
const fooStr = `Hello, ${barFunc}`; // "Hello, () => 'World'"
```

If it's `on`, require explicit `toString` or `String` call around `function`

Like this

```js
// yeah, we really want function source code here
const barFunc = () => 'World';
const fooStr = `Hello, ${String(barFunc)}`; // "Hello, () => 'World'"
```

**Additional context**

What about spreading this rule for other data types?

I think this is frustrating sometimes:

```js
const name = { foo: 'Foo' };
alert('Hello, ' + name); // Hello, [object Object] (╯°□°）╯︵ ┻━┻ 
```

What do you think?

ColCh:
#### PR checklist

- [x] Addresses an existing issue: fixes #4512
  - [x] Includes tests
- [x] Documentation update `???`

#### Overview of change:

adds implementation for rule discussed in #4512

#### CHANGELOG.md entry:

[new-rule] `strict-string-expressions` reports errors on type coercions found in string expressions


src/rules/strictStringExpressionsRule.ts
```diff
@@ -0,0 +1,99 @@
...
+            The following nodes are checked:
+
+            * String literals ("foo" + bar)
+            * ES6 templates (\`foo \${bar}\`)`,
+        type: "functionality",
+        typescriptOnly: true,
+        requiresTypeInfo: true,
+        options: [],
+        optionExamples: [true],
+        optionsDescription: "Not configurable.",
```
ColCh: I think there is no options at all

src/rules/strictStringExpressionsRule.ts
```diff
@@ -0,0 +1,99 @@
...
+        hasFix: true
+    };
+
+    public static CONVERSION_REQUIRED = 'Explicit conversion to string type required';
```
ColCh: error message is always same

src/rules/strictStringExpressionsRule.ts
```diff
@@ -0,0 +1,99 @@
...
+
+            * String literals ("foo" + bar)
+            * ES6 templates (\`foo \${bar}\`)`,
+        type: "functionality",
+        typescriptOnly: true,
+        requiresTypeInfo: true,
+        options: [],
+        optionExamples: [true],
+        optionsDescription: "Not configurable.",
+        hasFix: true
```
ColCh: fixer just wraps expression into `String(FOO)` call to make it explicit

src/rules/strictStringExpressionsRule.ts
```diff
@@ -0,0 +1,93 @@
...
+    public applyWithProgram(sourceFile: ts.SourceFile, program: ts.Program): Lint.RuleFailure[] {
+        return this.applyWithFunction(sourceFile, walk, undefined, program.getTypeChecker());
+    }
+}
+
+function walk(ctx: Lint.WalkContext<undefined>, checker: ts.TypeChecker): void {
+    const { sourceFile } = ctx;
+    ts.forEachChild(sourceFile, function cb(node: ts.Node): void {
+        switch (node.kind) {
+            case ts.SyntaxKind.BinaryExpression: {
```
ColCh: may be, this one can be omitted? due to `restrictPlusOperandsRule`

JoshuaKGoldberg:
@ColCh looks like you've got some lint failures from CI running this rule on TSLint itself. https://circleci.com/gh/palantir/tslint/16921?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

That looks like it's exposed an edge case or two. Should `| null` and `| undefined` be allowed in string expressions? I think so; it's common for error logging to have strings like:

```js
function complain(error?: Error | null) {
    console.error(`Oh dear: ${error}`);
}
```

`new Error().stack` being `string | undefined` is also pretty annoying.

If you really wanted, you could make an issue to add an option to disallow this, or file a followup issue after this PR... but IMO it's fine to just always allow these.

ColCh:
Yes, I agree with you. That `undefined` in union as well as other empty-in-runtime values like `never` or `null` creates so much problems :D

I think that better option is make those as configurable thing

src/rules/strictStringExpressionsRule.ts
ColCh: oops. at least one now

src/rules/strictStringExpressionsRule.ts
ColCh: I think not, because they do different things about `plus` operator

src/linter.ts
```diff
@@ -187,7 +187,7 @@ export class Linter {
             this.options.formatter !== undefined ? this.options.formatter : "prose";
         const Formatter = findFormatter(formatterName, this.options.formattersDirectory);
         if (Formatter === undefined) {
-            throw new Error(`formatter '${formatterName}' not found`);
+            throw new Error(`formatter '${String(formatterName)}' not found`);
```
ColCh: real case here 🎉
image

ColCh:
@JoshuaKGoldberg seems to pass now

I don't understand why `ci/circleci: testNext ` is failing but it's failing in master branch too

Screenshot 2019-07-28 at 22 09 03

so... please review :)

src/rules/strictStringExpressionsRule.ts
```diff
@@ -0,0 +1,159 @@
...
+            walk,
+            this.getRuleOptions(),
+            program.getTypeChecker(),
+        );
+    }
+
+    private getRuleOptions(): Options {
+        if (this.ruleArguments[0] === undefined) {
+            return {
+                [OPTION_ALLOW_EMPTY_TYPES]: true,
```
ColCh: empty types are enabled by default

ColCh: I'm not sure about name though.. can `never` be considered as `empty`?

JoshuaKGoldberg: Great semantic question. Technically `never` is a [https://en.wikipedia.org/wiki/Bottom_type](bottom type); however, `void`, `undefined`, etc. are separate types all on their own. `empty` seems like a good way to refer to all these.

ColCh:
Thank you!

src/rules/strictStringExpressionsRule.ts
ColCh: yeah, that was difficult to find common name for these

PR #4814, opened 2019-07-27
fix: add bigint support to restrict-plus-operand rule

### LINKED ISSUE #4741: "restrict-plus-operands" doesn't understand BigInt
### Bug Report

- __TSLint version__: 5.16.0
- __TypeScript version__: 3.4.0
- __Running TSLint via__: CLI

```ts
BigInt(1) + BigInt(2);
```

```json
{
  "rules": {
    "restrict-plus-operands": true
  }
}
```

#### Actual behavior
`Operands of '+' operation must either be both strings or both numbers, but found bigint + bigint`

#### Expected behavior
No errors.

vedadeepta:
#### PR checklist

- [x] Addresses an existing issue: fixes #4741
- [x] Includes tests
- [ ] Documentation update

#### Overview of change:
Add support for bigint in `restrict-plus-operand` rule. Checks if the node type is equal `ts.TypeFlags.BigInt` or `ts.TypeFlags.BigIntLiteral`.

Maybe make this optional because `bigint` is not supported below typescript 3.2 and is not supported for javascript target es2017 & below. More on this [typscript caveats](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-2.html#caveats-1).

#### CHANGELOG.md entry:

[enhancement] add BigInt support to `restrict-plus-operands` rule


vedadeepta:
Ok, but wouldn't that be a breaking change for devs who use `ts < 3.2` ? It will throw a error of `...., but found type any + any` instead of a proper error message for bigints.

PR #4833, opened 2019-08-21
4746: fix weird failures in strict-boolean-expressions for 'ignore-rhs' option

### LINKED ISSUE #4746: "ignore-rhs" causes weird failures
### Bug Report

- __TSLint version__: 5.16.0
- __TypeScript version__: 3.3.4000
- __Running TSLint via__: CLI

```ts
let bool1:boolean = true
let bool2:boolean = true
let obj:{}|undefined = {}

const func = () => {
  if(bool1 && obj && bool2) {
    return 1
  } else {
    return 2
  }
}
```

```json
{
  "rules": {
    "strict-boolean-expressions": [true, "allow-undefined-union", "ignore-rhs"]
  }
}
```

#### Actual behavior

Get error `This type is not allowed in the operand for the '&&' operator because it unions more than one truthy/falsy type. Allowed types are boolean or undefined-union.`

#### Expected behavior

It should pass. If I remove "ignore-rhs" from tslint.json it passes.

tanmoyopenroot:
#### PR checklist

- [x] Addresses an existing issue: fixes #4746 
- [x] bugfix
- [x] Includes tests
- [ ] Documentation update

#### Overview of change:
check expression only if `left` node is not a `BooleanBinaryExpression`

### DIFF
--- a/src/rules/strictBooleanExpressionsRule.ts
+++ b/src/rules/strictBooleanExpressionsRule.ts
@@ -153,7 +153,7 @@ function walk(ctx: Lint.WalkContext<Options>, checker: ts.TypeChecker): void {
                     // on the left side because it will be checked well enough on its own.  However,
                     // if ignore-rhs is on, we have to analyze the overall result of the left
                     // side no matter what, because its right side might not follow the rules.
-                    if (options.ignoreRhs || !isBooleanBinaryExpression(left)) {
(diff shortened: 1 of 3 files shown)


JoshuaKGoldberg:
Failures in testNext and testRc are unrelated.

PR #4839, opened 2019-08-25
Added allow-generics option to invalid-void rule

### LINKED ISSUE #4828: Async function return void cause `invalid void`
### Bug Report

- __TSLint version__: 5.19.0
- __TypeScript version__: 3.5.3
- __Running TSLint via__: CLI / VSCode 

```ts
const sayHello = async (): Promise<void> => {
  // do something
}
```

```json
default
```

#### Actual behavior
After install tslint@5.19.0, tslint reports it. In previous version tslint doesn't report this error.

tslint report 
```
void is not a valid type other than return types (invalid-void)tslint(1)
```
#### Expected behavior
I think `async function return  Promise<void>` is general usage.  tslint shouldn't report the error.
[invalid-void doc](https://palantir.github.io/tslint/rules/invalid-void/) seems don't contain this usage.

JoshuaKGoldberg:
#### PR checklist

- [x] Addresses an existing issue: fixes #4828 
  - [x] Includes tests
- [x] Documentation update

#### Overview of change:

Adds an `allow-generics` option to `invalid-void` for the sake of common uses such as `Promise<void>`. It's `true` by default, can be specified as `false`, and also allows a `string[]` whitelist: e.g. `"allow-generics": ["Promise"]`.

#### CHANGELOG.md entry:

[new-rule-option]: Added allow-generics to invalid-void


src/rules/invalidVoidRule.ts
```diff
@@ -15,35 +15,84 @@
...
+            [true, { [OPTION_ALLOW_GENERICS]: ["Promise", "PromiseLike"] }],
+        ],
         type: "maintainability",
         typescriptOnly: true,
     };
     /* tslint:enable:object-literal-sort-keys */
 
-    public static FAILURE_STRING = "void is not a valid type other than return types";
+    public static FAILURE_STRING_ALLOW_GENERICS =
+        "void is not a valid type other than generic or return types";
```
adidahiya:
```suggestion
        "void is only valid as a return type or generic type variable";
```

src/rules/invalidVoidRule.ts
```diff
@@ -15,35 +15,84 @@
...
+    public static FAILURE_STRING_NO_GENERICS = "void is not a valid type other than return types";
```
adidahiya:
```suggestion
    public static FAILURE_STRING_NO_GENERICS = "void is only valid as a return type";
```

src/rules/invalidVoidRule.ts
```diff
@@ -15,35 +15,84 @@
...
+    public static FAILURE_WRONG_GENERIC = (genericName: string) =>
+        `${genericName} may not have void as a generic type`;
```
adidahiya:
```suggestion
        `${genericName} may not have void as a type variable`;
```

JoshuaKGoldberg:
Thanks @adidahiya - sorry about the delay!

PR #4841, opened 2019-08-26
Fix bug where the strict ts flag wasn't recognised correctly by `no-unnecessary-type-assertion`

### LINKED ISSUE #4840: Rule no-unnecessary-type-assertion doesn't work if tsconfig has `"strict": true`
### Bug Report
Rule no-unnecessary-type-assertion doesn't work if your `tsconfig.json` has `"strict": true`. It was made so that it only works if you have `"strictNullChecks": true` but should also work for `"strict": true` since that covers all strict rules.

Bug introduction: https://github.com/palantir/tslint/commit/b836f285e63d7dc44908eae40ef5d6feb16e026a#diff-2237d1e758ab18614b7d976b89e56ea0R53
Strict rule explanation: https://www.typescriptlang.org/docs/handbook/compiler-options.html

Note, will fix this myself so PR incoming.

#### Reproduction using [TSLint Playground](https://palantir.github.io/tslint-playground)
Playground doesn't allow you to change the tsconfig?

_tsconfig.json_
```json
{
  "compilerOptions": {
    /* Enable all strict type-checking options. */
    "strict": true,
    /* Enable strict null checks. */
    // "strictNullChecks": true
  }
}
```

_Code_
```ts
const x = "y";
x!.length;
```

_with `tslint.json` configuration:_
```json
{
  "defaultSeverity": "error",
  "rules": {
    "no-unnecessary-type-assertion": true
  }
}
```

#### Actual behavior
No errors.

#### Expected behavior
An error!

guidsdo:
This commit fixes #4840

#### PR checklist

- [x] Addresses an existing issue: fixes #4840
  - [x] Includes tests
- [x] Documentation update

#### Overview of change:
Fixed simple bug. 

Nope.

#### CHANGELOG.md entry:

[bugfix] fix `no-unnecessary-type-assertion` no error when `"strict": true` in compiler flags

### DIFF
--- a/src/rules/noUnnecessaryTypeAssertionRule.ts
+++ b/src/rules/noUnnecessaryTypeAssertionRule.ts
@@ -44,13 +44,17 @@ export class Rule extends Lint.Rules.TypedRule {
         "This assertion is unnecessary since it does not change the type of the expression.";
 
     public applyWithProgram(sourceFile: ts.SourceFile, program: ts.Program): Lint.RuleFailure[] {
+        const strictChecksEnabled = !!program.getCompilerOptions().strict;
+        const strictNullChecksEnabled = program.getCompilerOptions().strictNullChecks === true;
+        const strictNullChecksNotDisabled = program.getCompilerOptions().strictNullChecks !== false;
+
         return this.applyWithWalker(
             new Walker(
                 sourceFile,
                 this.ruleName,
                 this.ruleArguments,
                 program.getTypeChecker(),
-                !!program.getCompilerOptions().strictNullChecks,
+                strictNullChecksEnabled || (strictChecksEnabled && strictNullChecksNotDisabled),
             ),
         );
     }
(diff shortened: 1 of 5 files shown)


src/rules/noUnnecessaryTypeAssertionRule.ts
```diff
@@ -44,13 +44,17 @@ export class Rule extends Lint.Rules.TypedRule {
         "This assertion is unnecessary since it does not change the type of the expression.";
 
     public applyWithProgram(sourceFile: ts.SourceFile, program: ts.Program): Lint.RuleFailure[] {
+        const strictChecksEnabled = !!program.getCompilerOptions().strict;
```
JoshuaKGoldberg: Love the variable names here. They're super informative!

One quick request - let's move the `program.getCompilerOptions()` into its own variable? Having it repeated a few times is relying on the assumption that the function is a quick lookup, and makes the code a little less readable IMO.

```suggestion
        const compilerOptions = program.getCompilerOptions();
        const strictChecksEnabled = !!compilerOptions.strict;
        // ...and the other two
```

guidsdo: Fixed

PR #4842, opened 2019-08-28
Support granular config in 'object-literal-shorthand'

### LINKED ISSUE #3778: [Feature Request] Add more granularity to object-literal-shorthand rule options
The `object-literal-shorthand` rule currently does not offer granularity for enforcing/disallowing shorthand separately for property assignment vs method declarations. 

I would prefer to allow (but not enforce!) shorthand for method declarations, but disallow shorthand for property assignments. 

Here's an example of how I would like to configure the `object-literal-shorthand` rule:

```ts
const someValue = 20;

const myObj = { 
    // error: shorthand property assignment not allowed
    someValue,
    // valid
    anotherValue: someValue,
    // valid
    someMethod(): number {
        return 42;
    },
    // also valid
    anotherMethod: (): number => {
       return 42;
    }
};
```

The motivation for this is that shorthand method declaration is very convenient and improves readability, while shorthand property assignment creates a coupling between the object's property name and a local variable name that can lead to bugs when refactoring code later.

pablobirukov:
#### PR checklist

- [x] Addresses an existing issue: fixes #3778 
- [x] Enhancement
  - [x] Includes tests

#### Overview of change:

Support granular configuration in `object-literal-shorthand`

#### CHANGELOG.md entry:

[new-rule-option] `object-literal-shorthand` supports `{"property"?: "never", "method"?: "never"}` as config options

### DIFF
--- a/src/rules/objectLiteralShorthandRule.ts
+++ b/src/rules/objectLiteralShorthandRule.ts
@@ -28,7 +28,19 @@ import * as ts from "typescript";
 
 import * as Lint from "..";
 
-const OPTION_NEVER = "never";
+const OPTION_VALUE_NEVER = "never";
+const OPTION_KEY_PROPERTY = "property";
+const OPTION_KEY_METHOD = "method";
+
+interface RawOptions {
+    [OPTION_KEY_PROPERTY]?: "never" | "always";
+    [OPTION_KEY_METHOD]?: "never" | "always";
+}
+
+interface Options {
+    enforceShorthandMethods: boolean;
+    enforceShorthandProperties: boolean;
+}
 
 export class Rule extends Lint.Rules.AbstractRule {
     /* tslint:disable:object-literal-sort-keys */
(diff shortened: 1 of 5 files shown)


src/rules/objectLiteralShorthandRule.ts
```diff
@@ -37,12 +49,37 @@ export class Rule extends Lint.Rules.AbstractRule {
         description: "Enforces/disallows use of ES6 object literal shorthand.",
         hasFix: true,
         optionsDescription: Lint.Utils.dedent`
-        If the \'never\' option is provided, any shorthand object literal syntax will cause a failure.`,
+        If the \'never\' option is provided, any shorthand object literal syntax will cause a failure.
+        With \`{"property": "never"}\` provided, the rule fails on property shothands only,
```
adidahiya:
```suggestion
        With \`{"property": "never"}\` provided, the rule only flags property shorthand assignments,
```

src/rules/objectLiteralShorthandRule.ts
```diff
@@ -37,12 +49,37 @@ export class Rule extends Lint.Rules.AbstractRule {
...
+        and respectively with \`{"method": "never"}\`, the rule fails only on method shorthands`,
```
adidahiya:
```suggestion
        and respectively with \`{"method": "never"}\`, the rule fails only on method shorthands.`,
```

src/rules/objectLiteralShorthandRule.ts
```diff
@@ -37,12 +49,37 @@ export class Rule extends Lint.Rules.AbstractRule {
         description: "Enforces/disallows use of ES6 object literal shorthand.",
         hasFix: true,
         optionsDescription: Lint.Utils.dedent`
-        If the \'never\' option is provided, any shorthand object literal syntax will cause a failure.`,
+        If the \'never\' option is provided, any shorthand object literal syntax will cause a failure.
```
adidahiya: Can you please mention at the beginning of this description that by default, with no options specified, that it enforces shorthand always?

Also, for clarity, I think we should support the following config: `{ "property": "always", "method": "never" }` (and vice versa). We should mention that "always" is assumed the default if the respective key is not specified in the config, but we should also allow users to be explicit about it so that it's easier to read tslint.json.

src/rules/objectLiteralShorthandRule.ts
```diff
@@ -52,74 +89,101 @@ export class Rule extends Lint.Rules.AbstractRule {
...
+        if (options.indexOf(OPTION_VALUE_NEVER) !== -1) {
+            return {
+                enforceShorthandMethods: false,
+                enforceShorthandProperties: false,
+            };
+        }
+        const optionsObject: RawOptions | undefined = options.find(
+            (el: string | RawOptions): el is RawOptions =>
+                typeof el === "object" &&
+                (el[OPTION_KEY_PROPERTY] === "never" || el[OPTION_KEY_METHOD] === "never"),
```
adidahiya:
```suggestion
                (el[OPTION_KEY_PROPERTY] === OPTION_VALUE_NEVER || el[OPTION_KEY_METHOD] === OPTION_VALUE_NEVER),
```

src/rules/objectLiteralShorthandRule.ts
```diff
@@ -52,74 +89,101 @@ export class Rule extends Lint.Rules.AbstractRule {
...
         );
+        if (optionsObject !== undefined) {
+            return {
+                enforceShorthandMethods: !(optionsObject[OPTION_KEY_METHOD] === "never"),
```
adidahiya: same here, use the constant

Issue #3592, opened 2017-12-18
Split out formatting rules into a separate package


adidahiya:
Following up on my comments here: https://github.com/palantir/tslint/pull/3395#issuecomment-347967428

I'd like to separate these rules and disable them in the `tslint:recommended` configuration so that it's easier to use prettier with tslint. Prettier does a great job of formatting and I'd rather invest time & energy in other tslint rules that are not formatting-related.

ajafff:
That means you want to publish a separate npm package like `@tslint/formatting`? Are these rules maintained in a separate repo or is this one converted to a monorepo?

Btw. if you want to publish packages with `@tslint` scope on npm: I accidentally registered that org and would hand it over to you.

adidahiya:
I'd like to convert this to a monorepo; I think tracking all issues together in one repo is a good workflow we should retain.

Thanks for the offer about the `@tslint` scope -- I haven't decided if we should use scoped packages or use names like `tslint-core`, `tslint-formatting`, etc... but it's certainly worth considering (we'll avoid conflicts if we use the scope).

JoshuaKGoldberg:
### 💀 _**It's time!**_ 💀

**TSLint is being deprecated** and no longer accepting pull requests for major new changes or features. See #4534. 😱

If you'd like to see this change implemented, you have two choices:
* **Recommended:** Check if this is available in [ESLint](https://eslint.org) + [typescript-eslint](https://typescript-eslint.io) ✅
* _Not Recommended: Fork TSLint locally_ 🤷‍♂️

👋 It was a pleasure open sourcing with you!

_If you believe this message was posted here in error, please comment so we can re-open the issue!_

JoshuaKGoldberg:
🤖 Beep boop! 👉  **TSLint is deprecated** 👈 _(#4534)_  and you should **switch to [typescript-eslint](https://typescript-eslint.io)**! 🤖 

🔒 This issue is being locked to prevent further unnecessary discussions. Thank you! 👋

PR #4871, opened 2019-10-05
Removed annoying formatting rules from recommended ruleset

### LINKED ISSUE #1520: tslint:recommended interface-name uses always-prefix
`tslint:recommended` defaults to `always-prefix`. However, the authoritative [sources](https://github.com/Microsoft/TypeScript/wiki/Coding-guidelines) [I can](https://basarat.gitbooks.io/typescript/content/docs/styleguide/styleguide.html) [find](https://www.typescriptlang.org/docs/handbook/interfaces.html) all discourage this behavior.

I cannot find documented rationale for tslint's choice to recommend `always-prefix` despite being at odds with the broader community's practice.

Please document the rationale for this, and/or consider changing the default to `never-prefix`.

### LINKED ISSUE #3248: interface-over-type-literal shouldn't be recommended
Yes, I know the Typescript docs say

> Because an ideal property of software is being open to extension, you should always use an interface over a type alias if possible.

but seriously, if I am making a simple type with zero intention of ever using it as an extendable thing, it makes no sense to code it as an interface.

### LINKED ISSUE #4305: Relax 'recommended' rules or make them auto-fixable
The recommended rules include some (imho) impossible rules to abide by. Sorting your object keys? Why would I want them alphabetic if they made more sense in some other ordering? Furthermore, it's not a fixable rule, it just complains. I am a huge fan of Typescript and consider linting a very important tool. However, rules like:

* object-literal-sort-keys
* ordered-imports
* interface-name

are too opinionated and prove a burden on the programmer who already has lots of problems to solve while developing a project. The set of 'recommended' rules should, in my opinion, be much more lax than that to reduce the barrier of entry to Typescript and tslint. They should be guiding and helpful, not opinionated and pedantic.

I propose removing the more 'extreme' rules from the recommended ruleset.

### LINKED ISSUE #4869: 'allow-leading-underscore' option for 'variable-name' rule
The 'recommended' configuration should include 'allow-leading-underscore' option for 'variable-name' rule, because this pattern is used when skipping parameters in callback functions.

JoshuaKGoldberg:
#### PR checklist

- [x] Addresses an existing issue: fixes #1520, fixes #3248, fixes #4305, fixes #4869. 
  - ~[ ] Includes tests~
- ~[ ] Documentation update~

#### Overview of change:

Removes a bunch of rules from `recommended.ts` around formatting and whitespace. Also adds `allow-leading-underscore` to `variable-name`.

#### CHANGELOG.md entry:

[rule-change] Removed formatting and whitespace rules from recommended ruleset


src/configs/recommended.ts
```diff
@@ -17,14 +17,9 @@
 
 export const rules = {
     "adjacent-overload-signatures": true,
-    align: {
-        options: ["parameters", "statements"],
-    },
     "array-type": {
-        options: ["array-simple"],
+        options: ["array"],
```
JoshuaKGoldberg: This is the only non-formatting change here. The default's inconsistency has been bugging me for a while... I can remove or revert this if you'd prefer?

adidahiya: Fair enough, the inconsistency is a little confusing...

src/configs/recommended.ts
```diff
@@ -43,37 +38,21 @@ export const rules = {
...
     "no-any": false,
     "no-arg": true,
     "no-bitwise": true,
     "no-conditional-assignment": true,
-    "no-consecutive-blank-lines": true,
     "no-console": true,
     "no-construct": true,
     "no-debugger": true,
     "no-duplicate-super": true,
-    "no-empty": true,
```
adidahiya: this isn't really a formatting rule, it's more about code style. I think we should keep it in the recommended ruleset

src/configs/recommended.ts
```diff
@@ -43,37 +38,21 @@ export const rules = {
...
-    "no-empty-interface": true,
```
adidahiya: same here, it's a code style rule. I think we should keep it.

dbartholomae:
When will this be released? :)

ThibaudAV:
😢 @JoshuaKGoldberg why these changes? it removes a lot of very appreciable rules in the so-called "recommended" rule set

Flowkap:
Great .. i was wondering why code suddenly did not have enforced rules. Trailingsemicolon and commans are for better git changelogs. We relied on the recommended to better document what we wanted to tweak... this was probably a wrong decision ... No wi need to clean up a lot of mess cause devs of course upgraded tslint package and got less strict new defaults.

dbartholomae:
For pure styling I would recommend prettier. This will also autofix your code :)

Flowkap:
For good reason were not. Those checks are not for styling but good reason as clean git history. I would approve some of the removals which are highly opinionated. But commans and curly braces as well as semicolons have more meaning and often clutter diffs.

adidahiya:
This change to the recommended rule set had been planned for a [very long time](https://github.com/palantir/tslint/issues/3592). [TSLint is deprecated](https://github.com/palantir/tslint/issues/4534) and we won't be making any more significant changes to its API.

PR #4889, opened 2019-11-07
fix(strict-comparison): Support comparing generic types with constraint

### LINKED ISSUE #4838: strict-comparison false positive (TSLint 5.19.0)
# Bug Report

*    TSLint version: 5.19.0
*    TypeScript version: 3.5.3
*    Running TSLint via: VSCode

## TypeScript code being linted

```ts
const MyType = string | boolean

class MyClass<T extends MyType> {
    myMethod(param1: T) {
        if(param1 === true) return 'Yes'  // <-- Cannot compare type 'object' to type 'boolean'. (strict-comparisons)
        if(param1 === false) return 'No'  // <-- Cannot compare type 'object' to type 'boolean'. (strict-comparisons)
        return param1
    }
}
```

## Actual behavior

strict-comparison error.

## Expected behavior

No error.

## Other

Similar to #4774

jacekjagiello:
#### PR checklist

- [x] Addresses an existing issue: fixes #4838 
  - [x] Includes tests
- [ ] Documentation update

#### Overview of change:
Fix strict-comparison false positive by supporting Gnenric types that have a constraint

* A Better way of finding constraint declaration(without passing around `sourceFile`)
* Better type support(avoiding using `as`)
* Test cases that involve generics with constraints, which this implementation does not support


src/rules/strictComparisonsRule.ts
```diff
@@ -276,6 +354,26 @@ function isEqualityOperator(node: ts.BinaryExpression): boolean {
     }
 }
 
+function getKindFromSyntaxKind(type: ts.SyntaxKind): TypeKind {
+    // tslint:disable:prefer-switch
```
JoshuaKGoldberg: Is there a reason you wanted an `if`/`else` here? The `prefer-switch` rule was enabled on the codebase as a stylistic preference to keep the code more consistent. A `switch` statement, `Map`, or `object` would generally be what we'd prefer.

test/rules/strict-comparisons/default/test.ts.lint
```diff
@@ -188,3 +188,17 @@ const func = (param: string | null = null) => {
 
 [CANNOT_USE]: Cannot use '%s' comparator for type '%s'.
 [CANNOT_COMPARE]: Cannot compare type '%s' to type '%s'.
+
+const MyType = string | boolean | number
+
+class MyClass<T extends MyType> {
```
JoshuaKGoldberg: I suspect there are edge cases we'll want to cover here. Would you mind adding the following?

* Generic `function () {}`s and `() => {}`s
* Closures with multiple generic types: i.e. `<T, U>`
* Generics with constraints that _do_ match up: i.e. `<T extends number, U extends number>`
* Generics with constraints that _don't_ match up: i.e. `<T extends number, U extends string>`

src/rules/strictComparisonsRule.ts
```diff
@@ -120,9 +120,68 @@ export class Rule extends Lint.Rules.TypedRule {
...
+    if (paramDeclaration === undefined) {
+        return undefined;
+    }
+
+    const constraint = getEffectiveConstraintOfTypeParameter(paramDeclaration);
+    if (constraint === undefined) {
+        return undefined;
+    }
+
+    if (ts.isTypeReferenceNode(constraint)) {
```
JoshuaKGoldberg: Ah, the old `ts.is*` trap 😄 you'll want to use `tsutils` for that:

```suggestion
    if (tsutils.isTypeReferenceNode(constraint)) {
```

src/rules/strictComparisonsRule.ts
```diff
@@ -120,9 +120,68 @@ export class Rule extends Lint.Rules.TypedRule {
     }
 }
 
+function getConstraint(t: ts.Type): ts.TypeReferenceNode | undefined {
+    if (t.getSymbol() === undefined) {
+        return undefined;
+    }
+
+    const paramDeclaration = t.symbol.declarations.find(
+        d => d.kind === ts.SyntaxKind.TypeParameter,
```
JoshuaKGoldberg: Yeah that `as` is a little unfortunate... `tsutils'` methods annotated with TypeScript's `is` return syntax should help with that:

```suggestion
        tsutils.isTypeParameter,
```

If strict equality checking against `undefined` gets in your way, you could either leave this as is, or use `filter` instead and check for `.length` I suppose...

src/rules/strictComparisonsRule.ts
```diff
@@ -120,9 +120,68 @@ export class Rule extends Lint.Rules.TypedRule {
...
+
+function findIdentifiers(expressions: ts.Expression[]): ts.Identifier[] {
+    return expressions
+        .map(e => {
+            if (isBinaryExpression(e)) {
+                return findIdentifiers([e.left, e.right]);
+            }
+            return e;
+        })
+        .reduce<ts.Identifier[]>((acc, val) => acc.concat(val as ts.Identifier[]), []);
```
JoshuaKGoldberg: A `flatMap` utility exists in `src/utils.ts`, maybe that would simplify this a bit?

src/rules/strictComparisonsRule.ts
```diff
@@ -276,6 +354,26 @@ function isEqualityOperator(node: ts.BinaryExpression): boolean {
     }
 }
 
+function getKindFromSyntaxKind(type: ts.SyntaxKind): TypeKind {
```
JoshuaKGoldberg: Later on, in the only use of `getKindFromSyntaxKind`, the returned value is cast `as ts.SyntaxKind`. Maybe that's what should be returned here?

JoshuaKGoldberg:
### 💀 _**It's time!**_ 💀

**TSLint is being deprecated** and no longer accepting pull requests other than security fixes. See #4534. ☠️ 

If you'd like to see this feature implemented, you have two choices:
* **Recommended:** Use [typescript-eslint](https://typescript-eslint.io) to lint your TypeScript code with ESLint ✅
* **Not Recommended:** Fork TSLint and add this feature in the forked version 🤷‍♂️

👋 It was a pleasure open sourcing with you!

PR #4900, opened 2019-12-04
Allow suppressing unnecessary null type asserts.

mprobst:
#### PR checklist

  - [x] Includes tests

#### Overview of change:

Previously, `noUnnecessaryTypeAssertionRule` could only suppress
specific types to cast to, but not `!`. This change adds a specific
check and branch to allow `!` assertions.

#### CHANGELOG.md entry:

[enhancement] `no-unnecessary-type-assertion` can now ignore `!` assertions.

### DIFF
--- a/src/rules/noUnnecessaryTypeAssertionRule.ts
+++ b/src/rules/noUnnecessaryTypeAssertionRule.ts
@@ -32,7 +32,7 @@ export class Rule extends Lint.Rules.TypedRule {
                 items: { type: "string" },
             },
         },
-        optionsDescription: "A list of whitelisted assertion types to ignore",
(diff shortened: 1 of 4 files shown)


JoshuaKGoldberg:
Hi @mprobst! I appreciate that you've put time into this PR and it's got to be (at least) annoying to be told 'no', but... TSLint is being deprecated and is no longer accepting these kinds of feature enhancements: #4534.

We would have normally asked that an issue be filed to request the feature as well, so we can discuss the right way to add it in before tackling code. You might wish to do so on https://github.com/typescript-eslint/tslint-to-eslint-config/issues. Good luck!

Issue #4914, opened 2020-02-25
TypeScript 3.8 Support


rrdelaney:
Hi! 👋

TSLint no longer builds using TypeScript 3.8, and we'd like to fix the source code up to at least be able to build TSLint, without adding new features. I talked with @JoshuaKGoldberg a bit about this last Friday, and we seemed to think this was a feasible idea 😄This issue is mostly just for tracking our effort, and fielding if TSLint would be open to PR's before we begin implementation.

Shinigami92:
TSLint is deprecated, https://github.com/typescript-eslint/typescript-eslint should be used as a successor

JoshuaKGoldberg:
Indeed, TSLint is deprecated, but we would want to still be able to build source code in the case of needing to release *security fix* -- as explicitly mentioned in #4534.

@rrdelaney we would definitely welcome PRs to fix the source. #4888 is probably a good start. You mentioned offline that the AST changes in TS 3.8 are also bad for TSLint. It'd be nice to have one issue for each group of source building breakage, though I understand it can be hard to group them without a deep investigation.

JoshuaKGoldberg:
@adidahiya what are your thoughts on releasing a patch version that fixes TS 3.8 crashes, and in general, releasing patch versions for breaking TypeScript changes?

It's my understanding that some larger framework users -e.g. Angular- don't yet have a fully supported option to switch to typescript-eslint (Angular reference: https://github.com/angular/angular-cli/issues/13732). Which means they're stuck with us for now.

I'd be in favor of expanding the schedule in #4534 to include security fixes _and fixes for crashes introduced by breaking TypeScript changes_.

Shinigami92:
Why invest time in a dead project instead of invest time to help projects like Angular finish the migration?

adidahiya:
@JoshuaKGoldberg I'm certainly open to it if it's not too much work... supporting 3.8 sounds like a good idea

JoshuaKGoldberg:
Good question @Shinigami92! A couple of main reasons:

* It'd be very little work to release a new patch version of TSLint with the fixes from this issue. Helping Angular finish the migration is quite a lot of work.
* Even when, say, a new version of Angular completely migrates to ESLint, many users will still be stuck on older versions that rely on TSLint.

[...]

PR #4916, opened 2020-02-29
Bumped supported versions of Node and TypeScript in CircleCI

JoshuaKGoldberg:
#### PR checklist

- [x] Addresses an existing issue: supports #4914 
- ~[ ] New feature, bugfix, or enhancement~
  - ~[ ] Includes tests~
- ~[ ] Documentation update~

#### Overview of change:

Removes the now-EOL Node 6 from CircleCI, and explicitly adds TypeScript versions 3.1-3.8 to tests.

I noticed `appveyor.yml` is still hanging around, so I removed that too.


.circleci/config.yml
```diff
@@ -118,7 +106,7 @@ jobs:
             - run: yarn test
 
     test2.9:
-        docker: *docker-node-6
+        docker: *docker-node-
```
adidahiya: missing `-8`, which might be tripping up circle in running the workflow

JoshuaKGoldberg: Ha, thanks. I'm surprised Circle didn't give a better error indicator for the invalid reference...

JoshuaKGoldberg:
I think this ranks up there with one of the most amusing open source CircleCI flows I've seen in a JavaScript/Node project...

Screenshot of the TSLint CircleCI flow with 18 parallel end steps

JoshuaKGoldberg:
@adidahiya I'm wondering if we could get away with maybe a third of those languages... say, 3.1, 3.4, and 3.8?

PR #4943, opened 2020-07-23
fix: add support for TypeScript 4.0

### LINKED ISSUE #4941: align rule (and possibly others) broken with TypeScript 4.0.0-dev
Latest master builds are failing: https://app.circleci.com/pipelines/github/palantir/tslint/8/workflows/ae8bbef2-5e19-45e6-b5f2-39f93e1330eb/jobs/18974/steps

alan-agius4:
Closes #4941

### DIFF
--- a/package.json
+++ b/package.json
@@ -40,11 +40,11 @@
         "mkdirp": "^0.5.3",
         "resolve": "^1.3.2",
         "semver": "^5.3.0",
-        "tslib": "^1.10.0",
+        "tslib": "^1.13.0",
         "tsutils": "^2.29.0"
     },
     "peerDependencies": {
-        "typescript": ">=2.3.0-dev || >=2.4.0-dev || >=2.5.0-dev || >=2.6.0-dev || >=2.7.0-dev || >=2.8.0-dev || >=2.9.0-dev || >=3.0.0-dev || >= 3.1.0-dev || >= 3.2.0-dev"
+        "typescript": ">=2.3.0-dev || >=2.4.0-dev || >=2.5.0-dev || >=2.6.0-dev || >=2.7.0-dev || >=2.8.0-dev || >=2.9.0-dev || >=3.0.0-dev || >= 3.1.0-dev || >= 3.2.0-dev || >= 4.0.0-dev"
     },
     "devDependencies": {
         "@octokit/rest": "^16.24.3",
@@ -73,7 +73,7 @@
         "tslint-config-prettier": "^1.18.0",
         "tslint-plugin-prettier": "^2.0.1",
         "tslint-test-config-non-relative": "file:test/external/tslint-test-config-non-relative",
-        "typescript": "~3.8.2",
+        "typescript": "~4.0.0-beta",
         "yarn-deduplicate": "^1.1.1"
     },
     "engines": {
--- a/src/rules/alignRule.ts
+++ b/src/rules/alignRule.ts
@@ -142,7 +142,9 @@ class AlignWalker extends Lint.AbstractWalker<Options> {
                     case ts.SyntaxKind.TupleType:
                         if (this.options.elements) {
                             this.checkAlignment(
-                                (node as ts.TupleTypeNode).elementTypes,
+                                // In TS 4 TupleTypeNode.elementTypes has been updated to elements
+                                // tslint:disable-next-line
+                                (node as any).elementTypes || (node as ts.TupleTypeNode).elements,
                                 OPTION_ELEMENTS,
                             );
                         }
--- a/src/rules/noRedundantJsdocRule.ts
+++ b/src/rules/noRedundantJsdocRule.ts
@@ -73,7 +73,7 @@ function walk(ctx: Lint.WalkContext): void {
                     );
                 }
                 break;
-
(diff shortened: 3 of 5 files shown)


palantirtech:
Thanks for your interest in [palantir/tslint](https://github.com/palantir/tslint), @alan-agius4! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

package.json
```diff
@@ -40,11 +40,11 @@
         "mkdirp": "^0.5.3",
         "resolve": "^1.3.2",
         "semver": "^5.3.0",
-        "tslib": "^1.10.0",
+        "tslib": "^2.0.0",
```
JoshuaKGoldberg, line 43: Is there a reason to bump the major version of tslint? This would be a breaking change for some consumers, so ideally we should try to avoid it if possible.

alan-agius4: Why is this a breaking change? Tslib is bound to the TS version used to compile the library, ie: tslint. (Unless I am missing some other use-case where tslib is used in tslint)

Since TS 3.9, tslib version 2 is recommended, to avail of new features. More context about this here: https://github.com/microsoft/tslib/pull/109

That said, if no one reported any issue I can revert this, but technically it should be 2.0.0.

JoshuaKGoldberg: Gotcha, thanks, that's helpful. The case I'm a little concerned about is a rule/plugin consumer accidentally relying on TSLint using a particular version of TSLint, though admittedly that is a pretty rare case.

@adidahiya, is this explicitly a concern or non-concern to your knowledge?

If no reasons to avoid the version are mentioned by anyone with historical context by the time TS 4.0 lands, I think we can safely merge.

alan-agius4: In that case the plugin author will be relying on perfect module hosting by package managers. Ie: the starts to align 😀

adidahiya: I'm fairly confident that an upgrade to tslib 2.0.0 is __not__ required here. If we publish tslint built with tslib 2.0.0, then our consumers will be required to use TS 3.9+, and that means our `peerDependencies` typescript range is invalid. I just did a large upgrade to tslib 2.0.0 across many repos and it was fairly disruptive, so I would like to avoid that in this PR. TS 3.9+ should work fine with the latest tslib 1.x version.

alan-agius4: @adidahiya, can you explain why updating tslib here should effect consumers of tslint?

Tslib is a direct dependency of tslint and is bound to the TypeScript version used to **build** tslint.

> means our peerDependencies typescript range is invalid.
I don’t think this is correct, if the user uses tslib for their application they will need to explicitly have it as a devDependency and they choose the version based on their TS version.

> I just did a large upgrade to tslib 2.0.0 across many repos and it was fairly disruptive
Can you please elaborate? I’d love to know more. Updating to tslib 2 shouldn’t be distinctive at all, if you use TS 3.9+ to compile your project.

alan-agius4: Ps: I don’t have any problem removing this, just want to have more details.

adidahiya: tslib 1.13 (_not_ 1.12) is compatible with TS <3.9 and 3.9+. So we should declare a dependency on that range, `~1.13.0`.

tslib is not simply used at build time, it is a runtime dependency. It provides helpers for TS language features which are not present in all JS runtime environments, for example object spread and `import *` syntax. So far, tslib and typescript have not declared an official "compatibility matrix" at runtime, and it was generally guaranteed that any version of tslib could work with any version of TS (unless otherwise stated). tslib 2.0 did a major version bump because they broke this contract, and any library which publishes a version which depends on tslib 2.0 at runtime _will not compile for consumers using TS <3.9_. Although it is true that most consumers of tslint simply use the CLI and don't care about the generated typings, there are TS consumers who use the programmatic API (e.g. custom rule plugins) and we should not break that API.

In my experience, when I did a large upgrade to tslib, I experienced some breaking changes in `import *` semantics at runtime which forced me to refactor a lot of code which modifying imports of external modules (specifically, in the Jest mocking system).

In any case, since tslint is deprecated, we should play it safe and not major bump our tslib dependency.

alan-agius4: Thanks @adidahiya for the detailed explanation. I’ll be doing the change tmr, as currently I am AFK.

alan-agius4: Tslib change done, though I still think this is incorrect. 

> tslib 2.0 did a major version bump because they broke this contract, and any library which publishes a version which depends on tslib 2.0 at runtime will not compile for consumers using TS <3.9. 

I don't think this is the case, tslib doesn't effect the emitted DTS hence it will not effect the compilation of the consumers of the library. But since tslint is deprecated I think it's okay to play it safe.

package.json
```diff
@@ -73,7 +73,7 @@
         "tslint-config-prettier": "^1.18.0",
         "tslint-plugin-prettier": "^2.0.1",
         "tslint-test-config-non-relative": "file:test/external/tslint-test-config-non-relative",
-        "typescript": "~3.8.2",
+        "typescript": "~4.0.0-dev.20200722",
```
adidahiya, line 76: can we use `~4.0.0-beta`?

